PR #8, opened 2026-04-14
Updated Dashboard Style

AleBles:
Two small changes QoL changes:
- central breakdown component, re-used for all available breakdowns
- dashboard reflows after terminal size changes


AgentSeal:
Thanks for the PR! The resize handling is a nice touch and something we want to add.

However, the generic Breakdown component won't work for us going forward. We just shipped per-panel custom logic (conditional coloring, computed columns like the 1-shot rate in Activity) that doesn't fit a one-size-fits-all abstraction. Each panel needs to grow independently.

Also, the 104-column cap was intentional - keeps things readable on wide terminals.

Going to close this one, but appreciate the contribution. The resize idea is solid and we'll likely add that separately.

PR #10, opened 2026-04-14
Dynamic Terminal Width

AleBles:
Ok trying this again, now I just replace `process.stdout.columns` with ink's `useWindowSize`, only addressing the dynamicness of the term size :), max size stays as-is

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,29 @@
 # Changelog
 
+## 0.4.0 - 2026-04-14
+
+### Added
+- Codex (OpenAI) support. Parses sessions from ~/.codex/sessions/ with full
+  token tracking, cost calculation, task classification, and tool breakdown.
+- Provider plugin system. Adding a new provider (Pi, OpenCode, Amp) is a
+  single file in src/providers/.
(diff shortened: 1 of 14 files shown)


AgentSeal (commented):
Thanks for the cleaner approach on this one!

The idea is right -- reactive terminal width instead of a one-shot read -- but this will crash at runtime. `useWindowSize()` is a React hook, and `getLayout()` is a plain function, not a component or custom hook. React will throw an "Invalid hook call" error.

Two ways to fix it:

**Option A: Pass columns down (minimal change)**
```tsx
function getLayout(columns?: number): Layout {
  const termWidth = columns || parseInt(process.env['COLUMNS'] ?? '') || 80
  // ... rest stays the same
}

// In each component that calls getLayout:
const { columns } = useWindowSize()
const { dashWidth, wide, halfWidth, barWidth } = getLayout(columns)
```

**Option B: Turn it into a custom hook**
```tsx
function useLayout(): Layout {
  const { columns } = useWindowSize()
  const termWidth = columns || parseInt(process.env['COLUMNS'] ?? '') || 80
  // ... rest stays the same
}
```

Option A is probably simpler since `getLayout` is called in three components and keeping it as a pure function makes it easier to test.

PR #13, opened 2026-04-14
feat: add Codex session support with Claude/Codex filter commands

canaokar:
Parse ~/.codex/sessions/**/*.jsonl alongside Claude Code logs and merge projects by cwd so cross-tool work on the same codebase collapses into one row.

Maps Codex tool names onto Claude's canonical names so the existing classifier works unchanged: shell_command -> Bash, apply_patch -> Edit, update_plan sets hasPlanMode, web_search_call -> WebSearch. Token math subtracts cached_input_tokens from input_tokens to match OpenAI billing.

  Adds three entry points:
  - `codeburn` — both sources (unchanged default)
  - `codeburn claude` — Claude Code only
  - `codeburn codex` — Codex only

Adds gpt-5, gpt-5-codex, gpt-5-mini, gpt-5-nano to pricing and short names. Honors CODEX_HOME env var.


AgentSeal:
Thanks for this -- great timing! We actually just shipped Codex support on main in commit 391a235 about 30 minutes ago.

Our implementation covers the same ground:
- Codex JSONL parsing from `~/.codex/sessions/`
- Tool name normalization to Claude canonical names
- `CODEX_HOME` env var
- Provider filtering (`--provider claude|codex|all` on all commands)
- Model pricing for GPT-5.x models

Plus a few extras:
- Provider plugin system (adding Pi/OpenCode/Amp is one file)
- TUI toggle (press `p` to cycle providers interactively)
- Duplicate token_count event detection
- 1-minute cache for instant provider switching
- Menubar per-provider breakdown
- 10 new tests

Since the core functionality overlaps, I'm going to close this one. But if you see anything in your approach that we missed or got wrong, please open an issue -- especially around tool mappings (you had `update_plan` -> hasPlanMode which is a nice catch we should look at adding).

Thanks for contributing!

PR #20, opened 2026-04-14
feat: project name normalization and session attribution

cordsjon:
## Summary

- **Project name extraction** — new `project-names.ts` module that converts encoded session directory names (e.g., `-Users-alice-projects-my-app`) into readable project names (`my-app`), replacing the raw directory name that currently shows in the dashboard
- **Alias-based dedup** — user-configurable `~/.config/codeburn/project-aliases.json` to merge variant names (e.g., `SVG-PAINT` + `30_SVG-PAINT` → single entry), solving the duplicate project entries problem
- **Session attribution** — `~/.config/codeburn/session-projects.jsonl` for manually attributing workspace sessions (started from a projects root) to specific projects, with a backfill script to retroactively process existing sessions

## Motivation

When users work from a projects root directory (e.g., `~/projects`), all sessions get classified as the same encoded path. Combined with Claude Code creating different encoded directories for the same project accessed via different paths, this fragments cost data across duplicate entries.

In our case, 51% of spend (\$4K/month) was lumped into a single "workspace" bucket with no project visibility, and one project appeared under 3 different names. After this change: workspace dropped to 1.3%, and all projects consolidated correctly.

## Files

| File | What |
|------|------|
| \`src/project-names.ts\` | New module — extract, alias, attribute pipeline |
| \`src/backfill-attribution.ts\` | New script — retroactive workspace attribution |
| \`src/parser.ts\` | Wire \`normalizeProject()\` into \`scanProjectDirs()\` |
| \`src/dashboard.tsx\` | \`shortProject()\` passes through normalized names |

## Config files (user-created)

\`\`\`json
// ~/.config/codeburn/project-aliases.json
{
  "aliases": {
    "SVG-PAINT": "30_SVG-PAINT",
    "keto-data": "50_KETO"
  }
}
\`\`\`

\`\`\`jsonl
// ~/.config/codeburn/session-projects.jsonl (one line per session)
{"session_id": "uuid-here", "project": "30_SVG-PAINT", "date": "2026-04-14"}
\`\`\`

## Test plan

- [x] \`npm run build\` passes (tsup bundle succeeds)
- [x] Verified with real data: 564 workspace sessions, 444 attributed, workspace cost 51% → 1.3%
- [ ] Review \`extractProjectName()\` edge cases with non-standard Claude Code paths
- [ ] Verify no regression in TUI dashboard rendering

### DIFF
--- a/src/backfill-attribution.ts
+++ b/src/backfill-attribution.ts
@@ -1,0 +1,227 @@
+/**
+ * Backfill session attribution for workspace sessions.
+ *
+ * Scans all JSONL session files in the workspace directory, extracts file
+ * paths from tool_use blocks and CWD fields, and determines the dominant
+ * project for each session. Writes results to session-projects.jsonl.
+ *
+ * Usage:
+ *   npx tsx src/backfill-attribution.ts [--dry-run]
+ *
+ * Or after build:
+ *   node dist/backfill-attribution.js [--dry-run]
+ */
+
+import { readFileSync, writeFileSync, existsSync, readdirSync, mkdirSync, appendFileSync, statSync } from 'fs'
+import { join, basename } from 'path'
+import { homedir } from 'os'
+
+const PROJECTS_BASE = join(homedir(), '.claude', 'projects')
+const CONFIG_DIR = process.env['CODEBURN_CONFIG_DIR'] || join(homedir(), '.config', 'codeburn')
+const OUTPUT_PATH = join(CONFIG_DIR, 'session-projects.jsonl')
+const DRY_RUN = process.argv.includes('--dry-run')
+
+/** System path segments to skip when identifying projects. */
+const SYSTEM_SEGS = new Set([
+  'users', 'user', 'home', 'projects', 'documents', 'desktop',
+  'volumes', 'mnt', 'opt', 'var', 'tmp',
+])
+
+/** Generic subdirectory names — too ambiguous to be project names. */
+const GENERIC_DIRS = new Set([
+  'src', 'dist', 'docs', 'scripts', 'tests', 'test', 'lib',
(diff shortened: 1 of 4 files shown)


AgentSeal:
Thanks for the PR @cordsjon. The underlying problem is real - encoded directory names are ugly in the dashboard. But this implementation has issues that block merging:

**No tests.** 468 lines of new code with zero test coverage. `extractProjectName()` has multiple fallback branches and none of them are tested. For a tool that tracks spending, untested parsing logic isn't something we can ship.

**Synchronous file I/O in the hot path.** `readFileSync`, `existsSync`, `statSync` are called inside `normalizeProject()`, which runs for every session file during parsing. The rest of the codebase is async. This blocks the event loop and will cause visible hangs on large session directories.

**Re-introduces a fixed bug.** The `parser.ts` changes still have `!f.startsWith('agent-')` which we removed in #25 because it was silently dropping 46% of API calls. Merging this PR would revert that fix. The backfill script has the same filter.

**Overengineered for the problem.** Three config files, a backfill script, module-level mutable cache state, a three-stage normalization pipeline. The core issue (readable project names) could be a single pure function with no config files.

**`extractProjectName` assumes a `/projects/` directory.** Users with code in `~/code/`, `~/dev/`, `~/work/` hit a fallback heuristic that guesses the last non-system path segment. That's fragile.

**Style.** Block comments and JSDoc on nearly every function. Codebase convention is no comments unless the WHY is non-obvious.

**Test plan incomplete.** Two items in the PR's own checklist are unchecked.

Closing this, but the problem itself is worth solving. A simpler approach would be a single `extractProjectName()` function (async, tested, no config files) that does a best-effort extraction from the encoded path. If you want to take another pass at that, we'd be happy to look at it.

PR #22, opened 2026-04-14
fix: prevent CSV formula injection in exports

serabi:
## Summary

Prevent CSV formula injection in exported reports by prefixing formula-like cells with an apostrophe before CSV escaping.

This covers cells that begin with:
- `=`
- `+`
- `-`
- `@`

Without this, opening an exported CSV in spreadsheet software can treat attacker-controlled values as formulas.

## Changes

- update `escCsv()` in `src/export.ts` to neutralize formula-like cells
- add `tests/export.test.ts` regression coverage for dangerous exported values

## Why

`codeburn export` writes project names, model names, shell commands, and other transcript-derived strings into CSV output. Quoting alone does not reliably prevent spreadsheet formula execution. Prefixing these cells with `'` is the usual mitigation.

## Authorship note

Codex helped write the patch and draft this PR. I reviewed the changes and verified the final result locally.

## Verification

- `npm test -- --run`
- `npm run build`
- `npx tsx src/cli.ts report`
- `npx tsx src/cli.ts today`

### DIFF
--- a/src/export.ts
+++ b/src/export.ts
@@ -3,10 +3,11 @@ import { resolve } from 'path'
 import { CATEGORY_LABELS, type ProjectSummary, type TaskCategory } from './types.js'
 
 function escCsv(s: string): string {
-  if (s.includes(',') || s.includes('"') || s.includes('\n')) {
-    return `"${s.replace(/"/g, '""')}"`
+  const sanitized = /^[=+\-@]/.test(s) ? `'${s}` : s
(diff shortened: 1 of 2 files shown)


AgentSeal:
Good catch — CSV formula injection is a real risk since we export user-derived strings (project names, shell commands, model names). Fix is clean and follows OWASP guidance. Tests are thorough. Merged to main, thanks @serabi!

PR #24, opened 2026-04-14
feat: add code review activity tracking

excellentdrums:
Is this interesting to anyone? Pretty simple stuff. Not sure if it's just noise.

Screenshot 2026-04-14 at 12 02 26 PM


AgentSeal:
Thanks for the PR @excellentdrums. Appreciate you flagging the uncertainty upfront - that's honest and helpful.

After looking at it, I think this one falls into the "noise" category you were wondering about. Two reasons:

1. "review" already exists in `RESEARCH_KEYWORDS`, so adding it to a separate `REVIEW_KEYWORDS` regex creates an overlap where classification depends on check order. "improve" has the same problem - "improve the error handling" reads more like refactoring than code review.

2. Code review as an activity category is pretty niche for how most people use Claude Code. Adding it fragments the activity breakdown without giving much actionable signal.

Going to close this for now but it's a fair idea to explore if we ever rethink the category system. Thanks for contributing.

excellentdrums:
Thanks for your thoughtful consideration! I'd already come to the same conclusion and was thinking along the same lines. I'd begun trying to find a "code reviewer language" keyword list, like "lgtm" or "approved" but it felt like more of a hack than a solution.

Also, not sure if you noticed but the keyword-based methodology already has some overlap issues and the cost aggregation at the Activity level doesn't always add up to the topline cost.

In any event, good stuff! Enlightening! Keep up the good work!

PR #30, opened 2026-04-15
add OpenCode provider

Broderick-Westrope:


AgentSeal:
Hey @Broderick-Westrope, really nice work here. The provider code is clean, the test coverage is thorough, and the subtask exclusion + fallback to OpenCode's own cost field are smart choices.

One concern before we merge: `better-sqlite3` is a native C++ addon that requires a build toolchain to install. Right now codeburn is pure JS with zero native deps, so `npx codeburn` works everywhere without friction. Adding a compilation step would break that for a chunk of users who don't have build tools set up.

Would you be open to swapping it for `sql.js`? It's SQLite compiled to WASM, so it runs anywhere Node runs with no native compilation. The provider logic stays the same, just the import and DB open call change slightly.

Happy to help with the swap if you'd prefer. Either way the rest of the PR is good to go.

Broderick-Westrope:
Hi @AgentSeal, I like the idea of codeburn, nice one! I wasn't so sure about the best approach to do sqlite in TypeScript land so I appreciate the pointer. I've made the switch to `sql.js`. If anything doesn't line up with your expectations feel free to fix it or let me know specifics. Thanks :)

AgentSeal:
Hey @Broderick-Westrope, merged your OpenCode provider to main. Thanks for the solid work -- the parsing logic, subtask exclusion, and cost fallback were all well done.

What changed from your PR:

- Swapped sql.js back to better-sqlite3 since we added a shared SQLite adapter (`src/sqlite.ts`) for Cursor in v0.5.0. OpenCode now uses the same adapter so there's one SQLite driver, not two.
- Lazy-loads the provider (same pattern as Cursor) so users without OpenCode or without better-sqlite3 installed see no difference.
- Added schema validation and stderr diagnostics on failures (matching Cursor's error handling).
- Added `bashCommands` to the provider type so shell command breakdowns work across all providers (your bash extraction code carried over).
- 31 tests, all using better-sqlite3 directly.

Would be great if you could pull main, run `npx codeburn report --provider opencode`, and post a screenshot showing it working with your OpenCode data. That way we can confirm it works end to end before the npm publish.

AgentSeal:
Closing this PR since the OpenCode provider is now on main (commit 2d114d9). Your parsing logic, test patterns, and data model work were the foundation -- we adapted it to use the shared SQLite adapter that landed in v0.5.0.

Thanks again for the contribution! Will tag you when 0.5.1 goes to npm.

Broderick-Westrope:
Easy, thanks. Seems to be working; don't judge the usage 😅 
CleanShot 2026-04-16 at 07 59 01@2x

Issue #27, opened 2026-04-14
Bug: Currency Change Not Working inside swiftbar


aniketmallick:
Was trying out the product, when saw any currency whichever I am trying to switch from the swiftbar is not working; tried refreshing but still nothing happened!

AgentSeal:
codeburn currency GBP          # set to British Pounds
codeburn currency AUD          # set to Australian Dollars
codeburn currency JPY          # set to Japanese Yen
codeburn currency              # show current setting
codeburn currency --reset      # back to USD

set the default currency for Codeburn, it would apply that setting to all, including swiftbar. It might take few minutes. We will be providing a fix for quick currency change in swiftbar soon

PR #33, opened 2026-04-15
fix(menubar): invoke binary directly + correct currency subcommand (#32, #27)

### LINKED ISSUE #32: Menubar "Open Full Report" and "Export CSV" actions not working on npm-installed codeburn
Hi, thanks for codeburn — the TUI dashboard is really nice to have.

Wanted to share something I ran into while setting up the SwiftBar menubar on a fresh `npm i -g codeburn`.

## What I observed
After `codeburn install-menubar`, clicking **Open Full Report** (or **Export CSV to Desktop**) in the menubar opens Terminal and shows:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/<user>/src/cli.ts' imported from /Users/<user>/
  url: 'file:///Users/<user>/src/cli.ts'
```

It looks like it's trying to run the TypeScript source from a dev checkout at `~/codeburn`, which a user who installed via npm wouldn't have.

## Environment
- codeburn 0.4.1 (via `npm i -g codeburn`)
- Node.js v22.18.0 (nvm)
- SwiftBar 2.0.1 (build 536), macOS 26.3.1

## What I think is going on
Looking at `src/menubar.ts`, `bin` gets resolved correctly around line 153, but lines 154-155 for the two action items build a `bash -c "cd ~/codeburn && npx tsx src/cli.ts ..."` command instead of using `${bin}`:

```ts
const bin = getCodeburnBin()
lines.push(`Open Full Report | terminal=true shell=/bin/bash param1=-c param2="cd '${home}/codeburn' && npx tsx src/cli.ts report; ..."`)
lines.push(`Export CSV to Desktop | terminal=false shell=/bin/bash param1=-c param2="cd '${home}/codeburn' && npx tsx src/cli.ts export -o '${home}/Desktop/codeburn-report.csv' 2>/dev/null"`)
```

There also seems to be a small interaction with how SwiftBar serializes `param2=` — the `cd '...' && ...` part in the long quoted string appears to only get `cd` delivered to `bash -c`, and the rest runs in the outer shell from `$HOME`. That's my guess for why the error path ends up at `~/src/cli.ts` rather than `~/codeburn/src/cli.ts`.

One shape that might sidestep both things is letting SwiftBar invoke the binary directly (each `paramN` is a separate argv, so no shell quoting involved):

```ts
lines.push(`Open Full Report | terminal=true shell=${bin} param1=report`)
lines.push(`Export CSV to Desktop | terminal=false shell=${bin} param1=export param2=-o param3=${home}/Desktop/codeburn-report.csv`)
```

But you'll have a much better sense than me of what fits with the rest of the codebase.

## Possibly related
I noticed #27 (currency picker) in passing — the submenu emits `codeburn config currency XXX` but the CLI is `codeburn currency XXX`. Mentioning in case it's useful context; happy to keep that separate if you'd rather.

## Workaround for anyone hitting this in the meantime
Editing `~/Library/Application Support/SwiftBar/plugins/codeburn.5m.sh` to post-process the output with `sed` has been working for me:

```bash
CODEBURN=$(command -v codeburn)
CSV_PATH="$HOME/Desktop/codeburn-report.csv"

"$CODEBURN" status --format menubar 2>/dev/null | sed -E \
  -e "s#^Open Full Report \|.*#Open Full Report | terminal=true shell=$CODEBURN param1=report#" \
  -e "s#^Export CSV to Desktop \|.*#Export CSV to Desktop | terminal=false shell=$CODEBURN param1=export param2=-o param3=$CSV_PATH#" \
  -e 's#codeburn config currency#codeburn currency#g' \
  || echo "-- | sfimage=flame.fill"
```

Thanks again for the project, and no rush on this one.

MukundaKatta:
## Summary

Closes #32 and #27.

The two action items at the bottom of the menubar plugin were built as:

\`\`\`
Open Full Report | terminal=true shell=/bin/bash param1=-c param2=\"cd '<HOME>/codeburn' && npx tsx src/cli.ts report; ...\"
Export CSV to Desktop | terminal=false shell=/bin/bash param1=-c param2=\"cd '<HOME>/codeburn' && npx tsx src/cli.ts export -o '<HOME>/Desktop/codeburn-report.csv' 2>/dev/null\"
\`\`\`

That breaks for npm-installed users in two ways:

1. **No \`~/codeburn\` checkout exists.** That path is only present on a dev clone — \`npm i -g codeburn\` doesn't ship a source tree.
2. **SwiftBar's \`param2=\` quoting eats the \`&&\`.** Only the \`cd\` half makes it into \`bash -c\`; the rest runs in the parent shell from \`$HOME\`, so the user sees \`Cannot find module '/Users/<u>/src/cli.ts'\` (the issue's exact symptom).

Fix: invoke the resolved \`\${bin}\` directly with discrete \`paramN=\` args. SwiftBar/xbar deliver each \`paramN=\` value as its own argv entry, so there's no shell quoting at all — and we no longer assume any checkout layout.

\`\`\`
Open Full Report | terminal=true shell=<bin> param1=report
Export CSV to Desktop | terminal=false shell=<bin> param1=export param2=-o param3=<HOME>/Desktop/codeburn-report.csv
\`\`\`

### Bonus: currency submenu (#27)

While here, the currency picker was emitting \`\${bin} config currency XXX\` — but the real subcommand in \`src/cli.ts\` is \`codeburn currency [code]\` (with \`--reset\` for USD). The \`config currency\` form silently fails on click. Fixed using the same direct-invoke pattern:

\`\`\`
--US Dollar (USD) * | terminal=false refresh=true shell=<bin> param1=currency param2=--reset
--British Pound (GBP) | terminal=false refresh=true shell=<bin> param1=currency param2=GBP
\`\`\`

### Test plan

- [x] \`npm run build\` — clean.
- [x] \`npm test -- --run\` — 28/28 pass.
- [x] Manually ran \`node ./dist/cli.js status --format menubar\` and inspected the rendered lines; matches the patterns above.

### AI-assisted disclosure

Drafted with Claude Code based on the diagnosis in #32 (the reporter had already pinpointed the param-quoting issue and proposed exactly this shape). I also verified the related #27 currency-subcommand mismatch in \`src/cli.ts\` and folded that fix in. I understand what the code does.

### DIFF
--- a/src/menubar.ts
+++ b/src/menubar.ts
@@ -151,8 +151,12 @@ export function renderMenubarFormat(
   lines.push('---')
   const home = process.env.HOME ?? '~'
   const bin = getCodeburnBin()
-  lines.push(`Open Full Report | terminal=true shell=/bin/bash param1=-c param2="cd '${home}/codeburn' && npx tsx src/cli.ts report; echo ''; echo 'Press any key to close...'; read -n1"`)
-  lines.push(`Export CSV to Desktop | terminal=false shell=/bin/bash param1=-c param2="cd '${home}/codeburn' && npx tsx src/cli.ts export -o '${home}/Desktop/codeburn-report.csv' 2>/dev/null"`)
+  // Invoke the resolved `codeburn` binary directly. SwiftBar/xbar deliver
+  // each `paramN=` value as its own argv entry, so there's no shell
+  // quoting involved — and we don't ship the user to a `~/codeburn`
+  // checkout that only exists when running from a dev clone (#32).
+  lines.push(`Open Full Report | terminal=true shell=${bin} param1=report`)
+  lines.push(`Export CSV to Desktop | terminal=false shell=${bin} param1=export param2=-o param3=${home}/Desktop/codeburn-report.csv`)
 
   // Currency submenu -- common currencies as clickable items.
   // Clicking one runs 'codeburn config currency XXX' and refreshes the plugin.
@@ -179,10 +183,14 @@ export function renderMenubarFormat(
   lines.push(`Currency: ${activeCurrency} | size=14`)
   for (const { code, name } of currencies) {
     const check = code === activeCurrency ? ' *' : ''
-    const cmd = code === 'USD'
-      ? `${bin} config currency --reset`
-      : `${bin} config currency ${code}`
-    lines.push(`--${name} (${code})${check} | terminal=false refresh=true shell=/bin/bash param1=-c param2="${cmd}"`)
+    // The real CLI subcommand is `codeburn currency [code]` (with `--reset`
+    // for USD), not `codeburn config currency` — the latter doesn't exist
+    // and silently fails when SwiftBar runs it. Fixes #27.
+    if (code === 'USD') {
+      lines.push(`--${name} (${code})${check} | terminal=false refresh=true shell=${bin} param1=currency param2=--reset`)
+    } else {
+      lines.push(`--${name} (${code})${check} | terminal=false refresh=true shell=${bin} param1=currency param2=${code}`)
+    }
   }
 
   lines.push(`Refresh | refresh=true`)


AgentSeal:
Nice fix, both bugs confirmed and the paramN approach is cleaner than shell-in-shell quoting. Merged to main, thanks @MukundaKatta!

PR #39, opened 2026-04-15
feat: add scrollable dashboard with sticky header and status bar

AleBles:
- Added ScrollPanel component (src/components/scroll-panel.tsx) - native Ink scrolling using overflow="hidden" + negative marginTop, no external dependencies (from https://gist.github.com/janis-me/1a55a8747f12e1c4fc86ebe2d05a0a55)                                                                     
  - Dashboard content (breakdowns, charts) now scrolls within a fixed viewport; Overview panel and StatusBar remain sticky
  - Scroll support for arrow keys and PgUp/PgDn                                                                                                                                                                                     
  - StatusBar split into two rows with separate navigation and period hints                                                                                                                                                                                                                                                                                             
  - Viewport height calculated from terminal rows minus fixed UI elements, adapts on resize  
  
  This makes the interactive terminal fixed within the available viewport, removing the need for mouse scroll and instead utilize keyboard keys to go trought the content which is consistent with existing naviation

### DIFF
--- a/src/components/scroll-panel.tsx
+++ b/src/components/scroll-panel.tsx
@@ -1,0 +1,113 @@
+// from https://gist.github.com/janis-me/1a55a8747f12e1c4fc86ebe2d05a0a55
+
+import { Box, DOMElement, measureElement, useFocus, useInput } from 'ink';
+import { useEffect, useReducer, useRef } from 'react';
+
+interface ScrollPanelState {
+  innerHeight: number;
+  height: number;
+  scrollTop: number;
+}
+
+type ScrollPanelAction =
+  | { type: 'SET_INNER_HEIGHT'; innerHeight: number }
+  | { type: 'SET_HEIGHT'; height: number }
+  | { type: 'SCROLL_DOWN' }
+  | { type: 'SCROLL_UP' }
+  | { type: 'PAGE_DOWN' }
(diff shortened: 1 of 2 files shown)


AgentSeal:
Hey @AleBles, thanks for this PR and for thinking about terminal sizing. Going to close this one for now.

A few reasons:

The bigger issue is an arrow-key conflict with our existing period switcher. Arrow keys already switch between Today / 7 Days / 30 Days / Month in the dashboard. The ScrollPanel uses `useInput` + `useFocus` to capture arrow keys for scrolling, which would silently break period switching. The StatusBar updates label `←→` as both "switch" and `↑↓` as "scroll" but one handler has to win.

Smaller things: the ScrollPanel is copied from an unlicensed gist (we're MIT), the import is missing the `.js` extension (breaks ESM build), `process.stdout.rows` doesn't react to vertical resize, and `FIXED_ROWS = 10` is hardcoded in a way that breaks if any header grows.

We have a few bigger features in the pipeline that'll reshape how the dashboard is laid out, and we'd rather land keyboard-scrollable viewport as part of that work -- with different keys for scroll (probably `j`/`k` or similar) and a proper resize hook.

Really appreciate you taking the time on this. If you want to contribute again, a provider (see #45 for Pi) or tests for existing code would be great landing points.

PR #44, opened 2026-04-15
feat: add GitHub Copilot provider

theodorosD:
- Parse ~/.copilot/session-state/*/events.jsonl
- Track model via session.model_change events
- Extract tools from assistant.message toolRequests
- Add fallback pricing for gpt-4.1, gpt-4.1-mini, gpt-4.1-nano, gpt-5-mini, o3, o4-mini
- Note: only outputTokens available (Copilot does not log input tokens)
- 9 tests covering JSONL parsing, model tracking, tool extraction,
  deduplication, and metadata
- Fix modelDisplayName to match longest key first (gpt-4.1-mini before gpt-4.1)
- Update provider-registry test for new coreProviders list

### DIFF
--- a/src/models.ts
+++ b/src/models.ts
@@ -42,6 +42,12 @@ const FALLBACK_PRICING: Record<string, ModelCosts> = {
   'gpt-5.4': { inputCostPerToken: 2.5e-6, outputCostPerToken: 10e-6, cacheWriteCostPerToken: 2.5e-6, cacheReadCostPerToken: 1.25e-6, webSearchCostPerRequest: WEB_SEARCH_COST, fastMultiplier: 1 },
   'gpt-5.4-mini': { inputCostPerToken: 0.4e-6, outputCostPerToken: 1.6e-6, cacheWriteCostPerToken: 0.4e-6, cacheReadCostPerToken: 0.2e-6, webSearchCostPerRequest: WEB_SEARCH_COST, fastMultiplier: 1 },
   'gpt-5': { inputCostPerToken: 2.5e-6, outputCostPerToken: 10e-6, cacheWriteCostPerToken: 2.5e-6, cacheReadCostPerToken: 1.25e-6, webSearchCostPerRequest: WEB_SEARCH_COST, fastMultiplier: 1 },
+  'gpt-5-mini': { inputCostPerToken: 0.4e-6, outputCostPerToken: 1.6e-6, cacheWriteCostPerToken: 0.4e-6, cacheReadCostPerToken: 0.2e-6, webSearchCostPerRequest: WEB_SEARCH_COST, fastMultiplier: 1 },
+  'gpt-4.1': { inputCostPerToken: 2e-6, outputCostPerToken: 8e-6, cacheWriteCostPerToken: 2e-6, cacheReadCostPerToken: 0.5e-6, webSearchCostPerRequest: WEB_SEARCH_COST, fastMultiplier: 1 },
+  'gpt-4.1-mini': { inputCostPerToken: 0.4e-6, outputCostPerToken: 1.6e-6, cacheWriteCostPerToken: 0.4e-6, cacheReadCostPerToken: 0.1e-6, webSearchCostPerRequest: WEB_SEARCH_COST, fastMultiplier: 1 },
+  'gpt-4.1-nano': { inputCostPerToken: 0.1e-6, outputCostPerToken: 0.4e-6, cacheWriteCostPerToken: 0.1e-6, cacheReadCostPerToken: 0.025e-6, webSearchCostPerRequest: WEB_SEARCH_COST, fastMultiplier: 1 },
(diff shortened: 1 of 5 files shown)


AgentSeal:
Hey @theodorosD, nice work. The provider is clean, the model change tracking is a good detail, and the pricing additions to models.ts are useful for all providers.

Things to fix:

**Rebase onto current main**

We merged OpenCode provider + bashCommands type in v0.5.1. Your index.ts and registry test will conflict. Rebase and keep only your copilot-specific changes.

**Add `bashCommands: []` to the yield**

`ParsedProviderCall` now requires a `bashCommands: string[]` field. Add `bashCommands: [],` to your yield object. If Copilot's JSONL includes the actual bash command strings in tool requests, you could wire up `extractBashCommands` from `../bash-utils.js` -- but `[]` is fine if the data isn't there.

**Add a factory function for testability**

Right now you export a singleton `copilot` object. Add a `createCopilotProvider(sessionStateDir?: string)` factory (same pattern as every other provider). This lets tests pass a temp dir and actually test session discovery. Your test file has a comment acknowledging this gap -- the workspace.yaml parsing and directory scanning are untested.

**Stronger event types**

Replace the `Record<string, unknown>` data field with typed variants:

```ts
type ModelChangeData = { newModel: string; previousModel?: string }
type UserMessageData = { content: string }
type AssistantMessageData = {
  messageId: string
  outputTokens: number
  toolRequests?: Array<{ name: string }>
}
```

This removes the `as string` / `as number` casts.

**Screenshot request**

Once updated, please run `npx codeburn report --provider copilot` and post a screenshot. We don't have Copilot session data to verify end-to-end.

theodorosD:
Screen Shot 2026-04-16 at 03 32 06 958 PM

AgentSeal:
Hey @theodorosD, thanks for the turnaround. Factory, typed events, bashCommands, and the screenshot all look good. 13 tests pass locally on a clean checkout.

Before I merge, a few things to address.

**1. Please drop `src/providers/kiro.ts` from this PR.**

The second commit says "fix: add missing bashCommands field to Kiro provider yield", but it actually adds a new 183 line Kiro provider that does not exist on main and is not registered in `providers/index.ts`. It also has an em dash in the header comment and an unused `basename` import. If you want to add Kiro, open a separate PR with the provider wired into `coreProviders`, tests, and a screenshot. Keeping this PR focused on Copilot makes it easier to review and revert if anything goes wrong.

**2. Document the zero input tokens limitation.**

Copilot only logs `outputTokens`, so `calculateCost(currentModel, 0, outputTokens, 0, 0, 0)` undercounts real API cost by a meaningful amount. The code is correct for the data you have, but a reader will not know that. Add a one line comment above the `calculateCost` call, and mention it in the PR description so users understand Copilot rows will sit lower than reality.

**3. Handle the case where `session.model_change` never fires.**

`currentModel = 'gpt-4.1'` at the top of the parser means if a Copilot session log is missing the initial `session.model_change` event, every assistant message gets attributed to gpt-4.1 with gpt-4.1 pricing. That is silent and wrong. Two options:

* Leave `currentModel` empty and skip calls until a model is known.
* Read the initial model from `workspace.yaml` if it carries one, then fall back.

A test for the missing model change path would be good.

**4. `workspace.yaml` regex needs to handle quotes and comments.**

`/^cwd:\s*(.+)$/m` captures too much. Examples that break:

* `cwd: "/home/user/app"` gives `app"` after `basename`.
* `cwd: /home/user/app  # project root` gives `# project root`.
* `cwd: '/path/with spaces/app'` gives `app'`.

Strip surrounding quotes and trailing `# ...` comments before calling `basename`, and add a test for the quoted case.

**Smaller items you can fold in or skip:**

* Pull the default `'gpt-4.1'` string into a named const at module scope.
* `delete_file` maps to `Edit`, which is not accurate. Map to `Delete` or leave it raw.
* `ToolRequest.toolCallId`, `ToolRequest.type`, `UserMessageData.interactionId`, `AssistantMessageData.interactionId`, and `ModelChangeData.previousModel` are declared but never read. Either remove them or add a short comment saying they document the on disk schema.
* The metadata test for `modelDisplayName` passes even if the longest match first sort is removed, because the insertion order already happens to favor the longer keys. A test that asserts `modelDisplayName('gpt-5-mini-2026-01-01')` returns `'GPT-5 Mini'` (not `'GPT-5'`) would lock that behavior in.

Once `kiro.ts` is dropped and items 2 through 4 are addressed, I am happy to merge.

theodorosD:
I hope the last push is good to go!

AgentSeal (approved):
All four round-2 items addressed: kiro.ts dropped, zero-input-tokens documented inline, currentModel starts empty with skip-until-known, workspace.yaml regex handles quotes and trailing comments. 7 new tests covering those paths plus the missing-model-change case. 16 Copilot tests total, 109 on branch. Thanks @theodorosD.

PR #46, opened 2026-04-15
feat: add Hermes Agent provider

AICodingJedi:
Adds a CodeBurn provider for the Hermes AI agent (https://github.com/AICodingJedi/hermes-agent).

Hermes stores session data in a SQLite database (~/.hermes/state.db) with cumulative token counts at the session level. This provider distributes tokens proportionally across assistant turns:
- Input tokens: equal split (shared conversation context)
- Output tokens: proportional by assistant content length
- Cache/reasoning tokens: proportional split

Features:
- Lazy-loaded via loadHermes() (same pattern as Cursor/OpenCode)
- Maps 25+ Hermes tools to CodeBurn display names
- Maps 11 model IDs to readable display names
- Deduplication keys prevent duplicate entries on re-scan
- Graceful fallback when better-sqlite3 is unavailable
- Supports HERMES_HOME env var for custom data directories

### DIFF
--- a/src/providers/hermes.ts
+++ b/src/providers/hermes.ts
@@ -1,0 +1,489 @@
+/**
+ * CodeBurn Provider Plugin for Hermes Agent
+ *
+ * Reads Hermes session data from SQLite (state.db) and JSONL trajectory files,
+ * distributes cumulative session-level tokens proportionally across assistant turns,
+ * and yields ParsedProviderCall objects for the CodeBurn TUI dashboard.
+ *
+ * Install: copy this file into codeburn/src/providers/ and register in index.ts,
+ *   or use as a standalone plugin (see README).
+ */
+
+import { readdir, readFile, stat } from 'fs/promises'
+import { basename, join } from 'path'
+import { homedir } from 'os'
+
+import type { Provider, SessionSource, SessionParser, ParsedProviderCall } from './types.js'
+
+// ---------------------------------------------------------------------------
+// Model display name mappings
+// ---------------------------------------------------------------------------
+const modelDisplayNames: Record<string, string> = {
+  'glm-5.1':                          'GLM 5.1',
+  'minimax/minimax-m2.7':             'MiniMax M2.7',
+  'nvidia/nemotron-3-super-120b-a12b': 'Nemotron 3 Super',
+  'google/gemma-4-26b-a4b-it':        'Gemma 4 26B',
+  'google/gemma-4-31b-it':           'Gemma 4 31B',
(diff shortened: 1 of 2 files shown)


AgentSeal:
Hey @AICodingJedi, thanks for adding Hermes support. The token distribution approach for cumulative session-level totals is a smart solve. A few things need fixing before we can merge:

**Must fix:**

1. **Use `src/sqlite.ts` instead of rolling your own loader.** We have a shared SQLite adapter that all providers use (Cursor, OpenCode). Your PR duplicates ~60 lines of loading logic. Import `isSqliteAvailable`, `getSqliteLoadError`, `openDatabase` from `../sqlite.js` and delete your custom `loadSqlite`, `isSqliteAvailable`, `getSqliteLoadError`, `openDatabase` functions plus the `SqliteDatabase` type. See `src/providers/opencode.ts` for the pattern.

2. **Add `bashCommands: string[]` to your yield.** We added this field to `ParsedProviderCall` in v0.5.1. Your PR will fail TypeScript without it. If Hermes has bash/terminal tool calls you can extract commands from, wire it up via `extractBashCommands` from `../bash-utils.js`. Otherwise just yield `bashCommands: []`.

3. **Move `calculateCost` to a top-level import.** You're doing `await import('../models.js')` inside the generator loop. Just add `import { calculateCost } from '../models.js'` at the top. No need for dynamic import here.

4. **Remove unused imports.** `readdir`, `readFile`, `basename` are imported but never used.

5. **Add tests.** We need at least: session discovery, archived/empty session handling, token distribution math, dedup, cost fallback, corrupt data. See `tests/providers/opencode.test.ts` for the pattern (creates real SQLite databases, 31 tests).

6. **Fix dedup key.** Your key includes derived token counts (`${turnInputTokens}:${turnOutputTokens}`). If the proportional split heuristic changes, the same messages get double-counted. Use `hermes:${session.id}:${turn.assistantMsg.id}` instead.

**Style (per our CLAUDE.md):**

7. Remove the section separator banners (`// ----`) and the file-header doc comment. We keep comments minimal -- only when the WHY is non-obvious.

8. Remove the `as Function` cast in `openDatabase` (goes away when you switch to `sqlite.ts`).

9. Add a newline at end of file (both `hermes.ts` and `index.ts`).

10. Don't move the `opencodeProvider`/`opencodeLoadAttempted` declarations in `index.ts`. Keep your changes to adding Hermes only -- no cosmetic reordering of existing code.

The core logic (session parsing, tool mapping, model names, token distribution) is solid. Once the above is addressed it should merge cleanly. Happy to help if you have questions on any of these.

Issue #12, opened 2026-04-14
Feature proposals from a power user (15+ projects, 170+ sessions/month)


sterngold:
First — great tool. Installed via `npx`, zero friction, immediately useful on a 15-project setup running 170+ sessions/month across Opus and Sonnet. Starred.

Here are things I'd find valuable as a heavy user:

## 1. Per-project drill-down

The project breakdown shows cost + session count. I'd love `codeburn report --project "My/projects"` to see the full dashboard (activity type, model split, tools, one-shot rate) scoped to just that project. When one project dominates spend (mine is 88%), the aggregate view hides what's happening in smaller ones.

## 2. Cost-per-session metric

Average cost per session alongside the project breakdown. A $200/day number is less actionable than knowing 3 sessions averaged $60 while the rest averaged $5.

## 3. Date range filtering on `report`

`codeburn report --from 2026-04-07 --to 2026-04-10`. Related to #5 (billing period) but more general. The export command could benefit from this too.

## 4. Model efficiency comparison

For users who route between Opus/Sonnet intentionally: one-shot rate broken down by model, not just by activity. If Sonnet's one-shot rate on coding is 90% vs Opus at 93%, that's a meaningful routing decision.

## 5. Session outlier detection

Flag sessions costing >2x the project average. A simple "top 5 most expensive sessions" with timestamp + project + dominant activity helps spot what went wrong (or right).

## 6. `--json` flag on `report`

The TUI is great for human eyes, but piping structured output to dashboards, scripts, or notes would extend the value significantly.

---

Happy to test any of these. Running macOS 26, Claude Code on Opus 4.6 primarily, 170 sessions/month with submodules and multiple MCP servers.

mallek:
Added --project and  --exclude on this PR to help address your number 1
https://github.com/AgentSeal/codeburn/pull/52

mallek:
Added --format json to address your number 6
#53

AgentSeal:
Hi @sterngold, circling back with status. Four of your six proposals shipped in 0.6.0 / 0.6.1.

### Shipped

**1. Per-project drill-down.** `--project <name>` and `--exclude <name>` on report, today, month, status, and export. Shipped in 0.6.1 (PR #52 by @mallek).

**2. Cost-per-session metric.** avg/s column in the By Project panel. Shipped in 0.6.0 (PR #51 by @lfl1337).

[...]

PR #51, opened 2026-04-16
feat: All Time period, avg/s per project, Top Sessions panel

lfl1337:
## What changed

- **All Time period (key `5`)** — shows all usage data since the first recorded session.
  Daily Activity adapts dynamically and shows every available day instead of a fixed window.
- **avg/s column in By Project** — average cost per session alongside total cost and session count.
- **Top Sessions panel** — highlights the 5 most expensive sessions across all projects,
  showing date, project, cost, and API call count.

## Why

Closes part of #12 (power user feature requests):
- Item 2: cost-per-session metric
- Item 5: session outlier detection (top 5 most expensive)

Also adds the All Time view that was partially stubbed out in `cli.ts` but never wired up.

## How to test

```bash
codeburn              # press 5 for All Time, check avg/s column and Top Sessions panel
codeburn -p all       # start directly in All Time
```

Tests: `npx vitest run` — 93/93 passing

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -50,10 +50,11 @@ function getDateRange(period: string): { range: DateRange; label: string } {
   }
 }
 
-function toPeriod(s: string): 'today' | 'week' | '30days' | 'month' {
+function toPeriod(s: string): 'today' | 'week' | '30days' | 'month' | 'all' {
   if (s === 'today') return 'today'
   if (s === 'month') return 'month'
   if (s === '30days') return '30days'
+  if (s === 'all') return 'all'
   return 'week'
 }
 
@@ -69,7 +70,7 @@ program.hook('preAction', async () => {
 program
   .command('report', { isDefault: true })
   .description('Interactive usage dashboard')
-  .option('-p, --period <period>', 'Starting period: today, week, 30days, month', 'week')
+  .option('-p, --period <period>', 'Starting period: today, week, 30days, month, all', 'week')
   .option('--provider <provider>', 'Filter by provider: all, claude, codex, cursor', 'all')
   .option('--refresh <seconds>', 'Auto-refresh interval in seconds', parseInt)
   .action(async (opts) => {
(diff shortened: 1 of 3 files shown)


AgentSeal:
Hey @lfl1337, thanks for this. Clean PR, stays inside the existing patterns, and the three features all tie back to concrete asks in #12. avg/s is the one I am most excited about, it turns the Project panel from a totals list into something actionable.

A few things before I merge.

**1. Please rebase onto main.**

main has moved since you opened this. Pi, OpenCode, and Copilot providers all landed. Your PR body says 93 tests passing, but current main sits at 106. After rebase, run `npx vitest run` and confirm the count is still green.

**2. Empty `firstTimestamp` on Top Sessions.**

The Copilot provider yields `timestamp: ''` when an event is missing one. In `TopSessions`, `session.firstTimestamp.slice(0, 10)` returns an empty string, so the row label starts with just a space before the project name. Not a crash, just ugly. Either filter sessions with empty timestamps, or show a placeholder like `----------` when missing.

**3. All Time on very large histories.**

`new Date(0)` pulls in every session ever. For users with months of Claude Code history this works, but it loads a lot. Not a blocker, but worth confirming it stays snappy on your own setup when you press `5`. If it drags, we can cap the day range.

**4. A couple of quick tests would be good.**

The top-5 selection and the avg/s math are both pure functions of ProjectSummary. A short test file covering:

* top 5 with fewer than 5 sessions total
* top 5 with tied costs (stable sort)
* avg/s with zero-session project returns \`-\`

would give us confidence the panel holds up on weird inputs.

**5. Screenshot?**

Would you mind dropping a screenshot of the dashboard with All Time active, showing the new avg/s column and the Top Sessions panel? Helps users in the PR and gives me something for the release notes.

Small things you can fold in or ignore:

* \`days !== undefined ? ... : ...\` in \`DailyActivity\` is now a tri-state (14, 31, undefined). A short comment above \`const days = period === 'all' ? undefined : ...\` explaining that undefined means all days would save the next reader a minute.
* The \`PROJECT_COL_AVG = 7\` constant makes sense, but the existing \`padStart(8)\` for cost and \`padStart(6)\` for sess are still inline magic numbers. Not introduced by you, just noting for a future cleanup pass.

Once rebased and the empty-timestamp case handled, happy to merge.

lfl1337:
Screenshot 2026-04-16 155202

AgentSeal (approved):
All five review items addressed cleanly: rebased onto main, empty firstTimestamp placeholder, three requested edge-case tests plus two extras, tri-state days comment, screenshot posted. Tests go from 93 to 98 on main. Thanks @lfl1337.

PR #52, opened 2026-04-16
feat: add --project and --exclude filters for project-level filtering

mallek:
Addresses #12 (item 1 — per-project drill-down).

## Summary

Adds two new repeatable CLI flags to all commands (`report`, `today`, `month`, `status`, `export`):

- `--project <name>` — include only projects matching name (case-insensitive substring match)
- `--exclude <name>` — exclude projects matching name (case-insensitive substring match)

Both flags are repeatable to match multiple projects:

```bash
codeburn report --exclude myapp --exclude tests
codeburn month --project api --project web
```

Filters can be combined with the existing `--provider` flag.

## Motivation

When tracking multiple projects, it's useful to isolate costs to a specific project or hide noisy projects from aggregate totals. The `--provider` flag exists for tool-level filtering, but there was no way to filter by project.

## Changes

- **`src/parser.ts`** — new `filterProjectsByName()` function supporting include/exclude arrays
- **`src/cli.ts`** — `--project` and `--exclude` options on all commands, repeatable via Commander collect pattern
- **`src/dashboard.tsx`** — thread filters through to interactive dashboard and reload callbacks
- **`README.md`** — documented project filtering with examples

## Test plan

- [x] `status --format json` totals match: `--project X` + `--exclude X` ≈ unfiltered total
- [x] Multiple values: `--exclude a --exclude b` excludes both
- [x] Multiple values: `--project a --project b` includes both
- [x] `--help` shows new flags on all commands
- [x] Clean build with `tsup`

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,20 @@ codeburn export --provider claude    # export Claude data only
 
 The `--provider` flag works on all commands: `report`, `today`, `month`, `status`, `export`.
 
+### Project filtering
+
+Filter results by project name (case-insensitive substring match). Both flags are repeatable:
+
+```bash
+codeburn report --project myapp                  # show only projects matching "myapp"
+codeburn report --exclude myapp                  # show everything except "myapp"
+codeburn report --exclude myapp --exclude tests  # exclude multiple projects
+codeburn month --project api --project web       # include multiple projects
+codeburn export --project inventory              # export only "inventory" project data
+```
+
+The `--project` and `--exclude` flags work on all commands and can be combined with `--provider`.
+
 ### Supported providers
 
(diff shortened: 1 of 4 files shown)


AgentSeal:
Merged, thanks for the contribution!

Added two follow-up commits on top before merging:
- Unit tests for `filterProjectsByName` covering include OR, exclude AND-negation, case-insensitivity, path vs name matching, empty-string edge case, and input immutability.
- Small fix so `status --format menubar` applies the filter to the per-provider today breakdown too (the main period blocks already did, that one loop was missed).

Landed in main as 87210da. Will be in the next release.

PR #59, opened 2026-04-16
Add OMP provider and model alias mapping

cgrossde:
### OMP provider support

[Oh My Pi](https://github.com/can1357/oh-my-pi) stores sessions at `~/.omp/agent/sessions/` in the same JSONL format as Pi. This PR adds OMP as a first-class provider so its sessions are discovered and tracked automatically.

- New `omp` provider reads from `~/.omp/agent/sessions/<sanitized-cwd>/*.jsonl`
- Shares all parsing logic with the existing Pi provider (same JSONL format, same tool names, same dedup strategy)
- Sessions carry `provider: 'omp'` so the `--provider omp` filter works
- Both providers remain independent — users running both Pi and OMP get separate entries in the dashboard

### Model alias mapping

Proxies and self-hosted LLM gateways often rewrite model names before they reach the client. When a model name doesn't match any LiteLLM entry, `calculateCost` returns 0 and the session shows `$0.00`. There was no way to fix this without a codeburn release.

This PR adds a user-configurable alias map:

```bash
codeburn model-alias "my-proxy-model" "claude-opus-4-6"  # add
codeburn model-alias --list                               # list
codeburn model-alias --remove "my-proxy-model"            # remove
```

Aliases are stored in `~/.config/codeburn/config.json` alongside the currency setting and applied at startup before any pricing lookup. User aliases take precedence over built-ins.

Built-in aliases ship for known proxy model name variants — specifically the `anthropic--claude-4.6-opus` double-dash dot-version format emitted by some Anthropic-compatible proxies — so common cases work out of the box without any configuration.

**Resolution order**: `getCanonicalName` strips structural noise first (pin suffixes, date stamps, `provider/` slash prefixes), then `resolveAlias` maps the result to the canonical key used by LiteLLM and the fallback pricing table.

## Files changed

- `src/providers/pi.ts` — OMP provider factory and singleton; `discoverSessionsInDir` parameterized by provider name
- `src/providers/index.ts` — OMP registered in core providers
- `src/models.ts` — `BUILTIN_ALIASES`, `setModelAliases`, `resolveAlias`; alias resolution wired into `getModelCosts` and `getShortModelName`
- `src/config.ts` — `modelAliases` field added to `CodeburnConfig`
- `src/cli.ts` — `model-alias` command; `setModelAliases` called in `preAction`
- `README.md` — OMP in provider table and requirements; new Model aliases section
- `tests/providers/omp.test.ts` — new
- `tests/models.test.ts` — new, covers builtin aliases, user aliases, override precedence, and existing model names

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
   <img src="https://raw.githubusercontent.com/getagentseal/codeburn/main/assets/dashboard.jpg" alt="CodeBurn TUI dashboard" width="620" />
 </p>
 
-By task type, tool, model, MCP server, and project. Supports **Claude Code**, **Codex** (OpenAI), **Cursor**, **cursor-agent**, **OpenCode**, **Pi**, and **GitHub Copilot** with a provider plugin system. Tracks one-shot success rate per activity type so you can see where the AI nails it first try vs. burns tokens on edit/test/fix retries. Interactive TUI dashboard with gradient charts, responsive panels, and keyboard navigation. Native macOS menubar app in `mac/`. CSV/JSON export.
+By task type, tool, model, MCP server, and project. Supports **Claude Code**, **Codex** (OpenAI), **Cursor**, **cursor-agent**, **OpenCode**, **Pi**, **[OMP](https://github.com/can1357/oh-my-pi)** (Oh My Pi), and **GitHub Copilot** with a provider plugin system. Tracks one-shot success rate per activity type so you can see where the AI nails it first try vs. burns tokens on edit/test/fix retries. Interactive TUI dashboard with gradient charts, responsive panels, and keyboard navigation. Native macOS menubar app in `mac/`. CSV/JSON export.
 
 Works by reading session data directly from disk. No wrapper, no proxy, no API keys. Pricing from LiteLLM (auto-cached, all models supported).
 
@@ -36,7 +36,7 @@ npx codeburn
 ### Requirements
 
 - Node.js 22+
-- Claude Code (`~/.claude/projects/`), Codex (`~/.codex/sessions/`), Cursor, OpenCode, Pi (`~/.pi/agent/sessions/`), and/or GitHub Copilot (`~/.copilot/session-state/`)
+- Claude Code (`~/.claude/projects/`), Codex (`~/.codex/sessions/`), Cursor, OpenCode, Pi (`~/.pi/agent/sessions/`), OMP (`~/.omp/agent/sessions/`), and/or GitHub Copilot (`~/.copilot/session-state/`)
 - For Cursor/OpenCode support: uses Node's built-in `node:sqlite` (Node 22+)
 
 ## Usage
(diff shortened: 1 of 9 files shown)


cgrossde:
@iamtoruk If you find some time, please have a look. Let me know if you would like to see some changes or prefer not to merge this.

iamtoruk:
Hey Chris, thanks for this. The OMP provider and the model-alias work are both useful, and the tests are thorough. I wanted to land it cleanly today so I did the rebase work on our side and pushed a reconciled branch to our origin:

https://github.com/getagentseal/codeburn/tree/feat/omp-support-model-aliases

Since main moved quite a bit while this PR was open (PR #117 just merged, and we went through Node 22+ / node:sqlite changes) there were six conflict files to work through. I kept your OMP provider and alias feature as-is and applied a few small fixups on top that I want to flag:

1. **Prototype-pollution hardening in `resolveAlias`.** The `??` chain let `Object.prototype` leak through when a model was literally named `__proto__`. Our existing prototype-pollution test suite caught it. Switched to `Object.hasOwn` checks.

2. **Dedup key and discovery cache now use `source.provider`.** The PR's `discoverSessionsInDir` accepts `providerName` but three call sites were still using the literal `'pi'`: the dedup key prefix, `saveDiscoveryCache`, and `parserVersion`. OMP sources now prefix with `omp:` and land in their own cache namespace. Paths already discriminated so this wasn't a correctness bug, just cosmetics, but it keeps the cache and dedup semantics matching what we did for cursor-agent.

3. **Tiny cleanup.** Combined the duplicated `pi.js` imports in `providers/index.ts` into one line, added a trailing newline to `pi.ts`.

Everything passes: 378 tests, cached-vs-uncached parity exact, and the `codeburn model-alias` flow works end-to-end.

**One thing I need from you before I merge:** the feature commit's author is set to `AgentSeal <hello@agentseal.org>`, which I think happened because our CLAUDE.md says "Commits from: AgentSeal". That guidance is for our own commits, not contributors. If you can rebase the commit with your own identity as author so you show up on the contributor graph, that would be great. One command on your end:

```
git commit --amend --author="Chris <2333584+cgrossde@users.noreply.github.com>" --no-edit
git push --force-with-lease
```

If that's awkward or you'd rather not, let me know and I'll merge from our branch instead and credit you in the CHANGELOG and merge commit message. Either way works for me.

Thanks again.

iamtoruk:
Landed in https://github.com/getagentseal/codeburn/commit/ac31883 with the conflict resolution and a few small fixups layered on top: prototype-pollution guard in resolveAlias, dedup key uses source.provider so OMP stays disjoint from Pi, combined pi.js imports, trailing newline on pi.ts. Thanks for the OMP provider and the model-alias work. Will credit in CHANGELOG when we cut the next release.

iamtoruk:
Merged via https://github.com/getagentseal/codeburn/commit/ac31883 (GitHub's auto-detect didn't fire because the merge landed via direct push rather than the PR UI).

cgrossde:
Awesome, nice to see this merged/integrated 👍

PR #69, opened 2026-04-17
Remove .claudeignore from README

### LINKED ISSUE #61: .claudeignore is not a claude feature
The optimizer mentions .claudeignore, but that isn't a documented (or, as far as I can tell, undocumented) feature of claude-code.

jakeefr:
Fixes #61. Removed the reference to .claudeignore as it is not an actual Claude Code feature, keeping the docs accurate.


AgentSeal:
Thanks @jakeefr, you were right about #61. Closing this in favor of #68 which handles the src-level removal comprehensively, and I'll roll the README cleanup into the 0.7.2 release patch alongside it. Appreciate the catch.

PR #77, opened 2026-04-18
Fix 10 bugs: model name matching, CSV injection, resource leaks, and more

hobostay:
## Summary

Found and fixed 10 bugs across 8 files during a thorough code audit of the codebase. All 230 existing tests pass.

### Bugs Fixed

**High Impact:**
- **`getShortModelName` wrong display names** (`models.ts`): Prefix matching used unsorted iteration, causing `gpt-4o-mini` to match `gpt-4o` first and display as "GPT-4o" instead of "GPT-4o Mini". Fixed by sorting keys by length descending.
- **`getModelCosts` overly aggressive fuzzy match** (`models.ts`): Bidirectional `startsWith` matching could return wrong pricing for short canonical names. Removed `key.startsWith(canonical)` direction.
- **Negative token counts in Codex** (`codex.ts`): Cumulative delta subtraction could produce negative values, leading to negative costs. Added `Math.max(0, ...)` clamping.
- **CSV injection vulnerability** (`export.ts`): `escCsv` only checked `=+-@` prefix but missed `\t`/`\r` which can also trigger formula execution in Excel/Sheets.

**Medium Impact:**
- **Zero-cost models silently dropped** (`models.ts`): Falsy check `!entry.input_cost_per_token` skipped free models with `0` cost. Changed to explicit `=== undefined` check.
- **Stream resource leak** (`fs-utils.ts`): `readSessionLines` didn't destroy the stream when the generator was abandoned early, leaking file descriptors. Added `finally { stream.destroy() }`.
- **Phantom empty-date bucket** (`day-aggregator.ts`): Turns with empty/missing timestamps created a phantom `""` date bucket in the daily cache. Added guard to skip such turns.
- **Export crash on empty periods** (`export.ts`): `periods[periods.length - 1].projects` threw `TypeError` if periods was empty. Added optional chaining fallback to `[]`.

**Low Impact:**
- **Config crash on malformed input** (`config.ts`): No validation of parsed JSON; a non-string `currency.code` caused confusing crash. Added type validation.
- **`categoryBreakdown` prototype pollution** (`parser.ts`): Used `{}` instead of `Object.create(null)`, inconsistent with all other breakdown maps. Fixed for consistency.
- **Codex model display name ordering** (`codex.ts`): Applied same sort-by-length-descending pattern used by `pi.ts` for correct prefix matching.

## Test plan
- [x] All 230 existing tests pass
- [ ] Verify `gpt-4o-mini` models display correctly in reports
- [ ] Verify CSV exports escape tab/CR-prefixed values
- [ ] Verify Codex sessions with session resets don't produce negative costs
- [ ] Verify daily cache doesn't contain empty-date entries

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -20,7 +20,10 @@ function getConfigPath(): string {
 export async function readConfig(): Promise<CodeburnConfig> {
   try {
     const raw = await readFile(getConfigPath(), 'utf-8')
-    return JSON.parse(raw) as CodeburnConfig
+    const parsed = JSON.parse(raw)
+    if (!parsed || typeof parsed !== 'object') return {}
+    if (parsed.currency && typeof parsed.currency.code !== 'string') delete parsed.currency
+    return parsed as CodeburnConfig
   } catch {
     return {}
   }
--- a/src/day-aggregator.ts
+++ b/src/day-aggregator.ts
@@ -34,12 +34,14 @@ export function aggregateProjectsIntoDays(projects: ProjectSummary[]): DailyEntr
 
   for (const project of projects) {
     for (const session of project.sessions) {
-      const sessionDate = dateKey(session.firstTimestamp)
-      ensure(sessionDate).sessions += 1
+      const sessionDate = dateKey(session.firstTimestamp || '')
+      if (sessionDate) ensure(sessionDate).sessions += 1
 
       for (const turn of session.turns) {
         if (turn.assistantCalls.length === 0) continue
-        const turnDate = dateKey(turn.assistantCalls[0]!.timestamp)
+        const rawTs = turn.assistantCalls[0]!.timestamp
+        if (!rawTs) continue
+        const turnDate = dateKey(rawTs)
         const turnDay = ensure(turnDate)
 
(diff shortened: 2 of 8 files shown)


iamtoruk:
Thanks for the audit, @hobostay. We reviewed all 10 items and cherry-picked the ones that are real, user-facing bugs:

**Merged (with regression tests):**
- Bidirectional fuzzy match in `getModelCosts` -- could return wrong pricing for short model names
- Falsy check in `parseLiteLLMEntry` silently dropping free/zero-cost models
- Stream resource leak in `readSessionLines` when generator is abandoned early
- CSV injection hardening for tab/CR prefixes
- Empty-periods crash guard in `exportCsv`/`exportJson`

**Not merged:**
- `getShortModelName` ordering -- the keys are already in correct order in the object literal, so the described bug does not manifest. Good defensive hardening but not worth the churn right now.
- `categoryBreakdown` prototype pollution -- theoretical, no practical impact
- Config validation, phantom date bucket, Codex negative tokens, duplicate read display -- minor edge cases we'll address separately if needed

Shipped as `a4d261a` on main. Appreciate the thorough look at the codebase.

Issue #93, opened 2026-04-18
Feature request: per-project daily drill-down + custom date range selector


jc370:
Hi team — love what you've built with CodeBurn. Really great visibility into AI coding costs. A few things that would make it even more useful:

### 1. Per-project daily cost breakdown
When you have multiple active projects, it's hard to understand day-by-day spend for a single project. A way to filter/drill into one project and see its daily trend would be really valuable — especially for projects used daily over weeks.

### 2. Custom date range selector
The current period tabs (Today / 7 Days / 30 Days / Month / All) are great but sometimes you want to look at a specific sprint, billing period, or date range. A custom start/end date picker would close that gap.

## Bug Report: Yesterday's data missing from menubar (UTC+ timezones)

Found a timezone bug in `toDateString` in `src/cli.ts`. It uses `.toISOString()` which returns UTC — so for anyone in a UTC+ timezone (UK, Europe, Asia), midnight local time resolves to the previous day in UTC. This shifts `yesterdayStr` back one day, silently dropping the most recent day from the menubar history payload.

**Fix:**
```typescript
// Before
function toDateString(date: Date): string {
  return date.toISOString().slice(0, 10)
}

// After
function toDateString(date: Date): string {
  const y = date.getFullYear()
  const m = String(date.getMonth() + 1).padStart(2, '0')
  const d = String(date.getDate()).padStart(2, '0')
  return `${y}-${m}-${d}`
}
```

Happy to raise a PR for the bug fix if useful. Thanks for building this!

jc370:
Quick clarification on the per-project drill-down request — I'm aware `--project` was shipped in 0.6.1 for the CLI and it works great. What I'd love is the same capability in the **menubar app GUI** — being able to select a project from a dropdown/filter in the popup and see that project's daily trend directly in the menubar without opening a terminal.

The menubar is where I naturally glance for a quick cost check, so having project-level visibility there (rather than switching to a terminal) would be the real win.

lfl1337:
Hey @jc370 — good find, your `toDateString` fix is correct. The history payload does silently drop the most recent day for anyone east of UTC.

Please go ahead and open that PR — the fix is worth landing quickly. Happy to review or help if you hit blockers. I'm not a maintainer myself so I can't merge, but another pair of eyes tends to help things along. Feel free to ping me here, in a draft PR, or on X (linked from my GitHub profile).

[...]

PR #96, opened 2026-04-18
fix(cli): use local time in toDateString to fix UTC+ date shift

jc370:
## Summary

`toDateString` called `toISOString()` which returns UTC. For any timezone east of UTC, midnight local time resolves to the previous calendar day in UTC — so `yesterdayStr` was calculated as one day too early, and the most recent day was silently dropped from the menubar history payload.

- Affects all users in UTC+ timezones (Europe, Asia, Australia, Africa)
- Reported in issue #93

## Fix

Build the `YYYY-MM-DD` string from `getFullYear()` / `getMonth()` / `getDate()` so the date reflects the user's local clock.

```ts
// Before
function toDateString(date: Date): string {
  return date.toISOString().slice(0, 10)
}

// After
function toDateString(date: Date): string {
  const y = date.getFullYear()
  const m = String(date.getMonth() + 1).padStart(2, '0')
  const d = String(date.getDate()).padStart(2, '0')
  return `${y}-${m}-${d}`
}
```

## Test plan
- [x] Verified menubar correctly shows yesterday's data when running in BST (UTC+1)
- [x] Fix confirmed by contributor @lfl1337 in issue #93

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -24,7 +24,10 @@ const MS_PER_DAY = 24 * 60 * 60 * 1000
 const BACKFILL_DAYS = 365
 
 function toDateString(date: Date): string {
-  return date.toISOString().slice(0, 10)
+  const y = date.getFullYear()
+  const m = String(date.getMonth() + 1).padStart(2, '0')
+  const d = String(date.getDate()).padStart(2, '0')
+  return `${y}-${m}-${d}`
 }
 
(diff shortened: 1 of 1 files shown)


iamtoruk:
Thanks for flagging this. The same local-time fix to toDateString is already on main as part of 0.8.5 (`src/cli.ts:30`). Closing since the behaviour you wanted is now shipping. Appreciate the catch.

PR #104, opened 2026-04-19
fix(sqlite): load node:sqlite in ESM runtime

aaronflorey:
## Summary
- replace eval-based SQLite loader with `createRequire(import.meta.url)` in `src/sqlite.ts`
- fix OpenCode/Cursor provider discovery when CLI runs in ESM and `require` is unavailable
- keep change minimal to avoid affecting provider parsing behavior

## Verification
- `npm run dev -- report`
- `npm run dev -- today`
- `npm test -- tests/providers/opencode.test.ts tests/provider-registry.test.ts`

### DIFF
--- a/src/sqlite.ts
+++ b/src/sqlite.ts
@@ -1,7 +1,11 @@
+import { createRequire } from 'node:module'
+
 /// Thin SQLite read-only wrapper over Node's built-in `node:sqlite` module (stable in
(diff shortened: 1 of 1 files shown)


iamtoruk:
Merged, thanks! Clean surgical fix. Verified locally: 238 tests pass, esbuild direct-eval warning gone, isolated ESM repro confirms createRequire works where eval('require') threw ReferenceError.

Issue #31, opened 2026-04-15
Add support for KiloCode and OpenCode


davmrtl:
Supporting KiloCode and OpenCode would be very helpful.

dswbx:
and qwen code would be nice

AgentSeal:
we will take care of this soon :)

AgentSeal:
Opencode support has been added @davmrtl  can you please test and post a screenshot here.

gojoGS:
> Opencode support has been added [@davmrtl](https://github.com/davmrtl) can you please test and post a screenshot here.

Hi, I've tried to use Codeburn with OpenCode, but I've ran into issues. Please check https://github.com/getagentseal/codeburn/pull/105 if you have the time.

iamtoruk:
@dswbx Qwen CLI support has been added! Can you test it out and post a screenshot here? Just run npx codeburn report --provider qwen

iamtoruk:
same to @davmrtl can you please check if kilocode and opencode both works fine in you case with latest codeburn version

PR #105, opened 2026-04-19
fix: restore SQLite-backed providers in ESM builds

gojoGS:
Even though OpenCode support was added in https://github.com/getagentseal/codeburn/issues/31, I wasn't able to use it in 0.7.3.

Codeburn opened, but didn't show any activities. I had used OpenCode extensively for the past 2 months and I had a valid and populated `opencode.db` located in `~/.local/share/opencode/`, so I wasn't sure the issue was on my end.

After investigating, I've found this manifested, because Codeburn failed to import the sqlite module and this failed silently. Thus no usage data was returned to be analyzed. Reworking module import fixed this issue for me.

Operating System: Windows 10
Shell: Powershell 7.6.0
Node version: v25.9.0
OpenCode version: 1.4.11
Codeburn version: 0.7.3

### DIFF
--- a/src/sqlite.ts
+++ b/src/sqlite.ts
@@ -3,6 +3,8 @@
 /// so the dependency graph no longer pulls in the deprecated `prebuild-install` package
 /// (issue #75). Works across Cursor and OpenCode session DBs, both of which we only read.
 
+import { createRequire } from 'node:module'
+
 type Row = Record<string, unknown>
 
 export type SqliteDatabase = {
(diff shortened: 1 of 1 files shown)


iamtoruk:
Thanks. The `createRequire(import.meta.url)` swap landed on main via PR #104 and is in 0.8.5 (`src/sqlite.ts:1-8`), same mechanism you have here. Closing. Appreciate the fix.

PR #111, opened 2026-04-20
fix(cli-date): reject month/day overflow in parseLocalDate

lfl1337:
## Summary

Follow-up to #80, which introduced `parseLocalDate`. The regex-only
validation there accepts structurally valid but semantically invalid
dates like `2026-13-45` or `2026-02-30`. The `Date` constructor
silently rolls them forward into a different valid date, so `--from` /
`--to` then produce a misleading `"must not be after"` error instead
of rejecting the input.

## Fix

After constructing the `Date`, compare the round-tripped
`year / month / day` back to the parsed numbers. If they diverge,
throw a clear `"month or day out of range"` error.

## Before

```
$ codeburn report --from 2026-13-45
Error: --from must not be after --to (got 2026-13-45 > undefined)
```

## After

```
$ codeburn report --from 2026-13-45
Error: Invalid date "2026-13-45": month or day out of range

$ codeburn report --from 2026-02-30
Error: Invalid date "2026-02-30": month or day out of range
```

## Tests

Two new cases in `tests/date-range-filter.test.ts` cover month and day
overflow. Full file 10/10 green.

### DIFF
--- a/src/cli-date.ts
+++ b/src/cli-date.ts
@@ -12,7 +12,11 @@ function parseLocalDate(s: string): Date {
     throw new Error(`Invalid date format "${s}": expected YYYY-MM-DD`)
   }
   const [y, m, d] = s.split('-').map(Number) as [number, number, number]
-  return new Date(y, m - 1, d)
+  const parsed = new Date(y, m - 1, d)
+  if (parsed.getFullYear() !== y || parsed.getMonth() !== m - 1 || parsed.getDate() !== d) {
+    throw new Error(`Invalid date "${s}": month or day out of range`)
+  }
+  return parsed
 }
 
 export function parseDateRangeFlags(from: string | undefined, to: string | undefined): DateRange | null {
--- a/tests/date-range-filter.test.ts
+++ b/tests/date-range-filter.test.ts
@@ -48,6 +48,16 @@ describe('parseDateRangeFlags', () => {
       .toThrow('Invalid date format')
   })
 
+  it('throws on month out of range', () => {
+    expect(() => parseDateRangeFlags('2026-13-01', undefined))
+      .toThrow('month or day out of range')
+  })
+
+  it('throws on day out of range for month', () => {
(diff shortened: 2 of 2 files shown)


Qodo-Free-For-OSS:
Hi, parseLocalDate now rejects inputs where Date round-tripping changes the parsed year, but JavaScript interprets years 0–99 as 1900–1999. Dates like "0099-01-01" will incorrectly throw "month or day out of range" despite matching the YYYY-MM-DD format contract.

**How to fix:** Handle years 0–99 explicitly

> ### Issue description
> `parseLocalDate` round-trip validation breaks for years `0000`–`0099` because `new Date(y, m-1, d)` interprets numeric years 0–99 as 1900–1999.
> 
> ### Issue Context
> The regex allows any 4-digit year, so users can pass `0099-01-01`. With the current check, these will throw `month or day out of range` even when month/day are valid.
> 
> ### Fix Focus Areas
> - src/cli-date.ts[10-20]
> 
> ### Suggested fix approach
> - Construct the date in a way that preserves years 0–99 (e.g., create `new Date(0)` then call `setFullYear(y, m-1, d)` and zero the time), then apply the same round-trip validation.
>   - Example:
>     - `const parsed = new Date(0); parsed.setFullYear(y, m-1, d); parsed.setHours(0,0,0,0);`
> - Alternatively, explicitly disallow years `0000`–`0099` with a clear error message if you don’t want to support them, but then also update the user-facing contract/docs/tests accordingly.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review. FYI, Qodo is free for open-source.*

iamtoruk:
Thanks for this @lfl1337 -- your overflow validation approach was exactly right. This fix has already landed on main (with the same round-trip technique you proposed here) as part of a broader cli-date cleanup. The current implementation also includes more test coverage: Feb 31, Apr 31, leap year validation (2024 vs 2025), and a clearer error message that shows the actual date components.

Closing as superseded. Appreciate the contribution.

PR #113, opened 2026-04-20
Show per-model token counts in TUI

bcchew-art:
## Summary

Surfaces per-model token totals in the `By Model` panel of the interactive dashboard. The token data (input, output, cacheRead, cacheWrite) was already computed in `session.modelBreakdown.tokens` and exposed in `codeburn report --format json`, but not rendered in the TUI. This adds a single compact `tokens` column between `cost` and `cache`.

- `formatTokens` extended to emit a `B` suffix for values at or above 1e9 so per-model cache totals (typically multi-billion) render as e.g. `4.09B` instead of overflowing.
- `ModelBreakdown` now aggregates `outputTokens` in addition to the existing input/cacheRead/cacheWrite, and shows the sum per row.
- Zero-token rows display `-` rather than `0`.

## Before / After

Before (no tokens column):

```
│ By Model                                                 │
│                              cost  cache  calls          │
│ ██████████ Opus 4.6       $879.94  96.1%   4817          │
│ ████░░░░░░ Opus 4.7       $373.89  94.8%   2325          │
```

After:

```
│ By Model                                                 │
│                              cost  tokens  cache  calls  │
│ ██████████ Opus 4.6       $879.94   1.14B  96.1%   4817  │
│ ████░░░░░░ Opus 4.7       $373.89  432.7M  94.8%   2325  │
│ █░░░░░░░░░ Sonnet 4.6      $93.68  156.3M  93.4%   3126  │
│ ░░░░░░░░░░ Haiku 4.5        $4.42   23.4M  92.8%    259  │
│ ░░░░░░░░░░ <synthetic>    $0.0000       -      -      1  │
```

## Test plan

- [x] `npx vitest run` passes (244/244, 6 new tests covering `formatTokens` B suffix and the per-model aggregation)
- [x] `npx tsx src/cli.ts` rendered and visually checked at 80, 120, and 160 column widths
- [x] Period switching (1-5) still re-renders the panel correctly
- [x] Zero-token rows show `-`, not `0.00K`

### DIFF
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -267,19 +267,21 @@ function ProjectBreakdown({ projects, pw, bw, budgets }: { projects: ProjectSumm
 }
 
 const MODEL_COL_COST = 8
+const MODEL_COL_TOKENS = 8
 const MODEL_COL_CACHE = 7
 const MODEL_COL_CALLS = 7
 const MODEL_NAME_WIDTH = 14
 
 function ModelBreakdown({ projects, pw, bw }: { projects: ProjectSummary[]; pw: number; bw: number }) {
-  const modelTotals: Record<string, { calls: number; costUSD: number; freshInput: number; cacheRead: number; cacheWrite: number }> = {}
+  const modelTotals: Record<string, { calls: number; costUSD: number; freshInput: number; output: number; cacheRead: number; cacheWrite: number }> = {}
   for (const project of projects) {
     for (const session of project.sessions) {
       for (const [model, data] of Object.entries(session.modelBreakdown)) {
-        if (!modelTotals[model]) modelTotals[model] = { calls: 0, costUSD: 0, freshInput: 0, cacheRead: 0, cacheWrite: 0 }
+        if (!modelTotals[model]) modelTotals[model] = { calls: 0, costUSD: 0, freshInput: 0, output: 0, cacheRead: 0, cacheWrite: 0 }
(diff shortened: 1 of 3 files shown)


ozymandiashh:
Correction: the earlier wording here was too vague. I rechecked #113 and this is the actual feature proof.

What is proven by the PR itself:
- `src/dashboard.tsx` renders a `tokens` column in the `By Model` panel between `cost` and `cache`.
- `tests/dashboard.test.ts` has a focused aggregation test that sums `inputTokens + outputTokens + cacheReadInputTokens + cacheCreationInputTokens` across sessions/projects for one model and expects the token label to render as `4.7M`.
- The zero-token test asserts that a model with no recorded tokens renders `-`, not a misleading zero value.
- `formatTokens` tests cover the suffixes used by the TUI column: `4.81B`, `168.4M`, `23.5K`, and raw values under 1000.

Commands rerun:
- `npx vitest run tests/dashboard.test.ts` - 11/11 tests passed.
- `npm run build` - passed.

Additional visual proof:
- Rendered the dashboard against local usage data with `COLUMNS=120 npx tsx src/cli.ts report --provider claude -p all`.
- The screenshot below is redacted: project/session names are replaced with `[redacted]`, and the unrelated `<synthetic>` row is removed.
- Aggregate costs/token/call values are intentionally visible here so the `By Model -> tokens` column can be validated against real output.

image

PR #155, opened 2026-04-25
fix(parser): keep last occurrence of streaming message id within file (#110)

### LINKED ISSUE #110: MCP tool calls and tool_use blocks missing from dashboard due to streaming write deduplication
## Bug

MCP server usage never appears in the dashboard breakdown, and other tool invocations
(Agent spawns, EnterPlanMode, etc.) are silently dropped from turns that involve
streaming responses.

## Environment

| | |
|---|---|
| **codeburn** | 0.8.1 |
| **Claude Code** | 2.1.114 |
| **Node.js** | 20.20.0 |
| **OS** | Ubuntu 24.04.2 LTS (Linux 6.17.0-1010-aws x86_64) |
| **Provider** | Claude Code sessions routed via Databricks AI Gateway (Bedrock backend) |

## Root cause

Claude Code writes the same `message.id` to the session JSONL multiple times as a
response streams in:

1. **First write** (`message_start`) — model + ID, empty content, partial usage
2. **Intermediate writes** — mid-stream updates
3. **Final write** (`message_stop`) — complete content including `tool_use` blocks,
   final token counts

`groupIntoTurns` in `src/parser.ts` deduplicates by `message.id` keeping the
**first** occurrence and skipping all subsequent entries. The final entry — the only
one containing `tool_use` blocks — is discarded.

Example from a real session JSONL where `mcp__playwright__browser_navigate` only
appears in the third entry for the same `message.id`:

```json
{"type":"assistant","message":{"id":"msg_01X...","model":"claude-opus-4-6","content":[],...}}
{"type":"assistant","message":{"id":"msg_01X...","model":"claude-opus-4-6","content":[],...}}
{"type":"assistant","message":{"id":"msg_01X...","model":"claude-opus-4-6","content":[{"type":"tool_use","name":"mcp__playwright__browser_navigate",...}],...}}
```

Only the first entry is parsed; the tool call is never counted.

## Impact

- MCP server breakdown panel never populates for affected turns
- `tool_use` blocks in streaming turns are missed (Agent, EnterPlanMode, Bash, etc.)
- Token counts may be slightly understated (final entry has the authoritative usage)

## Fix

Add a within-file pre-pass in `parseSessionFile` that keeps the **last** occurrence
of each `message.id` before handing entries to `groupIntoTurns`. The existing
cross-file `seenMsgIds` dedup in `groupIntoTurns` (keep first-seen) is correct and
should remain unchanged — the two concerns need to be handled separately.

```typescript
// In parseSessionFile, before groupIntoTurns:
const lastIdxById = new Map<string, number>()
for (let i = 0; i < filteredEntries.length; i++) {
  const msgId = getMessageId(filteredEntries[i])
  if (msgId) lastIdxById.set(msgId, i)
}
const dedupedEntries = filteredEntries.filter((entry, i) => {
  const msgId = getMessageId(entry)
  if (!msgId) return true
  return lastIdxById.get(msgId) === i
})
// pass dedupedEntries to groupIntoTurns instead of filteredEntries
```

## Verified

MCP calls (`mcp__playwright__browser_navigate`, `mcp__playwright__browser_snapshot`,
`mcp__github__get_file_contents`) confirmed present in JSONL but absent from dashboard
before fix, correctly attributed after. Total call counts unchanged — no
double-counting introduced. Full test suite: 273 pass, 0 regressions.

---

Please let me know if you want me to raise a PR for this as I am happy to. Awesome tool you have here and we use it every day since we discovered it.

ousamabenyounes:
## Summary

Fixes #110.

Claude Code writes the same `message.id` multiple times to the session JSONL as a response streams in. Only the final write carries the `tool_use` blocks (MCP servers, Agent, EnterPlanMode, …) and the authoritative token counts. The existing dedup in `groupIntoTurns` keeps the **first** occurrence by id, so for every streamed turn:

- MCP tool calls disappear from the breakdown
- `tool_use` blocks for Agent / EnterPlanMode / Bash are dropped
- Token counts are slightly understated

## Fix

Add a within-file pre-pass `dedupeStreamingMessageIds` in `parseSessionFile` that keeps the **last** occurrence of each `message.id`. The cross-file dedup in `groupIntoTurns` (against `seenMsgIds`) is correct for what it does — avoiding double-counting when the same session appears under multiple project dirs — and stays keep-first-seen.

The two concerns are now handled separately:

| Where | Concern | Policy |
|---|---|---|
| `parseSessionFile` (new) | Streaming writes within one file | keep last |
| `groupIntoTurns` (unchanged) | Same id seen across multiple files | keep first |

## Verification

- [x] Baseline: 355 pass, 0 fail
- [x] Post-fix: 359 pass, 0 fail (4 new tests, 0 regressions)
- [x] New test `tests/parser-streaming-dedup.test.ts` fails on the unfixed code (verified by stashing the fix and re-running)
- [x] `npx tsc --noEmit` clean
- [x] No bracket-assign on `{}`-init maps in `src/parser.ts` or `src/providers/` (semgrep guard)

## Files changed

| File | Change |
|---|---|
| `src/parser.ts` | new exported `dedupeStreamingMessageIds`; called from `parseSessionFile` before `groupIntoTurns` |
| `tests/parser-streaming-dedup.test.ts` | 4 new tests |

---
_Vibe Coded by Ousama Ben Younes_
_Developed With Ora Studio (Claude Code)_

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -119,6 +119,28 @@ function parseApiCall(entry: JournalEntry): ParsedApiCall | null {
   }
 }
 
+/// Streaming responses write the same `message.id` to the JSONL multiple times: an early
+/// `message_start` with empty content, optional intermediate updates, and a final
+/// `message_stop` carrying the full `tool_use` blocks plus authoritative usage. Within a
+/// single file we keep only the LAST occurrence of each id so `tool_use` blocks (MCP, Agent,
+/// EnterPlanMode, …) and final token counts reach `groupIntoTurns`. Cross-file dedup runs
+/// downstream against `seenMsgIds` and remains keep-first-seen.
+export function dedupeStreamingMessageIds(entries: JournalEntry[]): JournalEntry[] {
+  const lastIdxById = new Map<string, number>()
+  for (let i = 0; i < entries.length; i++) {
+    const id = getMessageId(entries[i]!)
+    if (id) lastIdxById.set(id, i)
+  }
+  if (lastIdxById.size === 0) return entries
+  const result: JournalEntry[] = []
+  for (let i = 0; i < entries.length; i++) {
+    const id = getMessageId(entries[i]!)
+    if (id && lastIdxById.get(id) !== i) continue
+    result.push(entries[i]!)
+  }
+  return result
+}
+
 function groupIntoTurns(entries: JournalEntry[], seenMsgIds: Set<string>): ParsedTurn[] {
   const turns: ParsedTurn[] = []
   let currentUserMessage = ''
@@ -289,7 +311,8 @@ async function parseSessionFile(
   if (entries.length === 0) return null
 
   const sessionId = basename(filePath, '.jsonl')
-  let turns = groupIntoTurns(entries, seenMsgIds)
+  const dedupedEntries = dedupeStreamingMessageIds(entries)
+  let turns = groupIntoTurns(dedupedEntries, seenMsgIds)
   if (dateRange) {
     // Bucket a turn by the timestamp of its first assistant call (when the cost was
     // actually incurred). Filtering entries directly produced orphan assistant calls
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -123,19 +123,29 @@ function parseApiCall(entry: JournalEntry): ParsedApiCall | null {
 /// `message_start` with empty content, optional intermediate updates, and a final
 /// `message_stop` carrying the full `tool_use` blocks plus authoritative usage. Within a
 /// single file we keep only the LAST occurrence of each id so `tool_use` blocks (MCP, Agent,
-/// EnterPlanMode, …) and final token counts reach `groupIntoTurns`. Cross-file dedup runs
-/// downstream against `seenMsgIds` and remains keep-first-seen.
+/// EnterPlanMode, …) and final token counts reach `groupIntoTurns`, but we preserve the
+/// FIRST occurrence's timestamp on that kept entry so downstream date-range bucketing /
+/// filtering still treats the assistant call as occurring when it began. Cross-file dedup
+/// runs downstream against `seenMsgIds` and remains keep-first-seen.
 export function dedupeStreamingMessageIds(entries: JournalEntry[]): JournalEntry[] {
+  const firstIdxById = new Map<string, number>()
   const lastIdxById = new Map<string, number>()
   for (let i = 0; i < entries.length; i++) {
     const id = getMessageId(entries[i]!)
-    if (id) lastIdxById.set(id, i)
+    if (!id) continue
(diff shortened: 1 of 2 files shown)


Qodo-Free-For-OSS:
Hi, dedupeStreamingMessageIds keeps only the last occurrence of each message.id, which makes parseApiCall/groupIntoTurns use the later streaming update’s timestamp (typically message_stop) as the call timestamp. This can incorrectly include/exclude turns in dateRange filtering and shift day bucketing near midnight because the code explicitly uses the first assistant call timestamp as the cost-incurrence time.

**How to fix:** Preserve first timestamp per id

> ### Issue description
> `dedupeStreamingMessageIds()` keeps the *last* entry for a given `message.id`, which changes the assistant call timestamp downstream to the later `message_stop` timestamp. This conflicts with `parseSessionFile` date bucketing/filtering logic that uses the first assistant call timestamp as the cost-incurrence timestamp.
> 
> ### Issue Context
> The goal (keeping final `tool_use` blocks and authoritative usage) is correct, but timestamp semantics matter for `dateRange` filtering and day bucketing.
> 
> ### Fix Focus Areas
> - src/parser.ts[122-142]
> - src/parser.ts[285-330]
> 
> ### Suggested fix
> Update `dedupeStreamingMessageIds` to keep the last entry’s content/usage, but preserve the earliest timestamp seen for that `message.id`:
> - First pass: track both `lastIdxById` and `firstTimestampById` (or earliest timestamp string) for each id.
> - Second pass: when you are about to push the kept (last) entry for an id, if `firstTimestampById` exists and differs from `entry.timestamp`, push a shallow clone `{ ...entry, timestamp: firstTimestamp }` (avoid mutating input in-place).
> - Update/add a unit test to assert that the kept entry retains the first timestamp for bucketing, while still retaining the final `tool_use`/usage from the last update.

---

ousamabenyounes:
Good catch on the timestamp semantics.

Pushed [`88baed8`](https://github.com/getagentseal/codeburn/pull/155/commits/88baed8). `dedupeStreamingMessageIds` now keeps the **last** entry's content (tool_use blocks + authoritative usage) but overrides its `timestamp` with the **first** entry's value, so downstream date-range bucketing / filtering in `parseSessionFile` continues to use the cost-incurrence time when the call started.

Tests updated to assert both invariants; added a non-mutation test on the input array.

Full suite: 360 / 360. `npx tsc --noEmit` clean.

iamtoruk:
Closing this. The streaming dedup fix itself is solid, but the PR also silently removes flushCodexCache() and flushAntigravityCache() calls from parseProviderSources, which would break cache persistence for codex and antigravity users (every run re-parses everything from scratch). These removals aren't mentioned in the PR description. We'll apply the dedup fix separately. Thanks for identifying the streaming issue though.

PR #160, opened 2026-04-27
fix(tests): use local-time timestamps in day-aggregator test

gcacace:
The test used 2026-04-09T23:59:00Z (UTC) which falls on April 10th in timezones ahead of UTC, causing dateKey() to return the wrong date.

Drop the Z suffix so timestamps are parsed as local time, matching dateKey() which uses getDate() (local). The original midnight-boundary scenario is preserved.


gcacace:
Closing as duplicate of #112.

PR #180, opened 2026-04-30
VSCode: correctly locate workspaceStorage when working inside a dev container

josteinaj:
In a dev container, the path used is:

```
/home/vscode/.vscode-server/data/User/workspaceStorage
```

In other words, the vscode configuration is stored under `~/.vscode-server/data` instead of `~/.config/Code`.

### DIFF
--- a/src/providers/copilot.ts
+++ b/src/providers/copilot.ts
@@ -318,6 +318,7 @@ function getVSCodeWorkspaceStorageDirs(): string[] {
   return [
     join(homedir(), '.config', 'Code', 'User', 'workspaceStorage'),
     join(homedir(), '.config', 'Code - Insiders', 'User', 'workspaceStorage'),
+    join(homedir(), '.vscode-server', 'data', 'User', 'workspaceStorage'),
   ]
 }
 


iamtoruk:
Thanks for this! Dev container support is a good catch.

One thing: PR #181 (just merged) refactored this function from returning a single path to returning an array of paths. Your PR will have a merge conflict against current main.

The good news is the fix becomes even simpler with the new approach. Instead of checking env vars, you can just add the path to the Linux array since `discoverVSCodeTranscripts` already handles non-existent directories gracefully (returns `[]`):

```ts
return [
  join(homedir(), '.config', 'Code', 'User', 'workspaceStorage'),
  join(homedir(), '.config', 'Code - Insiders', 'User', 'workspaceStorage'),
  join(homedir(), '.vscode-server', 'data', 'User', 'workspaceStorage'),
]
```

Could you rebase on main and update the diff? No env var check needed.

josteinaj:
Just my luck :sweat_smile:. I rebased and updated now. Indeed the solution is even simpler and more robust now :+1:.

iamtoruk:
Merged, thanks for the fix and for rebasing!

PR #189, opened 2026-05-01
fix(codex): read complete session_meta line during discovery

ozymandiashh:
## Problem

Codex sessions created with **Codex CLI 0.128+** whose first `session_meta` line exceeds 16 KB are silently excluded from totals. Symptom: the Codex tab in the menubar (and `codeburn report --provider codex`) shows €0 / 0 calls for "today" while an active Codex session has been running for hours.

## Root cause

`readFirstLine` (in `src/providers/codex.ts`) allocates a fixed 16 KB buffer:

```ts
const buf = Buffer.alloc(16384)
const { bytesRead } = await fh.read(buf, 0, 16384, 0)
...
const nl = text.indexOf('\n')
const line = nl >= 0 ? text.slice(0, nl) : text
```

Codex CLI 0.128+ writes the full `base_instructions` / system prompt inside the `session_meta` line, which now runs **22–27 KB** in practice. The 16 KB read never reaches a newline, the partial buffer fails `JSON.parse`, `readFirstLine` returns `null`, and `isValidCodexSession` rejects the file in `discoverSessionsInDir`. Affected sessions never reach the parser. (Cached sessions bypass discovery, which is why historical data stays intact.)

Concrete reproduction on my machine (Codex CLI 0.128.0):

```
sessions/2026/04/28/  → 30 .jsonl files
                       28 with first-line 22-27 KB  → silently dropped
                       2  with first-line ~7 KB     → parsed normally
~/.cache/codeburn/codex-results.json contains exactly those 2.
```

After this patch on the same data:
- 30-day Codex total: **€267 → €878** (the missing half was real spend)
- Today: **€0 / 0 calls → €17 / 188 calls** (the active 14 MB session is now counted)

## Fix

Stream the file via `readline` so the first line is read up to a 1 MiB cap, comfortably handling the current 22–27 KB Codex `session_meta` lines while still bounding memory if a malformed file ever has no newline:

```ts
const FIRST_LINE_READ_CAP = 1024 * 1024

const stream = createReadStream(filePath, {
  encoding: 'utf-8',
  start: 0,
  end: FIRST_LINE_READ_CAP - 1,
})
const rl = createInterface({ input: stream, crlfDelay: Infinity })
for await (const line of rl) { firstLine = line; break }
```

This avoids hardcoding any new buffer size that would break again the next time Codex grows its system prompt. `readline`'s async iterator re-throws underlying stream errors on Node 16+, so a missing/unreadable file falls through the existing `catch` block.

## Tests

Added in `tests/providers/codex.test.ts`:

- `accepts session_meta lines larger than 16 KB (Codex CLI 0.128+)` — 40 KB `base_instructions.text`; asserts both discovery and that `project` was correctly extracted from the long line. Fails on `main`.
- `handles a session_meta line that spans multiple stream chunks` — 120 KB line forces readline to assemble the line across `createReadStream`'s default 64 KiB highWaterMark.
- `handles a session_meta line without trailing newline` — single-line rollout file with no `\n` at EOF is still accepted.
- `rejects truncated/torn first-line writes without throwing` — partial mid-write payload is silently skipped.
- `returns no sessions for an empty rollout file` — explicit guard.

Full suite (`npx vitest run`) passes (419/419).

## Risk

- `readline` is already used elsewhere in the codebase (`src/fs-utils.ts` `readViaStream`).
- No change to parser semantics, model resolution, or cost calculation.
- Read window capped at 1 MiB per discovery validation; real Codex `session_meta` lines are 22–27 KB, ~40× headroom.

### DIFF
--- a/src/providers/codex.ts
+++ b/src/providers/codex.ts
@@ -1,4 +1,6 @@
-import { readdir, stat, open } from 'fs/promises'
+import { readdir, stat } from 'fs/promises'
+import { createReadStream } from 'fs'
+import { createInterface } from 'readline'
 import { basename, join } from 'path'
 import { homedir } from 'os'
 
@@ -70,21 +72,29 @@ function sanitizeProject(cwd: string): string {
 }
 
 async function readFirstLine(filePath: string): Promise<CodexEntry | null> {
-  let fh
+  // Codex CLI 0.128+ writes a session_meta line that can exceed 20 KB because
+  // it embeds the full base_instructions / system prompt. A fixed-size buffer
+  // would miss the trailing newline and reject the session as invalid.
+  // Stream the file via readline to read the first line regardless of length.
+  const stream = createReadStream(filePath, { encoding: 'utf-8' })
+  const rl = createInterface({ input: stream, crlfDelay: Infinity })
+  let firstLine: string | undefined
   try {
-    fh = await open(filePath, 'r')
-    const buf = Buffer.alloc(16384)
-    const { bytesRead } = await fh.read(buf, 0, 16384, 0)
-    if (bytesRead === 0) return null
-    const text = buf.toString('utf-8', 0, bytesRead)
-    const nl = text.indexOf('\n')
-    const line = nl >= 0 ? text.slice(0, nl) : text
-    if (!line.trim()) return null
-    return JSON.parse(line) as CodexEntry
+    for await (const line of rl) {
+      firstLine = line
+      break
+    }
   } catch {
     return null
   } finally {
-    await fh?.close()
+    rl.close()
+    stream.destroy()
+  }
+  if (!firstLine || !firstLine.trim()) return null
+  try {
+    return JSON.parse(firstLine) as CodexEntry
+  } catch {
+    return null
   }
 }
 
--- a/src/providers/codex.ts
+++ b/src/providers/codex.ts
@@ -80,19 +80,21 @@ async function readFirstLine(filePath: string): Promise<CodexEntry | null> {
   // Codex CLI 0.128+ writes a session_meta line that can exceed 20 KB because
   // it embeds the full base_instructions / system prompt. A fixed-size buffer
   // would miss the trailing newline and reject the session as invalid.
-  // Stream the file via readline so we can read the first line regardless of
-  // length, with `end` capping the read to keep memory bounded.
+  // Stream the file via readline so we can read the first line up to
+  // FIRST_LINE_READ_CAP, which keeps memory bounded if the file has no newline.
   const stream = createReadStream(filePath, {
     encoding: 'utf-8',
     start: 0,
     end: FIRST_LINE_READ_CAP - 1,
   })
+  // Silence stream errors so a late read-ahead error after we've already
+  // returned the first line cannot escape as an unhandled 'error' event.
+  // readline's async iterator re-throws underlying stream errors (ENOENT,
+  // EACCES, etc.) on Node 16+, which the catch below handles for the cases
+  // that matter for validation.
+  stream.on('error', () => {})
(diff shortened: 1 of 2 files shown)


iamtoruk:
Thanks for the contribution! Good catch on the first-line reading inefficiency.

However, the base code has changed since this PR was opened. The `readFirstLine` function in `src/providers/codex.ts` on `main` now uses `readSessionFile` (which reads the entire file via `fs/promises.readFile`) instead of the old 16KB `fs.read` buffer approach. So the truncation bug you identified is already fixed on main — just not in the most efficient way.

That said, your streaming `readline` approach is a better optimization than reading the entire file when we only need the first line. For large session files this would be a meaningful improvement.

If you'd like to continue with this, could you rebase on current `main`? The diff will look different since the code has changed, but the core idea (stream-read only the first line instead of slurping the whole file) is still valuable.

ozymandiashh:
Thanks for the quick read.

Quick double-check on `main` at `033da41`: `readFirstLine` in `src/providers/codex.ts:72-89` still uses the 16 KB buffer (`Buffer.alloc(16384)` + `fh.read(buf, 0, 16384, 0)`). [permalink](https://github.com/getagentseal/codeburn/blob/033da415c5892f85e2d68bc7fcc964531543e9fd/src/providers/codex.ts#L72-L89). `readSessionFile` is only used later in `createParser` once a session is already discovered — it doesn't run during validation. So the truncation bug still hits `discoverSessionsInDir` for any rollout whose first line exceeds 16 KB; on my machine that's every Codex CLI 0.128+ session.

You may be thinking of a different branch you have locally? Happy to rebase on whatever you point me at.

Either way the PR is currently based on `033da41`, so the diff is straight against current `main`. The streaming `readline` approach is a strict improvement either way: it's bounded to the 1 MiB cap, never slurps the full file (which can be tens of MB for a long-running session), and the new tests catch the regression class.

Let me know how you'd like to proceed.

iamtoruk:
You're right, apologies for the confusion. I was looking at local unpushed changes where `readFirstLine` was switched to `readSessionFile` — that hasn't landed on `main` yet. The 16 KB truncation bug is still present on the published `main` at `033da41`.

Your PR targets the correct code. Going to review the diff properly now.

iamtoruk (approved):
Reviewed properly this time. You're correct — the 16 KB truncation bug is live on main and this targets the right code.

The streaming readline approach is clean: bounded by the 1 MB cap, handles no-newline and multi-chunk cases, and the test coverage is thorough (>16 KB, >64 KB multi-chunk, truncated writes, empty files).

One heads-up: there are local changes to codex.ts that haven't been pushed yet (removing the caching layer, simplifying model names). I'll handle the merge conflicts on my side after this lands.

Merging — thanks for the fix.

ozymandiashh:
Thanks for the quick review and merge — appreciate the back-and-forth on the diagnosis. Glad the streaming approach made it in.

Really enjoy using codeburn day-to-day, so if you don't mind I'll keep an eye on the issue tracker and chip in where I can.

iamtoruk:
If you are a windows user please check out to  feat/windows-menubar-tauri and let me know if the windows menubar works... I am not able to cleanly validate everything on VM 

cd desktop
npm install
npm run tauri dev

ozymandiashh:
I'm on macOS unfortunately, can't give you a clean Windows validation. Happy to help with macOS-side issues or anything cross-platform if useful.

iamtoruk:
This is a valid fix. The 16 KB buffer silently drops any session where the first line exceeds that size, which is the common case on Codex CLI 0.128+.

Streaming via readline is the right approach - no arbitrary buffer size to break again later. Tests pass locally (417/417).

Will merge after rebase on current main.

PR #198, opened 2026-05-02
docs: add deno dx as a run directly example

Ardakilic:
Hello,

This PR introduces Deno DX running method that's got introduced in Deno v2.6

https://deno.com/blog/v2.6

I first tried brew tap method, but it installed a lot of dependencies, then I tried the npx method, but with deno.

Can confirm it works nicely, so I simply wanted to add a tiny note about this on README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ Or run directly without installing:
 ```bash
 npx codeburn
 bunx codeburn
+dx codeburn
 ```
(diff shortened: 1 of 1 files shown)


iamtoruk:
Thanks for the suggestion. We'll add the deno dx example ourselves. Closing this one out.

PR #200, opened 2026-05-03
Add Hermes Agent provider (hermes.ts)

Jasunltd:
Reads session data from ~/.hermes/sessions/session_*.json files. Hermes Agent (https://github.com/nousresearch/hermes-agent) stores complete session transcripts as JSON with messages, tool calls, and model metadata at the session level.

Key implementation details:
- Parses Hermes session JSON (role/user/assistant/tool messages)
- Extracts model from session-level metadata
- Maps Hermes tool names to CodeBurn standard tool set (terminal→Bash, read_file→Read, search_files→Grep, etc.)
- Extracts bash commands from terminal tool call arguments
- Estimates token counts from content length (same approach as Kiro/Copilot providers — Hermes doesn't store explicit counts)
- Classifies tools for the 13-category task classifier
- Deduplicates by session_id + message index
- Priced via LiteLLM cost database

Built with AI assistance (Claude Code / Hermes Agent). See: https://github.com/nousresearch/hermes-agent

### DIFF
--- a/src/providers/hermes.ts
+++ b/src/providers/hermes.ts
@@ -1,0 +1,361 @@
+import { readdir, stat } from 'fs/promises'
+import { basename, join } from 'path'
+import { homedir } from 'os'
+
+import { readSessionFile } from '../fs-utils.js'
+import { calculateCost } from '../models.js'
+import { extractBashCommands } from '../bash-utils.js'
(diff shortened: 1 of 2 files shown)


iamtoruk:
Thanks for putting this together. We already have Hermes support in our pipeline and it'll be available soon. Closing this one to avoid duplicate work, but appreciate the contribution.

PR #220, opened 2026-05-04
fix: update model aliases and enhance event parsing for Copilot provider

### LINKED ISSUE #219: Copilot: Models like claude-sonnet-4.6 are classified as Copilot (auto)
I have been using claude-sonnet-4.6  but codeburn shown Copilot (auto)
Image

NihalJain:
- Fixes #219


iamtoruk:
Thanks for the PR @NihalJain, good find on the model classification issue. Reviewed the changes in detail:

**Model aliases in `models.ts`** -- These are correct and needed. `claude-sonnet-4.6` (dot-version, tier-first) is a different format from what `getCanonicalName` produces, so explicit aliases are the right approach.

**Catch-all type on `LegacyCopilotEvent`** -- Good, consistent with the transcript format which already has this.

**`toolu_` prefix in tool call hints** -- Correct and properly ordered after the more specific `toolu_bdrk_` and `toolu_vrtx_` prefixes, so specificity is preserved.

**Two things I'd like adjusted before merging:**

1. **The `data.model` extraction in `parseLegacyEvents` (lines 89-92 of the diff)** -- Casting `event.data` broadly and extracting `model` from ANY event type is fragile. If a future event type has a `model` field meaning something different (e.g. a "3D model" reference), it would silently corrupt `currentModel`. Safer to scope this to specific event types, or at minimum only apply it when `currentModel` is still empty (i.e. as a fallback, not an override).

2. **Fallback change from `'copilot-auto'` to `COPILOT_OPENAI_AUTO`** -- This changes cost attribution from Claude pricing ($3/$15 per MTok) to OpenAI GPT pricing when the model can't be inferred. This is a meaningful behavioral change. `copilot-auto` was intentionally a neutral catch-all. If a user is on Anthropic models with no tool calls (so no ID prefixes to infer from), their costs would now be calculated at OpenAI rates. Can you share the reasoning here? If the intent is "Copilot defaults to OpenAI", that's reasonable for the fallback in `inferModelFromToolCallIds`, but should be paired with a similar default in `parseLegacyEvents` for consistency.

The `data.model` extraction in `inferModelFromToolCallIds` with weight 100 is fine -- explicit model IDs should outweigh heuristic inference.

iamtoruk (commented):
Retracting my earlier comment -- I jumped to review without checking with you first. Please disregard the previous feedback; we'll follow up separately once we've aligned internally on what we want from this PR.

Qodo-Free-For-OSS:
Hi, Adding a catch-all `LegacyCopilotEvent` variant (`type: string`) prevents TypeScript from narrowing on `event.type`, so accesses like `event.data.content` / destructuring `outputTokens` are no longer type-safe and can fail compilation or yield NaN costs at runtime when events don’t match the expected shape.

**How to fix:** Restore discriminated union narrowing

> ### Issue description
> `LegacyCopilotEvent` adds a catch-all variant with `type: string`, which breaks TypeScript’s discriminated-union narrowing for the known event literals. This makes `parseLegacyEvents` unsafe (and likely fails `tsc` under `strict`) and also permits malformed events to flow through and produce NaN costs.
> 
> ### Issue Context
> The parser relies on `event.type` checks to safely read `event.data.content` and `event.data.outputTokens`.
> 
> ### Fix Focus Areas
> - src/providers/copilot.ts[69-143]
> 
> ### Implementation notes
> Pick one:
> 1) Change the catch-all to exclude known literals, preserving narrowing:
> ```ts
> type UnknownLegacyEvent = {
>   type: Exclude<string, 'session.model_change' | 'user.message' | 'assistant.message'>
>   timestamp?: string
>   data: Record<string, unknown>
> }
> ```
> 2) Keep the catch-all but add explicit per-branch casts + runtime guards:
> - For `user.message`: verify `typeof data.content === 'string'`.
> - For `assistant.message`: verify `typeof data.messageId === 'string'` and `typeof data.outputTokens === 'number'` before using.
> Also ensure `calculateCost` is never called with non-finite token counts.

We noticed a couple of other issues in this PR as well - happy to share if helpful.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source). Free code review for open-source maintainers.*

NihalJain:
> Hi, Adding a catch-all `LegacyCopilotEvent` variant (`type: string`) prevents TypeScript from narrowing on `event.type`, so accesses like `event.data.content` / destructuring `outputTokens` are no longer type-safe and can fail compilation or yield NaN costs at runtime when events don’t match the expected shape.
> 
> **Severity:** action required | **Category:** correctness
> 
> **How to fix:** Restore discriminated union narrowing
> 
> **Agent prompt to fix** - you can give this to your LLM of choice:
> 
> > ### Issue description
> > `LegacyCopilotEvent` adds a catch-all variant with `type: string`, which breaks TypeScript’s discriminated-union narrowing for the known event literals. This makes `parseLegacyEvents` unsafe (and likely fails `tsc` under `strict`) and also permits malformed events to flow through and produce NaN costs.
> > ### Issue Context
> > The parser relies on `event.type` checks to safely read `event.data.content` and `event.data.outputTokens`.
> > ### Fix Focus Areas
> > 
> > * src/providers/copilot.ts[69-143]
> > 
> > ### Implementation notes
> > Pick one:
> > 
> > 1. Change the catch-all to exclude known literals, preserving narrowing:
> > 
> > ```ts
> > type UnknownLegacyEvent = {
> >   type: Exclude<string, 'session.model_change' | 'user.message' | 'assistant.message'>
> >   timestamp?: string
> >   data: Record<string, unknown>
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 2. Keep the catch-all but add explicit per-branch casts + runtime guards:
> > 
> > 
> > * For `user.message`: verify `typeof data.content === 'string'`.
> > * For `assistant.message`: verify `typeof data.messageId === 'string'` and `typeof data.outputTokens === 'number'` before using.
> >   Also ensure `calculateCost` is never called with non-finite token counts.
> 
> We noticed a couple of other issues in this PR as well - happy to share if helpful.
> 
> _Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source). Free code review for open-source maintainers._

Sure let me check on this

NihalJain:
Ah thanks for merging.

PR #226, opened 2026-05-05
feat(report): add per-model efficiency metrics

ozymandiashh:
## Summary
- add shared model efficiency aggregation for edit turns, one-shot rate, retry rate, and cost per edit
- surface per-model efficiency in the TUI By Model panel, JSON report output, and export models table
- add focused tests for model efficiency aggregation and CSV export fields

Part of #12, item 4.

## Validation
- `npx tsc --noEmit`
- `npx vitest run tests/model-efficiency.test.ts tests/export.test.ts tests/dashboard.test.ts`
- `npx vitest run`
- `npm run build`

## Pre-commit review
- Argus-style reviewer: FAIL on retry-rate units, fixed, then PASS
- Gemini 3.1 Pro Preview: PASS after retry-rate fix

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -10,6 +10,7 @@ import { buildMenubarPayload } from './menubar-json.js'
 import { getDaysInRange, ensureCacheHydrated, emptyCache, BACKFILL_DAYS, toDateString } from './daily-cache.js'
 import { aggregateProjectsIntoDays, buildPeriodDataFromDays, dateKey } from './day-aggregator.js'
 import { CATEGORY_LABELS, type DateRange, type ProjectSummary, type TaskCategory } from './types.js'
+import { aggregateModelEfficiency } from './model-efficiency.js'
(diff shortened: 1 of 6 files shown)


iamtoruk:
Superseded by #244, which carries your original commit forward and adds review fixes on top:

- rename `retryRate` → `retriesPerEdit` (CSV header `Retries/Edit`) — units were ambiguous (count vs %)
- TUI column header `1shot` → `1-shot` to match the existing By Activity panel
- min-sample guard in the TUI (suppresses one-shot rate when editTurns < 5) — needed because real-data N=1 rendered `100%` next to N=296 `76.7%`
- tighten `<synthetic>` handling: only skip a turn if *every* call is synthetic; attribute to the first non-synthetic call's model
- tests for multi-model and all-synthetic turns

Thanks for the patch — closing this in favor of #244 to keep the merge path clean.

PR #228, opened 2026-05-05
fix(parser): use Claude cwd for Windows project paths

### LINKED ISSUE #217: Windows: same project shown under 3+ paths (case + slash variants), splitting cost/session counts
## Summary
On Windows, a single Claude Code project shows up as multiple separate "projects" in CodeBurn output because case differences and slash conventions are not normalised before grouping. This splits cost, call count, and session count across what should be one bucket.

## Environment
- Platform: Windows 11 Pro 26200
- Node: v24.8.0
- CodeBurn: 0.9.6 (`npm i -g codeburn`)
- Shell: Git Bash (MSYS2) under VS Code

## Repro
Run `codeburn export --format json` against a session corpus that contains Claude Code projects logged under `~/.claude/projects/` from a Windows host. The same project surfaces with three distinct keys.

## Actual output (from my 30-day export)
A single repository at `C:\AI_LAB\OPENCLAW` produces three project rows:

```
323  "Project": "c//AI/LAB/OPENCLAW"
  6  "Project": "c:\AI_LAB\OPENCLAW"
  4  "Project": "C:\AI_LAB\OPENCLAW"
```

The same triple-split pattern appears for every project on the box (FLIPER, HEX, OKNO, JOB_HUNT, GeminiCLI/VSCodeExtension, …). Note also that `_` is being collapsed to `/` in the slashed variant: `AI_LAB` becomes `AI/LAB`, and `AI_Experiments` becomes `AI/Experiments`.

## Expected
All three should normalise to a single canonical project, e.g. `C:\AI_LAB\OPENCLAW` (or a forward-slash form, but consistent and case-folded for the drive letter).

## Suggested fix
When ingesting a Claude Code session path on Windows:
1. Lowercase the drive letter (`C:\` and `c:\` → `c:\`).
2. Pick one separator convention and stick to it — don't both keep `\` and rewrite as `/` depending on which session jsonl wrote the path.
3. Don't replace `_` with `/`. The `~/.claude/projects/` directory naming uses literal underscores and they're significant.
4. Group/normalise keys before the per-project rollup, so cost/calls/sessions are summed not split.

## Related
- Antigravity sessions surface as opaque ID-paths (`C//Users/sam/AppData/Roaming/Claude/local/agent/mode/sessions/378983c5/.../outputs`) rather than a clean "Antigravity" label. Probably a related path-resolution gap rather than an explicit normalisation bug, but worth bundling.

Happy to test a patch on the same dataset if useful.

ozymandiashh:
## Summary

Claude Code stores project sessions under sanitized directory slugs. On Windows that slug is lossy: a real path like `C:\AI_LAB\OPENCLAW` can appear as a directory such as `c--AI-LAB-OPENCLAW`. CodeBurn was trying to reconstruct the project path from that slug by replacing every `-` with `/`, which turns the example into `c//AI/LAB/OPENCLAW` and can split one real project into multiple rows when casing or slash style differs.

This PR makes the Claude parser use the canonical `cwd` already present inside the JSONL session data. That preserves the real Windows path, keeps underscores and drive letters intact, and groups case/slash variants of the same cwd into one project rollup.

## What changed

- Extract the first valid `cwd` from each Claude JSONL session and carry it alongside the parsed session summary.
- Use that `cwd` as `projectPath` for Claude project rollups instead of reconstructing paths from directory slugs.
- Group Claude sessions by a normalized path key so `C:\AI_LAB\OPENCLAW`, `c:\AI_LAB\OPENCLAW`, and `c:/AI_LAB/OPENCLAW/` collapse into one project.
- Preserve the existing slug-derived fallback for old or malformed sessions that do not contain `cwd`.
- Add regression coverage for the Windows path bug and for grouped case/slash variants.
- Add an Unreleased changelog note.

Closes #217.

## Why this is safe

The change is limited to Claude project rollup grouping. It does not change token parsing, pricing, date filtering, classification, or non-Claude providers. Existing sessions without `cwd` keep the old behavior.

## Validation

- `npx tsc --noEmit`
- `npx vitest run tests/parser-claude-cwd.test.ts tests/security/prototype-pollution.test.ts tests/parser-filter.test.ts tests/date-range-filter.test.ts`
- `npx vitest run`
- `npm run build`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@
 - **Session cost outlier detector.** New `optimize` finding flags sessions costing more than 2x their peer-session average within the same project. Ignores sub-$1 outliers to avoid noise. Requires at least 3 sessions per project for a baseline.
 
 ### Fixed (CLI)
+- **Windows Claude project paths.** Claude Code project rollups now prefer
+  the canonical `cwd` stored in session JSONL files instead of reconstructing
+  paths from lossy directory slugs, and group case/slash variants together.
+  Closes #217.
 - **`all` period semantics unified between CLI and dashboard.** The dashboard treated `--period all` as all-time (epoch start) while the CLI bounded it to the last 6 months. Both now consistently mean "Last 6 months". Period helpers (`Period`, `PERIODS`, `PERIOD_LABELS`, `toPeriod`, `getDateRange`) consolidated into `cli-date.ts`. Use `--from` / `--to` for unbounded historical ranges.
 
 ### Fixed (macOS menubar)
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -26,6 +26,11 @@ function unsanitizePath(dirName: string): string {
   return dirName.replace(/-/g, '/')
 }
 
+function normalizeProjectPathKey(projectPath: string): string {
+  const normalized = projectPath.trim().replace(/\\/g, '/')
+  return (normalized.replace(/\/+$/, '') || normalized).toLowerCase()
+}
+
 function parseJsonlLine(line: string): JournalEntry | null {
   try {
     return JSON.parse(line) as JournalEntry
@@ -246,6 +251,15 @@ export function extractMcpInventory(entries: JournalEntry[]): string[] {
   return Array.from(inventory).sort()
 }
 
+function extractCanonicalCwd(entries: JournalEntry[]): string | undefined {
+  for (const entry of entries) {
+    if (typeof entry.cwd !== 'string') continue
+    const cwd = entry.cwd.trim()
+    if (cwd) return cwd
+  }
+  return undefined
+}
+
 function buildSessionSummary(
   sessionId: string,
   project: string,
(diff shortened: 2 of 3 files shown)


iamtoruk:
Superseded by #245, which carries your original commit forward and adds a review fix on top:

**Bug found**: when a slug contained both cwd-bearing and no-cwd sessions, your scanProjectDirs created two ProjectSummary entries sharing the same dirName (one cwd-keyed, one slug-keyed). They later merged in `parseAllSessions` by `p.project`, and whichever was created first set `projectPath`. An alphabetically-early no-cwd session would silently win, leaving the lossy slug-derived path even though a clean canonical cwd was available on later sessions in the same directory.

**Fix**: added a consolidation step at the end of `scanProjectDirs` that folds slug-keyed entries into the cwd-keyed entry sharing the same dirName.

**Tests added**:
- mixed cwd / no-cwd in same slug → canonical projectPath wins regardless of iteration order
- null / empty / whitespace / missing cwd → slug-derived fallback still works

Closing in favor of #245. Thanks for the patch — the cwd-from-JSONL approach was the right call.

PR #236, opened 2026-05-05
fix(copilot): guard legacy event data shapes

ozymandiashh:
## Summary

`origin/main` currently fails `npx tsc --noEmit` in `src/providers/copilot.ts` after the legacy Copilot parser gained a catch-all event variant:

```ts
| { type: string; timestamp?: string; data: Record<string, unknown> }
```

That catch-all is useful for newer/unknown Copilot events, but it also widens the discriminated union. TypeScript can no longer prove that `event.data` has `content`, `outputTokens`, `messageId`, or `toolRequests` inside the `user.message` / `assistant.message` branches, so the parser trips strict type errors.

This PR keeps the catch-all behavior and adds small local field readers for legacy event data instead of destructuring from an unsafe union shape.

## What changed

- Add typed helpers for legacy event fields:
  - `legacyStringField()`
  - `legacyNumberField()`
  - `legacyToolRequests()`
- Use those helpers in `parseLegacyEvents()` for model IDs, user content, assistant IDs, token counts, and tool requests.
- Keep valid legacy event behavior the same: same model, dedup key, token counts, mapped tools, timestamps, user message, and session ID.
- Make malformed/newer event payloads safer: non-string messages become `''`, non-number `outputTokens` become `0`, and malformed tool request entries are ignored.
- Add a regression test covering generic/malformed legacy event shapes while preserving a valid assistant message that carries `data.model` directly.

## Why this shape

The parser already needs to tolerate unknown Copilot event types. Removing the catch-all would make the type error disappear, but it would also weaken the intent of the previous Copilot parsing change. Reading the few required fields from `Record<string, unknown>` keeps the parser tolerant while satisfying strict TypeScript.

## Validation

- `npx tsc --noEmit`
- `npx vitest run tests/providers/copilot.test.ts`
- `npx vitest run`
- `npm run build`

### DIFF
--- a/src/providers/copilot.ts
+++ b/src/providers/copilot.ts
@@ -72,6 +72,22 @@ type LegacyCopilotEvent =
   | { type: 'assistant.message'; timestamp?: string; data: { messageId: string; outputTokens: number; interactionId?: string; toolRequests?: LegacyToolRequest[] } }
   | { type: string; timestamp?: string; data: Record<string, unknown> }
 
+function legacyStringField(data: Record<string, unknown>, key: string): string {
+  const value = data[key]
+  return typeof value === 'string' ? value : ''
+}
+
+function legacyNumberField(data: Record<string, unknown>, key: string): number {
+  const value = data[key]
+  return typeof value === 'number' ? value : 0
+}
+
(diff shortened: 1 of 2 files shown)


iamtoruk:
Closing — this is stale and the fix already exists on main.

PR #240, opened 2026-05-05
feat(cli): add session replay

ozymandiashh:
## Summary

This adds `codeburn replay <session-id>`, a local session timeline for reviewing one recorded AI coding session turn by turn.

The existing dashboard and reports are good at aggregate cost visibility, but they do not make it easy to answer what happened inside a specific expensive or interesting session. `replay` fills that gap with a focused CLI view: prompt, cost, model, tools, shell commands, retries, edit flag, and token counts per turn.

## What changed

- add a new `codeburn replay <session-id>` command
- search sessions by exact session id or session id prefix
- support the same practical filters users expect elsewhere:
  - `--period`
  - `--from` / `--to`
  - `--provider`
  - `--project`
  - `--exclude`
- add `--json` for scripts and downstream analysis
- add `--no-prompts` for metadata-only replay output when users do not want prompt text in stdout
- render ambiguous prefix matches as a candidate list instead of picking one silently
- document the workflow and add an Unreleased changelog entry

## Output behavior

The text view prints a compact timeline with:

- project name and path
- session id and time window
- total spend, turn count, call count, and token totals
- per-turn category and cost
- user prompt, unless `--no-prompts` is set
- models used in the turn
- core tools, MCP tools, skills, and shell commands
- retry and edit flags

The JSON view returns the same structured replay data, with `userMessage: null` when `--no-prompts` is used.

## Privacy model

`replay` is local-only and does not upload data. By default it shows prompt text because the command is meant to help users understand a specific session.

`--no-prompts` hides user prompts in both text and JSON output, but it is not a full redaction mode. Shell commands and tool names remain visible because they are part of the execution timeline. The README calls this out explicitly so users do not mistake it for a share-safe export.

## Edge cases

- exact session id matches are preferred over prefix matches
- ambiguous prefixes fail with a visible candidate list and exit code 1
- empty queries return no matches in the library helper
- session id matching is case-insensitive
- candidate overflow pluralization is covered for more than 10 matches

## Validation

- `npx vitest run tests/replay.test.ts`
- `npm run build`
- `node dist/cli.js replay --help`
- `node dist/cli.js replay does-not-exist-hopefully --from 2999-01-01 --to 2999-01-02 --json` exits with code 1 and the expected not-found message
- `npx vitest run`
- `git diff --cached --check`

`npx tsc --noEmit` is still blocked by existing `src/providers/copilot.ts` type errors on `main`; this branch does not touch that provider path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 ## Unreleased
 
 ### Added (CLI)
+- **Session replay.** New `codeburn replay <session-id>` command renders a
+  local turn-by-turn timeline for a recorded session, including per-turn cost,
+  category, models, tools, MCP tools, skills, shell commands, retries, edit
+  flags, and token counts. Supports period, date range, provider, project,
+  exclude filters, `--json`, and `--no-prompts` for metadata-only output.
 - **Multiple subscription plans can be tracked at the same time.**
   `codeburn plan set` now stores plans in a provider-keyed `plans` map, so
   setting a Codex custom plan no longer overwrites an existing Claude plan.
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ codeburn export -f json         # JSON export
 codeburn optimize               # find waste, get copy-paste fixes
 codeburn optimize -p week       # scope the scan to last 7 days
 codeburn compare                # side-by-side model comparison
+codeburn replay <session-id>    # turn-by-turn session timeline
 codeburn yield                  # track productive vs reverted/abandoned spend
 codeburn yield -p 30days        # yield analysis for last 30 days
 codeburn models                 # per-model token + cost table (last 30 days)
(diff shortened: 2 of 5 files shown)


ozymandiashh:
Validation run on macOS arm64 with private details omitted. This is behavior proof for replay lookup and prompt hiding, not just build proof.

What was checked:
- A real session prefix resolved to a non-empty replay when scoped with provider/period options.
- `--json` output parsed successfully.
- `--no-prompts` returned replay data with user prompts hidden (`userMessage` values null in the structural check).
- A deliberately missing session id exited non-zero and reported no match, proving the failure path does not silently pick another session.

Commands:
- `npx vitest run tests/replay.test.ts` - 10/10 tests passed.
- `npx tsx src/cli.ts replay <redacted-session-prefix> --period all --provider claude --no-prompts --json` - JSON parsed, non-empty replay, prompts hidden.
- Negative lookup with a deliberately missing session id - non-zero exit and no-match message.

No project names, prompts, paths, session IDs, raw costs, usage totals, or private product details are included here.

ozymandiashh:
Closing this — it's an unsolicited feature PR that's gone stale, and per CONTRIBUTING these should start from an issue with maintainer buy-in. Cleaning up my open-PR queue; happy to revisit via an issue if there's interest.

PR #242, opened 2026-05-06
feat(optimize): detect context-heavy sessions

ozymandiashh:
## Summary

This adds a Context Bloat Analyzer finding to `codeburn optimize` so users can spot sessions where the effective input/cache footprint is much larger than the useful output. The goal is to make it obvious when a fresh thread or tighter context handoff would likely be cheaper than carrying everything forward.

- adds a new `detectContextBloat()` detector in the optimize pipeline
- measures effective input/cache tokens using the existing cache pricing assumptions: cache reads are discounted, cache writes use the cache-write multiplier
- flags sessions only when effective input/cache tokens are both large and disproportionate to output
- shows the top context-heavy sessions, their input/output ratio, and sharp growth from the previous project session
- caps extreme ratio display as `1000+:1` so zero-output sessions do not produce noisy numbers
- keeps candidate ordering deterministic for stable previews and tests
- documents the detector in README and CHANGELOG

## Detection model

The detector is intentionally conservative:

- effective input/cache floor: 75,000 tokens
- minimum effective input/output ratio: 25:1
- token-savings estimate: context above a healthier 15:1 ratio
- high impact when there are at least 3 candidates or at least 500K effective input/cache tokens across candidates
- growth is measured against the immediately previous session in the same project, even if that previous session is below the reporting threshold

This keeps raw cache-read-heavy sessions from being overstated while still catching sessions that are expensive in practice.

## Validation

- `npx vitest run tests/optimize.test.ts`
- `npm run build`
- `npx vitest run`
- `node dist/cli.js optimize --help`
- `git diff --check`

`npx tsc --noEmit` still fails on the existing Copilot provider type errors in `src/providers/copilot.ts`, outside this diff. The same failure is present on `origin/main`.

thats all i got for the moment :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@
   than the call's own cache buckets could contain. Threshold:
   >10 tools available, <20% coverage, observed in ≥2 sessions. Closes #2.
 - **Session cost outlier detector.** New `optimize` finding flags sessions costing more than 2x their peer-session average within the same project. Ignores sub-$1 outliers to avoid noise. Requires at least 3 sessions per project for a baseline.
+- **Context bloat detector.** New `optimize` finding flags sessions where
+  effective input/cache tokens are large and disproportionate to output.
+  Cache reads are discounted in the estimate to avoid overstating cheap cached
+  context. The report highlights top sessions by imbalance, notes sharp
+  growth from the previous project session, and suggests starting fresh with
+  only the current goal, relevant files, failing output, and constraints.
 
 ### Fixed (CLI)
 - **`all` period semantics unified between CLI and dashboard.** The dashboard treated `--period all` as all-time (epoch start) while the CLI bounded it to the last 6 months. Both now consistently mean "Last 6 months". Period helpers (`Period`, `PERIODS`, `PERIOD_LABELS`, `toPeriod`, `getDateRange`) consolidated into `cli-date.ts`. Use `--from` / `--to` for unbounded historical ranges.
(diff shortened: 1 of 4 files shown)


iamtoruk:
Superseded by #246, which carries your original commit forward and applies three review fixes on top:

**Found via real-data probe (22.6K sessions / $4.8K spend):**

1. **Heavy overlap with `detectSessionOutliers`.** Top-5 context-bloat sessions were identical to top-5 outliers — same sessions, two framings. Headline savings showed 62% of spend, double-counted. Fix: `detectSessionOutliers` now takes an optional `excludedSessionIds` set; context-bloat runs first and passes its IDs. Real-data outlier count dropped 96 → 19, top-5 lists are now disjoint, headline savings now an honest 10% of spend.
2. **Time-blind growth ratio.** "1131x previous session input" was sometimes just resume-after-a-long-gap, not bad context management. Fix: 7-day baseline window — growth callout suppressed when predecessor is older.
3. **Binary impact tiering.** `>=3 candidates → high` flattened a 300-session pile-up against a 3-session minor finding. Fix: three real tiers (high ≥10 or ≥500K total · low ≤2 AND <200K total · medium otherwise).

Plus 7 new tests covering medium/high boundaries, 1000+:1 cap with non-zero output, time-gap suppression, below-threshold anchor growth, and outlier-exclusion behavior.

Closing in favor of #246. The cache-aware effective-token approach was the right call — well-built detector, just needed the dedup with outliers and the time-gap polish to match real-world data shape.

PR #243, opened 2026-05-06
feat(provider): IBM bob provider support added

SunithaGudisagarIBM1:
https://github.com/getagentseal/codeburn/issues/248

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,7 @@ Arrow keys switch between Today, 7 Days, 30 Days, Month, and 6 Months (use `--fr
 | Qwen | `~/.qwen/projects/<project>/chats/` | Yes |
 | Goose | `~/.local/share/goose/sessions/sessions.db` (SQLite) | Yes |
 | Antigravity | `~/.gemini/antigravity/conversations/` | Yes |
+| IBM Bob | `~/Library/Application Support/IBM Bob/User/globalStorage/ibm.bob-code/tasks/` | Yes |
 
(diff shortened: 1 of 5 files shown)


SunithaGudisagarIBM1:
@iamtoruk can you take a look at this PR

SunithaGudisagarIBM1:
Hi @ozymandiashh, 
I have already implemented the issue #[248](https://github.com/getagentseal/codeburn/issues/248) in this PR, 
Can you please take this PR as i came with the issue and the implementation as well.
@iamtoruk @ozymandiashh I am interested to contribute to this open-source project, please consider this PR for merging.
Thank you

AgentSeal:
Thanks for working on this, and for opening the original issue requesting Bob support.

We've been reviewing both this PR and #303 against a real IBM Bob install. A few issues that make this one hard to merge as-is:

- **Merge conflicts** with current main (provider registry, README, and other files have changed significantly since this branch was created).
- **Model fallback lands on `cline-auto`** which is misleading for Bob users -- Bob deliberately hides its model name, so the fallback should be Bob-specific (e.g. `ibm-bob-auto`), not the generic Cline label.
- **Test fixtures fabricate model tags** (`claude-sonnet-4-6` in `api_conversation_history.json`) that don't exist in real Bob installs. The real-world case is always no model tag, and that path isn't tested.

PR #303 addresses these points and is closer to merge-ready, so we're going to move forward with that one after validating it against real data.

Appreciate you opening the issue and the implementation effort -- the discovery paths and shared parser reuse in your PR were solid.

PR #256, opened 2026-05-07
Fix runtime guard, SQLite diagnostics, and Antigravity Windows discovery

### LINKED ISSUE #232: SyntaxError: Invalid regular expression flags on Pi
Hello, 

I want to use the tool on my Pi 4 with Ubuntu. But I got the following error:

file:///usr/local/lib/node_modules/codeburn/node_modules/string-width/index.js:19
const zeroWidthClusterRegex = /^(?:\p{Default_Ignorable_Code_Point}|\p{Control}|\p{Format}|\p{Mark}|\p{Surrogate})+$/v;
                              ^

SyntaxError: Invalid regular expression flags
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:152:18)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:299:14)

Node.js v18.20.4

BR,

### LINKED ISSUE #249: Issue: CodeBurn doesn’t pick up Antigravity usage on Windows
Hi, I’m using `codeburn` `0.9.6` on Windows and Antigravity usage isn’t showing up at all.

I installed it with: npm install -g codeburn. When I run: codeburn report --provider antigravity --format json I get 0 calls and 0 sessions. The important part is that the Antigravity data is definitely there on disk. On my machine I have:

C:\Users\Admin\.gemini\antigravity\conversations

and that folder contains a lot of recent .pb files, including files from today. I also confirmed that the Antigravity language server is running on Windows. From looking through the installed CodeBurn code, it seems like the Antigravity provider is assuming a server setup that matches macOS/Linux, but not Windows. In particular, Windows Antigravity is exposing things like:

--extension_server_port
--extension_server_csrf_token

whereas CodeBurn seems to be expecting --https_server_port and a different RPC shape.

So this doesn’t look like missing logs or a bad path. It looks more like the Windows Antigravity provider path isn’t setup correctly yet.

In short:
- Antigravity conversation files exist
- Antigravity is actively running
- CodeBurn still reports zero usage on Windows

Happy to provide more details if useful, though in the same vein if I am doing something stupid please let me know!

### LINKED ISSUE #250: OpenCode/Cursor silently empty on Node 22.x < 22.13 (node:sqlite gated behind --experimental-sqlite)
## Bug

`codeburn status` shows zero calls and `$0.00` for OpenCode (and Cursor) on Node 22.11.0, even with a healthy `~/.local/share/opencode/opencode.db` (140 root sessions, 2k+ assistant messages with token data). All other providers (Claude Code, Codex, etc.) work correctly. The failure is silent — no warning printed in normal `status`/`report`/`today` output.

## Root cause

`dist/cli.js` calls `requireForSqlite("node:sqlite")` to load the built-in module. The `node:sqlite` module landed in Node 22.5 **behind the `--experimental-sqlite` flag** and only became available without a flag starting around Node 22.13 (and stable later). On Node 22.5–22.12 (no flag), the require throws `ERR_UNKNOWN_BUILTIN_MODULE: No such built-in module: node:sqlite`.

Codeburn catches this and sets `loadError`, but `discoverSessions()` in the OpenCode/Cursor providers just returns `[]` on `!isSqliteAvailable()` — no surfaced warning. End user sees an empty provider with no clue why.

## Reproduction

- Node 22.11.0 (or any 22.x prior to ~22.13 without `--experimental-sqlite`)
- codeburn 0.9.6 (`npx codeburn` or `brew install getagentseal/codeburn/codeburn`)
- Active OpenCode usage with messages in `~/.local/share/opencode/opencode.db`

```
$ node --version
v22.11.0
$ npx codeburn status --provider opencode
  Today  \$0.00  0 calls    Month  \$0.00  0 calls
$ node -e "require('node:sqlite')"
Error [ERR_UNKNOWN_BUILTIN_MODULE]: No such built-in module: node:sqlite
```

After upgrading to Node 22.22.2:

```
$ node --version
v22.22.2
$ npx codeburn status --provider opencode
  Today  \$10.39  114 calls    Month  \$10.39  114 calls
```

## Why this is hard to spot

- README still says \`better-sqlite3\` is installed automatically as an optional dependency, but the published 0.9.6 \`package.json\` has no \`optionalDependencies\`. Users searching for \`better-sqlite3\` issues hit a dead end.
- \`engines\` field is \`"node": ">=22"\`, so 22.0–22.12 appear supported but silently break SQLite providers.
- Closed issue #98 documents an earlier related silent-fail; the current 0.9.6 silent-fail mode is different (built-in module gate) but indistinguishable to a normal user.

## Suggested fixes (any one would resolve)

1. **Loud warning when SQLite-backed providers are configured but driver is unavailable.** The current `loadError` message exists but isn't surfaced unless a session parser is actually invoked. Print it during `discoverSessions()` if any DB files are found on disk but the driver can't load.
2. **Bump engines requirement** to whatever Node version makes `node:sqlite` available without a flag (≈22.13), or document this clearly in README under Requirements.
3. **Fall back to better-sqlite3** when `node:sqlite` is unavailable, and actually declare it in `optionalDependencies` (matching the README).
4. **Detect the specific `ERR_UNKNOWN_BUILTIN_MODULE` case** and emit a targeted message: "Your Node version (X) doesn't expose node:sqlite without --experimental-sqlite. Upgrade to Node 22.13+ or pass that flag."

## Environment

- codeburn 0.9.6
- Node 22.11.0 (broken) → Node 22.22.2 (working)
- macOS arm64
- OpenCode 1.14.39

ozymandiashh:
Closes #250.
Closes #249.
Closes #232.

## Summary

This PR fixes three recent runtime/provider failures that all presented as either silent zero-usage output or a crash before CodeBurn could explain what was wrong.

### Node runtime guard (#232)

CodeBurn now starts through a tiny `src/cli.ts` launcher that checks for Node `>=22.13.0` before dynamically importing the main CLI bundle.

That matters because Node 18 can currently install/run the package despite the engine mismatch, then crashes inside modern terminal dependencies with `SyntaxError: Invalid regular expression flags`. Early Node 22 also exposes `node:sqlite` inconsistently. The launcher keeps `dist/cli.js` small enough to parse on older Node versions and prints an actionable upgrade message before loading Ink/React/SQLite-backed provider code.

Packaging changes:

- Moved the existing CLI implementation to `src/main.ts`.
- Added `src/node-runtime.ts` with tested version parsing and guard messaging.
- Updated `tsup` to emit both `dist/cli.js` and `dist/main.js`.
- Kept `target: node20` intentionally so the launcher can run far enough on older Node to print the upgrade message.
- Raised `engines.node` and README requirements to `>=22.13.0`.

Verified build shape:

- `dist/cli.js` is 1.46 KB and only contains the runtime guard + dynamic import.
- `dist/main.js` contains the full CLI bundle.

### SQLite-backed provider diagnostics (#250)

Cursor, Goose, and OpenCode now check whether their DB data exists before checking SQLite availability. If data exists but `node:sqlite` cannot load, CodeBurn writes a clear stderr diagnostic instead of silently returning zero sessions.

This avoids the confusing Node 22.5-22.12 failure mode where OpenCode/Cursor usage exists on disk, but `node:sqlite` is gated behind `--experimental-sqlite` and the provider quietly reports nothing.

Also updated stale README text that still mentioned `better-sqlite3`; CodeBurn now documents the built-in `node:sqlite` requirement.

### Antigravity Windows discovery (#249)

Antigravity provider discovery now handles the Windows language-server shape reported in #249:

- Process discovery uses PowerShell / `Get-CimInstance Win32_Process` on Windows instead of POSIX `ps`.
- Server flag parsing supports both legacy and Windows names:
  - `--https_server_port`
  - `--csrf_token`
  - `--extension_server_port`
  - `--extension_server_csrf_token`
- Flag parsing supports both `--flag value` and `--flag=value` syntax.
- CSRF-like values are sanity-checked so unrelated process lines do not become false-positive servers.
- Connect-RPC response parsing accepts both wrapped and unwrapped payload shapes for model maps and trajectory metadata.

## Review fixes applied before PR

External review called out a few hardening points; this PR includes them:

- Added explicit CSRF token shape/length validation for Antigravity process parsing.
- Reduced Windows PowerShell discovery timeout from 8s to 5s.
- Forced PowerShell stdout to UTF-8.
- Guarded model-map extraction against malformed RPC payload entries.
- Added a Copilot legacy fixture for malformed assistant events while tightening TypeScript around that file.
- Added a `tsup` comment documenting why the launcher target intentionally remains below the runtime floor.

## Tests

- `npx tsc --noEmit`
- `npx vitest run tests/providers/antigravity.test.ts tests/providers/copilot.test.ts tests/node-runtime.test.ts`
- `npm run build`
- `npx vitest run` — 40 files, 539 tests passed
- `node dist/cli.js --version`
- `git diff --check`

## Pre-push review

- Argus-style local review: PASS
- Claude Opus 4.7 review: PASS after hardening follow-ups were applied
- Gemini 3.1 Pro Preview review: PASS; CLI emitted one model-capacity retry warning before returning the verdict

## What it does NOT touch

- Token aggregation and pricing math
- Dashboard rendering
- JSON stdout payload shape
- Provider discovery paths outside Cursor/Goose/OpenCode/Antigravity runtime checks


iamtoruk:
Closing this PR -- thanks for the work @ozymandiashh, but we need these split into separate PRs.

This bundles three unrelated fixes (#232, #249, #250) into one PR, which makes it hard to review and test independently. It also conflicts with 5-7 files on current main (especially the cli.ts restructuring), making the merge impractical at this point.

Per our contribution guidelines, please submit one PR per issue:

1. **#232 (Node runtime guard)** -- the launcher split and version check
2. **#250 (SQLite diagnostics)** -- the provider warnings when node:sqlite can't load
3. **#249 (Antigravity Windows)** -- the PowerShell discovery and flag parsing

Each one can be reviewed, tested, and merged on its own. Rebasing against current main will also be much easier with smaller, focused PRs.

PR #262, opened 2026-05-07
Fix: stop misattributing undated usage to 'Today'

darthrevanyunka:
## What this fixes
Codeburn was misreporting Cursor usage by assigning undated records to the current day. This made **Today** appear artificially high, and the "most active day" could shift depending on when reports or menubar refreshes were run.

## Root cause
Two parser paths generated synthetic "now" timestamps when source records were undated:
- Cursor bubble rows with missing `createdAt`
- Cursor `agentKv` sessions with no timestamp metadata

That rebucketed historical/undated usage into the runtime date.

## Why this is important
- Daily usage charts become unreliable
- Menubar Today values can be misleadingly large
- Teams tracking token spend by day get incorrect trend signals

## Changes
- Require real `createdAt` for bubble day bucketing (skip undated bubble rows)
- Stop defaulting `agentKv` timestamps to current time
- Parse timestamp candidates from `agentKv` content (`timestamp`, `createdAt`, and `time*` variants)
- Skip `agentKv` sessions when no reliable timestamp exists
- Add regression tests for all of the above

## Verification
- [x] `npm test -- tests/providers/cursor.test.ts`
- [x] `npm run build`

### DIFF
--- a/src/providers/cursor.ts
+++ b/src/providers/cursor.ts
@@ -50,6 +50,11 @@ type AgentKvRow = {
 type AgentKvContent = {
   type?: string
   text?: string
+  timestamp?: string | number
+  createdAt?: string | number
+  time?: string | number
+  timeCreated?: string | number
+  time_created?: string | number
   providerOptions?: {
     cursor?: {
       modelName?: string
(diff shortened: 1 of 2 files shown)


darthrevanyunka:
This PR is intended as a follow-up/complement to the recent Cursor provider fixes (including #257 and related work), focusing on one remaining day-bucketing edge case: undated records can still be assigned synthetic timestamps and drift into Today buckets.

What this PR tightens:

bubble rows: do not bucket rows without real createdAt into the current day
agentKv: avoid synthetic now bucketing; use embedded timestamp metadata when present, otherwise skip undated sessions
Why this matters:

prevents moving most-active-day behavior
keeps menubar and daily trend data consistent with actual event timing
Happy to split or trim scope if maintainers prefer this as a narrower patch.

iamtoruk:
Nice catch on the undated bucketing - tested locally and the fix is solid. Two small things before we merge:

1. `takeUserMessage()` (line 197) is now unused since you inlined the shift. Can you remove it?
2. Line 528 still passes `source.path` as a third arg to `parseAgentKv` but the signature only takes two params now. Harmless but should match.

Happy to merge once those are cleaned up.

darthrevanyunka:
All done and tested

iamtoruk:
Thanks for tracking this down @darthrevanyunka -- the root cause analysis was spot on, and the bubble-side fix is exactly what we needed.

We cherry-picked the bubble half of your fix into #321 (now merged). The two changes from your PR that landed:

1. SQL query now requires `createdAt IS NOT NULL` instead of including undated rows
2. Application-level guard skips rows with empty `createdAt` instead of defaulting to `new Date()`

We left the agentKv changes out for now because Cursor stores agentKv content as binary protobuf blobs, not JSON with timestamp fields. The existing file mtime fallback is currently the most reliable timestamp source for those rows. We validated this against a live Cursor database with ~15k agentKv rows -- none of them had parseable timestamp fields in their content.

Appreciate the thorough work on this. If you notice any edge cases we missed, feel free to open a follow-up.

PR #295, opened 2026-05-10
Fix mangled project paths in By Project and Top Sessions panels

meghraoa:
## Summary

The `By Project` and `Top Sessions` panels in the dashboard displayed corrupted project names whenever a directory contained a dash or a dot. For example, `~/Projects/foo-bar` rendered as `Projects/foo/bar`, and `~/Projects/site.github.io` rendered as `Projects/site/github/io` with the `Projects/` prefix dropped entirely on deeper paths.

## Root cause

`shortProject()` in `src/dashboard.tsx` received Claude Code's encoded slug (e.g. `-home-user-Projects-foo-bar`) and tried to recover the original path by splitting on `-` and joining with `/`. This is intrinsically lossy: there's no way to distinguish dashes that came from path separators from dashes that were part of the original directory name.

## Fix

`ProjectSummary.projectPath` is already populated from the canonical `cwd` extracted by `parser.ts` (`extractCanonicalCwd`, with fallback to the legacy decode when no JSONL `cwd` is found). Other consumers (`export.ts`, `cli.ts`) already use it. The dashboard was the last holdout still re-decoding the slug.

This PR:
- Switches both the `By Project` and `Top Sessions` panels to consume `projectPath` (the real cwd).
- Rewrites `shortProject` to operate on an absolute path: split on `/`, keep the last 3 segments, strip the home prefix and any `private/tmp/...` macOS noise.
- Renames the in-flight `projectName` field to `projectPath` for semantic accuracy.
- Exports `shortProject` so it can be unit-tested.

## Test plan

- [x] `npm test` passes (added 6 cases for `shortProject` covering dashes, dots, the home dir itself, deep nesting, and paths outside home).
- [x] Manual: ran the dashboard against `~/.claude/projects/`; directories with dashes and dots now render verbatim instead of being split.

### DIFF
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -247,16 +247,17 @@ function DailyActivity({ projects, days = 14, pw, bw }: { projects: ProjectSumma
   )
 }
 
-const _homeEncoded = homedir().replace(/\//g, '-')
-
-function shortProject(encoded: string): string {
-  let path = encoded.replace(/^-/, '')
-  if (path.startsWith(_homeEncoded.replace(/^-/, ''))) {
-    path = path.slice(_homeEncoded.replace(/^-/, '').length).replace(/^-/, '')
-  }
-  path = path.replace(/^private-tmp-[^-]+-[^-]+-/, '').replace(/^private-tmp-/, '').replace(/^tmp-/, '')
+export function shortProject(absPath: string): string {
+  const home = homedir()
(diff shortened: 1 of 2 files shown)


iamtoruk:
Merged via #320 with your fix as the base. Added homedir caching, Windows backslash normalization, and aligned the test helper field name. Thanks for the clean fix @meghraoa.

Issue #234, opened 2026-05-05
Feature Request: ChatGPT desktop app provider support


nkmd108:
### Summary

Codeburn currently has a `--provider chatgpt` flag but it returns zero data. The ChatGPT desktop app (macOS) stores local data at:

- `~/Library/Application Support/com.openai.chat/`
- - `~/Library/Application Support/com.openai.atlas/`
Both directories contain SQLite databases (`logs_2.sqlite`, `state_5.sqlite`) but they do not appear to contain token-level usage data in a format codeburn can parse.

### Expected Behavior

Running `codeburn report --provider chatgpt` should surface token usage and cost data from ChatGPT desktop app sessions, similar to how Claude Code sessions are read from `~/.claude/`.

### Current Behavior

`--provider chatgpt` returns 0 calls, 0 tokens, $0.00 cost despite active ChatGPT desktop app usage.

### Environment

- macOS (Apple Silicon)
- - Codeburn v0.9.6
- - - ChatGPT desktop app installed (com.openai.chat + com.openai.atlas)
### Notes

If the ChatGPT desktop app does not expose per-session token data locally, an alternative approach could be reading from the OpenAI API usage endpoint (requires an API key) or supporting the OpenAI Codex CLI session format as a proxy for GPT model usage.

Would be happy to share more details about the local file structure if helpful.

ozymandiashh:
Implemented this as a diagnostic spike in PR #314: https://github.com/getagentseal/codeburn/pull/314

The PR intentionally does not add a real `chatgpt` usage provider yet. It adds `codeburn doctor chatgpt-desktop`, which:

- checks the reported macOS storage roots: `~/Library/Application Support/com.openai.chat` and `~/Library/Application Support/com.openai.atlas`
- recursively finds local SQLite databases such as `logs_2.sqlite` / `state_5.sqlite`
- opens them through CodeBurn's read-only SQLite wrapper
- reports schema metadata only: database filenames, table/view counts, and usage-like column names
- redacts the home directory to `~` in text and JSON output
- does not read conversation rows, message text, or column values

The goal is to confirm whether ChatGPT Desktop exposes defensible local token/cost fields before shipping a provider. If the diagnostic shows stable columns like `prompt_tokens`, `completion_tokens`, `model`, or `cost_usd`, the next PR can add the real provider with fixtures. If it only shows state/log tables, this should stay blocked on upstream local storage/API data rather than estimating costs from message text.

[...]

PR #314, opened 2026-05-11
Add ChatGPT Desktop diagnostics

ozymandiashh:
## Summary

- add `codeburn doctor chatgpt-desktop` as a pre-provider diagnostic for ChatGPT Desktop / Atlas local storage
- scan the known macOS roots `~/Library/Application Support/com.openai.chat` and `~/Library/Application Support/com.openai.atlas`, with `CODEBURN_CHATGPT_DESKTOP_DIRS` for custom roots
- open discovered SQLite databases through CodeBurn's read-only SQLite wrapper and report schema metadata only: database filenames, table/view counts, and usage-like column names
- redact the home directory to `~` in text and JSON output so affected users can share diagnostic output without exposing their local username
- document why this does not register a `chatgpt` provider or estimate costs until token/cost fields are confirmed from real local schemas

Refs #234.

## Why this is a spike

Issue #234 reports ChatGPT Desktop/Atlas SQLite files such as `logs_2.sqlite` and `state_5.sqlite`, but token-level usage has not been confirmed in a parseable local format. A real provider needs defensible token and cost data, not message-length guesses. This PR gives affected users a safe way to answer the schema question first.

Example follow-up path:

- if the diagnostic finds stable columns like `prompt_tokens`, `completion_tokens`, `model`, or `cost_usd`, we can add a real provider with fixtures based on that schema
- if it only finds app state/log tables with no usage accounting, #234 should stay open as blocked on upstream local storage/API data instead of shipping approximate costs

## Validation

- `npm run build`
- `npx vitest run` (49 files / 666 tests)
- `npx tsx src/cli.ts doctor chatgpt-desktop`
- `npx tsx src/cli.ts doctor chatgpt --format json`
- `npx tsx src/cli.ts doctor nope` returns the expected unknown-target error
- `git diff --check`
- Claude Opus 4.7 effort max: PASS
- Gemini 3.1 Pro Preview: PASS

## Notes

- On this machine, the ChatGPT Desktop/Atlas storage roots are not present, so the diagnostic returns `storage-not-found` with redacted `~/Library/...` paths.
- `npx tsc --noEmit` still fails on the existing unrelated `src/models-report.ts(151,13)` `TaskCategory` type error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## 0.9.9 - Unreleased
+
+### Added (CLI)
+- **ChatGPT Desktop diagnostic.** New `codeburn doctor chatgpt-desktop`
+  command for issue #234. It probes `com.openai.chat` and `com.openai.atlas`
+  local SQLite storage read-only and prints schema metadata plus redacted local
+  paths: database filenames, table names, and usage-like column names. It
+  intentionally does not register a ChatGPT provider or estimate costs from
+  message text until token/cost fields are confirmed in the local schema.
+
 ## 0.9.8 - 2026-05-10
 
 ### Added (CLI)
--- a/README.md
+++ b/README.md
@@ -91,6 +91,17 @@ codeburn models --task feature         # filter to feature-development work
 codeburn models --provider claude      # filter to one provider
 ```
 
+### Diagnostics
+
+```bash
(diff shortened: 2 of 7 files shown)


iamtoruk:
Closing — this is research/investigation work that doesn't need to ship as a public CLI command. We'll inspect the ChatGPT Desktop SQLite files directly when building the provider.

PR #332, opened 2026-05-15
Cache menubar PATH discovery and add deterministic tests

### LINKED ISSUE #331: Menubar can’t find codeburn from user-managed Node PATH
## Problem
On macOS, the menubar app can fail with:

`env: codeburn: No such file or directory`

when `codeburn` is installed under user-managed Node paths (for example via nvm). GUI-launched apps often inherit a reduced `PATH`, so the binary is not found even though it works in Terminal.

## Impact
- Menubar refresh silently fails for affected users.
- The app appears broken despite a valid CLI install.
- Users with nvm/volta/asdf/npm-global layouts are disproportionately affected.

## Reproduction
1. Install `codeburn` in a user-managed Node location (e.g. nvm under `~/.nvm/versions/node/.../bin`).
2. Launch CodeBurn menubar app from Finder (not from an interactive shell).
3. Trigger refresh.
4. Observe error: `env: codeburn: No such file or directory`.

## Root Cause
The menubar process used GUI environment `PATH` values that can omit user-managed Node bin directories.

## Proposed / Implemented Approach
- Augment menubar process PATH with known locations (`/opt/homebrew/bin`, `/usr/local/bin`) and discovered user bins (`~/.volta/bin`, `~/.npm-global/bin`, `~/.asdf/shims`, and nvm Node version bins containing an executable `codeburn`).
- Cache discovered user PATH entries as a one-time computation so refreshes do not repeatedly rescan filesystem directories.
- Add deterministic tests for:
  - nvm executable/bin discovery behavior
  - PATH augmentation dedupe behavior

## Notes
No debug instrumentation should be kept in production for this fix line.

vaibhavarora14:
## Root cause
The macOS menubar process can inherit a reduced GUI `PATH`, so `codeburn` is not found when installed under user-managed Node paths (for example nvm).

## Fix summary
- Keep PATH augmentation logic in `CodeburnCLI` and make discovery testable via deterministic closures.
- Cache discovered user PATH entries as a one-time computation (`cachedDiscoveredUserPathEntries`) to avoid repeated filesystem scans on refresh.
- Preserve existing behavior of appending known bins and nvm bin directories that contain an executable `codeburn`.

## Tests added
- `CodeburnCLITests.discoversExpectedUserPathEntries`: validates static user bins + nvm executable-bin discovery.
- `CodeburnCLITests.augmentedPathAppendsWithoutDuplicates`: validates PATH augmentation and dedupe behavior.

## Manual verification
- `cd mac && swift build`
- `cd mac && swift test`

## Issue
- Closes #331

### DIFF
--- a/mac/Sources/CodeBurnMenubar/Security/CodeburnCLI.swift
+++ b/mac/Sources/CodeBurnMenubar/Security/CodeburnCLI.swift
@@ -13,6 +13,7 @@ enum CodeburnCLI {
     /// PATH additions for GUI-launched apps, which otherwise get a minimal PATH that misses
     /// Homebrew and npm global installs.
     private static let additionalPathEntries = ["/opt/homebrew/bin", "/usr/local/bin"]
+    private static let cachedDiscoveredUserPathEntries = discoverUserPathEntries()
 
     /// Returns the argv that launches the CLI. Dev override via `CODEBURN_BIN` is honoured only
     /// if every whitespace-delimited token passes `safeArgPattern`. Otherwise falls back to the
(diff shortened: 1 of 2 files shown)


iamtoruk:
Closing — we'll patch the PATH discovery directly on main with isSafe validation.

Issue #336, opened 2026-05-16
Not tracking sub agents (Fedora 44)


MorsCordis:
It doesn't appear that Codeburn is tracking tool calls performed by subagents with Claude CLI.

It also seems to treat each Gemini session as a single call, not tracking tool calls/usage particularly well.  Only coding and exploration are showing.

vmvarela:
The same gap exists for the OpenCode provider. OpenCode stores sessions in SQLite with a parent–child hierarchy: a root session can have child and grandchild sessions, each with their own Part rows. The current parser only queries direct messages on root sessions, so any token usage or tool calls that happen inside a child session are silently dropped.

The root cause is in how discoverSessions maps session sources and how the message query is structured — it doesn't traverse the session tree. A WITH RECURSIVE CTE on the Session table would capture the full subtree and attribute everything back to the root session correctly.

Planning to look into this for OpenCode specifically — would this be a welcome addition alongside #340

ozymandiashh:
This is now split across two ready PRs:

- #340 covers the original cross-provider tracking gap for Gemini/Claude/Codex-style agent activity. Gemini now emits one CodeBurn provider call per assistant message with usage instead of one aggregate call per session, Claude discovery includes direct `subagents/*.jsonl` files, and Codex agent tool normalization is covered by regression tests.
- #343 covers the OpenCode child-session case mentioned in the thread. It keeps root-only discovery to avoid double counting, then parses the full unarchived child/grandchild session subtree back under the root session.

Both PRs are currently clean/mergeable with GitHub checks passing.

Important limitation: I did not validate either PR against the original Fedora reporter environment or real private local subagent history. The validation uses synthetic fixtures and parser regression tests, so no local project names, prompts, paths, session IDs, usage values, or private product details are included.

iamtoruk:
Both PRs are now merged:

- #340: Gemini now emits one call per assistant message instead of one aggregate per session. Claude discovery scans project-level subagents/*.jsonl files. Codex agent tool normalization covered by tests.
- #343: OpenCode parser walks the child/grandchild session subtree via recursive CTE so agent sessions roll up to the root session.

Please retry on main and let us know if subagent tracking works for your setup.

MorsCordis:
This is pretty awesome guys. Fast turnaround, really impressive. 

[...]

PR #343, opened 2026-05-17
Track OpenCode child sessions under root sessions

ozymandiashh:
## Summary

Addresses the OpenCode portion of #336 by tracking agent/subagent work that OpenCode stores in child sessions.

OpenCode records sessions in SQLite with a `session.parent_id` hierarchy. Before this PR, CodeBurn discovered only root sessions and then parsed only messages whose `message.session_id` matched that root. That meant work done by OpenCode child or grandchild sessions could be invisible in CodeBurn, including token usage, tool calls, bash commands, and costs.

This PR keeps discovery root-only to avoid double counting, but changes parsing so a root session includes its unarchived child-session subtree.

## Root cause

The existing parser used root-session queries like this:

```sql
SELECT id, time_created, data
FROM message
WHERE session_id = ?
```

That works for a flat session, but not for OpenCode's parent-child session shape:

```text
root session
├─ root message rows
└─ child session
   ├─ child message rows
   └─ grandchild session
      └─ grandchild message rows
```

Because the query only matched the root session id, CodeBurn ignored the child and grandchild rows even though they belong to the same user-visible OpenCode task.

## What changed

- `message` and `part` reads now use a recursive `session_tree` CTE starting from the discovered root session.
- The recursive walk includes only unarchived descendants, so archived child sessions remain excluded.
- Child sessions are still excluded from top-level discovery via the existing `parent_id IS NULL` discovery query. This prevents double counting.
- Parsed child/grandchild assistant calls are attributed back to the root `sessionId`, so downstream session summaries stay grouped under the root task.
- Deduplication keys now use the real message session id: `opencode:<message-session-id>:<message-id>`. This avoids collisions between root and child messages.
- The preceding user prompt is tracked per message session, not in one global string. This prevents an interleaved root/child timeline from assigning the wrong prompt to a child assistant call.
- OpenCode provider docs and changelog now describe the root-only discovery plus subtree parsing behavior.

## Example behavior

Given this SQLite session tree:

```text
root
└─ child
   └─ grandchild
```

Before:

```text
CodeBurn parsed: root assistant calls only
Missing: child tools, grandchild bash commands, child/grandchild token usage
```

After:

```text
CodeBurn parses: root + child + grandchild assistant calls
Top-level sessions shown: root only
Call attribution: all calls grouped under root session
Dedup keys: preserve the real source session for each message
```

## Tests added

- Root + child + grandchild fixture where all three sessions contain assistant calls.
- Verifies all calls are attributed to the root `sessionId`.
- Verifies dedup keys preserve the real source session ids.
- Verifies per-session user prompts do not leak across root/child/grandchild messages.
- Verifies child tools such as `task` and grandchild bash commands are surfaced.
- Verifies archived child sessions are excluded from the recursive subtree.

## Validation

- [x] `npx vitest run tests/providers/opencode.test.ts tests/provider-registry.test.ts` - 49/49 tests passed.
- [x] `git diff --check` - passed.
- [x] Argus-style strict review - PASS.

## Scope and privacy notes

I did not use real local OpenCode history or reporter data. The validation uses synthetic SQLite fixtures only, so no local project names, prompts, paths, session IDs, usage values, or private product details are included.

This PR intentionally does not change OpenCode zero-usage router-call handling from #342. If #342 merges first, this branch may need a small rebase because both PRs touch `src/providers/opencode.ts`, but the behaviors are separate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,13 @@
   `Shell`, `ReadFile`, and `WriteFile`, and maps hidden managed Kimi Code
   model aliases to priced Kimi K2 entries.
 
+### Fixed (CLI)
+- **OpenCode child sessions are attributed to their root session.** The
+  OpenCode parser now walks the unarchived `session.parent_id` subtree so
+  child and grandchild agent sessions contribute token and tool usage under
+  the discovered root session while still excluding child sessions from
+  top-level discovery to avoid double counting.
+
 ## 0.9.9 - 2026-05-15
 
 ### Added (CLI)
--- a/docs/providers/opencode.md
+++ b/docs/providers/opencode.md
@@ -26,6 +26,10 @@ Per `<sessionId>:<messageId>`.
 
 - **Schema validation is loud.** When a required table is missing, the parser logs an actionable warning telling the user which table is gone and what version of OpenCode it expects. This is the right behavior; do not silently swallow these.
 - Source paths are encoded as `<dbPath>:<sessionId>`.
+- Discovery only emits root sessions (`parent_id IS NULL`) to avoid double
+  counting. Parsing a root session walks the unarchived `session.parent_id`
+  subtree, so child and grandchild agent sessions contribute their message,
+  token, and tool usage back to the root session.
 - Each message's `parts` are indexed; preserving the order matters for reasoning-token correctness.
 - Tokens are reported across `input`, `output`, `reasoning`, `cache.read`, and `cache.write`. Anthropic semantics.
 - External MCP tools are stored as `<server>_<tool>` names (for example
--- a/src/providers/opencode.ts
+++ b/src/providers/opencode.ts
@@ -13,6 +13,7 @@ import type {
 } from './types.js'
 
 type MessageRow = {
+  session_id: string
   id: string
   time_created: number
   data: Uint8Array | string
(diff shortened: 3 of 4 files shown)


vmvarela:
Nice fix for the child session attribution. One related issue this PR doesn't address: discoverFromDb still emits one SessionSource per root session with path set to "${dbPath}:${sessionId}". That composite string isn't a real filesystem path, so when parser.ts calls fingerprintFile(source.path) → stat(filePath), stat always fails and returns null. The cache reconciler then treats every OpenCode session as changed on every run, bypassing the session cache entirely.

The fix is to emit one SessionSource per DB file (path = dbPath) and move the root-session iteration inside createParser, which already has the open DB connection. That way fingerprintFile can stat the actual .db file and the cache works as intended.

ozymandiashh:
@vmvarela thanks for calling this out. #347 added a generic fingerprint fallback, and I opened #348 as the OpenCode provider-side follow-up using the shape you suggested: discovery now emits one source per real `opencode*.db` path, and the parser iterates unarchived root sessions inside that DB while preserving the child/grandchild subtree parsing from this PR.

I also added a synthetic cache proof in the PR showing the cache key moves from `<dbPath>:root-a` / `<dbPath>:root-b` to `<dbPath>`, with child calls preserved and legacy keys pruned.

PR #337, opened 2026-05-16
fix: invalidate Cursor cache after missing-createdAt fix

lincolnwan:
## Summary

- Bump Cursor result cache version from 3 to 4 so stale v3 caches are invalidated.
- Respect `CODEBURN_CACHE_DIR` for Cursor result cache, matching docs and enabling isolated cache tests.
- Add a regression test for old v3 Cursor caches created before the missing-`createdAt` fix.

## Why

This follows up on `fe2e622` ("Skip Cursor bubble rows that lack a createdAt timestamp") from PR #321; it invalidates caches generated before that parser fix.

That parser fix made fresh parsing skip Cursor bubble rows without `createdAt`, but existing `~/.cache/codeburn/cursor-results.json` files could still contain v3 cached calls timestamped with parse time.

If the Cursor DB mtime/size did not change, CodeBurn would continue trusting that stale cache and still report phantom current-day usage.

## Testing

- `npm test -- tests/cursor-cache.test.ts tests/providers/cursor.test.ts tests/providers/cursor-bubble-dedup.test.ts tests/providers/cursor-workspace-breakdown.test.ts`
- `npx tsc --noEmit`

## Notes

This does not change Cursor parsing logic. It only invalidates previously poisoned Cursor result caches and adds coverage for that behavior.

PR #345, opened 2026-05-17
Fix Cursor agentKv timestamp attribution

### LINKED ISSUE #325: Inaccurate usage dates for Cursor due to mtime fallback on store.db
**Description:**
Codeburn incorrectly reports historical Cursor AI usage as new usage if the underlying Cursor database files are modified by background processes. 

**Steps to Reproduce:**
1. Have historical Cursor usage stored in `~/.cursor/chats/<hash>/store.db`.
2. Trigger a database file modification (e.g., Cursor performs an internal migration or vacuum, updating the file's `mtime` without changing chat timestamps).
3. Run `codeburn` to check recent usage.

**Expected Behavior:**
Codeburn should report usage based on the internal timestamps of the chat records (e.g., December 2025).

**Actual Behavior:**
Codeburn reports all historical usage in that database as occurring on the day the file was modified (e.g., today).

**Root Cause:**
When Codeburn parses the Cursor SQLite databases, if it fails to extract or parse the internal message timestamp, it falls back to the file's `mtime` (`fileStat.mtime.toISOString()`). Since SQLite databases are frequently touched by the host application for maintenance, `mtime` is an unreliable indicator of when the AI request actually occurred. 

**Suggested Fix:**
Remove the `mtime` fallback. Rely strictly on the internal SQLite timestamps (`createdAt` metadata or similar inside `store.db`). If a timestamp genuinely cannot be parsed, it should either be skipped, flagged as unknown, or use the earliest known date, rather than surfacing as new billing usage.

ozymandiashh:
## Summary

Fixes #325 by removing the remaining Cursor timestamp fallback that could attribute historical usage to the SQLite database modification time.

This is separate from the undated `bubbleId:%` path fixed on `main`: bubble rows already require `createdAt`. The remaining risk was `agentKv`, where CodeBurn still dated every parsed agentKv session with `statSync(state.vscdb).mtimeMs` and fell back to `new Date()` if the stat failed.

## Root Cause

Cursor can modify `state.vscdb` for background maintenance, migrations, or unrelated writes. Using the database file `mtime` as the usage timestamp means an old agentKv conversation can be reported as new usage whenever the DB file changes.

Concrete example covered by the tests:

1. A Cursor `agentKv` session contains old user/assistant content.
2. The SQLite file mtime is changed to `2099-01-01`.
3. Before this fix, CodeBurn could date that old session using the DB mtime.
4. After this fix, CodeBurn only reports the session when the row carries an internal timestamp.

## What Changed

- Extracts internal Cursor agentKv timestamps from `createdAt`, `timestamp`, or `time`.
- Accepts ISO strings, numeric millisecond timestamps, and numeric second timestamps stored as JSON numbers or numeric strings.
- Keeps the first valid internal timestamp seen for a request/session.
- Skips agentKv sessions with no internal timestamp instead of using `state.vscdb` mtime or parse time.
- Bumps Cursor result cache version to recompute older cached results that may contain mtime-derived timestamps.
- Adds `CODEBURN_CACHE_DIR` support to the Cursor result cache so tests can isolate cache state from the user's real cache.
- Adds focused regression coverage for the mtime failure mode and valid internal timestamp paths.

## Relation To #337

#337 invalidates cached results after the undated bubble-row fix. This PR includes the same cache-version/cache-dir hunk because agentKv timestamp semantics also change and v3 caches can contain mtime-bucketed agentKv calls.

If #337 merges first, this PR can be rebased to keep the agentKv parser/tests and drop the duplicate cache hunk if needed.

## Validation

- [x] `npx vitest run tests/providers/cursor-agentkv-timestamp.test.ts tests/providers/cursor.test.ts tests/providers/cursor-bubble-dedup.test.ts tests/providers/cursor-workspace-breakdown.test.ts` - 24/24 tests passed.
- [x] `git diff --check` - passed.
- [x] Argus-style strict local review - PASS; scope is limited to Cursor parser/cache/changelog/tests, no private data or unrelated files.

## Notes

I did not validate with real local Cursor history and did not include any local project names, prompts, paths, session IDs, usage values, screenshots, or private product details. The proof uses synthetic SQLite fixtures specifically constructed to reproduce the timestamp-attribution failure mode.

The intentional tradeoff is that legacy agentKv rows without any internal timestamp are skipped. That is safer than reporting historical usage under the wrong day.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,13 @@
   child and grandchild agent sessions contribute token and tool usage under
   the discovered root session while still excluding child sessions from
   top-level discovery to avoid double counting.
+- **Cursor agentKv timestamps no longer use database mtime.** Cursor agentKv
+  rows now require an internal timestamp (`createdAt`, `timestamp`, or `time`)
+  before CodeBurn reports usage for that session. Rows without an internal
+  timestamp are skipped instead of being attributed to the mutable SQLite file
+  modification time, preventing historical Cursor usage from appearing under
+  today's date. Cursor result cache version bumped to recompute older cached
+  entries. Closes #325.
 
 ## 0.9.9 - 2026-05-15
 
--- a/src/cursor-cache.ts
+++ b/src/cursor-cache.ts
@@ -5,13 +5,11 @@ import { randomBytes } from 'crypto'
 
 import type { ParsedProviderCall } from './providers/types.js'
 
-// Bumped to 3 for the workspace-aware breakdown change: the cursor parser
-// now derives `sessionId` from the bubble row key (the real composer id)
-// rather than the empty `conversationId` JSON field, and the workspace
-// router relies on those composer ids to bucket calls per project.
-// Version 2 caches contain `sessionId: 'unknown'` for every call and would
-// route everything to the orphan project, so we invalidate them.
-const CURSOR_CACHE_VERSION = 3
+// Bumped to 4 for the Cursor timestamp hardening: agentKv calls now require
+// an internal row timestamp instead of using the mutable SQLite database mtime.
+// Version 3 caches can contain historical agentKv calls bucketed under the
+// database modification day, so they must be invalidated.
+const CURSOR_CACHE_VERSION = 4
 
 type ResultCache = {
   version?: number
@@ -23,7 +21,7 @@ type ResultCache = {
 const CACHE_FILE = 'cursor-results.json'
 
 function getCacheDir(): string {
-  return join(homedir(), '.cache', 'codeburn')
+  return process.env['CODEBURN_CACHE_DIR'] ?? join(homedir(), '.cache', 'codeburn')
 }
 
 function getCachePath(): string {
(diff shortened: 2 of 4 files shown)


ozymandiashh:
I rechecked this after rebasing and am not going to push the rebased branch as-is.

The blocker is the prior maintainer validation on #262: a live Cursor database with ~15k agentKv rows reportedly had no parseable timestamp fields, and the mtime fallback was kept intentionally for that path. Requiring `createdAt` / `timestamp` / `time` on agentKv rows would likely drop real agentKv-only usage entirely.

Recommendation: close or rescope this PR. The cache-only stale-result issue should stay separate from changing agentKv timestamp semantics.

iamtoruk:
Closing this one. Validated against real Cursor state.vscdb data:

```
sqlite3 state.vscdb "SELECT json_extract(value, '$.createdAt'), json_extract(value, '$.timestamp'), json_extract(value, '$.time') FROM cursorDiskKV WHERE key LIKE 'agentKv:blob:%' AND hex(substr(value,1,1))='7B' LIMIT 10"
```

All three fields are NULL across 5,048 agentKv rows. Cursor does not store internal timestamps in agentKv JSON, so this PR would silently drop all agentKv session data rather than fixing the timestamp attribution.

The mtime fallback is imperfect but at least reports the data. A fix here would need a different approach (e.g. deriving timestamps from row ordering + bubble cross-reference, or accepting mtime as a bounded estimate).

PR #348, opened 2026-05-18
Fix OpenCode session cache source paths

ozymandiashh:
## Summary

Fixes OpenCode session-cache keying by discovering one source per real `opencode*.db` file instead of one synthetic `<dbPath>:<sessionId>` source per root session.

The parser now opens the database source, iterates unarchived root sessions inside it, and keeps the recursive child-session parsing added in #343.

## Root cause

OpenCode discovery used paths like:

```text
/path/to/opencode.db:root-session-id
```

The shared session cache fingerprints source paths with `stat()`. That synthetic path is not a real file, so OpenCode cache entries could be treated as changed or stale on every run.

## What changed

- OpenCode discovery now emits `path: dbPath` for each `opencode*.db`.
- The parser iterates unarchived root sessions from the DB source.
- Root parsing still walks unarchived child/grandchild session subtrees.
- Legacy compound `<dbPath>:<sessionId>` source paths are still accepted.
- Old synthetic cache keys are pruned on the next parse because discovery now reports only the real DB path.
- Docs and changelog describe the new cache behavior.

## Validation

- [x] Differential synthetic repro against `origin/main` vs this branch.
- [x] `npx vitest run tests/providers/opencode-cache.test.ts tests/providers/opencode.test.ts tests/provider-registry.test.ts` - 54/54 tests passed.
- [x] `npx vitest run` - 876/876 tests passed.
- [x] `git diff --check origin/main...HEAD` - passed.
- [x] Argus-style strict review - PASS.

<details>
<summary>Cache proof</summary>

Synthetic OpenCode DB fixture:

```text
root-a
└─ child-a
root-b
```

Same fixture parsed on `origin/main` and this branch.

`origin/main` cache keys:

```text
/tmp/.../opencode.db:root-a
/tmp/.../opencode.db:root-b
```

This branch cache keys:

```text
/tmp/.../opencode.db
```

Behavior stayed equivalent:

```text
calls: 3
root sessions: 2
projects:
  tmp-project-a: 2 calls, 1 session
  tmp-project-b: 1 call, 1 session
child dedup key preserved:
  opencode:child-a:msg-child-a-assistant
warm unchanged run rewrote cache: false
legacy <dbPath>:<sessionId> keys pruned: true
```

</details>

## Scope and privacy notes

Validation uses synthetic SQLite fixtures only. No real local OpenCode history, prompts, paths, session IDs, usage values, or private project data are included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@
   model aliases to priced Kimi K2 entries.
 
 ### Fixed (CLI)
+- **OpenCode session cache keys now use real database paths.** Discovery now
+  emits one source per `opencode*.db` file and the parser iterates root sessions
+  inside that database, so shared session-cache fingerprinting stats the actual
+  SQLite file instead of a synthetic `<dbPath>:<sessionId>` identifier.
 - **OpenCode child sessions are attributed to their root session.** The
   OpenCode parser now walks the unarchived `session.parent_id` subtree so
   child and grandchild agent sessions contribute token and tool usage under
--- a/docs/providers/opencode.md
+++ b/docs/providers/opencode.md
@@ -4,11 +4,11 @@ OpenCode (sst/opencode).
 
 - **Source:** `src/providers/opencode.ts`
 - **Loading:** lazy (`src/providers/index.ts:59-75`)
-- **Test:** `tests/providers/opencode.test.ts` (676 lines, the largest provider test)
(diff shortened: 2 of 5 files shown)


ozymandiashh:
Closing to keep my open-PR queue clean — this branch has gone stale and now conflicts with main. Happy to rebase and reopen if the change is still wanted.

PR #352, opened 2026-05-18
Fix Gemini and Vibe one-shot rates

### LINKED ISSUE #351: Validate Mistral Vibe and Gemini one-shot rates
## Summary

MorsCordis reported that Mistral/Vibe and Gemini are showing 100% one-shot rates across tracked categories. They also noted that Mistral/Vibe does not appear to track cache hits, which can inflate estimated cost.

## Context

This came up as a follow-up on #336 after the subagent tracking work in #340 and #343.

The redacted Vibe logging structure shared there shows `meta.json.stats` with aggregate session-level fields such as `session_prompt_tokens`, `session_completion_tokens`, `context_tokens`, and `session_cost`, but no cache-read/cache-write token fields. `messages.jsonl` contains message and tool-call structure, but no per-assistant-message usage or cache fields.

## What to validate

- Mistral/Vibe one-shot rates: CodeBurn currently parses Vibe usage as one coarse session-level provider call, which may not give the classifier enough per-turn structure to produce reliable one-shot/retry rates.
- Gemini one-shot rates: Gemini now emits per-assistant-message calls, but the one-shot/retry classifier should still be validated against representative sessions because 100% across categories is suspicious.
- Mistral/Vibe cache accounting: if Vibe does not persist cache-read/cache-write token fields locally, CodeBurn cannot recover exact cache accounting from logs. This should be documented and, if possible, wired up later when Vibe exposes cache token fields.

## Notes

CodeBurn one-shot metrics are not based on cache hits. They are based on edit turns with zero detected retries/self-corrections, so one-shot validation and cache accounting should be treated as related observations but separate fixes.

Requested upstream Vibe fields that would help CodeBurn:

- cache read input tokens
- cache creation/write input tokens
- whether `session_prompt_tokens` includes cached tokens
- per-assistant-message usage, if available in a future Vibe release

ozymandiashh:
## Summary

Addresses #351 by fixing how non-Claude provider calls are cached for one-shot/retry classification.

The core issue was not cache hits. CodeBurn's one-shot rate is based on edit turns with zero detected retries/self-corrections. For Gemini and Mistral Vibe, the cached turn shape did not give the classifier enough structure to see retries inside a single user request.

This PR adds provider-level turn grouping so related assistant calls are cached under the same user turn. That lets the existing classifier see multi-message flows like `Edit -> Bash -> Edit` and count them as retries instead of reporting them as independent one-shot turns.

Closes #351.

## Root Cause

The classifier already knows how to detect retries when a turn contains multiple assistant calls:

```text
User: implement parser fix
Assistant call 1: Edit
Assistant call 2: Bash
Assistant call 3: Edit
```

That should be one edit turn with one retry, because the assistant edited, ran a command, then edited again.

Before this PR, provider calls emitted through `ParsedProviderCall` were converted into cached turns one call at a time. So Gemini assistant messages were cached like this:

```text
Turn 1: [Edit]  -> retries = 0, one-shot
Turn 2: [Bash]  -> no edit turn
Turn 3: [Edit]  -> retries = 0, one-shot
```

The classifier never saw the full `Edit -> Bash -> Edit` sequence in one turn, so the one-shot rate could look artificially perfect.

## Example: Gemini

Gemini already exposes per-assistant-message token/tool data, so no private user logs are needed to validate the fix.

Synthetic fixture shape:

```json
[
  { "type": "user", "content": "implement parser update in src/parser.ts" },
  { "type": "gemini", "id": "g1", "toolCalls": [{ "name": "edit_file" }] },
  { "type": "gemini", "id": "g2", "toolCalls": [{ "name": "run_command", "args": { "command": "npm test" } }] },
  { "type": "gemini", "id": "g3", "toolCalls": [{ "name": "edit_file" }] }
]
```

Before:

```text
3 cached turns, each with 1 assistant call
retry detector cannot see Edit -> Bash -> Edit
oneShotTurns can be inflated
```

After:

```text
1 cached turn, with assistant calls [g1, g2, g3]
retries = 1
oneShotTurns = 0
```

The new regression test asserts exactly that.

## Example: Mistral Vibe

Mistral Vibe's local logs are different from Gemini:

- `meta.json.stats` has cumulative session totals such as `session_prompt_tokens`, `session_completion_tokens`, and `session_cost`.
- `messages.jsonl` has user/assistant/tool message structure and assistant `tool_calls`.
- Current Vibe logs do not expose cache-read/cache-write token fields.

So this PR does not invent cache token counts. CodeBurn still reports Vibe cache token counts as `0` until Vibe persists those fields locally.

What this PR does instead:

1. Splits Vibe assistant messages into per-message provider calls so the classifier can see assistant message order.
2. Groups those calls under the current user turn via `turnId`.
3. Distributes cumulative session prompt/completion tokens across the assistant calls while preserving exact session totals.
4. Prefers `meta.json.stats.session_cost` when present, because that is the best local cost signal available and may already reflect Vibe-side accounting better than CodeBurn's price-derived estimate.
5. Preserves Vibe's provider-calculated cost through the session cache so `session_cost` is not lost during cached reads.

Synthetic Vibe fixture shape:

```json
// meta.json
{
  "stats": {
    "session_prompt_tokens": 300,
    "session_completion_tokens": 90,
    "session_cost": 0.123456
  }
}
```

```json
// messages.jsonl, simplified
{ "role": "user", "content": "implement parser update" }
{ "role": "assistant", "message_id": "a1", "tool_calls": [{ "function": { "name": "search_replace" } }] }
{ "role": "assistant", "message_id": "a2", "tool_calls": [{ "function": { "name": "bash", "arguments": "{\"command\":\"npm test\"}" } }] }
{ "role": "assistant", "message_id": "a3", "tool_calls": [{ "function": { "name": "write_file" } }] }
```

After parsing:

```text
1 cached turn, with assistant calls [a1, a2, a3]
retries = 1
oneShotTurns = 0
totalInputTokens = 300
totalOutputTokens = 90
totalCostUSD = 0.123456
```

## What Changed

- Add optional `ParsedProviderCall.turnId` so providers can mark calls that belong to the same user turn.
- Group parsed provider calls into cached turns by `(sessionId, turnId)` when `turnId` is present.
- Assign Gemini assistant calls to the current user turn.
- Split Mistral Vibe assistant messages into per-message calls grouped under the current user turn.
- Prefer `meta.json.stats.session_cost` for Mistral Vibe cost when present.
- Add optional cached `costUSD` for Vibe calls so provider-calculated Vibe cost survives cache round-trips.
- Bump session cache to v2 so existing cached Gemini/Vibe entries are re-derived with the new grouping.
- Document current Vibe cache limitations and the `session_cost` behavior.

### Behavior proof

The main proof is `tests/provider-turn-grouping.test.ts`, which exercises the exact suspicious one-shot shape rather than only checking that the suite is green.

Gemini fixture:

```text
input:  user prompt + Gemini assistant messages [edit_file, run_command, edit_file]
output: one parsed CodeBurn session turn with assistant calls [g1, g2, g3]
asserts: retries = 1
asserts: category oneShotTurns = 0
```

This proves the classifier can now see the full `Edit -> Bash -> Edit` sequence for Gemini instead of treating each assistant message as a separate one-shot candidate.

Mistral Vibe fixture:

```text
input:  meta.json session totals + session_cost + messages [search_replace, bash, write_file]
output: one parsed CodeBurn session turn with assistant calls [a1, a2, a3]
asserts: retries = 1
asserts: category oneShotTurns = 0
asserts: totalInputTokens = 300
asserts: totalOutputTokens = 90
asserts: totalCostUSD = 0.123456
```

This proves both sides of the Vibe fix: retry detection sees the multi-message edit flow, and Vibe's own `session_cost` survives parser/cache round-trip instead of being replaced by a price-derived estimate.

`tests/providers/mistral-vibe.test.ts` also includes a direct cost regression where the fixture sets an intentionally large token price but `session_cost = 0.381681`; the provider returns exactly `0.381681`, proving `session_cost` takes precedence over estimated pricing.

### Command results

- [x] `./node_modules/.bin/tsc --noEmit --pretty false`
- [x] `npx vitest run tests/provider-turn-grouping.test.ts tests/providers/gemini.test.ts tests/providers/mistral-vibe.test.ts tests/parser-gemini-cache.test.ts tests/session-cache.test.ts` - 68 tests passed
- [x] `npm run build`
- [x] `npm test -- --run` - 63 files / 874 tests passed
- [x] `git diff --check`
- [x] Claude Opus 4.7 effort max review: PASS
- [x] Gemini 3.1 Pro Preview review: PASS
- [x] GitHub checks: `check`, `assess`, and `semgrep` pass

## Notes

Validation uses synthetic fixtures only. No private prompts, project names, session IDs, or local user logs were used.

This PR does not claim exact Vibe cache token accounting. Current Vibe local logs do not include cache token fields, so exact cache read/write token reporting requires an upstream Vibe logging change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,16 @@
   model aliases to priced Kimi K2 entries.
 
 ### Fixed (CLI)
+- **Gemini and Mistral Vibe one-shot rates use provider turn grouping.**
+  Provider calls that belong to the same user turn are cached together, so
+  retry detection can see multi-message `Edit -> Bash -> Edit` flows instead
+  of counting each assistant message as an independent one-shot turn. Mistral
+  Vibe now splits assistant-message tool calls into turn-grouped calls while
+  preserving cumulative session token totals, and it prefers
+  `meta.json.stats.session_cost` over price-derived estimates when available
+  because current Vibe logs do not expose cache token fields. Session cache is
+  bumped to v2 so existing Gemini/Vibe cache entries are re-derived. Closes
+  #351.
 - **OpenCode child sessions are attributed to their root session.** The
   OpenCode parser now walks the unarchived `session.parent_id` subtree so
   child and grandchild agent sessions contribute token and tool usage under
--- a/docs/providers/mistral-vibe.md
+++ b/docs/providers/mistral-vibe.md
@@ -21,7 +21,11 @@ Subagent traces are stored under a parent session's `agents/` folder with the sa
 
 ## Caching
 
-None.
+Current Vibe local logs do not expose cache-read/cache-write token fields, so
+CodeBurn reports cache token counts as `0`. When `meta.json.stats.session_cost`
+is present, CodeBurn uses that session total instead of re-estimating from
+prompt/completion token prices because it is the best cache-aware cost signal
+available in the local log shape.
 
 ## Deduplication
 
@@ -29,8 +33,8 @@ Per `mistral-vibe:<session_id>`.
 
 ## Quirks
 
-- **Usage is cumulative per session.** Vibe does not write per-assistant-message token usage into `messages.jsonl`; token counts come from `meta.json.stats.session_prompt_tokens` and `session_completion_tokens`. CodeBurn emits one usage record per Vibe session.
-- **Cost prefers Vibe's own model prices.** `meta.json.stats.input_price_per_million` and `output_price_per_million` are used first, with the active model config as a fallback. LiteLLM pricing is only used when Vibe provides no price data.
+- **Usage is cumulative per session.** Vibe does not write per-assistant-message token usage into `messages.jsonl`; token counts come from `meta.json.stats.session_prompt_tokens` and `session_completion_tokens`. CodeBurn splits assistant-message tools into their user turns for classification and distributes the cumulative token/cost totals across those assistant calls so session totals remain unchanged.
+- **Cost prefers Vibe's own session total.** `meta.json.stats.session_cost` is used first. If it is missing, `meta.json.stats.input_price_per_million` and `output_price_per_million` are used with the active model config as a fallback. LiteLLM pricing is only used when Vibe provides no price data.
 - **Project names come from metadata.** Discovery uses `meta.json.environment.working_directory` and falls back to the session directory name if that field is missing.
 - **Tool calls come from messages.** Assistant `tool_calls[*].function.name` is normalized to the standard CodeBurn names (`bash` to `Bash`, `search_replace` to `Edit`, etc.). Bash commands are extracted from `function.arguments.command`.
 
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1517,6 +1517,32 @@ function providerCallToTurn(call: ParsedProviderCall): ParsedTurn {
 
 // ── Cache Conversion ───────────────────────────────────────────────────
 
+function providerCallToCachedCall(call: ParsedProviderCall): CachedCall {
+  return {
+    provider: call.provider,
+    model: call.model,
+    usage: {
+      inputTokens: call.inputTokens,
+      outputTokens: call.outputTokens,
+      cacheCreationInputTokens: call.cacheCreationInputTokens,
+      cacheReadInputTokens: call.cacheReadInputTokens,
+      cachedInputTokens: call.cachedInputTokens,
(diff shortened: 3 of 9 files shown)


iamtoruk:
Superseded by #355, which combines the turnId grouping from this PR (for Gemini/Vibe) with the toolSequence approach from #353 (for Kiro/Goose).

PR #356, opened 2026-05-18
Add MCP project profile advisor

ozymandiashh:
## Summary
- Adds an MCP project profile advisor to `codeburn optimize`.
- Flags MCP servers that are clearly useful in one or two hot projects but are also loaded into cold projects where they are never invoked.
- Estimates cold-project schema overhead with cache-aware accounting and recommends project-scoping instead of blindly removing the server.

## Why
Existing MCP optimize findings answer two narrower questions:

1. Is an MCP server configured but never used?
2. Is a large MCP server loaded with mostly unused tools?

This adds a third question: is the MCP useful, but scoped too broadly?

For example, a GitHub MCP server may be useful in the main code repo but unnecessary in docs/playground repos. Removing it globally would hurt the hot workflow; project-scoping it preserves the useful setup while reducing cold-project schema overhead.

## What changed
- Adds `detectMcpProfileAdvisor()` to `src/optimize.ts`.
- Builds per-server, per-project profile stats from existing parsed data:
  - loaded sessions come from `session.mcpInventory`
  - invocations come from `session.mcpBreakdown`
- Emits a finding when:
  - a server is loaded in at least three projects
  - at least 80% of invocations are concentrated in the top one or two hot projects
  - at least two cold sessions loaded the server without invoking it
- Suppresses servers already eligible for the existing low MCP tool coverage finding to avoid duplicate reports.
- Uses combined cache-aware token accounting when multiple candidate servers share the same cold sessions, so they cannot independently claim the same cache bucket.

## Validation
I validated the behavior by running the real `detectMcpProfileAdvisor()` export against controlled `ProjectSummary` fixtures via `npx tsx --eval`. This constructs the actual edge cases the feature is meant to handle and prints detector output.

```json
{
  "one_hot_project_profile": {
    "title": "1 MCP server should be project-scoped",
    "tokensSaved": 4000,
    "expectedTokensSaved": 4000,
    "containsHotProject": true,
    "containsColdProjects": true,
    "proof": "two cold projects each loaded 4 github MCP tools: 4*400 schema tokens * 1.25 cache-write pricing * 2 cold sessions = 4,000 effective tokens"
  },
  "two_hot_project_profile": {
    "title": "1 MCP server should be project-scoped",
    "containsBothHotProjects": true,
    "containsColdProjects": true,
    "proof": "top two projects can form the hot profile while cold projects are still recommended for project-scoping"
  },
  "shared_cold_session_cap": {
    "title": "2 MCP servers should be project-scoped",
[...]

PR #354, opened 2026-05-18
Add MCP and skill ROI optimize insights

ozymandiashh:
## Summary
- Adds two new `optimize` findings for capability-level MCP/skill analysis:
  - low edit ROI for invoked MCP servers and skills in implementation-like turns
  - retry impact for capabilities whose edit turns retry materially more than same-category baseline turns
- Keeps the findings diagnostic: CodeBurn asks users to inspect sessions and narrow/remove capabilities only after confirming the sessions justify it.
- Caps shared-turn token/cost/retry impact once when multiple candidate capabilities appear in the same turn, so co-occurring MCP + skill usage cannot inflate ranking.

## Why
`optimize` already reports unused MCP inventory and ghost skills, but it could not answer two higher-level questions:

1. "This MCP server/skill is being used, but is it helping implementation work?"
2. "When this capability is involved, does the agent need more retry loops than comparable work?"

Those are different from ghost/unused checks: a capability can be invoked frequently and still be low-signal for edit-producing work, or correlate with repeated fix/test loops. The new findings surface that as a review signal without pretending correlation is causation.

## What changed
- Adds turn-keyed capability aggregation over parsed session summaries:
  - MCP capabilities are grouped by server from canonical `mcp__<server>__<tool>` names.
  - Skill capabilities come from parsed `call.skills` only; generic turn labels such as `subCategory` are not treated as skills.
  - Capabilities are counted once per turn even if the same MCP server/tool appears multiple times.
- Adds a low-edit-ROI detector for implementation-like categories (`coding`, `debugging`, `feature`, `refactoring`, `testing`).
  - Example output shape:
    - `MCP docs: 1/4 implementation turns produced edits (25% edit rate), $2.00 touched`
    - `skill api-review: 0/3 implementation turns produced edits (0% edit rate), $1.50 touched`
- Adds a retry-impact detector using same-category edit baselines.
  - Example output shape:
    - `skill planner: 2.0 retries/edit turn vs 0.0 baseline in the same task categories (3 edit turns, baseline 3)`
- Reuses a single capability aggregation pass inside `scanAndDetect`, alongside existing MCP coverage aggregation.
- Adds regression tests for:
  - ROI sample thresholds
  - MCP/skill findings together
  - per-turn MCP deduplication
  - shared-turn savings caps for co-occurring MCP + skill candidates
  - no `subCategory` false skill labels
  - same-category retry baselines
  - retry sample thresholds

## Validation
[...]

PR #357, opened 2026-05-18
Add MCP and skill reliability report

ozymandiashh:
## Summary
- Adds an MCP/skill reliability report to `codeburn optimize`.
- Flags MCP servers and skills whose edit turns are disproportionately retry-heavy.
- Uses turn-level capability evidence and shared-turn token accounting so one retry-heavy turn is not counted twice when both an MCP server and a skill appear together.

## Why
Existing optimize findings can show broad MCP waste, context bloat, and low-worth sessions, but they do not answer a capability-level reliability question:

> When a specific MCP server or skill is used during edit work, does that work repeatedly need retries?

This is useful for MCP and skill tuning because the right action is often not removal. A retry-heavy skill may need clearer instructions. A retry-heavy MCP server may need narrower project scope, a smaller tool set, or better usage guidance. The finding is intentionally framed as correlation, not causation, so users inspect the sessions before changing config.

For example:

- `Skill reviewer` appears in 5 edit turns, and 3 of those edit turns need retries.
- `mcp__ci__run` maps to MCP server `ci`, and the same 3/5 edit-turn retry pattern appears there.
- If both the skill and MCP server appear in the same retry-heavy turns, the savings estimate counts those turns once instead of reporting two independent piles of waste.

## What changed
- Adds `detectCapabilityReliability()` to `src/optimize.ts`.
- Aggregates per-capability edit reliability from existing parsed turn data:
  - MCP server evidence comes from `call.mcpTools`, normalized from names like `mcp__ci__run` to server `ci`
  - skill evidence comes from `call.skills`
  - `turn.subCategory` is deliberately not treated as skill evidence, avoiding legacy or classifier-derived false labels
- Emits a finding only when a capability has enough evidence:
  - at least 5 edit turns
  - at least 3 retry-heavy edit turns
  - at least 50% of edit turns are retry-heavy
- Ignores read-only turns so the report stays focused on edit reliability.
- Estimates recoverable tokens from retry-heavy edit turns with a 50% ceiling.
- De-duplicates shared retry-heavy turns across all flagged capabilities before computing `tokensSaved`.
- Adds focused regression coverage in `tests/optimize.test.ts`.

## Validation
I validated the behavior by running the real `detectCapabilityReliability()` export against controlled `ProjectSummary` fixtures via `npx tsx --eval`. This exercises the actual detector path and prints the expected edge cases.

```json
{
  "skill_retry_report": {
    "title": "1 skill correlates with retry-heavy edits",
    "tokensSaved": 1500,
    "expectedTokensSaved": 1500,
    "includesSkill": true,
    "proof": "5 edit turns using Skill reviewer; 3 retry-heavy turns at 1,000 effective tokens each; shared recovery ceiling is 50%, so 3*1000*0.5 = 1,500 tokens"
  },
  "mcp_retry_report": {
    "title": "1 MCP server correlates with retry-heavy edits",
    "tokensSaved": 1500,
    "expectedTokensSaved": 1500,
    "includesMcpServer": true,
    "proof": "same retry pattern attributed from mcp__ci__run to MCP server ci"
  },
  "shared_mcp_skill_turn_cap": {
    "title": "2 MCP/skill capabilities correlate with retry-heavy edits",
    "tokensSaved": 1500,
    "expectedTokensSaved": 1500,
    "includesBoth": true,
    "proof": "MCP server ci and Skill reviewer share the same 3 retry-heavy turns; tokensSaved stays 1,500 instead of doubling to 3,000"
  },
  "healthy_guard": {
    "finding": null,
    "proof": "1/5 retry-heavy edit turns is below the 50% retry-rate threshold"
  },
  "subcategory_guard": {
    "finding": null,
    "proof": "turn.subCategory without actual call.skills metadata does not create a skill reliability finding"
  },
  "readonly_guard": {
    "finding": null,
    "proof": "read-only retry turns are ignored because the detector is scoped to edit reliability"
  }
}
```

What this proves:
- A retry-heavy skill emits a skill reliability finding with the expected `1,500` token estimate.
- A retry-heavy MCP tool invocation is attributed to its MCP server namespace.
- A shared MCP+skill retry pattern reports both capabilities but keeps `tokensSaved` at `1,500`, not the inflated `3,000`.
- Healthy capabilities below the retry-rate threshold emit no finding.
- `turn.subCategory` alone does not create a skill finding.
- Read-only retry turns are ignored.

Supporting checks:
- `./node_modules/.bin/tsc --noEmit --pretty false`
- `npx vitest run tests/optimize.test.ts` — 82 tests passed
- `npm run build`
- `npm test -- --run` — 62 files / 877 tests passed
- `git diff --check`
- Claude Opus 4.7, effort max review returned `PASS`
- Gemini 3.1 Pro Preview review returned `PASS`

## Notes
- The finding text explicitly says this is correlation, not proof of causation.
- The fix is a prompt to audit the retry-heavy capability, not an automatic MCP or skill config edit.
- This PR is independent of #354 and #356; it is based on current `main` and uses the existing parsed turn metadata.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ## Unreleased
 
 ### Added (CLI)
+- **MCP and skill reliability report.** `codeburn optimize` now detects MCP
+  servers and skills whose edit turns are disproportionately retry-heavy,
+  using turn-level MCP/Skill call evidence and a shared-turn token estimate so
+  one retry-heavy turn is not double-counted across multiple capabilities.
 - **Agent and subagent tracking coverage.** Gemini sessions now emit one
   provider call per assistant message with token usage instead of one aggregate
   call per session, preserving per-message tools, bash commands, timestamps,
--- a/src/optimize.ts
+++ b/src/optimize.ts
@@ -99,6 +99,15 @@ const WORTH_IT_LOW_MAX_CANDIDATES = 2
 const WORTH_IT_LOW_MAX_TOTAL_COST_USD = 10
 const WORTH_IT_HIGH_MIN_CANDIDATES = 10
 const WORTH_IT_HIGH_TOTAL_COST_USD = 50
+const CAPABILITY_RELIABILITY_MIN_EDIT_TURNS = 5
+const CAPABILITY_RELIABILITY_MIN_RETRY_TURNS = 3
+const CAPABILITY_RELIABILITY_MIN_RETRY_RATE = 0.50
+const CAPABILITY_RELIABILITY_RECOVERY_FRACTION = 0.50
+const CAPABILITY_RELIABILITY_PREVIEW = 5
+const CAPABILITY_RELIABILITY_LOW_MAX_CANDIDATES = 1
+const CAPABILITY_RELIABILITY_LOW_MAX_TOKENS = 50_000
+const CAPABILITY_RELIABILITY_HIGH_MIN_CANDIDATES = 5
+const CAPABILITY_RELIABILITY_HIGH_IMPACT_TOKENS = 200_000
 
 // ============================================================================
 // Scoring constants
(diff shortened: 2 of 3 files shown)


iamtoruk:
Merged via ac3a367 - rebased onto main, resolved CHANGELOG conflict, all 944 tests pass. Thanks for the contribution.

PR #366, opened 2026-05-20
fix: add DeepSeek v4 Pro/Flash pricing and fix snapshot survival through runtime cache

NihalJain:
## Summary

Fixes  #365!

Pricing sourced from api-docs.deepseek.com/quick_start/pricing:

  deepseek-v4-flash: $0.14/M input · $0.28/M output · $0.0028/M cache-read · free cache-write
  deepseek-v4-pro:   $1.74/M input · $3.48/M output · $0.0145/M cache-read · free cache-write

Note: LiteLLM PR [#27056 ](https://github.com/BerriAI/litellm/pull/27056)had incorrect cache-read prices (1.4e-8 for flash, 1.74e-7 for pro); corrected here to match the official docs (2.8e-9 and 1.45e-8 respectively).

Changes:
- bundle-litellm.mjs: add both models to MANUAL_ENTRIES so they survive future re-bundles until LiteLLM PR #27056 is merged
- models.ts: add display names; fix loadPricing() to merge bundled snapshot entries back after loading the 24-hour runtime cache file, so MANUAL_ENTRIES are never silently dropped
- providers/claude.ts: add shortNames for the Claude provider display
- tests/models.test.ts: regression tests for pricing values, provider-prefixed lookups, calculateCost with observed production token counts, and display names

## Testing

- [x] I have tested this locally against real data (not just unit tests)
- [x] `npm test` passes
- [x] `npm run build` succeeds
- [x] ` node dist/cli.js today` shows correct costs and session counts for this provider
- [x] ` node dist/cli.js today --provider claude` shows correct model names and pricing
- [x] Screenshot or terminal output attached below
  - Before: 
    - Screenshot 2026-05-20 at 9 58 21 PM
  - After: 
    - Screenshot 2026-05-20 at 10 06 02 PM

Issue #365, opened 2026-05-20
Cost is 0 for deepseek-v4-* models with claude


NihalJain:
<img width="422" height="42" alt="Image" src="https://github.com/user-attachments/assets/80efb4c7-1ad8-4fb5-80d6-f01525b4921c" />

ozymandiashh:
I looked at #366. It is directionally right: it adds fallback pricing for `deepseek-v4-*` and also handles the runtime LiteLLM cache path, which is the important root cause here.

One thing to fix before merging that PR: the current DeepSeek pricing page lists `deepseek-v4-pro` at the active 75% discounted rate until 2026-05-31 15:59 UTC, so the current values in #366 overprice Pro by 4x:

```text
#366 deepseek-v4-pro:
input  1.74e-6
output 3.48e-6
cache  1.45e-8

current documented discounted price:
input  4.35e-7
output 8.7e-7
cache  3.625e-9
```

I also opened #367 with the discounted Pro pricing and an end-to-end CLI regression for the reported behavior: the same Claude JSONL fixture with a fresh runtime LiteLLM cache missing these models reports cost 0 on `origin/main`, then reports non-zero costs on the branch.

So either #366 should update the Pro rates and add a CLI-level stale-cache regression, or #367 can be used as the follow-up fix.

NihalJain:
Oh wait I thought we use actual price for pricing.? I could update PR to have discounted if thats the case. @iamtoruk WDYT?

NihalJain:
We can add "end-to-end CLI regression for the reported behavior" test as follow up IMO

PR #367, opened 2026-05-20
Fix DeepSeek v4 Claude pricing through stale runtime cache

ozymandiashh:
## Summary

Addresses #365 by pricing DeepSeek v4 Claude-compatible sessions instead of reporting them as $0.

## Root cause

CodeBurn relies on LiteLLM pricing data, but `deepseek-v4-pro` and `deepseek-v4-flash` are not yet available in the bundled/runtime LiteLLM data used by current CodeBurn. A fresh 24-hour runtime pricing cache can also replace the bundled snapshot with data that still lacks these models, so adding snapshot entries alone is not enough.

## What changed

- Add temporary DeepSeek v4 manual pricing entries to the LiteLLM bundle until upstream LiteLLM includes them.
- Merge bundled snapshot fallback entries into cached or freshly fetched LiteLLM pricing only when the runtime data lacks a model.
- Add DeepSeek v4 display names for shared model reporting and the Claude provider.
- Add an end-to-end CLI regression fixture for Claude JSONL sessions with a stale runtime pricing cache.

## Validation

I ran the same Claude JSONL fixture against `origin/main` and this branch. The fixture contains two Claude-compatible assistant calls using the reported DeepSeek v4 model ids, plus a fresh runtime LiteLLM cache that does not contain those models.

Input shape:

```text
provider: claude
model: deepseek-v4-pro
input_tokens: 2,477,914
output_tokens: 762,994
cache_read_input_tokens: 258,556,928

model: deepseek-v4-flash
input_tokens: 1,552,573
output_tokens: 353,914
cache_read_input_tokens: 48,388,608
```

Before, on `origin/main`, CodeBurn parsed the calls and tokens but priced both models at zero:

```text
deepseek-v4-pro:   calls 1, cache reads 258,556,928, cost 0
deepseek-v4-flash: calls 1, cache reads 48,388,608, cost 0
TOTAL:             calls 2, cost 0
```

After, on this branch, the same input produces non-zero costs and the expected display names:

```text
DeepSeek v4 Pro:   calls 1, cache reads 258,556,928, cost 2.678966
DeepSeek v4 Flash: calls 1, cache reads 48,388,608, cost 0.451944
TOTAL:             calls 2, cost 3.13091
```

That before/after path is now covered by `tests/cli-deepseek-v4-pricing.test.ts`, which runs the CLI against the realistic Claude fixture and stale runtime cache.

## Notes

- DeepSeek v4 Pro uses the currently documented 75% discounted price that DeepSeek lists as active until 2026-05-31 15:59 UTC.
- This overlaps with #366, but adds the stale-runtime-cache CLI proof and uses the current discounted Pro price.


iamtoruk:
Hey @ozymandiashh  I have just noticed that there was a similar PR but yours was better for next time, rather than having a duplicate PR if we can suggest changes on the existing PR, it would be better.  :D

PR #370, opened 2026-05-21
Add Hermes Agent CLI provider

ozymandiashh:
## Summary

Closes #368 by adding Hermes-agent CLI support through Hermes' current SQLite state store at `~/.hermes/state.db`.

## Context

Hermes-agent now persists session metadata, messages, tool calls, token usage, and cost fields in `state.db`. CodeBurn did not have a Hermes provider, so `--provider hermes` could not discover or report those CLI sessions.

## What changed

- Add a lazy-loaded `hermes` provider backed by shared SQLite helpers.
- Discover Hermes root sessions from `~/.hermes/state.db`, with `HERMES_HOME` override support.
- Parse root and child session trees, including cycle-safe recursive traversal.
- Decode user/assistant messages, structured `\0json:` content, assistant `tool_calls`, fallback `role = 'tool'` rows, Bash commands, web searches, and MCP tool names.
- Preserve Hermes stored estimated/actual costs in cache while keeping locally calculated costs recomputable when pricing changes.
- Add Hermes to provider discovery, keyword metadata, and session-cache environment fingerprinting.

## Validation

I created a realistic Hermes `state.db` fixture with one CLI session:

```text
session: manual-root
source: cli
model: anthropic/claude-sonnet-4.6
input_tokens: 1200
output_tokens: 340
cache_read_tokens: 90
cache_write_tokens: 40
actual_cost_usd: 0.031
tool_calls: terminal(command: npm test ...), write_file(path: src/providers/hermes.ts)
```

Running CodeBurn against that fixture with `--provider hermes` produced the expected model, tokens, cost, tools, and shell command:

```text
calls: 1
cost: 0.031
tokens: input 1200, output 340, cacheRead 90, cacheWrite 40
model: Sonnet 4.6
tools: Bash 1, Write 1
shellCommands: npm 1
```

The same path is covered by `tests/cli-hermes-provider.test.ts`, which builds a temporary `~/.hermes/state.db` and runs the real CLI with `--provider hermes`.

Additional checks:

- `npm test -- --run tests/providers/hermes.test.ts tests/cli-hermes-provider.test.ts tests/provider-registry.test.ts` - 25 tests passed
- `npm test -- --run` - 67 files / 909 tests passed
- `npm run build`
- `git diff --check`

## Notes

- Older closed Hermes PRs (#46 and #200) targeted stale or incomplete storage assumptions; this implementation is based on the current Hermes `state.db` schema.
- Recursive session traversal is bounded/cycle-safe for malformed parent links.
[...]

PR #386, opened 2026-05-23
Add Hermes Agent provider

### LINKED ISSUE #368: support for Hermes-agent CLI
hello, I recently came across codeburn and have been using Hermes-agent cli for a couple of weeks now and would like to know if support is possible.

iquantecho:
Hey! I added Hermes Agent support to CodeBurn because I wanted to use it with my own Hermes sessions, and figured I’d share it in case it’s useful here too. No pressure at all if this isn’t the direction you want to take it.

## Summary

- Adds a Hermes Agent provider that reads session-level usage from Hermes SQLite state databases.
- Discovers both root and profile Hermes databases.
- Tracks input, output, cache read/write, reasoning tokens, stored costs, tools, shell commands, and inferred projects.
- Includes fixture-style provider tests for normal, legacy, root/profile, tool mapping, and path parsing cases.

## Testing

- `npm test -- tests/providers/hermes.test.ts`

I also checked it against my real local Hermes data before opening this.

Closes #368

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ## Unreleased
 
 ### Added (CLI)
+- **Hermes Agent provider.** Track token usage, cost, and tool breakdowns
+  for Hermes Agent sessions. Reads from `~/.hermes/state.db` and per-profile
+  databases. Supports session-level accounting with actual/estimated costs
+  from Hermes, falling back to CodeBurn's model pricing table.
 - **Tooling breakdowns in dashboard and menubar.** New panels showing core
   tools, MCP servers, and shell command usage per session and across periods.
 - **File-aware retry detection with typed ToolCall.** One-shot rate now tracks
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
   tool-name-based detection.
 
 ### Fixed (CLI)
+- **Hermes profile and Windows project parsing.** Hermes Agent state databases
+  now only treat exact `profiles/<name>/state.db` paths as named profiles,
+  avoiding sibling-directory prefix collisions, and `Current working directory:`
+  inference recognizes Windows drive paths.
 - **Codex 100% one-shot rate.** Codex function_call arguments are JSON strings,
   not objects, and `patch_apply_end` stores file paths in `changes` object keys.
   Both are now parsed correctly.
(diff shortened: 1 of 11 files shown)


iamtoruk:
Thanks for contributing this. The provider structure is solid and test coverage is good. Found a few things that need fixing before we can merge:

## Must fix

**1. Cost fallback bug (`hermes.ts:619`)**

```typescript
const actualCost = row.actual_cost_usd ?? row.estimated_cost_usd ?? 0
```

`??` does not trigger for `0`, only for `null`/`undefined`. So when `actual_cost_usd = 0` (explicit zero, not null) and `estimated_cost_usd = 0.50`, the chain returns `0`, skips the estimate entirely, and falls through to `calculatedCost`. Need explicit null checks:

```typescript
const costUSD =
  (row.actual_cost_usd ?? 0) > 0 ? row.actual_cost_usd!
  : (row.estimated_cost_usd ?? 0) > 0 ? row.estimated_cost_usd!
  : calculatedCost
```

**2. `inferProject` matches all message roles (`hermes.ts:480`)**

The `cwdPattern` regex runs against every message including assistant responses. An assistant quoting "Current working directory: /tmp/something" in its output would set the wrong project path. Filter to `role === 'user'` only (or `role === 'system'`).

**3. `CACHE_VERSION` bump is unnecessary (`session-cache.ts`)**

Bumping from 3 to 4 forces a full cache rebuild for every user on upgrade (all providers, not just hermes). The cache schema did not change. The `PROVIDER_PARSE_VERSIONS` entry you already added for hermes is the right mechanism here and is sufficient on its own. Please revert `CACHE_VERSION` back to 3.

## Should fix

**4. Discovery query has no LIMIT (`hermes.ts:506`)**

The `SELECT ... FROM sessions WHERE usage > 0` has no bound. A user with a large Hermes database would load all rows into memory. Add a `LIMIT 10000` or similar cap.

**5. `sanitizeProject` strips only one leading separator (`hermes.ts:320`)**

```typescript
trimmed.replace(/^[/\\]/, '')
```

Missing the `+` quantifier. A path like `//network/share` becomes `-network-share`. Should be `/^[/\\]+/`.

iamtoruk:
Thanks for adding Hermes Agent support. Heads up that there are three open PRs for the same provider (#442, #386, #370). After reviewing all three, #442 is the one being moved forward as the base, mainly because it is the leanest and follows the existing provider conventions most closely (HERMES_HOME override, read-only parameterized sqlite, the cursor-style path encoding). That is not a knock on this PR, just avoiding two Hermes providers landing.

We will consolidate on #442 (it has review feedback to address first). I will close this one to avoid duplication, but please do weigh in on #442 if there is anything from your implementation worth folding in. Appreciate the work.

PR #388, opened 2026-05-24
Pause menubar refresh after repeated stalls

dengsenyao:
## Summary
- add a small refresh backoff policy that pauses automatic force refresh after 3 consecutive watchdog stalls, then retries with exponential backoff from 30s up to 5m
- cancel Claude/Codex live quota single-flight tasks when force/manual refresh recovery cancels the parent refresh, and guard old quota tasks from clearing newer task handles
- show a menubar pause banner with a Retry action, and clear the pause after a successful refresh
- add explicit request timeout to the update check path; Claude/Codex usage and token refresh requests already had 30s request timeouts

## Why
The existing force-refresh watchdog cancelled only the outer `forceRefreshTask` and immediately let the next timer tick start the same force-refresh path again. If the underlying refresh path repeatedly stalled, the app could log `force refresh stuck ... cancelling and restarting` every watchdog interval forever while the menu data stayed stale.

The quota refresh path also used unstructured single-flight tasks. Cancelling the parent force refresh did not clear those handles, so a later refresh could wait on the same stuck single-flight task rather than starting cleanly.

## Tests
```sh
swift build
DYLD_FRAMEWORK_PATH=/Library/Developer/CommandLineTools/Library/Developer/Frameworks \
DYLD_LIBRARY_PATH=/Library/Developer/CommandLineTools/Library/Developer/usr/lib \
swift test --enable-swift-testing \
  -Xswiftc -F -Xswiftc /Library/Developer/CommandLineTools/Library/Developer/Frameworks \
  -Xlinker -F -Xlinker /Library/Developer/CommandLineTools/Library/Developer/Frameworks \
  -Xlinker -rpath -Xlinker /Library/Developer/CommandLineTools/Library/Developer/Frameworks \
  -Xlinker -rpath -Xlinker /Library/Developer/CommandLineTools/Library/Developer/usr/lib
```

### DIFF
--- a/mac/Sources/CodeBurnMenubar/AppStore.swift
+++ b/mac/Sources/CodeBurnMenubar/AppStore.swift
@@ -44,6 +44,7 @@ final class AppStore {
     var subscriptionError: String?
     var subscriptionLoadState: SubscriptionLoadState = ClaudeCredentialStore.isBootstrapCompleted ? .dormant : .notBootstrapped
     var capacityEstimates: [String: CapacityEstimate] = [:]
+    var refreshPauseMessage: String?
 
     var codexUsage: CodexUsage?
     var codexError: String?
@@ -217,6 +218,17 @@ final class AppStore {
         if clearCache {
             cache.removeAll()
         }
+    }
+
+    func pauseAutomaticRefresh(until: Date, consecutiveStalls: Int) {
+        let formatter = DateFormatter()
+        formatter.timeStyle = .short
+        formatter.dateStyle = .none
+        refreshPauseMessage = "Refresh paused until \(formatter.string(from: until)) after \(consecutiveStalls) stalled attempts. Click retry to resume."
+    }
+
+    func clearRefreshPause() {
+        refreshPauseMessage = nil
     }
 
     private let loadingWatchdogSeconds: TimeInterval = 60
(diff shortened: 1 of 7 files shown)


dengsenyao:
Additional local hard-hang verification:

I also ran the debug menubar app with a temporary `HOME` and a `CODEBURN_BIN` fake CLI that forks a child process which keeps stdout/stderr pipes open after the parent CLI is terminated. This simulates the failure mode where the CLI subprocess timeout fires but the awaited pipe drain never reaches EOF.

Observed log sequence:

```text
2026-05-24 15:59:26 CodeBurn: CLI subprocess timed out after 45s ... terminating
2026-05-24 16:00:11 CodeBurn: force refresh stuck for 90s - cancelling and restarting
...
2026-05-24 16:03:41 CodeBurn: force refresh stuck for 90s - cancelling and restarting
...
2026-05-24 16:05:11 CodeBurn: force refresh stuck for 90s - cancelling and restarting
...
2026-05-24 16:06:41 CodeBurn: force refresh stuck for 90s - cancelling and restarting
2026-05-24 16:06:41 CodeBurn: force refresh paused after 3 consecutive stalls until 2026-05-24 08:07:11 +0000
```

After the pause log, the next 30s observation window did not start another blind force refresh. This verifies the new behavior for a stable local hang: repeated force-refresh watchdog trips are bounded and transition to pause/backoff instead of continuing indefinitely.

PR #398, opened 2026-05-25
fix: group Cowork sessions by space name and unify cross-provider project entries

### LINKED ISSUE #369: [Feature/Enhancement] Worktrees of same project are counted separate  in "By Project" usage whereas they should be shown as same
The issue is git worktrees are shown as separately in "By Project" usage
Git worktrees belong to same project therefore either data should be summed up for all worktrees or they should be grouped under same project and codeburn can show separate worktree usage like in present just that they are grouped under same project

## Option A
Project usage = sum of all worktrees

## Option B

Project usage
    - worktree A usage
    - worktree B usage

# Current behavior

Image

cfuselli:
## Summary

This PR fixes two related grouping bugs that caused sessions from the same project to appear as separate entries in the dashboard.

### Bug 1: Cowork space sessions grouped by directory name instead of space name

**Root cause:** Claude Code's "cowork" (shared workspace) feature stores sessions in `~/.claude/claude_desktop_sessions/<spaceId>/<sessionId>.jsonl`. The directory name is an opaque UUID, not a human-readable project name. Previously, codeburn used the directory name as the project label, so every space appeared as a UUID like `8a3f2b1c-...` and cowork sessions were never merged with regular project sessions.

**Fix:** Read `~/.claude/desktop_spaces.json` to resolve each space UUID to its human-readable name (`space.cwd`), then use `resolveCanonicalProjectPath` on that path so the resulting project entry matches the regular session entry for the same repo.

### Bug 2: Cross-provider sessions for the same repo appearing as separate projects

Codex, Warp, and other providers encode project paths differently from Claude Code. This caused sessions for the exact same git repository to appear as 3–6 separate rows in the "By Project" panel instead of one.

**Three sub-cases fixed:**

#### 2a. Leading-slash stripping (Codex)
Codex's `sanitizeProject` strips the leading `/` from absolute paths: `/Users/alice/MyProject` becomes `Users/alice/MyProject`. The new `crossProviderKey` normaliser strips leading slashes before comparing, so both forms map to the same project entry.

#### 2b. Git worktrees for non-Claude providers
PR #375 (closes #369) already fixed worktree grouping for Claude sessions. This PR extends the same fix to Codex and other providers whose project paths are not resolved at the session-file level. A pre-merge step (`resolveOtherProjectWorktrees`) restores the leading `/` on Codex paths and runs them through `resolveCanonicalProjectPath` so worktree sessions fold into the main repo entry.

**Before:** `worktrees/abc1/MyProject` and `worktrees/def4/MyProject` appeared separately for Codex sessions even after #375.
**After:** Both collapse into `MyProject`.

#### 2c. Subdirectory sessions (all providers)
If a session's `cwd` is a subdirectory of a git repo (e.g. `/Users/alice/MyProject/src/feature`), it was treated as a separate project. `resolveCanonicalProjectPath` now walks up the directory tree until it finds a `.git` directory, so it correctly identifies `/Users/alice/MyProject` as the repo root.

**Before/After Example:**

| Before (6 rows)                        | After (1 row)   |
|----------------------------------------|-----------------|
| `/Users/alice/MyProject`               | `MyProject`     |
| `Users/alice/MyProject` *(Codex)*      |                 |
| `worktrees/abc1/MyProject` *(worktree)*|                 |
| `worktrees/def4/MyProject` *(worktree)*|                 |
| `MyProject/src/feature` *(subdir)*     |                 |
| `8a3f2b1c-...` *(cowork UUID)*         |                 |

## Changes

- `src/providers/claude.ts` — resolve cowork space UUIDs to human-readable names via `desktop_spaces.json`
- `src/parser.ts` — rewrite `resolveCanonicalProjectPath` to walk up directory tree; add `crossProviderKey` normaliser and `resolveOtherProjectWorktrees` pre-merge step
- `src/session-cache.ts` — bump `claude` parse version to `cowork-space-grouping-v1` to invalidate stale cache entries
- `tests/parser-claude-cwd.test.ts` — 6 new tests covering cowork grouping, worktree resolution, subdirectory resolution, Windows paths, and cross-provider merging

`npm test` passes with 14 new/existing tests; 20 pre-existing failures on `main` are unrelated to this PR.

### DIFF
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -298,7 +298,7 @@ const PROJECT_COL_AVG = 7
 const PROJECT_COL_BASE_WIDTH = 30
 const PROJECT_COL_WITH_OVERHEAD_WIDTH = 40
 
-function ProjectBreakdown({ projects, pw, bw, budgets }: { projects: ProjectSummary[]; pw: number; bw: number; budgets?: Map<string, ContextBudget> }) {
+function ProjectBreakdown({ projects, pw, bw, budgets, rows = 14 }: { projects: ProjectSummary[]; pw: number; bw: number; budgets?: Map<string, ContextBudget>; rows?: number }) {
   const maxCost = Math.max(...projects.map(p => p.totalCostUSD))
   const hasBudgets = budgets && budgets.size > 0
   const nw = Math.max(8, pw - bw - (hasBudgets ? PROJECT_COL_WITH_OVERHEAD_WIDTH : PROJECT_COL_BASE_WIDTH))
@@ -307,7 +307,7 @@ function ProjectBreakdown({ projects, pw, bw, budgets }: { projects: ProjectSumm
       <Text dimColor wrap="truncate-end">
         {''.padEnd(bw + 1 + nw)}{'cost'.padStart(8)}{'avg/s'.padStart(PROJECT_COL_AVG)}{'sess'.padStart(6)}{hasBudgets ? 'overhead'.padStart(10) : ''}
       </Text>
-      {projects.slice(0, 8).map((project, i) => {
+      {projects.slice(0, rows).map((project, i) => {
         const budget = budgets?.get(project.project)
         const avgCost = project.sessions.length > 0
           ? formatCost(project.totalCostUSD / project.sessions.length)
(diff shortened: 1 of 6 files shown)


iamtoruk (commented):
Good PR overall, solid tests (6 new, all pass), real problem solved. A few things to address:

### 1. Test failure (blocker)

`tests/providers/claude-config-dirs.test.ts` — "merges by sanitized slug even when sessions carry different canonical cwds" fails because the merge key changed from `p.project` (slug) to `crossProviderKey(p)` (normalized path). Two sessions with the same slug but different cwds now correctly separate instead of merging. The test was explicitly pinning the old behavior (see the comment at line 232-239). Since the PR intentionally changes this, update the test to expect 2 projects instead of 1.

```
npx vitest run tests/providers/claude-config-dirs.test.ts
# 1 failed: expected length 1 but got 2
```

### 2. Duplicated `getDesktopSessionsDir()` (should fix)

Identical 6-line function in both `src/providers/claude.ts:62` and `src/parser.ts:57`. Export it from `claude.ts` and import in `parser.ts`, or move to a shared module.

### 3. Dashboard `rows` values don't match the PR description (nit)

PR description says "14 for week/6-months, 31 for month/30-days, 20 for all-time, min 8 for today" but the code gives:
- `dayMode` → 8 ✓
- `all` → 14 (description says 20)
- `month`/`30days` → 14 (description says 31)
- `week` → 8 (description says 14)

Not a bug per se, but the description should match the code. 14 for everything non-day is probably fine.

### 4. `crossProviderKey` lowercases project names (minor)

```js
return path.includes('/') ? path : p.project.toLowerCase()
```

On case-sensitive filesystems (Linux), `/Users/alice/Foo` and `/Users/alice/foo` are different repos. Lowercasing could theoretically false-merge them. Low risk in practice but worth noting.

### 5. `spacesJsonCache` is never cleared (minor)

Module-level cache persists for the process lifetime. Fine for CLI one-shot commands but could serve stale data in the TUI dashboard during long sessions. Consider clearing on each `parseAllSessions` call, or adding a TTL.

iamtoruk:
Hey, nice work on this — Cowork grouping and cross-provider merging are real pain points.

One test is failing on our end:

```
tests/providers/claude-config-dirs.test.ts
  "merges by sanitized slug even when sessions carry different canonical cwds"
  expected length 1 but got 2
```

This is expected given your merge-key change (slug → path-based), but the test needs updating. That test was explicitly pinning the old slug-merge behavior — since you're intentionally changing it, flip the assertion to expect 2 separate projects.

Also `getDesktopSessionsDir()` is duplicated in both `claude.ts` and `parser.ts` — please extract to a shared spot.

Fix those two and this is good to merge. The dashboard rows and lowercase key nits from the review are optional.

cfuselli:
Thanks @iamtoruk, should be fixed now!

PR #399, opened 2026-05-26
fix(cli): restore antigravity-hook command

Leezj9671:
In commit 5800179, the Commander configuration for `antigravity-hook` and `agy-statusline-hook` commands in `src/main.ts` was accidentally deleted during conflict resolution. This commit restores them along with the missing imports.


iamtoruk (approved):
Verified: clean restoration of antigravity-hook commands accidentally deleted in #393 (commit 5800179). No modifications, no new deps.

PR #402, opened 2026-05-26
Add live Claude subscription quota command

ozymandiashh:
## Summary

Adds a new `codeburn quota` CLI command that surfaces the live 5-hour, weekly, weekly-Opus, and weekly-Sonnet utilisation Anthropic exposes through the OAuth usage endpoint. This is the same data the `claude` CLI itself renders in `/usage`, so it reflects the real subscription quota rather than CodeBurn's API-equivalent USD estimate against a static monthly budget.

The CLI reads the access token Claude Code already manages locally — `~/.claude/.credentials.json` on Linux/Windows and the macOS Keychain on macOS — refreshes it transparently when needed, and never asks the user for an API key. No proxying. No remote state.

The macOS menubar already does this via `mac/Sources/CodeBurnMenubar/Data/ClaudeSubscriptionService.swift` and `ClaudeCredentialStore.swift`. This PR ports that behaviour to the TypeScript CLI.

## Why

The existing `plan` / `plan-usage` path tells users how much they have spent in API-equivalent dollars against a configured monthly budget. It is useful for cost attribution but does not answer the question users actually ask first: "how close am I to my 5-hour rate limit?" or "how much of my weekly Sonnet quota is left?"

That answer is one HTTP request away. Anthropic returns it for every active Claude subscription. The macOS menubar already wires it up. Bringing it to the CLI closes the gap between the two surfaces and gives Linux / Windows users the same insight as the macOS menubar app.

### `src/claude-credentials.ts` (new)

Reads Claude OAuth credentials from the same two sources Claude Code itself writes to:

- `~/.claude/.credentials.json` on Linux / Windows / newer Claude Code installs
- macOS Keychain via `/usr/bin/security find-generic-password`, with the `\$USER` and unscoped account fallback that the Swift implementation already documents

Bootstraps once into our own cache at `~/.cache/codeburn/claude-credentials.v1.json` (mode `0600`, written atomically through `open(path, 'w', 0o600)` + `rename`), so we never re-prompt the macOS Keychain after the first read and we never write back into Claude's own credential store.

Refresh path:

- Proactive refresh when the current token is within five minutes of expiry. Transient refresh failures (429, 408, 5xx, network errors) fall back to the still-valid current token rather than crashing the caller.
- Explicit refresh after a 401 from a Claude-OAuth-protected endpoint. Persists rotated refresh tokens, and re-reads our cache file immediately before the HTTP call so a sibling process (another `codeburn` invocation, the menubar app) that just rotated the refresh token wins the race instead of us replaying the now-invalid token.
- `rebootstrapClaudeCredentials()` deletes our cache and re-reads the source. The quota layer calls this when the cached refresh token looks terminally dead, so a user who just re-ran `claude /login` picks up the new credentials without having to delete `~/.cache/codeburn/claude-credentials.v1.json` by hand.
- `ClaudeCredentialError.isTerminal` treats 429 and 408 from `platform.claude.com/v1/oauth/token` as transient, not as "reconnect required".

### `src/claude-quota.ts` (new)

Calls `GET https://api.anthropic.com/api/oauth/usage` with the OAuth access token, the `anthropic-beta: oauth-2025-04-20` header, and the `claude-code/2.1.0` User-Agent — same headers the Swift implementation uses.

- 200 → `SubscriptionQuota { tier, fiveHour, sevenDay, sevenDayOpus, sevenDaySonnet, fetchedAt }`. `utilization` is returned by Anthropic on a percent scale already (0..100, can exceed 100 when over-limit) so it is kept end-to-end without re-multiplication.
- 401 → one refresh + retry; a second 401 is treated as terminal so callers stop looping.
- 429 → persists the backoff window to `~/.cache/codeburn/claude-quota-backoff.json`. Honours the HTTP `Retry-After` header first (delta-seconds and HTTP-date forms both supported) and falls back to the `retry_after` JSON body field. Floored at 60 seconds.
- 5xx / other → `ClaudeQuotaError` with `code: 'http'`.

Returns `null` when no Claude credentials are present locally, so consumers can degrade silently.

### `src/main.ts`

New `codeburn quota` command:

- `codeburn quota` — text output with the four quota windows and reset timestamps.
- `codeburn quota --format json` — stable `SubscriptionQuotaJSON` shape suitable for menubar consumption or scripting.
- Exits non-zero when no credentials are present, when fetch fails, or when Anthropic is rate-limiting the endpoint, so it composes cleanly with shell scripts and CI checks.

Example:

```text
\$ codeburn quota
  Max 20x quota
  Fetched: 2026-05-26T11:52:42.578Z
  5-hour              3.0% (resets 2026-05-26T16:00:00Z)
  7-day              89.0% (resets 2026-05-26T15:00:00Z)
  7-day (Opus)     -
  7-day (Sonnet)      3.0% (resets 2026-05-26T15:00:00Z)
```

## Tests added

- `tests/claude-credentials.test.ts` — 23 tests covering blob sanitisation (Claude Code's mid-token line wraps), parser failure modes, source bootstrap vs cache preference, proactive refresh persistence, rotated refresh token persistence, terminal vs transient `isTerminal` classification (including 408 / 429 carve-out), transient fallback to current token, and `rebootstrapClaudeCredentials` re-reading the source.
- `tests/claude-quota.test.ts` — 13 tests covering tier classification, success mapping, 429 backoff persistence and short-circuit, 60-second floor on tiny `retry_after` values, 401-then-200 recovery with backoff clear, terminal 401-after-refresh, generic HTTP error surfacing, decode failure handling, `Retry-After` header preference over JSON body (delta-seconds and HTTP-date), and stable `SubscriptionQuotaJSON` shape.

## Validation

- [x] \`npm run build\` - passed.
- [x] \`npx vitest run\` - 70 files / 1031 tests passing.
- [x] \`git diff --check\` - passed.
- [x] Live end-to-end smoke against a real Claude Max 20x account on macOS (Keychain path): `codeburn quota` and `codeburn quota --format json` return the same values as `claude /usage`.
- [x] Gemini 3.1 Pro Preview review (via \`agy\`) — applied. Surfaced the stale-cache lock-in, concurrent refresh replay risk, 429-as-terminal misclassification, missing \`Retry-After\` header parsing, exit-code on missing credentials, and insecure two-step \`writeFile\`+\`chmod\`; all addressed in the current diff.
- [x] Claude Opus 4.7 effort max review — applied.

## Notes

- macOS Keychain access shells out to \`/usr/bin/security\`. The CLI does not link against the Security framework.
- The \`anthropic-beta: oauth-2025-04-20\` endpoint is not formally documented but is the same one \`claude\` CLI itself uses; the Swift menubar code has been calling it in production. If Anthropic changes the contract, the CLI degrades gracefully (\`code: 'decode'\` or \`code: 'http'\`) rather than producing wrong numbers.
- Out of scope here: wiring \`SubscriptionQuota\` into \`status --format menubar-json\`, the dashboard TUI, and the \`plan-usage\` overage rows. These are clean follow-ups now that the data layer is stable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 ## Unreleased
 
 ### Added (CLI)
+- **Live Claude quota.** New `codeburn quota` command (and `--format json`)
+  shows the real 5-hour, weekly, weekly-Opus and weekly-Sonnet utilisation
+  from Anthropic's OAuth usage endpoint, using the access token the local
+  `claude` CLI already manages. Reads credentials from
+  `~/.claude/.credentials.json` on Linux/Windows and the macOS Keychain on
+  macOS; refreshes the access token transparently on 401 and persists the
+  rotated token to `~/.cache/codeburn/`. 429 backoff is persisted so back-to-
+  back invocations honour Anthropic's retry window. Thanks @ozymandiashh.
 - **MCP project profile advisor.** `codeburn optimize` now flags MCP servers
   that are useful in one project but loaded into other projects where they are
   never invoked, with a project-scoping prompt that preserves the hot workflow
--- a/src/claude-credentials.ts
+++ b/src/claude-credentials.ts
@@ -1,0 +1,456 @@
+import { execFile } from 'child_process'
+import { randomBytes } from 'crypto'
+import { existsSync } from 'fs'
+import { mkdir, open, readFile, rename, unlink } from 'fs/promises'
+import { homedir, userInfo } from 'os'
+import { join } from 'path'
+import { promisify } from 'util'
+
+const execFileAsync = promisify(execFile)
+
+/**
+ * Reads Claude Code's OAuth credentials from local sources and refreshes them
+ * when needed. Mirrors the behaviour of `mac/Sources/CodeBurnMenubar/Data/ClaudeCredentialStore.swift`
+ * but stays cross-platform: the Swift version reads the macOS Keychain via
+ * the Security framework; this TypeScript port reads either the standard
+ * `~/.claude/.credentials.json` file (Linux / Windows / newer Claude Code
+ * installs) or shells out to `/usr/bin/security` on macOS.
+ *
+ * Refreshed tokens are persisted to our own cache file
+ * (`~/.cache/codeburn/claude-credentials.v1.json`) so that we do not rotate
+ * the Claude CLI's credentials. The cache file is written with mode 0600.
+ */
+
+const OAUTH_REFRESH_URL = 'https://platform.claude.com/v1/oauth/token'
+const OAUTH_CLIENT_ID = '9d1c250a-e61b-44d9-88ed-5944d1962f5e'
+const CLAUDE_CREDENTIALS_RELATIVE_PATH = '.claude/.credentials.json'
+const KEYCHAIN_SERVICE = 'Claude Code-credentials'
+const MAX_CREDENTIAL_BYTES = 64 * 1024
+const PROACTIVE_REFRESH_MARGIN_MS = 5 * 60 * 1000
+const CACHE_FILENAME = 'claude-credentials.v1.json'
+
+export type CredentialRecord = {
+  accessToken: string
+  refreshToken: string | null
+  expiresAt: Date | null
+  rateLimitTier: string | null
+}
+
+export type ClaudeCredentialErrorCode =
+  | 'no_source'
+  | 'decode_failed'
+  | 'no_refresh_token'
+  | 'refresh_http'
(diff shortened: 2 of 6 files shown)


iamtoruk:
Thanks for this. The credential handling and backoff logic are well done.

The menubar already surfaces live quota (5h, 7d, 7d-Opus, 7d-Sonnet) from the same endpoint, so a standalone CLI command duplicates what's shipping today. We're also working on something bigger: attributing quota consumption to individual sessions and tasks using local call data, which would make the raw numbers more actionable.

Closing this for now, but the credential/refresh infrastructure here is solid and we may pull from it when the attribution feature lands.

Issue #383, opened 2026-05-23
Add support to Windsurf


rotirotirafa:
Should be nice to have windsurf as option..

May i can actuate on this?

rkranjith79:
Could you pls do this

rotirotirafa:
Update on Windsurf provider implementation

I spent time today digging into how Windsurf stores session data locally so I could implement a proper provider for codeburn. Unfortunately I hit a wall and wanted to share what I found.

What I checked

~/Library/Application Support/Windsurf/User/globalStorage/state.vscdb (macOS) — only contains UI state, no cursorDiskKV table or session keys.
workspaceStorage/*/state.vscdb — has empty agentSessions.* and chat.ChatSessionStore.index keys.
~/.codeium/windsurf/cascade/*.pb — these are the actual conversation files, but they are encrypted protobuf (entropy ~7.8/8.0, no readable strings). No documented decryption method.
The local language_server process uses stdin/stdout LSP with the Windsurf parent process; no HTTP RPC endpoint like Antigravity exposes.
The Codeium web API (server.codeium.com) is Enterprise-only and requires a service_key.
Conclusion There is currently no accessible local source for Windsurf session data. Unlike Cursor (SQLite) or Claude (JSONL), Windsurf keeps everything encrypted.

What I did instead I kept the provider registered as a placeholder — it returns empty sessions with a clear note, but keeps the model/tool display-name maps ready. If Windsurf ever exposes local data (SQLite, JSONL, or an RPC endpoint), the provider will work immediately.

If anyone knows of another local data source I missed, I'm happy to revisit this

PR #407, opened 2026-05-27
feat(windsurf): register as placeholder provider with docs and tests

rotirotirafa:
Windsurf stores session data in encrypted .pb files and does not expose a local API. After investigating SQLite, RPC, and web API options, no accessible data source was found.

The provider is kept as a stub so it will work automatically once Windsurf exposes local session data.

- src/providers/windsurf.ts: minimal stub with display-name maps
- tests/providers/windsurf.test.ts: display name + empty session tests
- tests/provider-registry.test.ts: lazy-load + display name asserts
- docs/providers/windsurf.md: documents placeholder status
- docs/providers/README.md + README.md: add Windsurf to index

## Testing

- [x] `npm test` passes

### For new providers only:

- [] Screenshot or terminal output attached below proving it works with real data

image


iamtoruk:
Closing this placeholder Windsurf provider. Investigation (see #383) found Windsurf stores its conversations as encrypted protobuf with no accessible local token/usage source, so a registered provider can't actually read anything yet. If a readable data location surfaces, a real provider PR would be great, tracking the request in #383.

PR #415, opened 2026-05-29
feat(kiro): add Kiro CLI session support

leonardo-claudio:
## Summary

Extends the existing Kiro provider to read sessions from **Kiro CLI** (`~/.kiro/sessions/cli/`), in addition to the existing Kiro IDE support.

## Changes

- **Discovery:** scans `~/.kiro/sessions/cli/*.jsonl` for CLI sessions
- **Parser:** new JSONL parser for the CLI format (`Prompt`, `AssistantMessage`, `ToolResults` entries)
- **MCP detection:** tools not in the built-in set are prefixed as `mcp__<server>__<tool>` (server name read from `~/.kiro/settings/mcp.json`)
- **Project resolution:** uses `cwd` from session metadata
- **Tests:** 6 new tests covering CLI parsing, tool detection, MCP, dedup, and project naming
- **Docs:** updated `docs/providers/kiro.md`

## How Kiro CLI stores data

```
~/.kiro/sessions/cli/
├── <session-id>.json   # metadata (session_id, cwd, timestamps, model)
└── <session-id>.jsonl  # conversation (line-delimited Prompt/AssistantMessage/ToolResults)
```

## Testing

```bash
npx vitest run tests/providers/kiro.test.ts  # 38 tests pass (32 existing + 6 new)
npx tsc --noEmit                              # clean
```

### DIFF
--- a/docs/providers/kiro.md
+++ b/docs/providers/kiro.md
@@ -1,12 +1,14 @@
 # Kiro
 
-Kiro IDE chat history.
+Kiro IDE and CLI chat history.
 
 - **Source:** `src/providers/kiro.ts`
 - **Loading:** eager (`src/providers/index.ts:7`)
 - **Test:** `tests/providers/kiro.test.ts`
 
 ## Where it reads from
+
+### Kiro IDE
 
 VS Code-style globalStorage at `kiro.kiroagent`:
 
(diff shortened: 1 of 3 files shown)


iamtoruk:
Thanks for putting this together, and sorry for the slow turnaround. Timing worked against this one: while it was open, Kiro CLI support landed on main through #502, reading the same ~/.kiro/sessions/cli files.

The version on main takes a different approach in a few places we need to keep:

- it reads the real per-turn credit cost from user_turn_metadatas[].metering_usage and only estimates when that is missing, instead of always char-estimating at list rates
- it emits one call per turn with the turn's own end timestamp, so multi-day sessions land in the right day buckets
- it counts ToolResults entries as input context, which is often the largest share of a coding session

Because of that overlap the branch now conflicts with main, and merging it would regress cost accuracy, so I am closing it as superseded. One piece of yours is still worth landing: the extra tool-name mappings (code, subagent, and the fuller builtin set). If you want to send a small follow-up adding those to the current CLI parser, I would be glad to merge it. Appreciate the work here and hope to see more from you.

leonardo-claudio:
Thanks for the detailed feedback! Makes total sense — the approach in #502 is definitely more accurate with the per-turn metering. I've opened #624 as a small follow-up with just the tool-name mappings (`code`, `subagent`, `web_fetch`) and the `mcp__` passthrough. Let me know if anything needs adjusting.

PR #424, opened 2026-06-02
feat: add MiniMax-M3 model pricing support

octo-patch:
## Summary

Adds `MiniMax-M3` (released after M2.7) to `MANUAL_ENTRIES` in `scripts/bundle-litellm.mjs` and the `MiniMax M3` display name in `getShortModelName`. Snapshot regenerated via `npm run bundle-litellm`. M2.7 / M2.7-highspeed entries kept unchanged for back-compat with older sessions.

This follows the same pattern as #120, which added M2.7 / M2.7-highspeed pricing.

## Pricing

From [MiniMax paygo pricing](https://platform.minimax.io/docs/guides/pricing-paygo) (<=512K context):

| Model | Input | Output | Cache Write | Cache Read |
|---|---|---|---|---|
| MiniMax-M3 | $0.6/M | $2.4/M | n/a (0) | $0.12/M |

`cacheWriteCostPerToken` is pinned to `0` rather than `null` so the snapshot loader's `input * 1.25` fallback in `loadSnapshot` does not silently kick in for a model that does not actually bill for cache writes on paygo.

## Test plan

- [x] `npx vitest run tests/minimax.test.ts tests/models-hoist.test.ts tests/models.test.ts tests/models-report.test.ts` -> 138/138 pass (4 new M3 tests + existing 134 stay green)
- [x] `node scripts/bundle-litellm.mjs` regenerates `src/data/litellm-snapshot.json` with the M3 entry; diff vs. previous snapshot is exactly one added key (`MiniMax-M3`)
- [x] `npx tsc --noEmit` passes
- [x] Smoke check: `getModelCosts('MiniMax-M3')` returns the expected per-token rates and `getShortModelName('MiniMax-M3-20260101')` correctly strips the date suffix to `MiniMax M3`

## References

- Pricing page: https://platform.minimax.io/docs/guides/pricing-paygo
- Prior precedent: #120 (M2.7 / M2.7-highspeed)


iamtoruk:
Closing in favor of #457, which makes pricing automatic instead of hardcoded: it gap-fills from models.dev first-party makers and OpenRouter. MiniMax-M3 now resolves to the same $0.6/$2.4 official price with no per-model code, so it stays correct as new models ship. Thanks for flagging the gap.

PR #432, opened 2026-06-03
fix(cursor): period-aligned lookback; add Vercel AI Gateway provider

sisodiabhumca:
## Summary

- **Cursor:** Replace the fixed 35-day SQLite lookback with a period-aware time floor (`getCursorTimeFloor`) aligned to the active report range, capped at six months (same as CLI `all`). Bump `cursor-results` cache to v4 and store `lookbackFloor` so switching from e.g. week → month re-queries instead of serving a too-narrow cached slice.
- **Vercel AI Gateway:** Add optional `vercel-gateway` provider that reads usage from the [Custom Reporting API](https://vercel.com/docs/ai-gateway/capabilities/custom-reporting) when `AI_GATEWAY_API_KEY` or `VERCEL_OIDC_TOKEN` is set; skipped silently in the combined dashboard without credentials.

## Motivation

Users on Cursor saw **$0** for month/all reports because the DB query only scanned ~35 days while bubbles outside that window were dropped. Period-aligned lookback fixes month/week/all totals without scanning the entire multi-hundred-MB DB on every run.

Vercel gateway usage is cloud/API-based (not local logs); this complements existing IDE providers for teams routing models through Vercel AI Gateway.

## Test plan

- [x] `npm test` (1006 tests pass locally)
- [ ] `codeburn report --provider cursor -p month` shows non-zero usage when `state.vscdb` has bubbles in range
- [ ] `codeburn report --provider cursor -p week` then `-p month` — month total ≥ week (cache invalidates on wider floor)
- [ ] `codeburn report --provider vercel-gateway -p month` with `AI_GATEWAY_API_KEY` set returns API usage (or skips when unset)


iamtoruk (changes requested):
Good idea fixing the cursor lookback — that $0 on month/all was a real problem. And the Vercel gateway is a clean addition. A few things though:

**Vercel gateway silently produces zero results in production**

This is the big one. `discoverSessions()` returns `{ path: 'vercel-ai-gateway:report', ... }` — a virtual path. But `parseProviderSources` in parser.ts fingerprints every source with `stat()` on the path. `stat('vercel-ai-gateway:report')` fails → source gets skipped → `parse()` never runs. The unit test passes because it calls `createSessionParser().parse()` directly, bypassing that integration path. So the provider works in tests but does nothing in the real CLI.

Need to either skip fingerprinting for non-file-based providers, or have the Vercel gateway return a path that actually exists on disk.

**`getCursorTimeFloor` month overflow on the 31st**

`new Date(year, month - 6, day)` with `day = 31` and the target month having fewer days rolls into the next month. On August 31st, 6 months back gives you March 3 instead of February 28. Meanwhile `cli-date.ts` uses `day = 1` for the "all" period. So the CLI asks for data from Feb 1 but the cursor floor clips at March 3 — drops ~30 days silently. Use `day = 1` to match cli-date.ts.

**No fetch timeout on the Vercel API call**

This is the first provider that makes network requests. If the Vercel API hangs (misconfigured proxy, VPN, DNS drop), the entire codeburn run blocks forever. Add an `AbortSignal.timeout(15000)` or similar — the antigravity provider already has explicit timeouts for the same reason.

**Network errors crash the provider chain**

`fetchVercelGatewayReport` handles HTTP error responses (non-ok status) but doesn't catch `fetch()` itself throwing (DNS failure, connection refused). That exception propagates up through the parser and could abort parsing for remaining providers. Wrap the whole fetch in try/catch → return `[]`.

**Error response body written to stderr could leak info**

The error handler writes `detail.slice(0, 200)` from the raw Vercel response to stderr. If the API includes the bearer token or internal details in error responses, that ends up in CI logs. Just log the HTTP status code and a static message instead.

**Smaller stuff:**

- Test doesn't save/restore `VERCEL_OIDC_TOKEN` — the "empty discovery" test deletes it but afterEach doesn't put it back. Would leak env in CI if set.
- Cursor test only covers the happy path of `getCursorTimeFloor` (start within 6 months). Missing: start exceeding the cap, undefined dateRange, and the cache invalidation path.
- Drive-by Codex provider note in the README isn't related to this PR — should be a separate commit.
- Two features in one PR makes revert harder if one has issues, but not a blocker if the above are fixed.

Cache invalidation logic looks correct — the `lookbackFloor <= requestedFloor` string comparison works because ISO timestamps sort lexicographically. SSRF surface is clean — URL is hardcoded, params built from Date getters through URLSearchParams.

PR #433, opened 2026-06-03
feat: add JetBrains (IntelliJ/DataGrip) session discovery and parsing

zuberikea:
## Summary

Adds support for discovering and parsing GitHub Copilot chat sessions from JetBrains IDEs (IntelliJ IDEA, DataGrip, WebStorm, etc.).

## Problem

The Copilot provider only discovered sessions from:
- `~/.copilot/session-state/` (legacy VS Code format)
- VS Code workspace storage directories

Users of GitHub Copilot in JetBrains IDEs saw no data when running `codeburn`, even though sessions were being stored at `~/.copilot/jb/<session-id>/partition-*.jsonl`.

## Solution

- **`discoverJetBrainsSessions()`** - finds all `.jsonl` files under `~/.copilot/jb/`
- **`isJetBrainsFormat()`** - detects the JB event format (starts with `user.message_rendered` / `user.message` instead of `session.start`)
- **`parseJetBrainsEvents()`** - parses JB-specific events:
  - `assistant.message` with `text`/`content`/`thinking` fields
  - `tool.execution_start` for tool name extraction
  - Tool call ID prefixes (`toolu_vrtx_`, `call_`, etc.) for model inference
- **`inferJBProjectName()`** - extracts project name from file paths in tool execution arguments

## Testing

Tested locally with 105 JetBrains Copilot sessions. Before this change: 0 sessions detected. After: all sessions parsed correctly with cost, token counts, tool usage, and model detection.

### DIFF
--- a/src/providers/copilot.ts
+++ b/src/providers/copilot.ts
@@ -1,6 +1,6 @@
 import { existsSync } from 'fs'
 import { readdir, readFile, stat } from 'fs/promises'
-import { basename, dirname, join, posix, win32 } from 'path'
+import { basename, dirname, join, posix, sep, win32 } from 'path'
 import { homedir } from 'os'
 
 import { readSessionFile } from '../fs-utils.js'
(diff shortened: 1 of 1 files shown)


iamtoruk (changes requested):
Nice work getting JetBrains Copilot discovery in — been wanting this. A few things to fix before merging:

**`sep` not imported (breaks project inference entirely)**

`inferJBProjectName` uses `homedir().split(sep)` but `sep` isn't in the path import on line 3. This throws a ReferenceError that gets swallowed by the try/catch, so the function silently returns `null` every time. All JB sessions end up with the raw UUID as the project name instead of the actual repo name. Add `sep` to the import.

**Dedup key collision when messageId is missing**

The key is `copilot:jb:${sessionId}:${messageId}:${iterationNumber}`. If both `data.messageId` and `e.id` are undefined (which they will be for some event shapes), every such message gets the key `copilot:jb:<uuid>::0`. Only the first one survives dedup — the rest get dropped. This undercounts tokens and cost. Consider using an incrementing index as fallback when messageId is absent.

**`inferJBProjectName` reads the whole file, bypasses the size cap**

It calls `readFile()` directly instead of going through `readSessionFile()`, which has the 128MB safety guard. It only needs 100 lines but loads the entire file into memory. Then the same file gets read again during parsing. Use `readSessionFile` or better yet defer project inference to parse time when the content is already loaded.

**`isJetBrainsFormat` can false-positive on legacy files**

Matching on `user.message` as first event type is too broad — legacy files can start with that too. The dispatch order (transcript → JB → legacy fallback) means a legacy file starting with `user.message` would get routed to the JB parser and misparsed. Tighten this to only match on `user.message_rendered` or `partition.created`, which are JB-specific.

**Model inference duplicated, missing the `data.model` check**

The inline model inference reimplements `inferModelFromToolCallIds` but skips the explicit `data.model` field lookup that the transcript parser uses (weighted 100x). Worth reusing the existing function or at least adding the `data.model` check so JB sessions get the same quality of model detection.

**Dead first pass**

The "first pass" loop that sets `pendingUserMessage` is never read — the second pass uses its own `userMsg` variable. Can just delete it.

Also minor: no override param for `jbDir` in `createCopilotProvider`, so there's no way to test JB discovery with fixtures like the legacy/vscode paths support.

iamtoruk (approved):
All review feedback addressed cleanly: sep imported (crash fixed), dedup has an incrementing fallback, project inference reuses the already-loaded capped content (no double read / no cap bypass), isJetBrainsFormat no longer matches bare user.message, model inference unified into inferModelFromEvents with the data.model weighting, and the dead first pass is gone. tsc clean and the 25 copilot tests pass. Merging. I will add JetBrains test coverage and a small Windows path-split fix as a follow-up from our side.

PR #251, opened 2026-05-06
feat(accounts): add account control center

ozymandiashh:
## Summary

Rebases and updates the Account Control Center on current `main` while preserving the current CLI split: `src/cli.ts` remains the Node-version-safe launcher, and the account workflow lives in `src/main.ts`.

This PR builds on the multiple Claude config directory work and turns it into a user-facing attribution workflow for people who run work, personal, desktop, or client Claude accounts side by side.

It answers questions that aggregate cost views cannot answer:

- Which account generated this spend?
- Which projects and sessions drove it?
- Did a work/client project run on the wrong account?
- Is a configured subscription underused or over budget?
- Can usage be exported per account for reimbursement or accountability?

## What changed

- Claude discovery carries account metadata from `CLAUDE_CONFIG_DIRS` through project/session parsing.
- Parser and cache paths preserve `account`, `accountPath`, and `sourcePath`, so the same project path under two Claude accounts does not merge into one summary.
- New `codeburn accounts` command with text and JSON output.
- Account config stored in CodeBurn config:
  - `plan`
  - `monthlyUsd`
  - `budgetUsd`
  - `resetDay`
- `--account` filtering added to:
  - `report`
  - `today`
  - `month`
  - `status`
  - `export`
  - `accounts`
  - `optimize`
  - `compare`
- JSON reports include account rollups and account risks.
- CSV/JSON exports include account rows and account metadata.
- Dashboard and compare paths respect account filters while keeping current custom range and multi-plan behavior.

## Examples

```bash
codeburn accounts
codeburn accounts work
codeburn accounts set work --plan "Claude Max" --monthly-usd 100 --budget-usd 300
codeburn report --account work --format json
codeburn export --account client-a -f csv
```

A project present under both `work` and `personal` account roots is now reported separately instead of being silently merged. Unlabelled provider usage remains grouped as `unlabelled`, but it does not create duplicate-project noise against labelled Claude accounts.

## Account Signals

The account report surfaces:

- total cost and calls,
- sessions and project count,
- edit turns and one-shot turns,
- cost per session and cost per edit,
- top models and top projects,
- subscription utilization vs configured monthly price,
- budget utilization vs configured budget.

Risk signals include:

- duplicate project path across labelled accounts,
- work/client-looking paths on personal accounts,
- personal-looking paths on work/client accounts,
- underused subscription,
- over-budget account.

## Review Fixes

Pre-push review found and fixed these issues before the branch was updated:

[...]

PR #436, opened 2026-06-03
feat(menubar): configure CLAUDE_CONFIG_DIRS via Settings

### LINKED ISSUE #434: Add a Settings UI in the macOS menubar app to configure CLAUDE_CONFIG_DIRS
## Proposal

I'd like to add a small Settings UI to the macOS menubar app (`mac/`) that lets
users configure `CLAUDE_CONFIG_DIRS` so the menubar can aggregate usage across
multiple Claude config directories (e.g. work / personal accounts).

This builds directly on the CLI support merged in #288 (`getClaudeConfigDirs()`
in `src/providers/claude.ts`). Today that env var is honored by the CLI, but the
menubar is a GUI app that doesn't inherit the user's shell environment, so there
is no way to set it from the app.

## Planned approach

- A new "Config Directories" section under the existing **Claude** tab in
  Settings: a simple add/remove list of directory paths (with a folder picker).
- The list is persisted in `UserDefaults` and injected as `CLAUDE_CONFIG_DIRS`
  into every `codeburn` subprocess the app spawns (the existing
  `CodeburnCLI.makeProcess` path, plus the Terminal-launched `report`/`optimize`
  commands, with proper shell/AppleScript escaping).
- Empty list = current default behavior (`~/.claude`).

## Scope / non-goals

- This only affects **usage aggregation** (the data the CLI parses).
- Live quota / the "Connect" flow still reads the default account's credentials
  (Keychain / `~/.claude`); per-account live quota is intentionally out of scope.

## Relationship to #251

I noticed #251 (account control center) is in progress on the CLI side and
introduces a richer per-account attribution model. The change I'm describing is
menubar-only and doesn't touch the CLI data model, so it's compatible with both
the current merge behavior and #251's direction. But since it sits in the same
feature area, I wanted to check the preferred UX direction before opening a PR —
e.g. should the menubar UI eventually align with the account model from #251, or
is a minimal `CLAUDE_CONFIG_DIRS` editor acceptable as a first step?

Happy to adjust the approach based on your guidance. Thanks!

iamtoruk:
Closes #434.

## Summary

Adds a **Config Directories** section under the Claude tab in the macOS menubar Settings, so users can aggregate usage across multiple Claude config directories (e.g. work / personal accounts) from the GUI. The menubar is an accessory app that doesn't inherit the user's shell environment, so `CLAUDE_CONFIG_DIRS` was previously unreachable from the app.

## Approach: config file, not env injection

The original proposal in #434 was to inject `CLAUDE_CONFIG_DIRS` as an env var into every spawned subprocess, including the Terminal-launched `report`/`optimize` path. That path (`TerminalLauncher` → AppleScript `do script`) relies on a strict allowlist that excludes shell metacharacters — and arbitrary directory paths (`/Users/me/My $tuff/.claude`) can't satisfy it. Escaping arbitrary paths through AppleScript + shell is exactly the injection surface that code was written to close.

Instead, the list is persisted to the shared `~/.config/codeburn/config.json` (the same file the currency setting uses). The CLI reads it regardless of how it's launched, so **both** the menubar's data refresh **and** terminal-launched commands honor it consistently — with zero env injection into the shell/AppleScript paths.

## Changes

**CLI**
- `getClaudeConfigDirs()` now reads a `claudeConfigDirs` array from config.json as a fallback tier. Precedence: `CLAUDE_CONFIG_DIRS` env → `CLAUDE_CONFIG_DIR` env → **config.json** → `~/.claude`. Env still wins for per-shell overrides.
- `claudeConfigDirs?: string[]` added to the `CodeburnConfig` type.

**Menubar**
- `CLIClaudeConfig` mirrors `CLICurrencyConfig`'s flock-guarded write (shares the same lock, so a concurrent terminal `codeburn` write can't race the menubar).
- Settings UI: add/remove list with a native folder picker; forces a refresh on edit.

## Relationship to #251

This is intentionally the minimal "merge totals across dirs" approach (#208 option 1), preserving existing merge behavior. It does **not** add per-account attribution — that's #251's larger track. The config.json field is a clean seam #251's richer account model can build on; the #434 author explicitly designed this to be compatible with both.

## Testing

- 7 new tests in `tests/providers/claude-config-dirs.test.ts` (config.json precedence, `~` expansion, empty/non-array/non-string handling).
- Full suite: **1024 CLI tests + 31 Swift tests pass**. `tsc --noEmit` clean, both builds clean.

### DIFF
--- a/mac/Sources/CodeBurnMenubar/Data/CLIClaudeConfig.swift
+++ b/mac/Sources/CodeBurnMenubar/Data/CLIClaudeConfig.swift
@@ -1,0 +1,74 @@
+import Foundation
+
+/// Reads and writes the CLI's `claudeConfigDirs` list in
+/// `~/.config/codeburn/config.json`. The menubar is a GUI app that doesn't
+/// inherit the user's shell environment, so it can't rely on `CLAUDE_CONFIG_DIRS`
+/// to aggregate usage across multiple Claude config directories (work / personal
+/// accounts). Persisting to the shared config file instead means every `codeburn`
+/// invocation honors the list — whether the menubar spawns it directly or the
+/// user launches `codeburn report` in a terminal — without injecting environment
+/// variables through the shell/AppleScript paths (which only accept a strict
+/// metacharacter-free allowlist that arbitrary directory paths can't satisfy).
+///
+/// Shares the same on-disk flock as `CLICurrencyConfig` so a concurrent
+/// `codeburn` write from a terminal can't race the menubar and drop the other's
+/// changes (TOCTOU on config.json).
+enum CLIClaudeConfig {
+    private static var configDir: String {
+        (NSHomeDirectory() as NSString).appendingPathComponent(".config/codeburn")
+    }
+    private static var configPath: String {
+        (configDir as NSString).appendingPathComponent("config.json")
+    }
+    private static var lockPath: String {
+        (configDir as NSString).appendingPathComponent(".config.lock")
+    }
+
+    /// Returns the persisted config directories, or an empty array when none are
+    /// set (the CLI then falls back to `~/.claude`).
+    static func load() -> [String] {
+        guard
+            let data = try? SafeFile.read(from: configPath),
+            let json = try? JSONSerialization.jsonObject(with: data) as? [String: Any],
+            let dirs = json["claudeConfigDirs"] as? [Any]
+        else {
+            return []
+        }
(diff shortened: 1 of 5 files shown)


tvcsantos:
Just to give feedback on this. I pulled the test branch and i confirm it is working.

I would be nice to have a view for separte accounts. I guess that will be comming on #251 

Thanks!

PR #439, opened 2026-06-04
feat: add missing mac app icon

tvcsantos:
## Summary

On mac, the app was missing an icon. This PR adresses that issue and adds the missing icon, using the flame asset already in resources with a white background and no text. Just the flame.

Also added a small fix for linting issues by adding the missing types: ['node'] on tsconfig.

## Testing

- [x] `npm test` passes

### For new providers only:


iamtoruk:
Thanks for the icon. Landed it on main in #455, crediting you. The only thing I left out was the unrelated `tsconfig.json` `"types": ["node"]` change, since it is a project-wide change not needed for the icon (the project builds clean without it). Heads up that the same line is in your Devin PR #444 and should be dropped there too. Closing this in favor of #455.

PR #440, opened 2026-06-04
feat!: rename mac app name to CodeBurn

tvcsantos:
## Summary

I want to propose a rename of the mac app name to just CodeBurn. This PR addresses that. Also since this is a breaking change, when installing the app we also remove and kill the old one.

Note: intentionally left the process name with menubar, in order to keep cli and app processes with different names

## Testing

- [x] `npm test` passes

### For new providers only:


tvcsantos:
Hum okay thinking on this let us ignore this for now :)

PR #442, opened 2026-06-04
feat: add Hermes Agent provider

### LINKED ISSUE #368: support for Hermes-agent CLI

anthonyarmijo:
## Summary

Adds a Hermes Agent provider that reads session-level usage from Hermes SQLite state databases — both `~/.hermes/state.db` (default profile) and per-profile databases under `~/.hermes/profiles/<name>/state.db`.

**Supersedes #386. Closes #368.**

## What it does

- Discovers root and per-profile Hermes state databases
- Reads session-level token accounting (input, output, cache read/write, reasoning)
- Cost priority: stored `actual_cost_usd` > stored `estimated_cost_usd` > LiteLLM model pricing fallback
- Maps Hermes tools to CodeBurn's canonical tool names (terminal → Bash, delegate_task → Agent, browser_\* → Browser, etc.)
- Infers projects from `Current working directory:` lines in user/system messages
- Handles legacy databases that predate optional accounting columns

## Addresses @iamtoruk's review of PR #386

All five review items fixed:

| # | Issue | Fix |
|---|---|---|
| 1 | `??` doesn't trigger for `0` | Uses `> 0` checks for cost fallback priority |
| 2 | `inferProject` matches all message roles | Filters to `user` and `system` only |
| 3 | `CACHE_VERSION` bump unnecessary | Not bumped — uses `PROVIDER_PARSE_VERSIONS` entry only |
| 4 | Discovery query has no `LIMIT` | Added `LIMIT 10000` |
| 5 | `sanitizeProject` strips only one separator | Regex uses `/^[/\\\\]+/` quantifier |

Additional fixes beyond the review:
- `parseTimestamp` used falsy check `!raw` (rejects `0`) → changed to `raw == null`
- Added comment explaining reasoning tokens are priced at output rate (matching Gemini convention)
- `totalReasoningTokens` aggregated in `SessionSummary` via `parser.ts`

## Testing

- 9 fixture-based provider tests covering discovery, parsing, legacy schemas, cost fallback, tool mapping, profile isolation, and Windows path inference
- Provider registry test updated to include hermes
- Tested against real Hermes data: 557MB state.db with 198 sessions across 2 profiles

## Files changed

| File | Change |
|---|---|
| `src/providers/hermes.ts` | New — 460 lines |
| `tests/providers/hermes.test.ts` | New — 480 lines |
| `docs/providers/hermes.md` | New — provider documentation |
| `src/providers/index.ts` | +2 lines — import + coreProviders entry |
| `src/types.ts` | +1 line — `totalReasoningTokens` |
| `src/session-cache.ts` | +2 lines — `HERMES_HOME` env var + parse version |
| `src/parser.ts` | +6 lines — reasoning token aggregation + cost preservation |
| `docs/providers/README.md` | +1 line — Hermes row |
| `tests/provider-registry.test.ts` | +1 line — hermes in expected list |
| `CHANGELOG.md` | +7 lines — Unreleased entry |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Added (CLI)
+- **Hermes Agent provider.** Track token usage, cost, and tool breakdowns
+  for Hermes Agent sessions. Reads from `~/.hermes/state.db` and per-profile
+  databases. Supports session-level accounting with actual/estimated costs
+  from Hermes, falling back to CodeBurn's model pricing table. Supersedes
+  #386, closes #368.
+
 ### Fixed (CLI)
 - **Antigravity hook stale path repair.** `codeburn antigravity-hook install`
   now installs the statusLine command through a persistent `codeburn` binary
--- a/docs/providers/README.md
+++ b/docs/providers/README.md
@@ -16,6 +16,7 @@ For the architectural picture, see `../architecture.md`.
 | [Copilot](copilot.md) | JSONL | `src/providers/copilot.ts` | `tests/providers/copilot.test.ts` |
 | [Droid](droid.md) | JSONL | `src/providers/droid.ts` | `tests/providers/droid.test.ts` |
 | [Gemini](gemini.md) | JSON / JSONL | `src/providers/gemini.ts` | none |
+| [Hermes Agent](hermes.md) | SQLite | `src/providers/hermes.ts` | `tests/providers/hermes.test.ts` |
 | [IBM Bob](ibm-bob.md) | JSON | `src/providers/ibm-bob.ts` | `tests/providers/ibm-bob.test.ts` |
 | [KiloCode](kilo-code.md) | JSON | `src/providers/kilo-code.ts` | `tests/providers/kilo-code.test.ts` |
 | [Kiro](kiro.md) | JSON | `src/providers/kiro.ts` | `tests/providers/kiro.test.ts` |
--- a/docs/providers/hermes.md
+++ b/docs/providers/hermes.md
@@ -1,0 +1,67 @@
+# Hermes Agent
+
+Hermes Agent CLI profiles.
+
+- **Source:** `src/providers/hermes.ts`
+- **Loading:** eager (`src/providers/index.ts`)
+- **Test:** `tests/providers/hermes.test.ts`
+
+## Where it reads from
+
+| Source | Path |
+|---|---|
+| Default Hermes profile | `$HERMES_HOME/state.db` if set, otherwise `~/.hermes/state.db` |
+| Named Hermes profiles | `$HERMES_HOME/profiles/<profile>/state.db` |
+
+## Storage format
+
+SQLite. The provider reads Hermes' aggregate `sessions` token/cost counters and the matching `messages` rows for user prompt and tool-call context.
+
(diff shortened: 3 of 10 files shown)


iamtoruk (changes requested):
Thanks for this. Of the three open Hermes Agent PRs (this, #386, #370), this is the leanest and most convention-following, so it is the one I would like to move forward. The provider itself is well built: it follows the createXProvider + HERMES_HOME override convention, opens the DB read-only with parameterized queries, the `#hermes-session=` path scheme mirrors the existing cursor `#cursor-ws=` pattern, the session-cache registrations are conventional, and the tests use real sqlite fixtures. It builds clean and the suite passes on the branch. A few things to fix before merge, then it is in good shape.

## Must fix
1. **Do not swallow SQLite busy errors (data loss).** `parseProviderSources` relies on a transient DB-lock error propagating out of `parser.parse()` so it can skip the file and retry next run, leaving the prior cache intact. Other sqlite providers depend on this, and your own `validateSchema` deliberately re-throws busy errors. But the broad `catch` in `createParser.parse` (and in `discoverFromDb`) swallows the busy error and returns with no result, so the file gets cached with empty turns and is then served as "unchanged" on every later run. A single transient lock makes a Hermes session's data permanently disappear until the DB file changes. Please re-throw when `isSqliteBusyError(err)` in both catch blocks.

2. **Validate reported cost for finiteness.** Cost selection uses `(actual_cost_usd ?? 0) > 0 ? ...`, which rejects NaN and negatives but lets `Infinity` through (a REAL column can hold it). Because `hermes` is on the cost-preserve allowlist in parser.ts, that value is cached and used directly, so one Infinity poisons every aggregate total persistently. Guard with `Number.isFinite(x) && x > 0`, mirroring mistral-vibe's `safeNumber`.

3. **Split out the `totalReasoningTokens` core-type change.** This PR adds a required `totalReasoningTokens` to the core `SessionSummary` type plus parser aggregation, affecting all providers. But nothing reads it (not displayed, not exported), and Hermes does not need it (reasoning is already billed per call via `calculateCost`). Required-but-unset at the existing `SessionSummary` construction sites is also a latent type issue that only passes because tests are excluded from tsc. Please drop it from this PR, or land it separately as a change that actually wires up a consumer (and make it optional).

## Smaller
- The shared `fingerprintFile` splits a source path on the FIRST `#`, while your decoder uses the LAST, so a `HERMES_HOME`/profile path containing `#` is silently skipped. Niche but silent data loss.
- Only preserve the cost through the cache when Hermes actually reported one. For sessions that fell back to `calculateCost`, preserving the cached value means a later pricing-table update will not refresh them.
- The per-session `messages` query has no `LIMIT`; a pathological session could load a lot into memory. Discovery is capped at 10000, which silently drops older sessions for very heavy profiles.

## Rebase
Currently conflicts with main in CHANGELOG.md, src/providers/index.ts, and tests/provider-registry.test.ts (from recent merges). Please rebase.

Re the "supersedes #386" note: a CHANGELOG line cannot close another author's PR, so #386 and #370 will be closed separately once this lands.

umbra-sh:
@anthonyarmijo  any updates on this?

Issue #443, opened 2026-06-05
Incorrect costing for copilot usage since the switchover to metered billing


stephanjohnson:
Great project ❤️ 

Now that copilot is using metered billing... I want to use this a bit more actively 😄 

I did update the installation, but I am not sure if the information I am seeing is correct.

<img width="1019" height="1148" alt="Image" src="https://github.com/user-attachments/assets/cdcf55bb-2a10-4897-80a6-77f84d792d60" />

vs the GitHub dashboard

<img width="1315" height="611" alt="Image" src="https://github.com/user-attachments/assets/d9328a22-a5b4-4aa2-bf7b-04bf69049d54" />

I tried running codeburn from the application directory vs globally from my home directory. It made no difference.

Am I doing something wrong?

dat-intellis:
I think there's something wrong with how it calculates or includes subagents. I ran a session on sonnet 4.6, spawned a subagent on opus 4.8, and it doesn't show up anywhere. The calculation is definitely off somehow.

iamtoruk:
Thanks for the detailed report (and the screenshots — really helpful) 🙏

**What's happening:** codeburn estimates Copilot cost from **tokens × the underlying model's API price** (Copilot's logs don't record a real charge, so it infers the model from the session and prices the tokens at Anthropic/OpenAI API rates). The GitHub dashboard, on the other hand, bills by **premium requests / metered usage**, which is a *request-count × per-model multiplier* metric — a completely different unit from token cost. So the two numbers are expected to diverge; codeburn is showing "what these tokens would cost at API rates," not GitHub's billed amount.

This got more visible because of GitHub's billing changes. To build a model that actually matches your dashboard, I need to know which billing system you're on, since GitHub **switched from premium-request billing to usage-based "AI Credits" on June 1, 2026** (annual Pro/Pro+ plans stay on the legacy premium-request model for now):

1. **Plan:** Copilot Pro, Pro+, Business, or Enterprise?
2. **Billing system:** Does your usage page show **"premium requests"** (legacy) or **"AI Credits / usage-based"** (new)? Monthly vs annual subscription is usually the tell.
3. **One calibration point:** for a period, what does the GitHub dashboard show (premium-request count and/or the $ figure) vs what codeburn shows — even for a single model? That lets me validate the model against your real numbers.
4. **Which models** do you mostly drive in Copilot (e.g., GPT-4.1, Claude Sonnet, Opus)? codeburn currently infers the family heuristically, which I want to verify.

[...]

PR #449, opened 2026-06-06
docs(copilot): clarify token-estimate vs metered billing (interim for #443)

iamtoruk:
Interim baseline for #443 — **not** the full metered-billing cost model (that's gated on reporter calibration; see below).

## What this does
- Adds a **Pricing** section to `docs/providers/copilot.md` explaining why codeburn's Copilot number won't match the GitHub dashboard: codeburn estimates `tokens × underlying-model API rate`, while GitHub bills by metered **premium requests** (legacy, annual plans) or **AI Credits** (the usage-based model GitHub moved to on 2026-06-01).
- Sets `costIsEstimated: true` on Copilot calls (both parsers).

## Why it's only a baseline
The real fix — modeling GitHub's metered billing so the dollar figure matches the dashboard — can't be done blind:
- GitHub switched billing systems **5 days ago** (premium-request → AI Credits). We don't yet know which system the reporter is on.
- The premium-request multiplier table has **changed silently** in the past (Sonnet 1×→9×, Opus 10×→27×), so there's no stable spec to test against except a real user's dashboard.
- Making a Copilot-specific cost survive the cache round-trip needs changes to the core cost-authority path (`parser.ts` drops + recomputes provider cost), which warrants its own careful PR.

So we asked the reporter (and the subagent commenter) for their plan + billing system + a calibration data point on #443. Once we have that, the real metered-billing model gets built **on their numbers** and they validate the branch before merge.

## Status
Draft / **do not merge yet** — kept open as the honest-labeling baseline. tsc clean; `copilot.test.ts` 25/25 pass.

### DIFF
--- a/docs/providers/copilot.md
+++ b/docs/providers/copilot.md
@@ -36,6 +36,15 @@ Copilot does not always tag the model on each message. The parser infers it from
 
 See `copilot.ts:176-213`.
 
+## Pricing (read this before comparing to the GitHub dashboard)
+
+**CodeBurn's Copilot cost is a token-based estimate and will not match your GitHub Copilot bill.** They measure different things:
+
+- **CodeBurn** prices Copilot usage as `tokens × the underlying model's API rate`. Copilot's logs don't record a billed amount and often don't tag the model, so the parser infers the model family from the tool-call-ID prefix (see Model inference) and prices the tokens at Anthropic/OpenAI **API** rates. Calls are flagged `costIsEstimated`.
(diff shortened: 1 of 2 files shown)


iamtoruk:
Closing this stale interim draft. #443 (Copilot metered-billing accuracy) stays open and is the tracking issue for the real fix; this doc-only stopgap isn't worth carrying separately. Can revisit the wording when #443 is addressed.

PR #472, opened 2026-06-10
fix(mac): restore right-click status-item menu on macOS 27

theparlor:
## Summary

- Fixes the right-click status-item menu (Settings / Refresh Now / Check for Updates / Quit) being unreachable on **macOS 27.0**. Left-click / popover is unaffected. This worked on macOS 26.
- **Root cause:** macOS 27 no longer routes *any* right-mouse event to an `NSStatusItem` button's `target`/`action`. With the `sendAction(on:)` mask widened to every mouse type on an instrumented build, only `.leftMouseDown`/`.leftMouseUp` are ever delivered — right-mouse-down and right-mouse-up are consumed by the system's status-item host (`NSMenuBarNavigationSceneExtension`). So the `event.type == .rightMouseUp` branch in `handleButtonClick` is dead code on 27.
- **Fix:** present the menu from a global `addGlobalMonitorForEvents(matching: .rightMouseDown)` hit-tested against the status-item window, using `NSMenu.popUp(positioning:at:in:)` instead of the `statusItem.menu = menu; button.performClick` trick (which depends on the click→action path 27 changed). The legacy action path is retained for macOS ≤ 26, guarded by a 0.3 s debounce so a single right-click never double-presents.

## Testing

This change is confined to `mac/Sources/CodeBurnMenubar/CodeBurnApp.swift` (the Swift menubar) — it does not touch the TS CLI, so the `npm` checks below are N/A to this diff.

- [x] `swift build -c release` succeeds (Swift 6.2.3)
- [x] Verified live on **macOS 27.0 (build 26A5353q)**, Apple Silicon

Verification method: instrumented build with `NSLog` probes in the button action and menu path.

- 5 right-clicks with the original code → **0** action fires (handler never invoked).
- Mask widened to all mouse types → only `.leftMouseDown` (rawValue 1) delivered; **no** right-mouse events.
- With this fix → global monitor receives `rightMouseDown`, hit-test matches the status-item window, `NSMenu.popUp` presents the menu. One right-click → menu appears. Confirmed by a maintainer-style manual test on the affected machine.

Note: no AI co-author trailer is present on the commit (per CONTRIBUTING / `block-claude-coauthor`).

### DIFF
--- a/mac/Sources/CodeBurnMenubar/CodeBurnApp.swift
+++ b/mac/Sources/CodeBurnMenubar/CodeBurnApp.swift
@@ -32,6 +32,8 @@ struct CodeBurnApp: App {
 final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate {
     private var statusItem: NSStatusItem!
     private var popover: NSPopover!
+    private var rightClickMonitor: Any?
+    private var lastContextMenuPresentedAt: Date = .distantPast
     fileprivate let store = AppStore()
     let updateChecker = UpdateChecker()
     /// Held for the lifetime of the app to opt out of App Nap and Automatic Termination.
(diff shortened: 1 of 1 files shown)


theparlor:
A couple of process notes I want to own up front:

1. **I should have opened an issue first.** CONTRIBUTING asks to comment on an issue and get a maintainer's nod before sending code, and I jumped straight to a PR. Apologies for skipping that. I couldn't find an existing open issue for this specific macOS 27 right-click regression (the closed Tahoe items are the separate "status item won't render" bug), but that's not a reason to bypass the process. Happy to file a tracking issue and link it here, or to close this if it duplicates work already in progress or takes an approach you'd rather not.

2. **I've only verified on macOS 27.0 (build 26A5353q).** The new global-monitor path is confirmed working there. The retained legacy `.rightMouseUp` action path + the 0.3 s debounce for macOS ≤ 26 are **not** tested — I don't have a 26 machine to check against. My intent was to keep the old behavior intact on 26 and have the debounce absorb any double-fire if the global monitor also triggers there, but that's reasoning, not evidence. If someone on 26 can confirm the right-click menu still presents exactly once, that would close the gap. If you'd prefer, I can instead gate the legacy path behind an explicit OS-version check rather than relying on the debounce.

Glad to adjust the approach based on what you'd like to see.

PR #475, opened 2026-06-10
fix(menubar): support installer HTTP proxies

### LINKED ISSUE #473: Support company HTTP proxies in menubar installer downloads
#### Problem

`codeburn menubar` installs the macOS menubar app by downloading release metadata, the app zip, and checksum from GitHub. In company networks, direct outbound GitHub access may be blocked and traffic must go through an HTTP proxy.

Current installer network calls do not respect proxy environment variables, so install/update can fail behind corporate proxies.

Relevant calls:
- release lookup: `src/menubar-installer.ts:133-144`
- checksum download: `src/menubar-installer.ts:147-155`
- app zip download: `src/menubar-installer.ts:169-180`

#### Proposed change

Add proxy-aware fetch support for the menubar installer:

- honour `HTTPS_PROXY` / `https_proxy` for HTTPS requests
- fall back to `HTTP_PROXY` / `http_proxy`
- honour `NO_PROXY` / `no_proxy` for host/domain bypass
- route installer GitHub API, checksum, and zip downloads through that proxy
- add tests for proxy selection and `NO_PROXY` bypass

#### Implementation notes

Current local change uses `undici` `ProxyAgent` because Node does not provide a built-in `node:http.ProxyAgent`.

Touched areas:
- proxy resolution and fetch wrapper: `src/menubar-installer.ts:35-61`
- proxy-aware installer calls: `src/menubar-installer.ts:133-173`
- tests: `tests/menubar-installer.test.ts:101-116`
- dependency: `package.json:49-57`

#### Acceptance criteria

- `codeburn menubar` works when `HTTPS_PROXY=http://company-proxy:port` is set.
- `NO_PROXY=.github.com` bypasses proxy for GitHub hosts.
- Existing menubar installer behaviour remains unchanged when no proxy env vars are set.
- Targeted installer tests pass.
- TypeScript compile passes.

#### Verification from local change

- `npm test -- run tests/menubar-installer.test.ts` passed: 9 tests.
- `npx tsc --noEmit` passed.
- `npm run build` was blocked by unrelated network failure in `scripts/bundle-litellm.mjs` before compilation.

sleicht:
## Summary
Add proxy-aware downloads to the menubar installer so corporate networks can install the macOS menubar app through standard proxy environment variables.

## Context
`codeburn menubar` downloads release metadata, the app zip, and checksum from GitHub. Company networks may block direct GitHub access and require traffic through an HTTP proxy.

Closes #473

## Changes
- Route menubar installer network calls through an Undici `ProxyAgent` when proxy env vars are set.
- Honour `HTTPS_PROXY` / `https_proxy` for HTTPS requests, falling back to `HTTP_PROXY` / `http_proxy`.
- Honour `NO_PROXY` / `no_proxy` for host and domain bypass.
- Add tests for proxy selection and `NO_PROXY` bypass.

## Key Implementation Details
The PR uses `undici@^7.27.2` rather than v8 because the project supports Node `>=22.13.0` and Undici v8 requires Node `>=22.19.0`.

## Testing
```bash
npm test -- run tests/menubar-installer.test.ts
npx tsc --noEmit
```

Both commands passed locally on the clean PR branch.

`npm run build` remains blocked in this environment by network failure in `scripts/bundle-litellm.mjs` before compilation.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "ink": "^7.0.0",
         "react": "^19.2.5",
         "strip-ansi": "^7.2.0",
+        "undici": "^7.27.2",
         "zod": "^3.25.76"
       },
       "bin": {
@@ -3431,6 +3432,15 @@
       "integrity": "sha512-yDJTmhydvl5lJzBmy/hyOAA0d+aqCBuwl818haVdYCRrWV84o7YyeVm4QlVHStqNrrJSTb6jKuFAVqAFsr+K3Q==",
       "dev": true,
       "license": "MIT"
+    },
+    "node_modules/undici": {
+      "version": "7.27.2",
+      "resolved": "https://registry.npmjs.org/undici/-/undici-7.27.2.tgz",
+      "integrity": "sha512-uZsKNuzQxDMUY6M3pIMvy5tvlGmtq8XJ2oLAkfRKGNu+1VQAIvLy2xIVG5ATZl5wDXl/tddByAWCizRbOme+TA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=20.18.1"
+      }
     },
     "node_modules/undici-types": {
       "version": "6.21.0",
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "ink": "^7.0.0",
     "react": "^19.2.5",
     "strip-ansi": "^7.2.0",
+    "undici": "^7.27.2",
     "zod": "^3.25.76"
   },
   "devDependencies": {
--- a/src/menubar-installer.ts
+++ b/src/menubar-installer.ts
@@ -6,6 +6,7 @@ import { homedir, platform, tmpdir } from 'node:os'
 import { join } from 'node:path'
 import { pipeline } from 'node:stream/promises'
 import { Readable } from 'node:stream'
+import { ProxyAgent, fetch as undiciFetch } from 'undici'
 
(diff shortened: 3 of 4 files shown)


iamtoruk:
Thanks for this, proxy support is genuinely useful and the implementation is clean. The env var precedence and the NO_PROXY matching look right, and using a per-request dispatcher instead of setGlobalDispatcher is the right call.

One thing to sort before merge: the package-lock entry for undici resolved to a private registry rather than the public npm one:

```
"resolved": "https://artifactory.gcp.sanet17.ch/artifactory/api/npm/sanitas-npm/undici/-/undici-7.27.2.tgz"
```

Looks like npm picked that up from your company proxy during install. It would break `npm ci` for anyone who cannot reach that host (CI, other contributors), and it leaks your internal registry URL into the repo. The integrity hash matches the real public undici@7.27.2, so the package itself is fine, it just needs to resolve from registry.npmjs.org.

Could you regenerate the lockfile against the public registry and recommit? Roughly:

```
npm install --registry=https://registry.npmjs.org/
```

Minor and not blocking: a quick `codeburn menubar` run on a normal network would be good to confirm the switch to undici's fetch did not change the non-proxy download path, since the tests cover proxy resolution but not the actual download.

Thanks again for the contribution.

sleicht:
Thank you for the review, @iamtoruk. I fixed the lock file and rebased onto `main`.

PR #485, opened 2026-06-12
fix(cursor): page bubble scans by requested range

### LINKED ISSUE #482: Cursor provider should not truncate reports at 250k bubbles
## Problem

When Cursor has a large local database, codeburn emits:

```text
codeburn: Cursor database has 3,25,568 bubbles, scanning the most recent 2,50,000. Older sessions may be missing from this report.
```

That means lifetime or long-range reports can silently omit older Cursor sessions once the local bubble table grows past the hard scan cap.

## Expected behavior

Reports should not drop older sessions just because the Cursor database has more than 250,000 bubbles. The provider should either query the relevant date/window directly, page through the rows it needs, or otherwise make the cap apply only where it cannot affect the requested report.

## Actual behavior

The provider scans only the most recent 250,000 bubbles and warns that older sessions may be missing.

## Notes

I checked existing issues, PRs, and discussions for the exact warning text and related phrases (`Cursor database has`, `Older sessions may be missing`, `scanning the most recent`, `250000`, `2,50,000`) and did not find this specific issue reported.

Related but not identical:

- #326
- #163
- #169

vaibhavarora14:
## Summary

- Replaces the historic newest-250k Cursor bubble cap with paged scans bounded by the requested date range.
- Caches parsed Cursor calls by DB fingerprint and requested lower bound so wider reports cannot reuse narrower cached results.
- Adds regression coverage for older in-window bubbles and requested-range filtering.

## Target issues

Closes #482

## Testing

- [x] `npm test -- --run tests/providers/cursor-bubble-scan-window.test.ts`
- [ ] `npm test` passes

### DIFF
--- a/src/cursor-cache.ts
+++ b/src/cursor-cache.ts
@@ -5,12 +5,10 @@ import { randomBytes } from 'crypto'
 
 import type { ParsedProviderCall } from './providers/types.js'
 
-// Bumped to 3 for the workspace-aware breakdown change: the cursor parser
-// now derives `sessionId` from the bubble row key (the real composer id)
-// rather than the empty `conversationId` JSON field, and the workspace
-// router relies on those composer ids to bucket calls per project.
-// Version 2 caches contain `sessionId: 'unknown'` for every call and would
-// route everything to the orphan project, so we invalidate them.
+// Bumped to 4 for the paged Cursor bubble scan: version 3 caches could have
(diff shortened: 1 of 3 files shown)


iamtoruk:
Superseded: main already pages Cursor bubble scans by the requested ROWID window (src/providers/cursor.ts, the paged ROWID-descending variant) and issue #482 is closed. Closing as already-addressed. Appreciate the fix.

PR #488, opened 2026-06-13
fix(codex): attribute MCP tool-call end events

### LINKED ISSUE #478: Codex MCP usage is missing from recent lookback windows
I’m seeing Codex MCP usage show up in long lookback ranges, but not in recent ones like `today`, `week`, or `month`.

I believe this is a parsing issue in the Codex provider. Recent Codex sessions still contain MCP activity, but the MCP completion event now appears as `event_msg` with `type: "mcp_tool_call_end"`. Codeburn appears to be counting the surrounding token usage, but not attributing the MCP tool call itself.

**What I see:**
- Codex sessions are detected normally.
- Token usage is recorded.
- MCP usage is missing from recent windows.
- Older history still includes MCP usage.

**What I expect:**
- MCP calls from recent Codex sessions should be attributed the same way as older sessions.

I have a local session example that reproduces this, but I’m not attaching the raw JSONL publicly because it contains workspace and session details. I can share a redacted excerpt if that helps.

If useful, I can also help turn this into a minimal repro snippet with just the relevant event shapes.

fightingst:
## Summary

- Normalize recent Codex MCP `function_call` entries with `namespace: "mcp__..."` into CodeBurn's canonical `mcp__server__tool` tool names.
- Parse `event_msg` / `mcp_tool_call_end` events when Codex emits only the completion event, while de-duping paired events by `call_id` so one MCP invocation is not counted twice.
- Add Codex provider regression coverage for both paired `function_call` + `mcp_tool_call_end` logs and end-only logs.

## Target issues

Closes #478

## Testing

- [x] `npm test -- --run tests/providers/codex.test.ts`
- [x] `npx tsc --noEmit --pretty false`
- [x] Inspected a real local Codex JSONL sequence with `function_call` namespace + matching `mcp_tool_call_end` `call_id`
- [x] `npm test` passes

### DIFF
--- a/src/providers/codex.ts
+++ b/src/providers/codex.ts
@@ -40,6 +40,36 @@ const toolNameMap: Record<string, string> = {
   read_dir: 'Glob',
 }
 
+function normalizeMcpSegment(segment: string): string {
+  return segment
+    .replace(/[^a-zA-Z0-9_]+/g, '_')
+    .replace(/_+/g, '_')
+    .replace(/^_+|_+$/g, '')
+}
+
+function mcpToolName(server: string, tool: string): string | null {
+  const normalizedServer = normalizeMcpSegment(server)
+  const normalizedTool = normalizeMcpSegment(tool)
+  if (!normalizedServer || !normalizedTool) return null
+  return `mcp__${normalizedServer}__${normalizedTool}`
+}
+
+function codexFunctionCallToolName(payload: CodexEntry['payload']): string {
+  const rawName = payload?.name ?? ''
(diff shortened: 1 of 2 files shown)


iamtoruk:
Superseded: main already attributes Codex MCP calls via the event_msg/mcp_tool_call_end path (src/providers/codex.ts:409). Closing this implementation. Note issue #478 stays open, if Codex MCP usage still under-reports on the latest version, it needs a fresh diagnosis rather than this branch, which predates the current handling.

PR #489, opened 2026-06-13
fix(devin): preserve raw submodel IDs in model rollups

bmcdonough:
## Summary
- preserve raw Devin model IDs during parser session model rollup instead of collapsing to short names
- align model-efficiency aggregation with the same Devin-aware model keying
- return raw Devin model display names and add provider/unit/CLI regression coverage
- add a design note documenting root cause, proposed scope, and Claude comparison

#### Test plan
- [x] npx vitest run tests/providers/devin.test.ts tests/model-efficiency.test.ts tests/cli-devin-model-variants.test.ts

Generated with [Devin](https://cli.devin.ai/docs)

### DIFF
--- a/docs/design/devin-full-model-name-reporting.md
+++ b/docs/design/devin-full-model-name-reporting.md
@@ -1,0 +1,118 @@
+# Devin Full Model Name Reporting
+
+- **Date:** 2026-06-12
+- **Status:** Proposed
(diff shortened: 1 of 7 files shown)


bmcdonough:
closing in favor of https://github.com/getagentseal/codeburn/pull/585

Issue #497, opened 2026-06-17
feat(menubar): Daily budget alerts should respect metric (Cost or Tokens)


reneleonhardt:
It would be great if the daily budget alert would respect the metric selection.

At the moment only a cost is selectable, like $25.

But if the user switches the default metric from `Cost` to `Tokens`, the user should be able to choose a specific token amount, preferably a number input field to allow any daily amount like 56M for MiniMax M3 to stay within the monthly subscribed quota of 1.7B tokens for example.

In addition, the current Cost choices don't respect the currency setting.

<img width="447" height="446" alt="Image" src="https://github.com/user-attachments/assets/ffec3663-9e9b-4038-8cba-9c8e344bdbfc" />

reneleonhardt:
As menubar space is very sparse with hundreds of competing icons, a more compact representation would be very welcome.

For example, with the Tokens metric and a monthly subscription I don't much care about input / output distinction, a combined value like `🔥 20M` would be much smaller and more meaningful to me.

iamtoruk:
Thanks for the suggestion. Two parts here:

1. Compact combined token view: this already exists. In Settings, set the display metric to "Total Tokens" and the menubar shows a single combined value (flame + total), e.g. a compact "20M", instead of the separate up/down split. Options are Cost ($), Tokens (up/down), Total Tokens, and Icon Only.

2. Daily budget respecting the metric (the title): this was cost-only, so on a token metric there was no token-based alert. That is now fixed. When the display metric is Tokens or Total Tokens, the daily budget is set in tokens (Off / 1M / 5M / 10M / 25M / 50M / 100M), the flame turns yellow when today's tokens pass it, and the banner reads "Daily budget of NM tokens exceeded". The cost metric keeps its dollar budget. The two budgets are stored separately so switching metric never reinterprets a threshold.

This ships in the next menubar release. Closing as both parts are addressed, but reopen if the token view or alert does not behave as you expect.

reneleonhardt:
Thank you very much, I will check all changes after the next release! ❤️

One thing: `Total tokens` appends ` tok`, resulting in `🔥 0 tok` at the beginning of a new day for example.
Would it be possible to remove the tok suffix? The user doesn't need it, he knows if he changed the default settings, and doesn't mistake `🔥 20M` for $20M I'm sure 😄

iamtoruk:
[...]

PR #506, opened 2026-06-18
feat(menubar): custom daily budget amount

iamtoruk:
Follow-up to #505 (metric-aware daily budget for #497): the budget could only be a fixed preset. This adds manual entry.

## Change (macOS menubar Settings)
- Adds a **"Custom…"** option to the Daily budget picker. Selecting it reveals a text field:
  - Cost metric: dollars (with a `$` prefix).
  - Token metrics: amount in **millions of tokens** (stored as `value * 1_000_000`).
- The typed value is written to the same `dailyBudget` / `dailyTokenBudget` fields, so the flame tint and the "budget exceeded" banner use it with no further changes.
- On opening Settings, a previously saved non-preset value reselects "Custom…" and pre-fills the field.
- Empty/invalid input is treated as 0 (off); negatives are clamped.

## Validation
- `swift build` (mac package) passes.
- macOS app, so not covered by the JS suite. Runtime check: pick "Custom…", type an amount, confirm it persists, the picker keeps showing "Custom…" after reopening, and the alert triggers at the typed threshold.

### DIFF
--- a/mac/Sources/CodeBurnMenubar/Views/SettingsView.swift
+++ b/mac/Sources/CodeBurnMenubar/Views/SettingsView.swift
@@ -33,6 +33,30 @@ struct SettingsView: View {
 private struct GeneralSettingsTab: View {
     @Environment(AppStore.self) private var store
 
+    // "Custom…" budget entry state, one per metric (cost in dollars, tokens in
+    // millions). When custom is active the picker shows "Custom…" and a field
+    // appears for an exact amount.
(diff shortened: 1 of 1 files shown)


reneleonhardt:
Thank you so much for reacting super fast! ❤️

There are no automatic builds so there are no attachment artifacts to try myself.
If possible next time, please attach a screenshot so see the change in action.

Issue #414, opened 2026-05-29
Support Non GPT Azure Foundry Models


cloudant42:
Hi,

I am using code burn mainly with models hosted in Azure Foundry.

It is working perfectly for all GPT models so far, that I use with the codex cli, but I noticed that I get only partial information for other models hosted in foundry such as Kimi2.6 or deepseek models that I use with open code.

I think in general the API response that is used to share the token usage is similar for all models in Azure foundry. At the moment the only thing I see in codeburn are the number of calls, but no information related to token usage.

Would be awesome if this is also working. Thanks!

iamtoruk:
Thanks for the detailed report. I dug into this, and the missing token usage is an OpenCode-side issue rather than something CodeBurn can recover.

CodeBurn reads token usage straight from what OpenCode stores per message (`message.tokens` / `usage`). For models served through an openai-compatible provider (which is how Azure Foundry non-GPT models like Kimi and DeepSeek are configured), OpenCode currently writes zero tokens even though the Foundry API returns usage in the response. So there is nothing in OpenCode's session files for CodeBurn to display, which is why you see call counts but no tokens. When OpenCode does record usage (for example its native `openai` provider), CodeBurn shows it correctly.

This is a known OpenCode bug: the usage is present in the API response, but OpenCode fails to parse and persist it for openai-compatible providers (it ends up as 0). See https://github.com/anomalyco/opencode/issues/423.

Next step: update OpenCode and CodeBurn and retest. That issue was fixed for some openai-compatible providers (LM Studio, Ollama); if Azure Foundry now records tokens, CodeBurn will pick them up automatically. If it still shows zero, the fix needs to land in OpenCode. Happy to double-check if you can paste one assistant message from `~/.local/share/opencode/storage/message/<session>/<id>.json` showing its `tokens` field.

PR #523, opened 2026-06-19
feat(opencode): read file-based JSON sessions (OpenCode 1.1+)

iamtoruk:
## Problem
OpenCode 1.1+ stores sessions as file-based JSON under `~/.local/share/opencode/storage/{session,message,part}/` instead of a SQLite database. The provider only globbed for `opencode*.db`, so on a current OpenCode install it discovered zero sessions and reported no usage at all. Confirmed locally: a live OpenCode 1.1.65 session with ~67K tokens of context was completely invisible to CodeBurn (0 sessions discovered).

## Change
- New `opencode-file-parser.ts`: file-based discovery and parsing. Reads sessions from `storage/session/**/*.json`, tokens, cost, and model from `storage/message/<id>/*.json`, and tools and bash commands from `storage/part/<id>/*.json`. No SQLite dependency.
- `opencode.ts` prefers file-based storage when present and falls back to the SQLite DB so pre-migration installs keep reporting.
- Shared message-to-call logic (token extraction, tool and bash parsing, cost, skip rules) extracted into `session-message.ts`, so the file path, the SQLite path, and Kilo Code all behave identically.

## Verification
- Against a live OpenCode 1.1.65 session: 0 to 3 sessions, 33 calls, 87K input/output/reasoning tokens plus 1.17M cache reads, $0.56 estimated, model gpt-5.3-codex-spark, tools Bash/Glob/Read, user messages captured. Errored turns (such as a rejected model request) are correctly skipped.
- 7 new file-based tests. Existing OpenCode (43) and Kilo Code (4) SQLite tests pass unchanged.
- Full suite 1239/1239, tsc clean.

Related to #414. That issue also covers a narrower gap (no tokens for non-GPT Azure Foundry models in the SQLite path); this PR fixes the broader case where current OpenCode is not read at all. The SQLite fallback path is unchanged.

### DIFF
--- a/src/providers/opencode.ts
+++ b/src/providers/opencode.ts
@@ -3,6 +3,7 @@ import { homedir } from 'os'
 
 import { getShortModelName } from '../models.js'
 import { discoverSqliteSessions, createSqliteSessionParser, type SqliteProviderConfig } from './sqlite-session-parser.js'
+import { discoverOpenCodeFileSessions, createOpenCodeFileSessionParser } from './opencode-file-parser.js'
 import type { Provider, SessionSource, SessionParser } from './types.js'
 
 const toolNameMap: Record<string, string> = {
@@ -39,6 +40,7 @@ function getSqliteConfig(dataDir?: string): SqliteProviderConfig {
 
 export function createOpenCodeProvider(dataDir?: string): Provider {
   const sqliteConfig = getSqliteConfig(dataDir)
+  const resolvedDataDir = getDataDir(dataDir)
 
   return {
     name: 'opencode',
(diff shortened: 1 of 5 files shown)


src/providers/opencode.ts
```diff
@@ -53,11 +55,19 @@ export function createOpenCodeProvider(dataDir?: string): Provider {
       return toolNameMap[rawTool] ?? rawTool
     },
 
+    // OpenCode 1.1+ stores sessions as file-based JSON; older builds used a
```
rafan, line 58: issue: this breaks codeburn report for new versions of opencode. It was actually the other way around: the older version uses file-based session storage, new version uses DB. So when both exist, codeburn can not  discover new usage. 

Ref: the file->DB migration happened in [1.2.0](https://github.com/anomalyco/opencode/releases/tag/v1.2.0) in Feb.

Issue #525, opened 2026-06-19
devin schema version changed to ATIF-v1.7


bmcdonough:
@tvcsantos 

I noticed devin support within codeburn stopped working.
There is a new schema version `ATIF-v1.7` in my devin transcripts.
The current devin support is only built for `ATIF-v1.4`.

Looks like this happened with devin [v2026.5.26-0](https://docs.devin.ai/cli/changelog/stable#2026-5-26-0)

bmcdonough:
resolved with release [0.9.15](https://github.com/getagentseal/codeburn/commit/b4d7269a7f60e824b86eaf602955ea9ac185c62d)

PR #526, opened 2026-06-19
fix(devin): support ATIF v1.7 transcripts

bmcdonough:
## What
- update the Devin provider parser to handle ATIF v1.7 transcript shape (//step-level  + )
- keep compatibility with existing Devin transcript fields by normalizing legacy and newer cost/token fields
- add coverage for ATIF v1.5, v1.6, and v1.7 transcript variants
- refresh Devin provider docs for supported ATIF versions and field mappings

## Why
Recent Devin CLI transcripts use newer ATIF shapes (including v1.7) that are not strictly metadata-scoped. This change ensures codeburn can ingest these sessions end-to-end while preserving existing behavior.

 RUN  v3.2.6 /Users/william.mcdonough/github/bmcdonough/codeburn

 ✓ tests/providers/devin.test.ts (14 tests) 60ms

 Test Files  1 passed (1)
      Tests  14 passed (14)
   Start at  15:18:26
   Duration  600ms (transform 180ms, setup 0ms, collect 244ms, tests 60ms, environment 0ms, prepare 111ms)
   
 - resolves issue #525

### DIFF
--- a/docs/providers/devin.md
+++ b/docs/providers/devin.md
@@ -49,8 +49,13 @@ The config file is:
 ~/.config/codeburn/config.json
 ```
 
+The macOS Settings window writes this value from the Devin tab. There is no
+environment-variable override and no default rate. Do not hardcode a universal
+ACU price; Devin ACU pricing is account/contract dependent.
+
 When the rate is missing or invalid, `discoverSessions()` returns `[]` and the
-parser yields no calls.
+parser yields no calls. Devin remains registered as a provider, but it does not
+appear in CLI/UI results until configured.
 
 ## Storage format
 
--- a/docs/providers/devin.md
+++ b/docs/providers/devin.md
@@ -4,7 +4,7 @@ Cognition Devin CLI local usage tracking.
 
 - **Source:** `src/providers/devin.ts`
 - **Loading:** eager (`src/providers/index.ts`)
-- **Test:** `tests/providers/devin.test.ts` (336 lines)
+- **Test:** `tests/providers/devin.test.ts`
 
 ## Where it reads from
 
@@ -14,7 +14,7 @@ Devin CLI data lives under:
 ~/.local/share/devin/cli/
 ```
 
-The MVP usage source is transcript JSON:
+Usage comes from transcript JSON:
 
 ```text
 ~/.local/share/devin/cli/transcripts/*.json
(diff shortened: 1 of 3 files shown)


docs/providers/devin.md
```diff
@@ -49,127 +49,111 @@ The config file is:
 ~/.config/codeburn/config.json
 ```
 
-The macOS Settings window writes this value from the Devin tab. There is no
```
tvcsantos, line 52: Don't really understand what was the reasoning to remove this 🤔

docs/providers/devin.md
```diff
@@ -49,127 +49,111 @@ The config file is:
...
-- `metadata.metrics.cache_creation_tokens`
-- `metadata.metrics.cache_read_tokens`
-- `metadata.created_at`
-- `metadata.generation_model`
-- `metadata.request_id`
-- `tool_calls[].function_name`
-
-User-input steps (`metadata.is_user_input === true`) are skipped. Non-user
-steps are included only if they have positive ACU usage or positive token usage.
+CodeBurn supports ATIF transcript variants used by Devin across **ATIF-v1.5**, **ATIF-v1.6**, and **ATIF-v1.7** (and remains backward-compatible with older Devin transcripts that still use v1.4-style field names).
```
tvcsantos: As it is I would not frame as "CodeBurn supports ATIF". The existing code is tied to devin. For that claim to be true. The code would need to be on a separate place and be shared by other providers, which at this point is not.

docs/providers/devin.md
```diff
@@ -49,127 +49,111 @@ The config file is:
...
+
+The parser does not hard-fail on `schema_version`; it parses any object root
```
tvcsantos: schema_version has been mandatory always since day 1 on ATIF RFC. so IMO we don't need the defensive programming because i am assuming Devin complies with the specification.

docs/providers/devin.md
```diff
@@ -49,127 +49,111 @@ The config file is:
...
+
+- **sessionId:** `session_id` -> `trajectory_id` -> transcript filename
+- **model:** `step.extra.generation_model` -> `step.metadata.generation_model` -> `step.model_name` -> `agent.model_name` -> `sessions.model` -> `devin`
+- **timestamp:** `step.metadata.created_at` -> `step.timestamp` -> `sessions.last_activity_at` -> `sessions.created_at`
 
 ## Pricing
 
-`metadata.committed_acu_cost` is per step, not cumulative. The provider converts
-each step with:
+`costUSD` is always provider-supplied and uses configured ACU conversion:
```
tvcsantos, line 107: Still per step and not cumulative, right?

docs/providers/devin.md
```diff
@@ -49,127 +49,111 @@ The config file is:
...
 
 ```text
 costUSD = committed_acu_cost * devin.acuUsdRate
 ```
 
+If a step only has `committed_credit_cost`, CodeBurn converts credits to ACU
+using Devin's current export convention:
+
+```text
+committed_acu_cost = committed_credit_cost / 10000
```
tvcsantos, line 117: What is the source of this? Can you provide a reference?

Also IMO this should not be hardcoded, but rather configurable. Because Devin can change it at any time, and most likely on enterprise accounts can even depend on the agreed plan.

docs/providers/devin.md
```diff
@@ -49,127 +49,111 @@ The config file is:
...
+| `created_at`        | timestamp fallback                                                                                           |
+| `last_activity_at`  | preferred session timestamp fallback                                                                         |
+| `hidden`            | skip hidden sessions                                                                                         |
 
 ## Deduplication
 
 `devin:<sessionId>:<step.step_id>`
 
-The provider name is part of the key via the `devin:` prefix.
+When `step_id` is missing, parser falls back to 1-based step index.
```
tvcsantos, line 144: `step_id` accoding to ATIF RFC has never been optional and is always mandatory. So it will never be missing

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
 import { isPositiveNumber, safeNumber } from "../parser.js";
 
 type AgentTrajectory<T extends Step> = {
-  schema_version: string;
+  schema_version?: string;
```
tvcsantos: never optional. was always mandatory from the first version of ATIF

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
   session_id?: string;
-  agent: Agent;
-  steps: T[];
+  trajectory_id?: string;
+  agent?: Agent;
```
tvcsantos: mandatory since ATIF first version

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
+  steps?: T[];
```
tvcsantos: mandatory since ATIF first version. When not steps. The array is empty but is still there

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
 };
 
 type Agent = {
-  name: string;
-  version: string;
+  name?: string;
```
tvcsantos: mandatory since ATIF first version.

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
+  version?: string;
```
tvcsantos: mandatory since ATIF first version

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
   model_name?: string;
+  tool_definitions?: unknown;
```
tvcsantos: not unknown, but rather an array following openai functional calling schema

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
 };
 
 type ToolCall = {
-  tool_call_id: string;
-  function_name: string;
+  tool_call_id?: string;
```
tvcsantos, line 42: field is mandatory, not optional

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
+  function_name?: string;
```
tvcsantos, line 43: field is mandatory, not optional

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
+  function?: {
```
tvcsantos, line 44: does not exist. only exists on tool definition

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
+type DevinStepExtra = {
+  committed_acu_cost?: number;
+  committed_credit_cost?: number;
+  generation_model?: string;
 };
 
 type Step = {
-  step_id: number;
-  source: string;
+  step_id?: number | string;
```
tvcsantos, line 86: step_id is a number according to ATIF RFC

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
+  source?: string;
```
tvcsantos, line 87: source is required and not optional

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
+  timestamp?: string;
   model_name?: string;
-  message: string;
+  message?: unknown;
```
tvcsantos, line 90: message is mandatory and not optional.

The type if we want to accept versions above 1.6+ should be array or a string.

When is an array then it has the following schema

{ type: string, text?: string, source?: unknown }

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
+  metrics?: DevinMetrics;
```
tvcsantos, line 91: should not be tied to DevinMetrics. Step type is supposed to be generic and unrelated with Devin.

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
+  metadata?: DevinMetadata;
```
tvcsantos, line 92: metadata is not part of ATIF at step level and this seems to be custom to devin. Hence should not be declared in this type.

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
+  extra?: DevinStepExtra;
```
tvcsantos, line 93: extra is part of the spec but is an object with no bound contract.

Since Step type is supposed to be generic for ATIF it should not be tied to devin in any way

src/providers/devin.ts
```diff
@@ -15,53 +15,77 @@ import { readSessionFile } from "../fs-utils.js";
...
   tool_calls?: Array<ToolCall>;
 };
 
-type DevinStep = Step & { metadata?: DevinMetadata };
+type DevinStep = Step;
```
tvcsantos, line 97: The preivous version was intented since this type was defined to provide the custom DevinStep that extends Step with custom attributes and information

src/providers/devin.ts
```diff
@@ -96,10 +120,15 @@ const DEVIN_PROVIDER_NAME = "devin";
 const DEVIN_PROVIDER_DISPLAY_NAME = "Devin";
 const DEVIN_TRANSCRIPTS_SUBDIR = "transcripts";
 const DEVIN_SESSIONS_DB = "sessions.db";
+const DEVIN_CREDITS_PER_ACU = 10_000;
```
tvcsantos, line 132: What is the source for this? IMO this should be configurable, since Devin can change, and also might depend on subscription plan

src/providers/devin.ts
```diff
@@ -96,10 +120,15 @@ const DEVIN_PROVIDER_NAME = "devin";
...
 
 function parseTranscript(raw: string): DevinAgentTrajectory | null {
   try {
-    return JSON.parse(raw) as DevinAgentTrajectory;
+    const parsed = JSON.parse(raw) as unknown;
+    if (!parsed || typeof parsed !== "object" || Array.isArray(parsed)) {
+      return null;
+    }
+    return parsed as DevinAgentTrajectory;
```
tvcsantos, line 140: No need for the complex defensive programming. The transcript must be allways from ATIF format

src/providers/devin.ts
```diff
@@ -195,7 +306,13 @@ function getModelName(
...
 
 function getToolNames(step: DevinStep): string[] {
-  return (step.tool_calls ?? []).map((call) => call.function_name);
+  const tools: string[] = [];
+  const toolCalls = Array.isArray(step.tool_calls) ? step.tool_calls : [];
+  for (const call of toolCalls) {
+    const toolName = call.function_name ?? call.function?.name;
+    if (toolName) tools.push(toolName);
+  }
+  return tools;
```
tvcsantos, line 351: 1. tool_calls is always an array
2. function.name is not part of the ATIF schema for tool_calls. only function_name is the valid attribute. So no need for the extra logic

src/providers/devin.ts
```diff
@@ -267,7 +384,7 @@ class DevinSessionParser implements SessionParser {
     if (!raw) return;
 
     const transcript = parseTranscript(raw);
-    if (!transcript?.steps) return;
+    if (!transcript?.steps || !Array.isArray(transcript.steps)) return;
```
tvcsantos, line 423: steps is always an array so no need for the check

src/providers/devin.ts
```diff
@@ -280,14 +397,15 @@ class DevinSessionParser implements SessionParser {
 
     for (let index = 0; index < transcript.steps.length; index++) {
       const step = transcript.steps[index];
-      if (step.metadata?.is_user_input) continue;
+      if (!step || typeof step !== "object" || Array.isArray(step)) continue;
+      if (isUserInputStep(step)) continue;
```
tvcsantos, line 438: step is always defined and is an object with several mandatory fields. So no need for the check. It can be simplified to

```suggestion
      if (isUserInputStep(step)) continue;
```

src/providers/devin.ts
```diff
@@ -280,14 +397,15 @@ class DevinSessionParser implements SessionParser {
...
 
-      const usage = getUsage(step.metadata);
+      const usage = getUsage(step);
       if (!usage) continue;
 
       const timestamp = getTimestamp(step, session) ?? "";
-
-      const deduplicationKey = `devin:${sessionId}:${step.step_id}`;
+      const stepId = `${step.step_id ?? index + 1}`;
```
tvcsantos, line 444: step id is mandatory so no need for this

src/providers/devin.ts
```diff
@@ -110,37 +139,117 @@ function parseNumericTimestamp(value: number): string {
...
+  if (isPositiveNumber(metadataCreditCost)) {
+    return metadataCreditCost / DEVIN_CREDITS_PER_ACU;
+  }
+
+  const extraCreditCost = safeNumber(step.extra?.committed_credit_cost);
+  if (isPositiveNumber(extraCreditCost)) {
+    return extraCreditCost / DEVIN_CREDITS_PER_ACU;
+  }
+
+  return 0;
```
tvcsantos, line 192: The logic can be grouped in two blocks. ACU costs and Credit costs. The computational effort vs readability in my opinion does not pay all the repetition for the early returns. Also isPositiveNumber subsumes safeNumber.

```suggestion
  const acuCost = [
    step.metadata?.committed_acu_cost,
    step.extra?.committed_acu_cost,
  ].filter(cost => isPositiveNumber(cost));
  
  const creditCost = [
    step.metadata?.committed_credit_cost,
    step.extra?.committed_credit_cost
  ].filter(cost => isPositiveNumber(cost))
  .map(cost => cost / DEVIN_CREDITS_PER_ACU);
  
  return acuCost.shift() || creditCost.shift() || 0;
```

src/providers/devin.ts
```diff
@@ -110,37 +139,117 @@ function parseNumericTimestamp(value: number): string {
...
+  if (inputTokens === 0 && promptTokens > 0) {
+    inputTokens = Math.max(
+      0,
+      promptTokens - cacheReadInputTokens - cacheCreationInputTokens,
+    );
+  }
+
+  const outputTokens = safeNumber(
+    metrics?.output_tokens ?? metrics?.completion_tokens,
+  );
```
tvcsantos, line 226: I would suggest to extract each one to a separate method for maintanability and also separation of concerns, making also this method `getUsage` easier to read.

src/providers/devin.ts
```diff
@@ -110,37 +139,117 @@ function parseNumericTimestamp(value: number): string {
   return new Date(millis).toISOString();
 }
 
-function getUsage(
-  metadata: DevinMetadata | undefined | null,
-): DevinUsage | null {
-  if (!metadata) return null;
-  const metrics = metadata.metrics;
+function normalizeMessageText(message: unknown): string {
```
tvcsantos, line 151: I suggest first to check the schema from ATIF to avoid all this complex code. and to a proper message, taking into account ATIF schema.

tvcsantos (commented):
Thanks for the PR.

In order to better match ATIF format you can check:
- https://github.com/harbor-framework/harbor/blob/main/rfcs/0001-trajectory-format.md
- https://www.harborframework.com/docs/agents/trajectory-format#schema-versions

docs/providers/devin.md
bmcdonough: restored lines 52-58

docs/providers/devin.md
bmcdonough: I changed the wording, let me know your thoughts.

src/providers/devin.ts
bmcdonough: restored to being required

src/providers/devin.ts
bmcdonough: left trajectory_id optional.
restored agent to be required.
restored steps to be required

docs/providers/devin.md
bmcdonough: changed the wording and set schema_version to be required

bmcdonough:
converted to [DRAFT] while I work on the issues identified in the excellent review by [tvcsantos](https://github.com/tvcsantos)

src/providers/devin.ts
bmcdonough: set name, version to required.
set tool_definitions to array.

bmcdonough:
closed in favor of https://github.com/getagentseal/codeburn/pull/570

PR #543, opened 2026-06-21
feat(providers): add Hermes Agent usage provider

ozymandiashh:
## Summary

Adds a usage provider for Hermes Agent (the NousResearch desktop agent), following the existing provider pattern (modeled on the ZCode provider). It reads Hermes' local SQLite store, attributes per session tokens and cost, and registers the provider with its models, docs and tests.

## Motivation

Hermes Agent usage was not tracked at all, so its spend was invisible in codeburn. Hermes records everything needed locally, so a read-only provider can surface it alongside the other agents.

## Data source

`~/.hermes/state.db` (SQLite), table `sessions`. Columns consumed: `model`, `input_tokens`, `output_tokens`, `cache_read_tokens`, `cache_write_tokens`, `reasoning_tokens`, `estimated_cost_usd`, `actual_cost_usd`, `cost_status`, `started_at`, `ended_at`, `cwd`.

Unlike ZCode (which folds cached tokens into `input_tokens`, OpenAI style), Hermes keeps cache read and cache write tokens in their own columns, so they are mapped directly rather than merged.

## Cost resolution

Hermes often records its own cost, so the provider honors it with a clear precedence:

1. A real measured cost: a positive `actual_cost_usd` is used as is.
2. The agent's own estimate: `estimated_cost_usd` is used when `cost_status = 'estimated'`.
3. No usable stored cost: the session is priced through codeburn's model pricing table via `calculateCost`, including reasoning tokens.

## Correctness notes

- The database is opened once and all eligible rows are read from that single connection. Discovery does not emit one source per row, so parsing a long history does not open and close a connection per session (which would risk file descriptor exhaustion on large stores).
- The row filter keeps a session when any token column is positive or when `estimated_cost_usd` or `actual_cost_usd` is positive. This means cost-only sessions (zero tokens but a recorded cost) are retained rather than silently dropped.
- Reasoning tokens are passed through into the fallback pricing path. `calculateCost` gains an optional, backward compatible `reasoningTokens` argument priced at the output token rate; existing callers are unaffected.
- Each call carries its own `project` / `projectPath` derived from the session's `cwd`, so usage attributes to the right working directory. Rows are ordered by `COALESCE(ended_at, started_at, 0), id` for stable output.

## Tests

- A fixture seeded SQLite store shaped like `~/.hermes/state.db` with assertions for: a normal session (correct model and input/output/cache tokens with a computed cost), a cost-only session (zero tokens, positive recorded cost) being retained, a reasoning-token session being priced correctly in the fallback path, and many rows parsing from a single connection open.
- `npm test`: full suite green. `npm run build`: ok. `tsc --noEmit`: clean.

## Scope

Adds `src/providers/hermes.ts`, registers it in `src/providers/index.ts`, wires pricing in `src/models.ts`, adds `tests/providers/hermes.test.ts`, and documents it in `docs/providers/hermes.md` with an entry in `docs/providers/README.md`, mirroring how ZCode was added.

### DIFF
--- a/docs/providers/README.md
+++ b/docs/providers/README.md
@@ -44,6 +44,7 @@ For the architectural picture, see `../architecture.md`.
 | [Warp](warp.md) | SQLite | `src/providers/warp.ts` | `tests/providers/warp.test.ts` |
 | [Vercel AI Gateway](vercel-gateway.md) | REST API | `src/providers/vercel-gateway.ts` | `tests/providers/vercel-gateway.test.ts` |
 | [ZCode](zcode.md) | SQLite | `src/providers/zcode.ts` | `tests/providers/zcode.test.ts` |
+| [Hermes](hermes.md) | SQLite | `src/providers/hermes.ts` | `tests/providers/hermes.test.ts` |
 
 ### Shared
 
--- a/docs/providers/hermes.md
+++ b/docs/providers/hermes.md
@@ -1,0 +1,98 @@
+# Hermes
+
+Nous Research Hermes Agent (desktop + CLI), recording per-session usage in a single global SQLite database.
+
+- **Source:** `src/providers/hermes.ts`
+- **Loading:** lazy (`src/providers/index.ts`). Lazy because we read Hermes's SQLite database with `node:sqlite`.
+- **Test:** `tests/providers/hermes.test.ts` (7 tests, fixture-based)
+
+## Where it reads from
+
+Hermes keeps a single global SQLite database for all sessions, CLI and desktop alike.
+
+| Source | Path |
+|---|---|
+| Hermes state db | `~/.hermes/state.db` |
+
+## Storage format
+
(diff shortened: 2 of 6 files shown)


iamtoruk:
Thanks @ozymandiashh, and sorry for the overlap here. Several Hermes provider PRs were in flight at once (#386, #442, #469, and this one).

We landed a SQLite-based Hermes provider in #544 (rebased from #442, then validated end to end against a real `~/.hermes/state.db` with a multi-agent review pass that caught and fixed a `SQLITE_BUSY` cache bug, and confirmed `sessions.cwd` project grouping). Since this PR covers the same ground, closing it in favor of #544.

One thing worth salvaging if it is not already covered: your `src/models.ts` pricing additions. If the bundled pricing tables are missing any Hermes model prices you added here, a small follow-up PR with just those would be very welcome.

Really appreciate the contribution and your consistent work across the provider set.

Issue #515, opened 2026-06-19
Menubar v0.9.13 shows "Couldn't load Today" / dataCorrupted on macOS 26.5


funeconomist:
**Environment**
- codeburn CLI: 0.9.12
- Menubar: v0.9.13
- macOS: 26.5.1 (Build 25F80)
- Node: v25.3.0
**What happens**
After installing via `codeburn menubar`, the menubar app launches but 
immediately shows:
> Couldn't load Today
> decode(DecodingError.dataCorrupted: Data was corrupted. Debug description: 
> The given data was not valid JSON.. Underlying error: Error Domain=NSCoco...
Clicking Retry does not resolve it.
**What I checked**
- `codeburn today --format json` works and outputs valid JSON (~1s)
- `codeburn status --format menubar-json` works and outputs valid JSON (~1s)
- `~/.cache/codeburn/menubar-status.json` was NOT created by the menubar app 
  on launch — had to write it manually
- `~/Library/LaunchAgents/com.codeburn.refresh.plist` was NOT installed by 
  `codeburn menubar`
- App is not sandboxed, no quarantine attributes
- Killed and relaunched the app — same error persists
**Expected**
Menubar loads today's usage data on launch.
**Suspected cause**
The menubar app may not be successfully running the CLI as a subprocess 
(possibly due to the app's subprocess PATH not including /opt/homebrew/bin), 
and falls back to reading a cache file that doesn't exist yet, producing the 
decode error.

ozymandiashh:
Dug into this on the Swift side. The `DecodingError.dataCorrupted` you see comes from the LIVE CLI path, not the cache or PATH:

- In `DataClient.fetch` (mac/Sources/CodeBurnMenubar/Data/DataClient.swift), a non-zero CLI exit throws `nonZeroExit(code:stderr:)`, and only `try JSONDecoder().decode(MenubarPayload.self, from: result.stdout)` throws `dataCorrupted`. So the CLI is being found and is exiting 0, but its **stdout is empty or not valid JSON in the GUI-subprocess context** (a `JSONDecoder` on empty `Data()` yields exactly "The given data was not valid JSON").
- This means the PATH theory likely isn't it: a PATH miss would surface as `nonZeroExit` (127), not `dataCorrupted`. `CodeburnCLI` already probes `/opt/homebrew/bin` + `/usr/local/bin`.
- The missing cache file isn't the source either: `MenubarStatusCache` returns `nil` for a missing file (it never reaches a decode).
[...]

PR #546, opened 2026-06-21
fix(menubar): surface CLI stdout/stderr on payload decode failure (#515)

ozymandiashh:
Addresses #515. When the CLI exits 0 but its stdout fails to decode as `MenubarPayload` (the `dataCorrupted` case the reporter hit on Node v25, where a banner leaked onto stdout before the JSON), the app surfaced only an opaque `decode(DecodingError.dataCorrupted ...)`. The only way to diagnose it was to manually wrap the CLI and capture its raw output.

This makes the failure self-diagnosing. `DataClientError.decode` now carries a bounded (<= 4 KB) preview of the raw stdout, the stderr, and the total stdout byte count, rendered through a `CustomStringConvertible` conformance. The existing `String(describing:)` sink in `AppStore` then shows the offending bytes instead of an opaque message, so a recurrence (a future Node banner, a truncated payload, an empty stdout) is self-evident from the error text.

Adds `DataClientDecodeErrorTests` covering:
- the surfaced message includes the stdout preview and the stderr, and
- the preview is capped at 4 KB (bytes beyond the cap are not embedded).

Also confirmed, with output from the built CLI, that the `menubar-json` path already writes only the JSON payload to stdout (`src/main.ts` does a single `console.log(JSON.stringify(payload))`), so no CLI change was needed here. If a future Node version reintroduces a stdout banner, this error now reports exactly what landed there.

Scope: `mac/Sources/CodeBurnMenubar/Data/DataClient.swift` plus the new test. `swift build` passes locally; the full `swift test` suite runs in CI.

### DIFF
--- a/mac/Sources/CodeBurnMenubar/Data/DataClient.swift
+++ b/mac/Sources/CodeBurnMenubar/Data/DataClient.swift
@@ -6,13 +6,14 @@ import Foundation
 /// Pipe file descriptors pinned forever.
 private let maxPayloadBytes = 20 * 1024 * 1024
 private let maxStderrBytes = 256 * 1024
+private let maxDecodeStdoutPreviewBytes = 4 * 1024
 private let spawnTimeoutSeconds: UInt64 = 45
 private let maxConcurrentSpawns = 6
 
 enum DataClientError: Error {
     case spawn(String)
     case nonZeroExit(code: Int32, stderr: String)
-    case decode(Error)
+    case decode(Error, stdoutPreview: String, stderr: String, stdoutBytes: Int)
     case timeout
     case outputTooLarge
 }
@@ -46,7 +47,12 @@ struct DataClient {
         do {
             return try JSONDecoder().decode(MenubarPayload.self, from: result.stdout)
         } catch {
-            throw DataClientError.decode(error)
+            throw DataClientError.decode(
+                error,
+                stdoutPreview: decodeFailureStdoutPreview(from: result.stdout),
+                stderr: result.stderr,
+                stdoutBytes: result.stdout.count
+            )
         }
     }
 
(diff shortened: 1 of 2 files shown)


iamtoruk:
Ahahaha same timing.. seems like we both are active at same moment :D

iamtoruk:
Thanks @ozymandiashh, and apologies for the overlap here. An equivalent self-diagnosing decode fix landed in #547 at nearly the same time you opened this: `DataClientError.decode` now carries a bounded stdout preview, byte count, and stderr via `CustomStringConvertible`, the same approach you took here (we capped at 2 KB vs your 4 KB, functionally the same). #547 also unblocked the Swift test target, which had stopped compiling after #510 added `codexCredits` to `CurrentBlock`.

Your diagnosis in #515 was the real work and is what made either fix possible, so full credit there. Your `DataClientDecodeErrorTests` is a touch more thorough than what #547 shipped; happy to fold the extra coverage in as a follow-up.

Closing as superseded by #547. Thanks again for both the diagnosis and the fix.

iamtoruk:
@ozymandiashh  can you please check the codeburn dashboard and codeburn device feature they seem major one once you confirm, I will release them. Its already on main

ozymandiashh:
Sure, working on it, I think there might be a bug, trying to figure it out. Ill keep you posted.

ozymandiashh:
I verified the dashboard/devices release path on the current branch and opened #554 for the one server-lifecycle bug I found.

What passed:
- Dashboard loads real data for Today / 7 days / 30 days / Month / All.
- Manual Chrome check confirmed each period reaches the expected label and does not stay stuck in Loading.
- Devices/share endpoints stay alive after bad input.
- Targeted tests, TypeScript, and build pass.

What #554 fixes:
- bad `period` or date query params on `/api/usage`, `/api/devices`, and share usage requests now return JSON 400 instead of letting a server path call `process.exit(1)`.
- CLI behavior remains unchanged: invalid `--period` still exits non-zero.

I do not have a second Mac, so the only thing I could not physically validate is mDNS discovery between two real machines. Pairing/transport/API behavior was exercised locally.

Also opened #553 for the #515 Antigravity token-count/menubar hardening follow-up.

ozymandiashh:
Dashboard/devices verification update: I opened #554 for the two issues found while checking the release candidate.

What I verified manually:
- Dashboard period tabs load distinct, correct data for Today, 7 days, 30 days, Month, and All.
- Invalid `/api/usage` and `/api/devices` period params return JSON 400 and the server remains alive.
- Devices no longer show stale previous-period data while a new period is selected.
- Loopback device pairing/share/remove works end-to-end. I do not have a second Mac, so I could not physically verify mDNS discovery across two separate Macs.

Automated validation on #554:
- 58 focused tests passing across cli-date, dashboard, and sharing
- TypeScript check passing
- Full build passing
- GitHub checks green

PR #553, opened 2026-06-22
fix(menubar): clamp unsafe Antigravity token counts

ozymandiashh:
Addresses #515.

This hardens the Antigravity ingestion path so unsafe or unstable token counts cannot break the menubar payload or Swift decoding.

What changed:
- Parse Antigravity token counts through a safe integer path, including safe numeric strings and bigints, while rejecting unsafe, negative, fractional, and malformed values.
- Preserve output-only and split-output metadata correctly, including the case where total output is present, thinking output is present, and response output is omitted.
- Add a final token-sanitization pass before menubar JSON emission.
- Make the Swift menubar payload decoder tolerant for token fields while keeping non-token fields strict.
- Invalidate stale Antigravity caches after the token-safety change and include SQLite `-wal` / `-shm` sidecars in the DB cache fingerprint.

Validation:
- `npm test -- tests/providers/antigravity.test.ts tests/session-cache.test.ts tests/daily-cache.test.ts tests/menubar-json.test.ts --run` — 127 passing
- `npx tsc --noEmit`
- `npm run build`
- `cd mac && swift build`


ozymandiashh:
Updated after the cache-preservation review pass.

One extra edge case is now covered: if an existing `session-cache.json` contains an unsafe Antigravity token count, `loadCache()` now drops only the invalid provider/file section and preserves valid sections such as durable Copilot cache entries. That avoids losing durable telemetry while still flushing the corrupt Antigravity data.

Validation after the update:
- `npm test -- tests/session-cache.test.ts tests/providers/antigravity.test.ts tests/menubar-json.test.ts tests/daily-cache.test.ts --run` — 134 tests passed
- `npx tsc --noEmit` — passed
- `cd mac && swift build` — passed
- `git diff --check` — passed

PR is mergeable and current checks are green.

iamtoruk:
Closing: the target issue #515 is resolved/closed, and at ~1,000 lines this now conflicts with the menubar payload/aggregator rewrites on main. If unsafe Antigravity token counts still surface, an issue with a repro plus a focused clamp PR would be welcome.

PR #554, opened 2026-06-22
fix(web): reject invalid dashboard periods without exiting

ozymandiashh:
Follow-up from the dashboard/devices release check.

A bad `period` query should not be able to terminate a long-running web/share server. The CLI should still fail loudly for invalid `--period`, but HTTP handlers need to return an error response and keep serving.

What changed:
- Added a non-exiting period parser for server paths.
- `/api/usage` and `/api/devices` now return JSON 400 for invalid period/date query params instead of exiting the process.
- The share server maps invalid usage-query inputs to HTTP 400 instead of treating them as internal errors.
- CLI behavior remains unchanged: invalid `--period` still prints the error and exits non-zero.

Validation:
- `npm test -- tests/cli-date.test.ts tests/dashboard.test.ts tests/sharing --run` — 57 passing
- `npx tsc --noEmit`
- `npm run build`
- Manual Chrome check across Today / 7 days / 30 days / Month / All: each period reached the expected label with no stuck Loading state.
- Manual API check: invalid `/api/usage` and `/api/devices` periods returned 400 JSON; `/api/identity` still returned 200 afterwards.

### DIFF
--- a/dash/src/App.tsx
+++ b/dash/src/App.tsx
@@ -1,5 +1,5 @@
 import { useEffect, useMemo, useState, type ReactNode } from 'react'
-import { keepPreviousData, useQuery, useQueryClient } from '@tanstack/react-query'
+import { useQuery, useQueryClient } from '@tanstack/react-query'
 
 import {
   approvePairing,
@@ -354,7 +354,6 @@ export function App() {
   const { data, isError, error, refetch } = useQuery({
     queryKey: ['devices', period, provider],
     queryFn: () => fetchDevices(period, provider),
-    placeholderData: keepPreviousData,
     // When devices are paired, re-pull periodically so a device that briefly
     // dropped (asleep/network blip) reappears on its own instead of staying
     // gone until you switch tabs.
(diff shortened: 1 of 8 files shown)


ozymandiashh:
Updated and re-verified after rebasing on current `main`.

Final validation:
- `npm test -- tests/cli-date.test.ts tests/sharing/transport.test.ts --run` — 26 tests passed
- `npx tsc --noEmit` — passed
- `npm run build` — passed (CLI + dashboard)
- manual Chrome dashboard QA after the final build:
  - `/api/usage?period=not-a-period` returns 400
  - `/api/devices?period=not-a-period` returns 400
  - server remains alive after invalid input (`/api/usage?period=today` returns 200)
  - Month / Today / 7 days / 30 days / All all load to their expected period labels without staying in Loading
  - observed final-load timings on my large local dataset: Month ~0.5s, Today ~3.5s, 7 days ~7.1s, 30 days ~9.1s, All ~10.1s

I also checked the dashboard visually with screenshots for each period after the rebase. I do not have a second Mac available, so I could not physically validate mDNS discovery between two machines; local pairing/transport/API behavior was exercised.

PR is mergeable and current checks are green.

Issue #561, opened 2026-06-22
Spend budgets with alerts (overview display + scriptable --check)


ozymandiashh:
## Summary

codeburn is great at showing where AI spend goes, but it stops at observation. I would like it to help me ACT on spend: set a budget and get told when I am over it. I have this implemented and validated against my own usage, and I am filing first to align on the UX and scope before the PR lands.

There is no spend-budget feature today (`context-budget.ts` is the model context window, unrelated).

## Proposed UX

Set a budget per cadence, in your display currency:

```bash
codeburn budget --monthly 100     # also --weekly / --daily
codeburn budget --list
codeburn budget --remove monthly
```

Two ways it surfaces:

1. **In `overview`**, a single line under the totals, only when the period matches a configured budget (today -> daily, last-7-days -> weekly, this-month -> monthly):

   ```
   Monthly budget: €67.40 of €100.00 (67%)  projected €98.20 by month end
   ```

   colored green under 80%, amber from 80%, red at 100%+. The projection is a simple linear run-rate for the in-progress period.

2. **`codeburn budget --check`** for scripting: it prints each configured budget's status and exits non-zero if any is over budget, so you can wire it into a shell prompt, a pre-commit, or a cron/CI step and get a real alert without codeburn shipping any cross-platform notification machinery.

```bash
codeburn budget --check || notify-send "Over AI budget"
```

## Design notes (where I would value your steer)

- **No notification daemon.** The alert mechanism is the `--check` exit code, which keeps the feature small and composable and leaves the choice of how to be notified to the user. Happy to revisit if you would rather it integrate elsewhere (e.g. the menubar).
- **Budgets are in the display currency** (the one `currency` sets and `overview` shows), compared against converted spend.
- **The overview budget line is suppressed when a filter is active** (`--provider` / `--project` / `--exclude`), because a global budget next to a filtered subset would be misleading.
- **Period windows match `overview`** exactly: `budget --check` reuses the same date-range helper per cadence, so the numbers you see in `overview` and in `--check` agree.
- **Default behavior is unchanged**: with no budget configured, `overview` output and every other command are byte-for-byte identical.

If this UX looks right I will open the PR (one new `src/budget.ts`, a `budget` command, and an overview line, with unit/CLI/render tests). Glad to adjust naming, thresholds, or where it surfaces.

iamtoruk:
[...]

PR #562, opened 2026-06-22
feat(budget): spend budgets with overview display and scriptable --check

ozymandiashh:
Implements #561.

## What this adds

A spend-budget feature, so codeburn does not just show where AI spend goes but lets you act on it: set a daily/weekly/monthly budget, see budget usage with a run-rate projection in `overview`, and get a scriptable over-budget signal from `budget --check`. There is no spend-budget feature today (`context-budget.ts` is the model context window, unrelated).

```bash
codeburn budget --monthly 100      # also --weekly / --daily, in your display currency
codeburn budget --list
codeburn budget --remove monthly
codeburn budget --check            # exits non-zero if any configured budget is over


## How it surfaces

1. **`overview`** shows one line under the totals when the period matches a configured cadence (today -> daily, last-7-days -> weekly, this-month -> monthly):

   ```
   Monthly budget: €67.40 of €100.00 (67%)  projected €98.20 by month end
   ```

   green under 80%, amber from 80%, red at 100%+, with a linear run-rate projection for the in-progress period.

2. **`budget --check`** prints each configured budget's status and exits 1 if any is over budget. It is well suited to cron or CI:

   ```bash
   codeburn budget --check || notify-send "Over AI budget"
   ```

   It can also be used in pre-commit hooks or a shell prompt, but each configured tier reruns discovery, so a frequently rendered prompt may be expensive.

## Design decisions

- **The alert is the `--check` exit code**, not a notification daemon. That keeps the feature small and composable and leaves the choice of how to be notified to the user.
- **Budgets are in the display currency** (the one `currency` sets and `overview` shows). Spend is summed in USD and converted once via `convertCost` before comparison, so the line reads in the same units the user typed.
- **Budgets count the full would-be API cost, including subscription-proxied spend.** They use `project.totalCostUSD`, without subtracting `totalProxiedCostUSD`, which is consistent with the overview Cost total.
- **The overview budget line is suppressed when a filter is active** (`--provider`, `--project`, `--exclude`). A global budget shown next to a filtered subset would read as falsely-green, so it is hidden rather than misleading.
- **Projection applies to the monthly tier only.** The daily and weekly windows are rolling and complete by definition (today, last 7 days), so a run-rate projection is meaningless there; only the in-progress calendar month projects.
- **`budget --check` uses the same date-range source as `overview`** per cadence (daily -> `today`, weekly -> `week`, monthly -> `month`), so the spend window and the numbers agree between the two surfaces.
- **Default behavior is unchanged.** With no budget configured, `overview` output and every other command are byte-for-byte identical. The feature is purely additive.

## Implementation

- `src/budget.ts` (new): a pure `computeBudgetStatus` returning `{ spent, budget, pct, projected, state }`, where `projected = spent * totalDays / elapsedDays` and `state` is `under` (<80%), `warn` (>=80%), or `over` (>=100%). No I/O, fully unit-tested.
- `src/config.ts`: a new optional `budget` field (`{ daily?, weekly?, monthly? }`, display-currency amounts).
- `src/main.ts`: the `budget` command (set with finite > 0 validation, `--list`, `--remove`, `--check` that aggregates current-period spend via the existing session pipeline) and the overview wiring that builds the budget status for the matching tier.
- `src/overview.ts`: the budget line in `renderOverview`, colored by state, with the projection only while the period is in progress. Whole-number percentages are floored so the display cannot read 100% before the raw state is over.

## Verification

- Unit tests (`tests/budget.test.ts`): threshold boundaries, exact projection math, invalid-input guards.
- CLI tests (`tests/cli-budget.test.ts`, isolated HOME): set/list/remove; invalid amount exits 1; `--check` exits 1 over and 0 under-or-none; 99.x% floors to 99% with WARN and exit 0; the budget line is suppressed under filters; `--check` and `overview -p week` report the same weekly spend window.
- Render tests (`tests/overview.test.ts`): the line appears with the right percent and projection when a budget is set, is absent otherwise, and 99.6% displays as 99% while state remains warn.
- `npx vitest run tests/budget.test.ts tests/cli-budget.test.ts tests/overview.test.ts` passes with 19 tests. `npx tsc --noEmit` is clean.
- Full `npx vitest run` reaches 1,749 passing tests and 5 skipped tests, but the existing `tests/cli-status-menubar.test.ts` selected-config-source case exceeds its 5-second timeout under full parallel load. That file passes all 13 tests in isolation, including the same case in 3,496ms.

## Files

- `src/budget.ts` (new), `src/config.ts`, `src/main.ts`, `src/overview.ts`
- `tests/budget.test.ts` (new), `tests/cli-budget.test.ts` (new), `tests/overview.test.ts`

### npx tsc --noEmit

```text
$ perl -e '$seconds = shift @ARGV; alarm $seconds; exec @ARGV' 600 npx tsc --noEmit
exit: 0


### npx vitest run tests/budget.test.ts tests/cli-budget.test.ts tests/overview.test.ts

```text
$ perl -e '$seconds = shift @ARGV; alarm $seconds; exec @ARGV' 300 npx vitest run tests/budget.test.ts tests/cli-budget.test.ts tests/overview.test.ts
exit: 0

 RUN  v3.2.6 /Users/husamsoboh/codeburn/.architect/wt/fix0716-02

 ✓ tests/budget.test.ts (3 tests) 3ms
 ✓ tests/overview.test.ts (9 tests) 41ms
 ✓ tests/cli-budget.test.ts (7 tests) 15241ms
   ✓ codeburn budget command > saves, lists, and removes a monthly budget  2617ms
   ✓ codeburn budget command > rejects an invalid amount without writing a budget  1440ms
   ✓ codeburn budget command > exits 1 when the current month is over budget  1794ms
   ✓ codeburn budget command > floors a 99.x percent budget check without reporting over  1468ms
   ✓ codeburn budget command > exits 0 when the current month is under budget or no budget is configured  2496ms
   ✓ codeburn budget command > keeps overview budget lines only on unfiltered overviews  3264ms
   ✓ codeburn budget command > uses the same weekly spend window for budget --check and overview -p week  2161ms

 Test Files  3 passed (3)
      Tests  19 passed (19)
   Start at  18:55:09
   Duration  16.20s (transform 319ms, setup 65ms, collect 428ms, tests 15.29s, environment 1ms, prepare 364ms)

### DIFF
--- a/src/budget.ts
+++ b/src/budget.ts
@@ -1,0 +1,62 @@
+const MS_PER_DAY = 24 * 60 * 60 * 1000
+
+export type BudgetTier = 'daily' | 'weekly' | 'monthly'
+export type BudgetState = 'under' | 'warn' | 'over'
+
+export type BudgetStatus = {
+  spent: number
+  budget: number
+  pct: number
+  projected: number
+  state: BudgetState
+}
+
+export type BudgetStatusInput = {
+  spent: number
+  budget: number
+  elapsedDays: number
+  totalDays: number
+}
+
+function requireFinitePositive(name: string, value: number): void {
+  if (!Number.isFinite(value) || value <= 0) {
+    throw new RangeError(`${name} must be a finite number greater than 0`)
+  }
+}
+
+function requireFiniteNonNegative(name: string, value: number): void {
+  if (!Number.isFinite(value) || value < 0) {
+    throw new RangeError(`${name} must be a finite number greater than or equal to 0`)
+  }
+}
+
+function toDayIndex(d: Date): number {
+  return Math.floor(Date.UTC(d.getFullYear(), d.getMonth(), d.getDate()) / MS_PER_DAY)
+}
+
+export function diffCalendarDays(from: Date, to: Date): number {
+  return toDayIndex(to) - toDayIndex(from)
+}
+
+export function daysInMonth(date: Date): number {
+  return new Date(date.getFullYear(), date.getMonth() + 1, 0).getDate()
+}
+
+export function computeBudgetStatus(input: BudgetStatusInput): BudgetStatus {
+  requireFiniteNonNegative('spent', input.spent)
+  requireFinitePositive('budget', input.budget)
+  requireFinitePositive('elapsedDays', input.elapsedDays)
+  requireFinitePositive('totalDays', input.totalDays)
+
+  const pct = (input.spent / input.budget) * 100
+  const projected = input.spent * input.totalDays / input.elapsedDays
+  const state: BudgetState = pct >= 100 ? 'over' : pct >= 80 ? 'warn' : 'under'
+
+  return {
+    spent: input.spent,
+    budget: input.budget,
+    pct,
+    projected,
+    state,
+  }
+}
--- a/src/config.ts
+++ b/src/config.ts
@@ -44,6 +44,12 @@ export type CodeburnConfig = {
   // can show "saved $X by running locally". Distinct from modelAliases which
   // rewrites actual spend.
   localModelSavings?: Record<string, string>
+  // Spend budgets are stored in the configured display currency, not USD.
+  budget?: {
+    daily?: number
+    weekly?: number
+    monthly?: number
+  }
   // Absolute directory prefixes whose Claude Code sessions are routed through a
   // subscription-backed LLM proxy (e.g. GitHub Copilot via ANTHROPIC_BASE_URL;
   // tools like claude-code-over-github-copilot / claudegate). The JSONL records
--- a/src/main.ts
+++ b/src/main.ts
@@ -5,7 +5,7 @@ import { exportCsv, exportJson, type PeriodExport } from './export.js'
 import { findUnpricedModels, loadPricing, setModelAliases, setPriceOverrides, setLocalModelSavings, setProxyPaths, normalizeProxyPath } from './models.js'
 import { parseAllSessions, filterProjectsByName, filterProjectsByDateRange, clearSessionCache } from './parser.js'
 import { allProviderNames, getAllProviders } from './providers/index.js'
-import { convertCost } from './currency.js'
+import { convertCost, formatCost } from './currency.js'
 import { renderStatusBar } from './format.js'
 import { toDateString } from './daily-cache.js'
 import { dateKey } from './day-aggregator.js'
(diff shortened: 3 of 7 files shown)


iamtoruk (changes requested):
The feature itself reviewed well: tight scope, atomic config writes through saveConfig, sound timezone and rollover math, and thorough boundary tests. It also coexists cleanly with the guard feature, different granularity and storage, so no conflict there.

The only hard blocker is that the branch no longer applies: main has moved and src/overview.ts changed in exactly the regions this PR touches (the models.js import line and the block after Savings, plus the formatCost/formatTokens area from the locale-grouping change). GitHub reports it as conflicting. Please rebase onto current main and re-run the suite.

While you are in there, one design note worth a decision or at least a line in the PR body: the budget counts the full would-be API cost including subscription-proxied spend (project.totalCostUSD, not totalCostUSD minus totalProxiedCostUSD). For proxy users the budget can trip while out-of-pocket is far under. Internally consistent with the overview Cost total, so keeping it is defensible, but say so explicitly.

Two nits, non-blocking: display rounds the percent while the state uses the raw value, so 99.6 percent renders as (100 percent) in warn color rather than over; and --check re-runs discovery once per configured tier, fine for cron or CI but heavy if someone wires it into a shell prompt as the body suggests, maybe soften that suggestion.

ozymandiashh:
Rebased onto current main (48a9241) and force-pushed as c63f148. Changes since your review:

- **Conflicts resolved**: the overview.ts drift (models.js import, post-Savings block, locale-grouping formatters) is merged keeping both main's en-US grouped formatting with the unpriced-model warning block and the budget line after it.
- **Proxied-spend note**: the PR body now states explicitly that budgets count the full would-be API cost including subscription-proxied spend, consistent with the overview Cost total.
- **Percent boundary fixed**: display floors the raw percent on both surfaces, so 99.6 percent renders as 99 percent in warn and 100 percent appears only when the state is over; regression tests cover the boundary on overview and on budget --check.
- **Shell-prompt suggestion softened** in the body: --check reruns discovery per configured tier, fine for cron or CI, expensive for a frequently rendered prompt.

Suite is green locally on the rebased tree (the cli-status-menubar 5s timeout is a pre-existing load-sensitivity, passes 13/13 in isolation and on a quiet machine).

iamtoruk (approved):
Re-reviewed. Rebase is clean and the budget block now sits correctly after the unpriced-models section, the proxied-cost semantics are documented in the body, and both nits got real fixes with regression tests (floor instead of round on both surfaces). One optional line for the body since main moved after your push: budget spend also includes estimated costs (the #688 markers are display-only metadata), which is consistent with the overview total. Ready to merge.

Issue #566, opened 2026-06-23
feat(menubar): aggregate metrics across all shared/synced devices in menubar indicator


pulver22:
When utilising the multi-device sync feature (e.g., syncing a local laptop/MacBook with a remote workstation via codeburn devices add), the web-based/terminal dashboard correctly displays the aggregated, cumulative token counts and costs under the "Combined" row.

However, the desktop menubar indicator applet only displays metrics reflecting the local machine's standalone usage, ignoring data synchronised from peer devices.

I would like the menubar indicator to mirror the behaviour of the main combined dashboard. Ideally, it should display the total global/combined metrics across all connected and paired instances, or offer a configuration toggle to switch views between:
* Local Machine Only
* Combined (All Devices)

ozymandiashh:
on it

ozymandiashh:
PR: #567

Good request. The combined view already exists in the terminal (`codeburn devices`), but the menubar reads `codeburn status --format menubar-json`, which only ever builds **local** usage and never pulls paired devices, so the indicator can't show combined yet.

I'm splitting this in two so the data layer lands cleanly before the UI:

**This PR (data layer):** exposes combined + per-device usage through the menubar-json contract.
- Extracts the per-device summary + combined reduce that was inline in `renderDevices` into a reusable `summarizeDeviceUsage()`, so the JSON and the terminal share one source of truth (terminal output unchanged).
- Adds an optional `combined` block to `MenubarPayload` (per-device list + combined totals, incl. `deviceCount`/`reachableCount` so the UI can show "N of M devices").
- Adds `--scope local|combined` to `status --format menubar-json` (default `local`, fully backward compatible). `combined` pulls paired devices (offline peers degrade to error rows, never block).
- Two correctness guards: combined refuses `--days` and `--provider`/`--project`/`--exclude`, because the device-sharing query only carries a period/range and can't represent non-contiguous days or per-provider/project filters, so peers would return unfiltered usage. Cache tokens are window-scoped to the selected period (they live in a 365-day daily history, so an unscoped sum would pollute a "today" view).

**Follow-up (Swift):** the actual menubar UI, exactly your second option, a Local / Combined toggle, plus the caching/timeout/staleness handling so a combined refresh never blocks the badge on a slow or sleeping peer.

Once the data PR is in I'll wire the toggle. If you'd rather the toggle default to Combined vs Local, say so and I'll match it.

ozymandiashh:
[...]

PR #567, opened 2026-06-23
feat(menubar): expose combined multi-device usage in menubar-json (#566)

ozymandiashh:
## Summary

Addresses #566 (data layer). The macOS menubar reads `codeburn status --format menubar-json`, which only reflects the **local** machine. Combined multi-device usage existed only in the terminal `codeburn devices` command. This exposes combined + per-device usage through the menubar-json contract so the menubar can mirror the combined dashboard. The menubar UI toggle itself is a follow-up Swift PR.

## What changed

- **Refactor:** extract the per-device summary + combined reduce that was inline in `renderDevices` (src/sharing/host.ts) into a reusable pure `summarizeDeviceUsage(results, window?)`. `renderDevices` now formats from it with byte-identical output (existing tests unchanged). Error/unreachable devices are excluded from the combined sums but kept in `perDevice`; `deviceCount` counts all, `reachableCount` counts those that responded.
- **Contract:** optional `combined?: CombinedUsage` on `MenubarPayload` (per-device list + combined totals incl. calls/sessions). Absent by default, so existing consumers are unaffected.
- **CLI:** `--scope local|combined` on `status` (default `local`). `combined` builds the local payload, pulls paired devices via `pullDevices` (per-peer failures isolated into error rows, never block), and attaches the summary.
- **Correctness guards:** reject `--scope combined` with `--days` (non-contiguous, not representable over the sharing `UsageQuery`) and with `--provider`/`--project`/`--exclude` (the sharing query carries no filters, so peers would report unfiltered usage). Window-scope the combined cache-token sum to the selected period (cache lives in the 365-day daily history; `current` carries no cache counts).

## Testing

- `npx vitest run tests/sharing/host.test.ts tests/menubar-json.test.ts tests/cli-status-menubar.test.ts` -> **28/28**. New cases: `summarizeDeviceUsage` combined totals + error-row exclusion + window-scoped cache; `MenubarPayload` omits `combined` by default and accepts the documented shape; CLI attaches `combined` for `--scope combined` (and omits it for local); CLI rejects `--scope combined` with `--days`/`--provider`/`--project`/`--exclude`.
- `npx tsc --noEmit` clean; semgrep hot-path rule -> 0 findings.
- Full suite: the only failing test is the pre-existing `overview.test.ts` thousands-separator case, which fails on non-US locales independent of this change (reproduces on `main`).

## Notes

- Combined is a coarse rollup (cost / calls / sessions / tokens / cache) + per-device breakdown, matching the terminal "Combined" row, not the full per-model/per-project breakdown.
- Follow-up Swift PR: the menubar Local/Combined toggle + caching/timeout so a combined refresh never blocks the badge on a slow peer.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -17,10 +17,11 @@ import { renderOverview } from './overview.js'
 import { runWebDashboard } from './web-dashboard.js'
 import { hostname } from 'os'
 import { runShareServer } from './sharing/share-run.js'
-import { addRemote, linkRemote, pullDevices, renderDevices } from './sharing/host.js'
+import { addRemote, linkRemote, pullDevices, renderDevices, summarizeDeviceUsage } from './sharing/host.js'
 import { browse } from './sharing/discovery.js'
 import { promptChoice } from './sharing/prompt.js'
 import { loadRemotes, saveRemotes } from './sharing/store.js'
+import type { UsageQuery } from './sharing/share-server.js'
 import { formatDateRangeLabel, parseDateRangeFlags, parseDayFlag, parseDaysFlag, getDateRange, toPeriod, type Period } from './cli-date.js'
 import { runOptimize } from './optimize.js'
 import { renderCompare } from './compare.js'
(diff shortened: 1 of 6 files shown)


ozymandiashh:
@iamtoruk I think you might be interested in these, I think you are eligible. 
https://claude.com/contact-sales/claude-for-oss 
https://openai.com/ro-RO/form/codex-for-oss/

iamtoruk:
@ozymandiashh For Anthropic, we have already been selected, and I am in their OSS program. For OpenAI, I have applied. Let's see.

ozymandiashh:
congrats

PR #570, opened 2026-06-26
fix(devin): add missing support for ATIF v1.7

### LINKED ISSUE #525: devin schema version changed to ATIF-v1.7

tvcsantos:
## Summary

Devin now outputs in ATIF v1.7 format. This PR addresses that supporting the new format, while keeping backward compatible with the existing versions.

This PR relates and completes #526. Added @bmcdonough as co-author since the work was based on his work on the PR and also since he was actually the one finding the issues. Thanks for that @bmcdonough.

Closes #525 

## Testing

- [x] `npm test` passes

### For new providers only:

### DIFF
--- a/docs/providers/devin.md
+++ b/docs/providers/devin.md
@@ -4,7 +4,7 @@ Cognition Devin CLI local usage tracking.
 
 - **Source:** `src/providers/devin.ts`
 - **Loading:** eager (`src/providers/index.ts`)
-- **Test:** `tests/providers/devin.test.ts` (336 lines)
+- **Test:** `tests/providers/devin.test.ts`
 
 ## Where it reads from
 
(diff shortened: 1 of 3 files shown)


bmcdonough:
@tvcsantos Thank you for your efforts getting this over the finish line.  I haven't had time to get back to this.

iamtoruk:
Thanks for this, @tvcsantos (and @bmcdonough) — solid, well-tested ATIF v1.7 support. 🙌

I pushed one small follow-up commit on top with a few review nits:

- **Dropped the `@modelcontextprotocol/sdk` `JsonSchemaType` import** and the `ToolDefinition`/`FunctionDefinition` types it served. They only existed to type `Agent.tool_definitions`, which the parser never reads — so this removes an unnecessary dependency from the Devin provider. `tool_definitions` is now `unknown`.
- **Removed `isImageContentPart`** — it was unused, and its `"image" in part` check could never be true (image parts carry `source`/`type`, not `image`).
- `isTextContentPart` now uses the `type` discriminant instead of property presence.
- **`getMetricsFromStep`**: falls back to legacy `metadata.metrics` when `step.metrics` is present but empty, so a partial metrics object can't silently zero out usage.
- Added two tests (empty-`step.metrics` fallback, image-only message) and fixed the title typo.

Validation: `tsc --noEmit` clean, full suite **1313/1313 passing**, `tsup` build OK. Merging now.

tvcsantos:
> * **Dropped the `@modelcontextprotocol/sdk` `JsonSchemaType` import** and the `ToolDefinition`/`FunctionDefinition` types it served. They only existed to type `Agent.tool_definitions`, which the parser never reads — so this removes an unnecessary dependency from the Devin provider. `tool_definitions` is now `unknown`.

Makes sense, it was not used anyway 👍 

> * **Removed `isImageContentPart`** — it was unused, and its `"image" in part` check could never be true (image parts carry `source`/`type`, not `image`).

Ups :D, also it was not working as expected yup 👍 

> * `isTextContentPart` now uses the `type` discriminant instead of property presence.

Yup that is it. The method was broken. Should have been like you did 👍 

> * **`getMetricsFromStep`**: falls back to legacy `metadata.metrics` when `step.metrics` is present but empty, so a partial metrics object can't silently zero out usage.
> * Added two tests (empty-`step.metrics` fallback, image-only message) and fixed the title typo.

Makes sense 👍 

Thanks for reviewing it.

Issue #574, opened 2026-06-29
Token amounts not accurate


Joon:
We are rolling out codeburn for agentic coding tracking for our developer population and have feedback from many of them that the numbers shown are not accurate. 

I tested this morning and codeburn is definitely not picking up all usage.

We use Cursor as the coding tool, and most of our devs use macbooks.

Test evidence:
Cursor token usage screenshot from admin portal for the current user (Millions of tokens shown):

<img width="1830" height="1118" alt="Image" src="https://github.com/user-attachments/assets/fdf8d486-6dc4-4fe0-87cd-c93916dcdd9f" />

Codeburn daily usage screen for the current user (thousands of tokens shown):

<img width="1830" height="1118" alt="Image" src="https://github.com/user-attachments/assets/f155eb4f-5200-4d84-b6bb-4549131f1b8d" />

iamtoruk:
Thanks again for the clear report, the screenshots made this easy to chase. Here's the honest picture, plus a branch to test.

**What's happening:** Cursor stores each chat in a local SQLite DB, but on current builds the per-message token count is written as `{input: 0, output: 0}`. So codeburn was estimating Cursor tokens rather than reading them, which is why the numbers came in low.

Cursor does record its own tokenizer-accurate **context size** per conversation in `composerData.promptTokenBreakdown` (the number behind the in-app context-window bar). codeburn just wasn't reading it.

The honest limit: we searched the entire local Cursor footprint (global and per-workspace `state.vscdb`, `ai-code-tracking.db`, logs, LevelDB, `~/.cursor`). The exact billed figures are **not stored on disk** anywhere. Only the input/context count is local; output tokens, cache tokens, and the cumulative-per-turn total your admin portal bills are computed server-side and never written locally. That is inherent to usage-based billing.

**What we changed (branch in #575, draft):**

- **Input tokens** now come from `promptTokenBreakdown` instead of an estimate. On a test machine, one day's Cursor input went from ~44.9k to ~168.5k tokens, matching the real per-conversation context.
- **Tools + bash:** codeburn now surfaces the agent's tool calls (Read, Grep, Glob, Shell) and the actual shell commands it ran, pulled from the local agent log.
- **Pricing:** `composer-2.5` was missing a price mapping (showed $0); it is now priced like `composer-2`.

It gets **much closer** but will not match the portal exactly: it is the latest context snapshot (not cumulative-per-turn), output stays a reply-text estimate, and cache tokens are not on disk.

[...]

PR #575, opened 2026-06-29
fix(cursor): use Cursor's real context tokens for input (#574)

iamtoruk:
## What

Targets #574 (Cursor token amounts way under the admin portal).

Current Cursor builds store the per-bubble `tokenCount` as `{0,0}`, so codeburn could only estimate Cursor input from visible text (plus an `agentKv` content-char pass that double-counted the same conversation). Cursor actually records its own tokenizer-accurate context size per conversation in `composerData.promptTokenBreakdown.totalUsedTokens`, the number behind the in-app context-window bar. This reads that and credits it once per conversation for input.

## Result (measured on a real local DB)

| | input tokens | output tokens |
|---|---|---|
| before | 44,873 | 18,087 |
| after | **168,486** | 8,716 |

The new input matches the sum of the real per-conversation context.

## Known limits (this is why it's a draft, please validate)

We searched the entire local Cursor footprint (global + workspace `state.vscdb`, `ai-code-tracking.db`, logs, LevelDB, `~/.cursor`). The exact figures are **not stored locally**, only Cursor's client-side input/context count is. So:

- It's the **latest context snapshot** per conversation, not cumulative-per-turn, so a long multi-turn chat still undercounts versus the admin portal.
- **Output** is a reply-text estimate (no output count exists on disk).
- **Cache** tokens are server-side only.
- `composer-2.5` and other Cursor model names need price mappings (cost shows 0 for those).
- Exact parity with the admin portal needs an opt-in Cursor API key (planned follow-up).

`agentKv` parsing is retained for a tools + bash-command breakdown in a follow-up.

## For testers

Compare codeburn's Cursor numbers on this branch against your Cursor admin portal. This read-only query reports the real on-disk context total (aggregate only, no chat content) so you can see local vs codeburn vs admin:

```bash
DB="$HOME/Library/Application Support/Cursor/User/globalStorage/state.vscdb"
sqlite3 "$DB" "SELECT COALESCE(SUM(CAST(json_extract(value,'\$.promptTokenBreakdown.totalUsedTokens') AS INT)),0) AS real_context_tokens FROM cursorDiskKV WHERE key LIKE 'composerData:%';"
```
(If it says "database is locked," quit Cursor and rerun.)

### DIFF
--- a/src/providers/cursor.ts
+++ b/src/providers/cursor.ts
@@ -1,10 +1,11 @@
-import { existsSync, readdirSync, readFileSync } from 'fs'
+import { existsSync, readdirSync, readFileSync, statSync } from 'fs'
 import { join } from 'path'
 import { homedir } from 'os'
 
 import { calculateCost } from '../models.js'
+import { extractBashCommands } from '../bash-utils.js'
 import { readCachedResults, writeCachedResults } from '../cursor-cache.js'
-import { isSqliteAvailable, getSqliteLoadError, openDatabase, blobToText, type SqliteDatabase } from '../sqlite.js'
+import { isSqliteAvailable, isSqliteBusyError, getSqliteLoadError, openDatabase, blobToText, type SqliteDatabase } from '../sqlite.js'
 import type { DateRange } from '../types.js'
 import type { Provider, SessionSource, SessionParser, ParsedProviderCall } from './types.js'
 
@@ -48,7 +49,7 @@ type BubbleRow = {
   output_tokens: number | null
   model: string | null
   created_at: string | null
-  conversation_id: string | null
+  request_id: string | null
   user_text: Uint8Array | string | null
   text_length: number | null
   bubble_type: number | null
@@ -59,11 +60,18 @@ type BubbleRow = {
 }
 
 type AgentKvRow = {
-  key: string
   role: string | null
   content: Uint8Array | string | null
   request_id: string | null
-  content_length: number
+  model: string | null
+}
+
+// SQLITE_BUSY must reach parser.ts, whose busy path skips the source without
+// caching; swallowing it here would stamp a silently degraded parse into the
+// results cache under an unchanged DB fingerprint (Cursor writes via WAL, so
+// contention does not change the main file's stat).
+function rethrowBusy(err: unknown): void {
+  if (isSqliteBusyError(err)) throw err
 }
 
 const CHARS_PER_TOKEN = 4
@@ -298,7 +306,7 @@ const BUBBLE_QUERY_BASE = `
     json_extract(value, '$.tokenCount.outputTokens') as output_tokens,
     json_extract(value, '$.modelInfo.modelName') as model,
     json_extract(value, '$.createdAt') as created_at,
-    json_extract(value, '$.conversationId') as conversation_id,
+    json_extract(value, '$.requestId') as request_id,
     CAST(substr(json_extract(value, '$.text'), 1, 500) AS BLOB) as user_text,
     length(json_extract(value, '$.text')) as text_length,
     json_extract(value, '$.type') as bubble_type,
@@ -309,11 +317,10 @@ const BUBBLE_QUERY_BASE = `
 
 const AGENTKV_QUERY = `
   SELECT
-    key,
     json_extract(value, '$.role') as role,
     CAST(json_extract(value, '$.content') AS BLOB) as content,
     json_extract(value, '$.providerOptions.cursor.requestId') as request_id,
-    length(value) as content_length
+    json_extract(value, '$.providerOptions.cursor.modelName') as model
   FROM cursorDiskKV
   WHERE key LIKE 'agentKv:blob:%'
     AND hex(substr(value, 1, 1)) = '7B'
@@ -356,7 +363,7 @@ const BUBBLE_QUERY_PAGE = `
     json_extract(value, '$.tokenCount.outputTokens') as output_tokens,
     json_extract(value, '$.modelInfo.modelName') as model,
     json_extract(value, '$.createdAt') as created_at,
-    json_extract(value, '$.conversationId') as conversation_id,
+    json_extract(value, '$.requestId') as request_id,
     CAST(substr(json_extract(value, '$.text'), 1, 500) AS BLOB) as user_text,
     length(json_extract(value, '$.text')) as text_length,
     json_extract(value, '$.type') as bubble_type,
@@ -373,7 +380,8 @@ function validateSchema(db: SqliteDatabase): boolean {
       "SELECT COUNT(*) as cnt FROM cursorDiskKV WHERE key LIKE 'bubbleId:%' LIMIT 1"
     )
     return rows.length > 0
-  } catch {
+  } catch (err) {
+    rethrowBusy(err)
     return false
   }
 }
@@ -406,7 +414,9 @@ function buildUserMessageMap(db: SqliteDatabase, timeFloor: string): Map<string,
         map.set(composerId, { messages: [text], pos: 0 })
       }
     }
-  } catch {}
+  } catch (err) {
+    rethrowBusy(err)
+  }
   return map
 }
 
@@ -438,7 +448,8 @@ function scanBubblesPaged(
     let batch: BubbleRow[]
     try {
       batch = db.query<BubbleRow>(BUBBLE_QUERY_PAGE, [beforeRowId, BATCH])
-    } catch {
+    } catch (err) {
+      rethrowBusy(err)
       break
     }
     if (batch.length === 0) break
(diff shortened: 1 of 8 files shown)


ozymandiashh (dismissed):
## Review: fix(cursor): use Cursor's real context tokens for input

The direction is correct: reading `composerData.promptTokenBreakdown.totalUsedTokens` for real input instead of the old agentKv char estimate is the right call. The agentKv removal eliminates the double-count. However, I found several issues during review and have a branch with fixes ready.

### Issues Found & Fixed

**Critical:**
1. **Cache version not bumped** (`src/cursor-cache.ts:14`) — `CURSOR_CACHE_VERSION` stays at 4 despite the parser semantics changing radically (parseAgentKv removed, real context tokens added). Users with existing v4 cache will see stale double-counted agentKv calls and none of the new token behavior. Fix: bump to 5.

2. **Real input tokens priced at wrong model** (`src/providers/cursor.ts:636-642`) — Real context tokens are credited onto the type-1 user bubble, which carries no `modelInfo.modelName`. `resolveModel(null)` falls back to `claude-sonnet-4-5` and `modelForDisplay(null)` shows `cursor-auto`. So an Opus conversation's full input cost gets billed at Sonnet rates under the wrong label. Fix: pre-pass building `composerId -> model` from assistant bubbles, used as fallback when `row.model` is null.

**Bugs:**
3. **`Shell` missing from `BASH_TOOLS`** (`src/classifier.ts:20`) — Cursor's agent uses `Shell` as the tool name, but the classifier only recognized `Bash`, `BashTool`, `PowerShellTool`. Cursor agent turns with shell commands wouldn't classify as bash/build/test.

4. **`buildUserMessageMap` keyed by empty `conversationId`** (`src/providers/cursor.ts:325,397`) — The query uses `json_extract(value, '$.conversationId')` which is empty in current Cursor builds. User messages never matched composer IDs, so `takeUserMessage` returned empty strings. Fix: extract composerId from `bubble_key` like `parseBubbles` does.

5. **Null coalescing treats valid `0` as null** (`src/providers/cursor.ts:480`) — `r.used ?? r.ctx ?? 0` falls through to `ctx` when `totalUsedTokens` is `0` (a valid count). Fix: explicit `r.used !== null ? r.used : (r.ctx ?? 0)`.

6. **Tools dropped for multi-turn conversations** (`src/providers/cursor.ts:674`) — `agentTools` only attached when `creditedHere = true` (first credited turn). Tools from turns 2+ silently dropped. Fix: decouple tool attachment from input credit decision.

7. **Stale comment** (`src/providers/cursor.ts:777`) — Says "parseAgentKv is kept for the tools/bash breakdown in a follow-up" but the function is removed and `loadAgentToolsByComposer()` is active.

### Additional Notes
- `BUBBLE_REQUESTID_QUERY` and `loadAgentToolsByComposer` do full-table JSON scans with no date/window cap, which partially defeats the large-DB paging protection. Worth adding a time filter in a follow-up.
- No tests existed for the new `composerData` / `requestId` / tool extraction paths. I've added 6 tests covering real token crediting, once-per-conversation, fallback, contextTokensUsed, tool/bash attribution, and model attribution.

### Fix Branch
All fixes are on `fix/cursor-real-tokens-review-fixes` in my fork (`ozymandiashh/codeburn`). 69 tests pass (42 original + 6 new + 21 classifier), `tsc --noEmit` clean. Happy to open a PR against this branch if preferred.

ozymandiashh (commented):
Inline review comments. Also noted (not inline since the code was unchanged by this PR): `buildUserMessageMap` at line 397 still keys by `json_extract(value, '$.conversationId')` which is empty in current Cursor builds - user messages never match composer IDs. And `Shell` is missing from `BASH_TOOLS` in `src/classifier.ts` - Cursor agent turns with shell commands won't classify as bash. Full fix branch at ozymandiashh/codeburn:fix/cursor-real-tokens-review-fixes.

src/providers/cursor.ts
```diff
@@ -470,13 +459,106 @@ function scanBubblesPaged(
...
+  FROM cursorDiskKV
+  WHERE key LIKE 'composerData:%'
+`
+
+function loadComposerInputTokens(db: SqliteDatabase): Map<string, number> {
+  const map = new Map<string, number>()
+  try {
+    const rows = db.query<{ composer_id: string; used: number | null; ctx: number | null }>(COMPOSER_TOKENS_QUERY)
+    for (const r of rows) {
+      const tokens = r.used ?? r.ctx ?? 0
```
ozymandiashh: **Bug:** `r.used ?? r.ctx ?? 0` treats a valid `totalUsedTokens: 0` as null and falls through to `ctx`. Use an explicit null check: `r.used !== null ? r.used : (r.ctx ?? 0)`.

src/providers/cursor.ts
```diff
@@ -520,36 +602,52 @@ function parseBubbles(
...
+      let creditedHere = false
+
+      // Current Cursor leaves tokenCount at {0,0}. Use the conversation's real
+      // context size (promptTokenBreakdown) for input, credited once per
+      // conversation, and the reply text for output. Fall back to the
+      // visible-text estimate only when no breakdown was recorded (older builds).
       if (inputTokens === 0 && outputTokens === 0) {
         const textLen = row.text_length ?? 0
-        if (textLen === 0) continue
         if (row.bubble_type === 1) {
```
ozymandiashh, line 790: **Critical:** Real input tokens are credited onto this type-1 user bubble, but user bubbles carry no `modelInfo.modelName`. The pricing falls back to `claude-sonnet-4-5` and display to `cursor-auto`, so the full context cost of an Opus conversation gets billed at Sonnet rates under the wrong label. Fix: build a `composerId -> model` map from assistant bubbles and use it as fallback when `row.model` is null.

src/providers/cursor.ts
```diff
@@ -573,7 +671,12 @@ function parseBubbles(
       const languages = extractLanguages(blobToText(row.code_blocks))
       const hasCode = languages.length > 0
 
-      const cursorTools: string[] = hasCode ? ['cursor:edit', ...languages.map(l => `lang:${l}`)] : []
+      const agentTurn = creditedHere ? agentTools.get(conversationId) : undefined
```
ozymandiashh: **Bug:** Tools/bash only attach when `creditedHere` is true (first credited turn). Multi-turn conversations silently lose tools from turns 2+. Fix: decouple tool attachment from the input credit decision - just use `agentTools.get(conversationId)` directly.

src/providers/cursor.ts
```diff
@@ -470,13 +459,106 @@ function scanBubblesPaged(
...
+  }
+  return map
+}
+
+type AgentTools = { tools: string[]; bash: string[] }
+
+// Cursor logs the agent's tool calls (Read, Grep, Shell, ...) in agentKv blobs
+// keyed by requestId. Bubbles carry the same requestId plus the composerId, so
+// joining the two attributes each conversation's tools and Shell commands.
+const BUBBLE_REQUESTID_QUERY = `
```
ozymandiashh: **Warning:** `BUBBLE_REQUESTID_QUERY` and `loadAgentToolsByComposer` do full-table JSON scans with no date/window cap. On multi-GB Cursor DBs this reintroduces the kind of full-scan stall the paged bubble path was designed to avoid. Worth adding a time filter in a follow-up.

src/providers/cursor.ts
```diff
@@ -801,9 +774,14 @@ function createParser(
           // seenKeys is not mutated by calls that the workspace filter is
           // about to drop. Cross-source dedup happens at yield time.
           const localSeen = new Set<string>()
-          const { calls: bubbleCalls } = parseBubbles(db, localSeen, timeFloor)
-          const { calls: agentKvCalls } = parseAgentKv(db, localSeen, dbPath)
-          allCalls = [...bubbleCalls, ...agentKvCalls]
+          // promptTokenBreakdown carries Cursor's real per-conversation input
+          // count, so it supersedes the old agentKv content-char estimate,
+          // which double-counted against the bubble stream. parseAgentKv is
```
ozymandiashh: Stale comment: says "parseAgentKv is kept for the tools/bash breakdown in a follow-up" but the function is removed and `loadAgentToolsByComposer()` is active now.

ozymandiashh:
Fix branch with all review issues addressed: #576. All fixes tested (69 tests pass, tsc clean). The fix PR targets this branch directly so it can be merged into this PR before merging to main.

Joon:
Hi team. I tested this change today and it did not work as I would expect. Testing protocol:

- Cloned this repo, checked out the branch fix/cursor-real-tokens
image

- ran `npm run build` and then `node dist/cli.js`
image


- Compared the codeburn tokens to the tokens I could see in the cursor admin console

Codeburn details:
image


Cursor console details for me (note that all work today happened on this machine, no cloud agents in use at the moment): 
image

Please advise if the issue is with my testing methodology or if there is still an issue with how Codeburn discovers usage. 

Thanks!

ozymandiashh:
Hi Joon — thanks for the careful testing, and you're right that the numbers don't line up. It's not your methodology; it's a structural limitation.

The admin console and what codeburn reads locally are two different measurements:

- The **console** is a server-side, per-request *cumulative* bill — every turn re-sends the growing conversation context (billed again each turn), split into **input + output + cache-write + cache-read** at per-model rates.
- **Locally**, Cursor only stores a single client-side context-window snapshot (`composerData.promptTokenBreakdown.totalUsedTokens`) — the latest context size, credited once per conversation. Cache read/write and real output counts aren't on disk at all.

So a local-only read structurally *undercounts* the console, and the gap grows with conversation length. #575 is a genuine improvement (input ~45k → ~168k, much closer to reality), but exact parity from local data isn't achievable — the per-turn cumulative input, cache, and real output simply aren't stored locally.

The parity path is to read the same server-side data the console reads, via Cursor's Admin API: `POST https://api.cursor.com/teams/filtered-usage-events` returns per-event `inputTokens / outputTokens / cacheWriteTokens / cacheReadTokens` and reconciles to `/teams/spend` and the dashboard within ~1%. The catch is access — that API is **team-scoped** (an `admin:*` key minted from team settings), so a solo/Pro user on one machine can't generate one. We'd gate it as an **opt-in** source and keep the local read as a clearly-labeled estimate. (Individuals can pull the same breakdown from the dashboard CSV — `export-usage-events-csv` — via their own session cookie, but that's an unofficial endpoint.)

Separately, I pushed a Composer **pricing** fix in #576: those house models were priced as Sonnet 4.6 ($3/$15), but Cursor publishes Composer-2/2.5 at $0.5/$2.5 (and 1.5/1 are different too), so cost was overcounted ~6x — now using Cursor's real published rates.

Joon:
Thank you for the explanation, I appreciate your thoughtful replies and the fast turn-around time! 

One question I had that your comment has not addressed: I had ChatGPT traffic today (as seen on the Cursor console) but codeburn didn't collect that. This was on a long-lived session that I started yesterday. The codeburn "today" view doesn't show any ChatGPT traffic - would those tokens accrue to yesterday instead because it's a long-lived agent?

iamtoruk:
Good question, and it's a different issue from the cumulative vs local gap, so worth answering on its own.

How #575 dates Cursor usage: each call is stamped with its own message time (the bubble's `createdAt`), not the session's start. The one wrinkle is that a conversation's real input (the context-window number) is credited once, on its first user turn, not spread across every turn. The parser reads the conversation's current total context size and attaches all of it to that single bubble.

So for a long-lived session your hypothesis is basically right. If today's activity didn't open a new user turn in that conversation (an agent continuing on its own, or follow-ups Cursor stored without a fresh `type:1` user bubble dated today), there's no today-stamped bubble to carry the input, and it lands on the day of the last user turn, i.e. yesterday. The daily split for a long multi-day conversation is lumpy by design rather than spread across the days the work actually happened.

A two-minute check to tell which case you're in, no SQL needed:

1. Run your usual report for **today**, then for a **2-day** window.
2. If the GPT tokens show up in the 2-day window but not in today, it's the dating-on-first-turn effect above (they're sitting on yesterday).
3. If they show up in neither, it's a second gap: `promptTokenBreakdown.totalUsedTokens` is Cursor's Composer context meter, and a GPT ("ChatGPT") session may not populate it, so we fall back to a text estimate and skip turns whose text lives in the agent stream. That one is fixable (estimate-from-text for non-Composer models). Tell me what you see and I'll take it.

One more thing: #576 just merged into this branch (cache-version bump so you won't serve a stale cache, accurate Composer pricing, and tool-attribution fixes), so please `git pull`, `rm -rf ~/.cache/codeburn`, rebuild, and re-run before the check above.

PR #576, opened 2026-06-30
fix: review fixes for PR #575 (cache version, model attribution, user msg join, tool classification)

ozymandiashh:
## Review fixes for PR #575

Addresses issues found during code review of #575. All fixes are tested (69 tests pass, `tsc --noEmit` clean).

### Fixes

1. **Cache version bump** (`src/cursor-cache.ts`) — `CURSOR_CACHE_VERSION` 4 -> 5. Without this, users with existing cache see stale double-counted agentKv calls and none of the new real-token behavior.

2. **Model attribution for real input tokens** (`src/providers/cursor.ts`) — Real context tokens are credited on user bubbles (type=1) which carry no `modelInfo`. Added a pre-pass building `composerId -> model` from assistant bubbles so pricing/display uses the conversation's actual model instead of defaulting to `cursor-auto`/`claude-sonnet-4-5`.

3. **`buildUserMessageMap` keyed by empty field** (`src/providers/cursor.ts`) — Was using `json_extract(value, '$.conversationId')` which is empty in current Cursor builds. Now extracts composerId from `bubble_key` matching `parseBubbles`.

4. **`Shell` added to `BASH_TOOLS`** (`src/classifier.ts`) — Cursor's agent uses `Shell` as the tool name; it was missing from the bash tool set.

5. **Null coalescing fix** (`src/providers/cursor.ts`) — `r.used ?? r.ctx` treated valid `totalUsedTokens: 0` as null. Now uses explicit null check.

6. **Multi-turn tool attribution** (`src/providers/cursor.ts`) — Tools/bash were only attached on the first credited turn (`creditedHere`). Decoupled so all turns get their tools.

7. **Stale comment** — Updated comment that referenced parseAgentKv being kept for a follow-up.

8. **New tests** (`tests/providers/cursor-real-tokens.test.ts`) — 6 tests covering real token crediting, once-per-conversation, fallback estimation, contextTokensUsed fallback, tool/bash attribution, and model attribution.

### DIFF
--- a/src/classifier.ts
+++ b/src/classifier.ts
@@ -17,7 +17,7 @@ const URL_PATTERN = /https?:\/\/\S+/i
 
 export const EDIT_TOOLS = new Set(['Edit', 'Write', 'FileEditTool', 'FileWriteTool', 'NotebookEdit', 'cursor:edit'])
 const READ_TOOLS = new Set(['Read', 'Grep', 'Glob', 'FileReadTool', 'GrepTool', 'GlobTool'])
-export const BASH_TOOLS = new Set(['Bash', 'BashTool', 'PowerShellTool'])
+export const BASH_TOOLS = new Set(['Bash', 'BashTool', 'PowerShellTool', 'Shell'])
 const TASK_TOOLS = new Set(['TaskCreate', 'TaskUpdate', 'TaskGet', 'TaskList', 'TaskOutput', 'TaskStop', 'TodoWrite'])
 const SEARCH_TOOLS = new Set(['WebSearch', 'WebFetch', 'ToolSearch'])
 
--- a/src/cursor-cache.ts
+++ b/src/cursor-cache.ts
@@ -11,7 +11,11 @@ import type { ParsedProviderCall } from './providers/types.js'
 // router relies on those composer ids to bucket calls per project.
 // Version 2 caches contain `sessionId: 'unknown'` for every call and would
 // route everything to the orphan project, so we invalidate them.
-const CURSOR_CACHE_VERSION = 4
+// Version 5: parseAgentKv was removed (it double-counted against bubbles);
+// real context tokens from composerData.promptTokenBreakdown now drive
+// input, and agentKv is used only for the tools/bash breakdown. Cached v4
+// results contain stale agentKv calls and lack the real token figures.
+const CURSOR_CACHE_VERSION = 5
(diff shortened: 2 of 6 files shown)


iamtoruk:
@ozymandiashh, can you please email me at research@eywa.to, I would love to connect with you regarding new OS project I will be launching.

iamtoruk:
Really appreciate this, it's a sharp review and most of it I want to take.

The cache-version bump especially was a genuine miss on my side: without it, existing users keep serving stale double-counted results. The model-attribution pre-pass is also the right fix for real input landing on user bubbles (which carry no `modelInfo`) and showing as `cursor-auto`/Sonnet instead of the real model. The `$.conversationId` join fix and `Shell` in `BASH_TOOLS` are both correct too.

Two notes before this can go in: one revert, one no-op.

**(6) Multi-turn tool attribution: please revert.** `agentTools.get(conversationId)` is the *whole conversation's* aggregated tool/bash list. The `creditedHere` gate attached it once per conversation on purpose; removing it re-attaches the full aggregate to every emitted bubble, so tools/bash get N-counted. Measured on a real DB (same one, cache cleared):

| | this branch | with #6 |
|---|---|---|
| bash commands | 10 | 100 |
| Read | 57 | 423 |
| Shell | 10 | 100 |

So roughly 7-10x inflation. If the goal is genuinely per-turn tools (which would be nicer), it needs `loadAgentToolsByComposer` to return per-turn buckets keyed by `requestId` rather than the per-conversation aggregate; as written it just duplicates. Restoring the `creditedHere` gate keeps the counts correct. A multi-bubble fixture would lock this down, the new tests are single-bubble, which is why it passed.

**(5) Null coalescing: no-op, optional to drop.** `r.used ?? r.ctx ?? 0` already does the right thing: `??` only falls through on `null`/`undefined`, not on `0`, so a valid `totalUsedTokens: 0` stays `0` (and is then filtered by `tokens > 0`). The `!== null` version is equivalent in practice and slightly less robust if the field ever comes back `undefined` instead of `null`. Just flagging the rationale.

Everything else (1-4, 7, 8) is great. Revert #6 (and optionally #5) and add a multi-turn tool test, and I'll merge this into the Cursor branch. Thanks again, these are real improvements.

ozymandiashh:
Ill work on it and I also emailed you

PR #582, opened 2026-07-01
feat(web): mobile responsive web dashboard

### LINKED ISSUE #581: feat(web): Mobile responsive web dashboard
## Problem

The web dashboard (`codeburn web`) has no mobile responsive adaptation. On mobile devices:

- Fixed 240px sidebar takes most of the screen width
- Stat cards grid (4 columns) overflows
- Charts are compressed and unreadable
- No way to collapse/toggle the sidebar
- Filter buttons overflow horizontally

## Proposed Solution

Add responsive design using Tailwind CSS breakpoints (md: / 768px):

1. **Sidebar to off-canvas drawer**: Hidden by default on mobile, slides in from left with backdrop overlay when toggled
2. **Hamburger menu**: Appears in header on mobile to toggle sidebar
3. **Stat cards**: `grid-cols-2` base stays 2 columns on mobile
4. **Bottom panels**: `lg:grid-cols-2` becomes single column on mobile
5. **Chart area**: Horizontally scrollable on mobile
6. **Header**: Controls wrap and scroll horizontally
7. **Touch targets**: Minimum 36px height for all interactive elements

All changes are pure Tailwind responsive classes + minimal React state (`sidebarOpen`). No new dependencies. Desktop view unchanged.

Happy to submit a PR if the approach sounds good.

ele-yufo:
Closes #581

## What

Add mobile responsive support for the web dashboard (`codeburn web`). The dashboard currently has no mobile adaptation — fixed 240px sidebar, 4-column stat cards, no responsive breakpoints.

## Changes (30 insertions, 6 deletions — all in `dash/src/App.tsx`)

| Change | Implementation |
|--------|---------------|
| **Hamburger button** | SVG button in header brand area, hidden on `md:` (768px+) via `md:hidden` |
| **Sidebar drawer** | `fixed inset-y-0 left-0 z-50` on mobile with `transition-transform duration-300`, slides in/out via `-translate-x-full` / `translate-x-0`. On `md:` stays `md:relative md:translate-x-0` |
| **Backdrop overlay** | `fixed inset-0 z-40 bg-black/40 backdrop-blur-sm md:hidden`, click to close |
| **Sticky header** | Added `sticky top-0 z-40` for mobile scroll |
| **Responsive content** | `flex-col md:flex-row` — stacks vertically on mobile |
| **Full-width main** | Added `w-full` to main |
| **Auto-close sidebar** | All SideLink onClick handlers also call setSidebarOpen(false) |

## Design decisions

- **`md:` breakpoint (768px)** as the mobile/desktop boundary — consistent with Tailwind defaults
- **No new dependencies** — pure Tailwind responsive classes + 1 React state
- **Desktop unchanged** — all responsive classes use `md:` prefix, desktop layout identical
- **Uses existing `cn()` utility** for conditional class merging

## Testing

- Desktop (1280px): sidebar visible, 4-col cards, no regression
- Mobile (375px): hamburger visible, sidebar hidden, 2-col cards, single-column panels
- Sidebar toggle: opens/closes with backdrop
- Device selection: auto-closes sidebar on mobile

### DIFF
--- a/dash/src/App.tsx
+++ b/dash/src/App.tsx
@@ -347,6 +347,7 @@ export function App() {
   const [view, setView] = useState<string>('all')
   const [unit, setUnit] = useState<Unit>('cost')
   const [searchOpen, setSearchOpen] = useState(false)
+  const [sidebarOpen, setSidebarOpen] = useState(false)
   const [responded, setResponded] = useState<Set<string>>(new Set())
 
   const qc = useQueryClient()
@@ -427,8 +428,20 @@ export function App() {
   return (
     <div className="min-h-screen bg-outer-background p-2.5">
       <div className="flex h-[calc(100vh-20px)] flex-col gap-2.5">
-        <header className="flex h-12 shrink-0 items-center gap-4 rounded-md border border-border bg-card px-5 shadow-[0_2px_8px_rgba(0,0,0,0.03)]">
+        <header className="flex h-12 shrink-0 items-center gap-4 rounded-md border border-border bg-card px-5 shadow-[0_2px_8px_rgba(0,0,0,0.03)] sticky top-0 z-40">
           <div className="flex items-center gap-2">
+            <button
+              type="button"
+              onClick={() => setSidebarOpen(!sidebarOpen)}
+              className="md:hidden flex items-center justify-center w-9 h-9 rounded-md border border-border bg-interactive-secondary text-foreground"
+              aria-label="Toggle sidebar"
+            >
+              <svg viewBox="0 0 24 24" width="20" height="20" fill="none" stroke="currentColor" strokeWidth="2" strokeLinecap="round">
+                <line x1="3" y1="6" x2="21" y2="6" />
+                <line x1="3" y1="12" x2="21" y2="12" />
+                <line x1="3" y1="18" x2="21" y2="18" />
+              </svg>
+            </button>
             <img src="/codeburn-logo.png" alt="CodeBurn" className="h-6 w-6" />
             <span className="text-lg font-semibold tracking-[-0.02em] text-foreground">
               Code<span className="text-[#e8553a]">Burn</span>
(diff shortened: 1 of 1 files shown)


ozymandiashh:
Thanks for taking this on, the off-canvas drawer mechanic here works well. A few things I ran into while comparing against the current `dash/src/App.tsx` that might be worth addressing:

- **Header controls still overflow on mobile.** This diff only adds the hamburger and `sticky top-0 z-40` to the header; the `ml-auto flex items-center gap-2` container (5 period buttons + 2 unit buttons + provider `<select>`) has no overflow handling, so at 375px the page still scrolls sideways (about 328px of horizontal overflow on the unchanged code), and adding the hamburger to that row makes it slightly tighter. `max-md:min-w-0 max-md:overflow-x-auto` on that container plus `shrink-0` on the button groups keeps them scrollable instead of pushing the page wider.
- **A couple of unscoped classes reach desktop.** `sticky top-0 z-40` on the header and `z-50` on the aside are added without an `md:`/`max-md:` scope, and tailwind-merge doesn't drop them because they don't conflict with the `md:` overrides. So at >=768px the header and aside gain position/stacking they didn't have before. Scoping those to mobile (or dropping the `sticky`, since the outer container is fixed-height and doesn't scroll) keeps the desktop layout untouched.
- **A11y.** The hamburger could expose `aria-expanded`/`aria-controls`; the backdrop is a `<div>` (a `<button>` makes it keyboard-dismissable); and the closed drawer stays in the tab order since it's only translated off-screen, so keyboard/screen-reader users tab through the hidden links. Adding `visibility:hidden` on the closed state (alongside the transform) drops them from the tab and accessibility trees.
- **Minor.** The "Search local devices" button doesn't call `setSidebarOpen(false)`, so the drawer stays open behind the search modal (which shares `z-50` with the drawer).

For cross-reference, I opened #587 which takes the same drawer approach but also covers the header overflow, 36px touch targets, and the a11y items, and scopes every mobile rule with `max-md:` so the desktop view stays byte-identical. Might be useful to compare the two. Either way, appreciate you pushing on the mobile experience.

PR #584, opened 2026-07-01
fix(dash): use period-scoped cache tokens instead of summing entire history

### LINKED ISSUE #583: Web dashboard Cache read/write cards sum entire history instead of current period
## Bug Description

The web dashboard (`codeburn web -p 30days`) shows incorrect values for the **Cache read** and **Cache write** stat cards. They display totals from the entire `history.daily` array (up to 365 days of backfill), rather than only the data within the user-selected period (e.g. 30 days).

## Steps to Reproduce

1. Run `codeburn report -p 30days` in the terminal
2. Run `codeburn web -p 30days` and open the web dashboard
3. Compare the "cached" value in the terminal with the "Cache read" card in the web UI

## Expected Behavior

Both views should show the same Cache read value for the same period (30 days).

## Actual Behavior

- **Terminal** (`report -p 30days`): shows `1391.1M cached`
- **Web** (`web -p 30days`): shows `Cache read: 2.07B`

The web value is ~1.5x larger because it sums over ~365 days of history data.

## Root Cause

In the web frontend (`dist/dash/assets/index-*.js`), the Cache read/write cards compute their values by reducing over the entire `e.history.daily` array:

```js
l = (e?.history.daily) ?? []
o = l.reduce((b,x) => b + x.cacheWriteTokens, 0)  // Cache write card
c = l.reduce((b,x) => b + x.cacheReadTokens, 0)    // Cache read card
```

However, `history.daily` is populated from `BACKFILL_DAYS = 365` days of cached data (for the trend chart), not limited to the current period. Meanwhile, other cards (Cost, Calls, Sessions, Tokens in/out) correctly use `e.current` which is properly scoped to the selected period.

## Suggested Fix

The frontend should use `e.current.tokens.cacheRead` and `e.current.tokens.cacheWrite` (which are correctly scoped to the period in `buildMenubarPayload`) instead of summing `history.daily`.

Alternatively, filter `history.daily` to only include dates within the selected period before summing.

## Environment

- codeburn version: 0.9.14
- OS: macOS (Darwin 25.3.0)
- Node: v24.13.0

Llldmiao:
## Summary

The web dashboard's **Cache read** and **Cache write** metric cards were summing the entire `history.daily` array (up to 365 days of backfill data intended for the trend chart) instead of using the current period's values. This caused them to show inflated numbers compared to `codeburn report`.

For example, with `codeburn web -p 30days` vs `codeburn report -p 30days`:
- Terminal shows: `1391.1M cached`
- Web showed: `Cache read: 2.07B` (summing ~365 days)

## Changes

1. **`src/menubar-json.ts`**: Added `cacheReadTokens` and `cacheWriteTokens` to the `MenubarPayload.current` type and populated them from `PeriodData` (which is already correctly scoped to the selected period).

2. **`dash/src/lib/api.ts`**: Added the new fields to the `Current` type and initialized them in `normalizePayload` (backward-compatible — defaults to 0 for older payloads).

3. **`dash/src/App.tsx`**: Changed both `DeviceView` and `CombinedView` to read cache tokens from `current` instead of reducing over `history.daily`.

4. **`tests/menubar-json.test.ts`**: Added a test asserting that `cacheReadTokens` and `cacheWriteTokens` are correctly passed through to the payload.

## Testing

All 1359 existing tests pass. The new test verifies the fix explicitly.

Fixes #583

### DIFF
--- a/dash/src/App.tsx
+++ b/dash/src/App.tsx
@@ -75,8 +75,8 @@ function Stat({ label: lbl, value }: { label: string; value: string }) {
 function DeviceView({ payload, isRemote, unit }: { payload?: Payload; isRemote: boolean; unit: Unit }) {
   const c = payload?.current
   const daily = payload?.history.daily ?? []
-  const cacheWrite = daily.reduce((s, d) => s + d.cacheWriteTokens, 0)
-  const cacheRead = daily.reduce((s, d) => s + d.cacheReadTokens, 0)
+  const cacheWrite = c?.cacheWriteTokens ?? 0
+  const cacheRead = c?.cacheReadTokens ?? 0
   const toolBars: BarItem[] = c
     ? Object.entries(c.providers).filter(([, v]) => v > 0).sort((a, b) => b[1] - a[1]).map(([k, v]) => ({ name: k, value: v, display: usd(v) }))
     : []
@@ -256,10 +256,8 @@ function CombinedView({ devices, unit }: { devices: DeviceUsage[]; unit: Unit })
     if (!c) continue
     inTok += c.inputTokens
     outTok += c.outputTokens
-    for (const e of d.payload?.history.daily ?? []) {
-      cacheWrite += e.cacheWriteTokens
-      cacheRead += e.cacheReadTokens
-    }
+    cacheWrite += c.cacheWriteTokens ?? 0
+    cacheRead += c.cacheReadTokens ?? 0
     for (const [k, v] of Object.entries(c.providers)) providers.set(k, (providers.get(k) ?? 0) + v)
     for (const m of c.topModels) models.set(m.name, (models.get(m.name) ?? 0) + m.cost)
     for (const a of c.topActivities) activities.set(a.name, (activities.get(a.name) ?? 0) + a.cost)
--- a/dash/src/lib/api.ts
+++ b/dash/src/lib/api.ts
@@ -28,6 +28,8 @@ export type Current = {
   inputTokens: number
   outputTokens: number
   cacheHitPercent: number
+  cacheReadTokens: number
+  cacheWriteTokens: number
   codexCredits: number
   topActivities: Array<{ name: string; cost: number; turns: number; oneShotRate: number | null }>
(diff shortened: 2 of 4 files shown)


ozymandiashh:
The dashboard-side fix reads correctly, switching the cache cards from summing `history.daily` to the period-scoped `current` is exactly right. Two things I hit while checking it out:

- **`npm run typecheck` fails in `dash/` on this branch.** Removing the two `.reduce` calls leaves `const daily = payload?.history.daily ?? []` in `DeviceView` unused, and `dash/tsconfig.json` has `noUnusedLocals: true`, so tsc errors:
  ```
  src/App.tsx(77,9): error TS6133: 'daily' is declared but its value is never read.
  ```
  CI only runs Semgrep and `vite build` uses esbuild (which ignores the flag), so it builds fine, but a local typecheck or an IDE will flag it. Deleting that line clears it.
- **The same over-count exists on another surface.** `src/sharing/host.ts` `summarizeOneDevice` also sums `history.daily` for cache, and that feeds the `codeburn devices` CLI table and the combined cross-device totals. So after this change the web cards are correct but `codeburn devices` still shows the inflated numbers for any period shorter than the 365-day backfill. Porting the same `current`-first read there closes it, ideally with a fallback to the windowed daily sum for older peers, since sharing is a cross-version protocol and an old peer won't send the new `current.cache*` fields.

For cross-reference, I have #586 up that fixes both surfaces, adds the backward-compatible fallback for old peers, and includes tests asserting the cards are independent of the 365-day backfill. Might be worth comparing. Thanks for digging into this one, the repro in the description made it easy to trace.

iamtoruk:
Thanks for this, and for the clear write-up with the 2.07B vs 1391.1M example, that is exactly the bug. It turned out #586 fixes the same issue (#583) and also covers the devices/sharing summary path (src/sharing/host.ts), which has the identical daily-history summation that this PR did not touch, so we merged that one to get both code paths fixed in a single change. Your diagnosis was spot on and reached it independently. Closing as superseded by #586, appreciate the fix.

PR #585, opened 2026-07-01
fix(devin): report friendly GPT model names with effort tier

ozymandiashh:
## What

Devin transcripts label each step with a `generation_model` in dash form plus an
effort suffix (`gpt-5-3-codex-xhigh`). `getShortModelName` is keyed in dot form
(`gpt-5.3-codex`) and its version-boundary check matched the dash id against the
base `gpt-5` entry, collapsing every Devin GPT variant to `GPT-5`.

This normalizes the dash form to canonical dot form, maps it through the
short-name table, and surfaces the effort tier, e.g. `GPT-5.3 Codex (xhigh)`.
It falls back to the friendly `model_name` when `generation_model` is an opaque
`MODEL_*` id, and reads `extra.generation_model` so ATIF v1.7 transcripts behave
like v1.4. Devin rows in the JSON report and model-efficiency now key on the
friendly name.

## Robustness

The dash-to-dot rewrite is restricted to a single-digit minor at a token
boundary, and the friendly-name path defers to `getShortModelName` when a suffix
token is purely numeric. This keeps dated snapshot ids such as
`gpt-4-1106-preview` from being mislabeled as a fake version (`GPT-4.1106
Preview`); they now pass through raw.

## Test

Adds a Devin model-variant matrix covering the effort tiers, the `codex` suffix
with and without effort, the `MODEL_*` fallback, ATIF v1.7 via `extra`, and the
`gpt-4-1106-preview` snapshot regression. Full suite green.

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/Users/husamsoboh/codeburn/node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/Users/husamsoboh/codeburn/node_modules
--- a/src/model-efficiency.ts
+++ b/src/model-efficiency.ts
@@ -1,5 +1,5 @@
 import { getShortModelName } from './models.js'
-import type { ProjectSummary } from './types.js'
+import type { ParsedApiCall, ProjectSummary } from './types.js'
 
 export type ModelEfficiency = {
   model: string
@@ -19,6 +19,10 @@ function rate(num: number, den: number): number | null {
   return Math.round((num / den) * 1000) / 10
 }
 
+function modelKey(call: ParsedApiCall): string {
+  return call.provider === 'devin' ? call.model : getShortModelName(call.model)
+}
+
 export function aggregateModelEfficiency(projects: ProjectSummary[]): Map<string, ModelEfficiency> {
   const byModel = new Map<string, MutableModelEfficiency>()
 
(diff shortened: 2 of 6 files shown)


iamtoruk:
This is a great fix, thank you. The effort-tier surfacing is exactly right, and I especially like that the dated-snapshot guard keeps `gpt-4-1106-preview` from being misread as a version. The test coverage across dash/dot forms, `MODEL_*` fallback, and `extra` vs `metadata` is thorough. CI is green on the logic.

One thing blocks the merge though: commit `995d75d` accidentally committed a `node_modules` **symlink** at the repo root:

```
new file mode 120000
node_modules -> /Users/husamsoboh/codeburn/node_modules
```

It points at an absolute path on your machine, so it leaks your local username and would be a dangling symlink for everyone else. It also breaks `git`/`npm` workflows on checkout (it stopped `gh pr checkout 585` from working here, since git won't replace a real `node_modules/` with the symlink). It slipped past `.gitignore` because the ignore rule is `node_modules/` with a trailing slash, which matches a directory but not a symlink.

Could you drop it and re-push? A follow-up commit is simplest (it'll squash away on merge anyway):

```
git rm node_modules
git commit -m "chore: remove accidental node_modules symlink"
git push
```

Once that's gone this is good to merge.

ozymandiashh:
Good catch, thank you. Dropped the accidental `node_modules` symlink in a follow-up (3898720): `git rm node_modules` + commit. It'll squash away on merge. Confirmed it's the only branch affected. Should be good to merge now.

ozymandiashh:
Also hardened `.gitignore` (c70a512): added a slash-less `node_modules` alongside `node_modules/` so a stray symlink can't slip past the ignore rule the way this one did. Both squash away on merge.

bmcdonough:
@ozymandiashh
Thank you so much for the thoughtful solution.  Identifying the hyphen to period was the better solution.

PR #594, opened 2026-07-02
feat(providers): add Zed agent provider

### LINKED ISSUE #480: Add Zed agent provider (threads.db / zstd JSON schema documented)
## Summary

Add a CodeBurn provider for [Zed](https://zed.dev)'s built-in AI agent.

In #196 a maintainer noted Zed's agent usage is invisible to CodeBurn and that "Zed provider support is on the roadmap." That issue has since been closed (it was really about Cursor misattribution + the classifier). There doesn't appear to be a tracking issue for the actual Zed provider, so opening one here.

I use Zed daily and have inspected the real on-disk format on macOS, so I can build and validate this against real sessions per the contributing guide. Filing first to confirm the approach before opening a PR.

## Where Zed stores agent data

Unlike file-per-session providers (Codex, Pi), Zed uses a **single SQLite database** with one row per agent thread:

```
~/Library/Application Support/Zed/threads/threads.db
```

Schema (table `threads`):

```sql
CREATE TABLE threads (
    id TEXT PRIMARY KEY,
    summary TEXT NOT NULL,
    updated_at TEXT NOT NULL,      -- ISO 8601
    data_type TEXT NOT NULL,       -- e.g. "zstd"
    data BLOB NOT NULL,            -- zstd-compressed JSON thread
    parent_id TEXT,
    folder_paths TEXT,
    folder_paths_order TEXT,
    created_at TEXT
);
```

The `data` BLOB is **zstd-compressed JSON** (magic bytes `28 B5 2F FD`). Decompressed, each thread JSON contains everything needed for pricing:

- `model`: `{ "provider": "anthropic", "model": "claude-opus-4-8" }`
- `request_token_usage`: map of `request_id -> { input_tokens, output_tokens, cache_creation_input_tokens, cache_read_input_tokens }`
- `cumulative_token_usage`: same shape, thread total (can be empty on in-progress threads, so summing `request_token_usage` is the reliable path)
- `messages` with tool calls, plus `updated_at`, `title`, `profile`, etc.

Token semantics match Anthropic's (separate cache-creation vs cache-read fields), so pricing maps cleanly onto the existing LiteLLM engine.

## Proposed approach

- New `src/providers/zed.ts` modeled on the SQLite-backed providers (Cursor / Forge / Warp) rather than the file-per-session shape.
- Lazy-load `better-sqlite3` (already an optional dep for Cursor/OpenCode) so non-Zed users aren't affected.
- Decompress each `data` blob with zstd, sum `request_token_usage` per thread, resolve `model.model`, emit one call per request entry.
- Map Zed's provider/model strings onto existing pricing; add a small display-name + tool-name map.
- Ship a fixture-based test under `tests/providers/` using a redacted real thread.

## Questions before I start

1. Is the **Cursor/Forge SQLite provider** the pattern you'd want me to follow for a single-DB, many-rows source?
2. Preference for the zstd dependency — add a small zstd lib, or shell out / reuse anything already vendored?
3. Anything specific you want in the fixture (redaction expectations for `messages`/prompts)?

Happy to implement once you confirm the approach. I can include terminal proof of `npm run dev -- today --provider zed` against my real Zed history in the PR.

iamtoruk:
Closes #480.

Adds a provider for Zed's built-in agent. Full credit to @chatzinikolakisk for documenting the on-disk format in the issue; the schema was additionally verified byte-for-byte against a real Zed install.

## How it works

- Single source: `threads.db` (macOS `~/Library/Application Support/Zed/threads/threads.db`, with Linux and Windows paths mapped). One row per thread; the `data` blob is zstd-compressed JSON.
- Each thread's `request_token_usage` map is emitted as one call per request with exact `input/output/cache_creation/cache_read` token fields (Anthropic semantics), priced through the existing LiteLLM engine, deduped as `zed:<threadId>:<requestId>`.
- In-progress threads with an empty per-request map fall back to `cumulative_token_usage` so they still count.
- Decompression uses Node's built-in `zlib.zstdDecompressSync` (no new dependency). It ships from Node 22.15; since the package floor is 22.13, the provider feature-detects and skips with a notice on older Nodes, mirroring the `node:sqlite` guard.
- Registered as a lazy provider like the other SQLite-backed ones, so `--provider zed` validates and non-Zed users load nothing extra.

## Testing

- Six fixture tests build a real `threads.db` (exact schema) with zstd-compressed thread JSON: per-request emission with exact token fields, cumulative fallback, non-zstd and malformed-blob skipping, cross-parse dedup, zero-usage threads, and missing-DB discovery.
- Full suite passes (1,393 tests). CLI smoke against a real (empty) Zed install runs clean.

## Known limits

- Per-request timestamps do not exist in the map, so all of a thread's calls carry the thread's `updated_at`; day-level attribution for long-running threads is approximate.
- The local install used for verification has an empty `threads` table, so numbers were validated via fixtures, not real history. @chatzinikolakisk, if you are up for it, running `codeburn overview --provider zed` on this branch against your real usage and posting the result would be a great validation.
- `folder_paths` is not yet mapped to per-project attribution; all Zed usage lands under a single `zed` project for now.

### DIFF
--- a/src/providers/index.ts
+++ b/src/providers/index.ts
@@ -171,11 +171,26 @@ async function loadZcode(): Promise<Provider | null> {
   }
 }
 
+let zedProvider: Provider | null = null
+let zedLoadAttempted = false
+
+async function loadZed(): Promise<Provider | null> {
+  if (zedLoadAttempted) return zedProvider
+  zedLoadAttempted = true
+  try {
+    const { zed } = await import('./zed.js')
+    zedProvider = zed
+    return zed
+  } catch {
+    return null
+  }
+}
+
 const coreProviders: Provider[] = [claude, cline, codebuff, codex, copilot, devin, droid, gemini, hermes, ibmBob, kiloCode, kiro, kimi, mistralVibe, mux, openclaw, openDesign, pi, omp, qwen, rooCode, zerostack, grok]
 
 // Lazily loaded providers, listed by name so --provider validation works even
 // when an optional module fails to load. Must stay in sync with getAllProviders.
-const lazyProviderNames = ['antigravity', 'forge', 'goose', 'cursor', 'opencode', 'cursor-agent', 'crush', 'warp', 'vercel-gateway', 'zcode']
+const lazyProviderNames = ['antigravity', 'forge', 'goose', 'cursor', 'opencode', 'cursor-agent', 'crush', 'warp', 'vercel-gateway', 'zcode', 'zed']
 
 // Canonical set of every provider name (core + lazy), used to validate the
 // --provider CLI flag. Computed lazily so importing this module never depends on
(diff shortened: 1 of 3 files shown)


iamtoruk:
Validated end-to-end against a real Zed thread now (fresh install, real agent chat): the provider's output matches the store token-for-token on all four fields (input 37,786 / output 3,594 / cacheRead 176,640 / cacheWrite 0). Two findings from the live data are folded in: the per-request map undercounts (cumulative was ~3x the map sum), so threads now top up to the exact cumulative counter with a remainder entry, and legacy uncompressed json rows are read alongside zstd. Format details additionally verified against Zed's source (crates/agent/src/db.rs: DataType, DbThread, TokenUsage, SerializedLanguageModel).

PR #599, opened 2026-07-02
Fix cursor project parsing

Enclavet:
## Summary

- Fixes Cursor usage being lumped under a single `cursor` project instead of the real workspace. Recent Cursor builds renamed the per-workspace composer list key from `composer.composerData` to `composer.composerHeaders` (identical `{ allComposers: [{ composerId }] }` shape); `loadWorkspaceMap` only read the old key, so every composer orphaned into the `cursor` catch-all.
- `loadWorkspaceMap` now reads **both** keys and merges their `allComposers` lists — restoring per-project attribution on current Cursor while staying backward-compatible with older installs.
- Adds a regression test covering the new `composer.composerHeaders` key and the legacy `composer.composerData` key.

## Testing

- [x] `npm test` passes *(cursor suite 13/13; build clean.

### For new providers only:

- [ ] N/A — this fixes the existing Cursor provider, not a new one

**Before (old `composer.composerData`-only query):** all usage orphaned 

projects: [ 'cursor' ] [cursor] calls=22 cost=$0.2506

**After (this PR):** usage attributed to the real workspace

discovered projects: [ '-Users-me-projects-superzoom', 'cursor' ] [-Users-me-projects-superzoom] calls=21 in=67472 out=7284 cost=$0.1653 [cursor] calls=1 in=545 out=5572 cost=$0.0852

### DIFF
--- a/src/providers/cursor.ts
+++ b/src/providers/cursor.ts
@@ -193,22 +193,31 @@ function loadWorkspaceMap(workspaceStorageDir: string): WorkspaceMapping {
       continue
     }
     try {
+      // Cursor renamed the per-workspace composer list from
+      // 'composer.composerData' to 'composer.composerHeaders' in newer builds
+      // (identical { allComposers: [{ composerId }] } shape). Read both keys
+      // and merge so the composer -> workspace mapping keeps working across
+      // Cursor versions. Without this, on builds that only write
+      // 'composer.composerHeaders' every composer falls through to the
+      // 'cursor' orphan bucket and per-project attribution is lost.
       const rows = db.query<{ value: string }>(
-        "SELECT value FROM ItemTable WHERE key='composer.composerData'",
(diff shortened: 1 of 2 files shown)


ozymandiashh (approved):
Looks good and safe to merge. The fix is minimal, preserves the legacy `composer.composerData` path, and correctly reads the newer `composer.composerHeaders` key so project attribution works across Cursor versions.

Non-blocking suggestions:
- Update the workspace mapping comment in `src/providers/cursor.ts` to mention both `composer.composerData` and `composer.composerHeaders`.
- Consider adding one regression test for the transition case where both keys are present in the same workspace DB, to lock in the intended merge behavior.

I ran the cursor provider tests and build locally. The broader suite has unrelated pre-existing failures on the same files when checked against `origin/main`, so I do not consider those a regression from this PR.

PR #608, opened 2026-07-03
feat(copilot): track GitHub Copilot JetBrains IDE usage

### LINKED ISSUE #211: Add token usage for copilot in Intellij IDEA
Is it possible to add in the providers entry for GitHub Copilot the detection of the sessions opened in IntelliJ IDEA ?
I use copilot both in VSCode and Intellij IDEA and for now codeburn only detects my token usage in VSCode

NihalJain:
## What & why

The JetBrains Copilot plugin (IntelliJ, PyCharm, RubyMine, …) stores its chat/agent sessions under `~/.config/github-copilot/<ide>/<kind>/<storeId>/` — a location none of the existing Copilot sources (CLI JSONL, VS Code chat sessions/transcripts, OTel SQLite) read. As a result all JetBrains Copilot usage was silently uncounted in every CodeBurn report. This adds a reader for that store so those sessions are discovered, priced, and attributed to the right project.

Fixes https://github.com/getagentseal/codeburn/issues/211

## How it works

- **Reader.** The store's session content is a Nitrite `.db` — an H2 MVStore of Java-serialized documents. It is scanned as `latin1` for byte-offset stability: no Java deserializer, no new dependency, and it is not SQLite so `node:sqlite` is not involved.
- **Reply text.** Assistant replies live in nested-escaped `{"__first__":{"type":"Subgraph"…}}` blobs. The text is recovered by unescaping one level at a time and, at the depth where the Markdown record's `data` field is a well-formed one-level-escaped JSON document, reading it structurally — so a reply containing its own quotes is never truncated or duplicated (which would otherwise inflate the estimate).
- **Tokens/cost.** The store records no token counts, so output tokens are estimated from the reply text (`CHARS_PER_TOKEN = 4`, re-decoded latin1→utf8 so multibyte replies count by codepoint) and every call is marked `costIsEstimated`. Failed generations (error status, no reply) are billed $0.
- **Sessions.** One `.db` holds many chat tabs; turns are grouped back to their conversation GUID so the UI shows one session per tab, deduped by reply content per conversation.
- **Project attribution**, most authoritative first:
  1. the plugin-recorded `projectName` field (JetBrains Copilot 1.12+), joined across kind dirs by store id — the billable turns live in `chat-agent-sessions`, but the label is usually written into the sibling `chat-sessions`/`chat-edit-sessions` store. Read length-delimited and re-decoded latin1→utf8 so non-ASCII repo names round-trip.
  2. the `.git` repo root of a referenced `file://` path.
  3. a generic `copilot-jetbrains` bucket when neither signal exists. The conversation title is a chat-thread name, not a project, so it is kept out of the project field and surfaced as the session label instead.

Override the JetBrains github-copilot root with
`CODEBURN_COPILOT_JETBRAINS_DIR`.

## Docs

- `docs/providers/copilot.md` — full JetBrains section (store layout, latin1 scan, reply extraction, projectName precedence + cross-kind join).
- `docs/providers/README.md` — Copilot storage updated to note the Nitrite .db.

## How to verify

- `npm test -- copilot` and `npx tsc --noEmit` (fixtures reproduce the real nested-escaped .db framing, including quote- and multibyte-bearing replies).
- End to end against a real install: `CODEBURN_CACHE_DIR=$(mktemp -d) node dist/cli.js status --provider copilot \ --period all --format menubar-json` — JetBrains sessions appear By-Project under their real repo names.
- Set `CODEBURN_COPILOT_JETBRAINS_DIR` to a fixture root to parse a controlled store without touching the real config dir.

## Testing

- [x] `npm test` passes

### For new providers only:


NihalJain:
Screenshot (Before) with npx codeburn@0.9.15:
Screenshot 2026-07-03 at 6 33 13 PM


Screenshot (After) this patch:
Screenshot 2026-07-03 at 6 33 54 PM

NihalJain:
@iamtoruk

iamtoruk:
This is excellent work. The structural read of the escaped blobs, the agent-mode handling that never bills the user's prompt as output, the errored-turn $0 path, and the cross-kind projectName join are all the right calls, and the docs are genuinely useful. Verified locally: full suite and typecheck pass on this branch, and again after merging current main into it.

I pushed one maintainer commit (d2edf8c) with two fixes so this can merge now:

1. Content-derived dedup keys. The keys were positional (conversation + scan index). Copilot is a durable provider: cached turns are never deleted, and a re-parse appends any key it has not seen. MVStore compaction can rewrite the store with blobs in a different byte order, and a rewrite that moves a new blob ahead of an old one hands the new turn the old turn's key (skipped as already seen) and re-emits the old turn under a fresh index, double-billing it while the new turn goes uncounted. The keys now derive from the reply content (sha256 prefix plus a per-hash occurrence counter, which also keeps errored turns with empty replies distinct). There is a regression test that reproduces the compaction scenario and fails on the positional scheme.

2. Test isolation for the new env var. CODEBURN_COPILOT_JETBRAINS_DIR is now in the env-isolation cleared list, so a developer who has it set in their shell cannot leak their real JetBrains store into the pre-existing copilot fixture tests.

Follow-ups worth their own issue, none blocking: discovery re-reads every .db in full on each scan for the projectName join (worth memoizing by path and mtime, the menubar rescans often); the conversation-title regex only matches printable ASCII, so non-Latin titles fall out of session grouping; the model token list will need topping up as Copilot adds models (it currently lacks a few recent ones); and the git walk-up never resolves Windows file URIs, though the projectName tier covers Windows users on current plugins. Happy to file that issue with details if you want to pick any of them up.

One accepted limitation worth knowing about: the store has no per-turn timestamps, so turns are stamped with the file mtime, and the first import of a long-lived active store lands its history on one day. The durable cache then freezes each turn's first-seen stamp, so it self-corrects going forward.

Thanks for closing the oldest gap in the provider matrix. Merging once CI is green.

iamtoruk (approved):
Verified locally: full suite and typecheck on the branch head, again after merging current main, and the new compaction regression test fails on the pre-fix positional keys and passes on the content-derived ones.

PR #609, opened 2026-07-03
fix(overview): pin number grouping to en-US so output is locale-indep…

NihalJain:
…endent

`codeburn overview` formatted token/call/session counts with a bare `toLocaleString()`, which groups digits according to the host's locale. On an Indian-locale machine (en-IN) a value like 2,002,000,000 renders as 2,00,20,00,000, so the rendered summary — and its snapshot test — differed by machine. (formatCost already avoided this via an explicit thousands regex.)

Pin `formatTokens` to 'en-US' and route the remaining count columns through a new `formatCount` helper that does the same, so the overview output is byte-identical everywhere.

Verify: `npm test -- overview` (previously failed on non-US locales; the "thousands separators" case now passes regardless of $LANG / ICU default).

## Testing

- [x] `npm test` passes

### For new providers only:

### DIFF
--- a/src/overview.ts
+++ b/src/overview.ts
@@ -14,7 +14,14 @@ function formatCost(usd: number): string {
   return baseCost(usd).replace(/(\d)(?=(\d{3})+(\.|$))/g, '$1,')
 }
 function formatTokens(n: number): string {
-  return Math.round(n).toLocaleString()
+  // Pin the locale so grouping is deterministic regardless of the host's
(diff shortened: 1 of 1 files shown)


iamtoruk (approved):
Verified: reproduced the failure on main under an en-IN locale (grouping renders 2,00,20,00,000), confirmed this branch fixes it, and re-ran the overview tests after merging current main into the branch locally. All call sites in overview.ts are covered. The same bare toLocaleString() pattern exists in six other files; that sweep is follow-up material, not this PR's scope.

PR #613, opened 2026-07-03
test(cursor): cover dual composer key workspace routing

ozymandiashh:
## Summary

Follow-up to #599. The Cursor workspace map now reads both `composer.composerData` and `composer.composerHeaders`, but the regression coverage only exercised those keys in separate workspace DBs. This adds an end-to-end fixture where both rows exist in the same `workspaceStorage/<hash>/state.vscdb` and verifies the parser routes both composers into the real workspace source.

Without this case, a regression could still keep the workspace source visible via the legacy `composer.composerData` row while leaking `composer.composerHeaders` composers into the `cursor` orphan bucket.

## What changed

- Refactored the Cursor workspace breakdown fixture helper so one workspace DB can contain multiple composer-list rows.
- Kept the legacy `createWorkspaceDir(...)` wrapper, so existing tests still write `composer.composerData` exactly as before.
- Added a parser-level regression test for a workspace containing both `composer.composerData` and `composer.composerHeaders`, including an orphan composer assertion to prove the headers composer does not fall through to `cursor`.

## Testing

- `npm test -- --run tests/providers/cursor.test.ts tests/providers/cursor-workspace-breakdown.test.ts` -> 23 passed.
- `npm run build` -> succeeds.

### DIFF
--- a/tests/providers/cursor-workspace-breakdown.test.ts
+++ b/tests/providers/cursor-workspace-breakdown.test.ts
@@ -71,9 +71,14 @@ function createGlobalDb(composerIds: string[]): string {
   return dbPath
 }
 
+type WorkspaceComposerRow = {
+  key: 'composer.composerData' | 'composer.composerHeaders'
+  composerIds: string[]
+}
+
 /// Creates one workspaceStorage/<hash>/ subdir with workspace.json (folder URI)
-/// and state.vscdb (composer.composerData listing the supplied composerIds).
-function createWorkspaceDir(hash: string, folderUri: string, composerIds: string[]): void {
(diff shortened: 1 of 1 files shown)


ozymandiashh:
Superseded by #599 (commit `400ca08`, merged 2026-07-02), which shipped the `composer.composerHeaders` dual-key workspace mapping in `src/providers/cursor.ts` **and** added regression tests for both keys in `tests/providers/cursor.test.ts`. This standalone test is now redundant duplicate coverage. Closing to keep the regression tests for this behavior consolidated in one file.

PR #618, opened 2026-07-04
fix(kiro): Fix Kiro IDE session parsing — context.messages, .kiro-server path, workspace-sessions

Enclavet:
## Problem

The Kiro provider returns **0 calls** on both Linux remote dev boxes and Windows despite hundreds of session files existing on disk. Three independent issues combine to make the provider non-functional:

1. **`extractText` misses the `entries` key** — Kiro IDE stores message content in `context.messages[].entries` (not `.content`), so the recursive text extractor never finds it and returns 0 chars for every file.

2. **`parseModernExecution` only checks top-level conversation arrays** — Current Kiro builds store messages at `data.context.messages`, not `data.messages`. The parser never looks inside `data.context`.

3. **`getKiroAgentDir` missing `.kiro-server` path on Linux** — Remote dev boxes (common at AWS/Amazon) store Kiro data at `~/.kiro-server/data/User/globalStorage/kiro.kiroagent` instead of `~/.config/Kiro/...`. Without the fallback, discovery returns 0 sources on Linux.

4. **Newer Kiro builds write to `workspace-sessions/`** — Post-June 2026 Kiro stores session state in `kiro.kiroagent/workspace-sessions/<base64>/<sessionId>.json` with a `history[].message` format that the provider doesn't discover or parse.

## Fix

1. Add `'entries'` to `extractText()` recursion keys
2. Check `data.context[key]` before `data[key]` for conversation arrays in `parseModernExecution`
3. Add `existsSync` check for `~/.kiro-server/data/...` in `getKiroAgentDir()` (Linux only)
4. New discovery path for `workspace-sessions/<base64>/*.json` + parser for `history[].message` format
5. Extract tool names from `usageSummary[].usedTools` when present (structured billing data)
6. Add `chatSessionId` to session ID resolution chain
7. Add Kiro-specific tool name mappings (`executeBash` → Bash, `fsWrite` → Edit, `grepSearch` → Grep, etc.)

## Impact

| Environment | Before | After |
|---|---|---|
| Linux remote dev box | 0 calls | 7,637 calls, $940 |
| Windows | 0 calls | 937 calls, $367 |

## Testing

- 6 new test cases covering `context.messages` with entries, `usageSummary` tool extraction, execution index skipping, workspace-sessions discovery/parsing, stub-only sessions, and `sessions.json` skipping
- All 42 kiro tests pass
- Verified on Windows (cleared `~/.cache/codeburn/session-cache.json` required after upgrade since stale cache stores empty parse results from the broken code)

## Note for users upgrading

After installing this version, delete `~/.cache/codeburn/session-cache.json` (or equivalent) to force a fresh parse. The old cache stores "0 results" for every Kiro IDE file from when the parser was broken — the fingerprint-based cache won't re-parse unchanged files without clearing it.

### DIFF
--- a/src/providers/kiro.ts
+++ b/src/providers/kiro.ts
@@ -1,6 +1,6 @@
 import type { Dirent } from 'fs'
 import { existsSync } from 'fs'
-import { readdir, readFile } from 'fs/promises'
+import { readdir, readFile, stat } from 'fs/promises'
 import { basename, dirname, extname, join } from 'path'
 import { homedir } from 'os'
 
@@ -585,7 +585,7 @@ function createParser(source: SessionSource, seenKeys: Set<string>): SessionPars
       const historyArr = record['history']
       if (Array.isArray(historyArr) && typeof record['sessionId'] === 'string') {
         const sessionId = record['sessionId'] as string
-        const modelRaw = (record['selectedModel'] as string) ?? ''
+        const modelRaw = stringField(record, ['selectedModel'])
         let modelId = normalizeModelId(modelRaw)
         if (modelId === 'auto' || !modelId) modelId = 'kiro-auto'
 
@@ -601,7 +601,8 @@ function createParser(source: SessionSource, seenKeys: Set<string>): SessionPars
           if (!rec) continue
 
           // Track if this session references execution files (which are parsed separately)
-          if (typeof rec['executionId'] === 'string') hasExecutionRefs = true
+          const execBacked = typeof rec['executionId'] === 'string'
+          if (execBacked) hasExecutionRefs = true
 
           const msg = asRecord(rec['message'])
           if (!msg) continue
@@ -610,7 +611,11 @@ function createParser(source: SessionSource, seenKeys: Set<string>): SessionPars
           if (role === 'user' && text) {
             inputChars += text.length
             pendingUserMessage = text.slice(0, 500)
-          } else if (role === 'assistant' && text && text !== 'On it.') {
+          } else if (role === 'assistant' && !execBacked && text && text !== 'On it.') {
+            // An item carrying an executionId is execution-backed: its content is
+            // counted from the execution file, so counting it here would double-count.
+            // 'On it.' is the observed placeholder text Kiro writes for such stubs
+            // when the executionId rides a separate history item.
             outputChars += text.length
             hasRealAssistantContent = true
           }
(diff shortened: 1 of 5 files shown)


ozymandiashh:
Thanks for jumping on the Kiro parsing fix. There are a few things I need fixed before I can merge this:

1. Add cache invalidation for the Kiro parser fix.

Kiro is not in `PROVIDER_PARSE_VERSIONS` in `src/session-cache.ts:110-122`, and `computeEnvFingerprint()` only includes a parser version when the provider has an entry there (`src/session-cache.ts:136-141`). For unchanged files, `parseProviderSources()` reuses cached entries unless `cachedFileNeedsProviderReparse()` says otherwise (`src/parser.ts:1992-1998`), and that currently returns `false` for Kiro (`src/parser.ts:1910-1924`).

So if a user already has a Kiro source file cached as `turns: []` from the old broken parser, this PR can still show 0 calls until they manually clear cache. Please add a Kiro parser version, for example `kiro: 'ide-parsing-v1'`, or equivalent invalidation.

Also please add a regression test that goes through `parseAllSessions()` with an existing Kiro zero-turn `session-cache.json` entry. The current Kiro tests parse provider sources directly, so they do not cover the stale-cache path.

2. Avoid double-counting workspace-session stubs and execution files.

Discovery includes both nested execution files and workspace-session JSON files (`src/providers/kiro.ts:768-780`, `src/providers/kiro.ts:789-805`). Execution parsing uses `kiro:${sessionId}:${executionId}` (`src/providers/kiro.ts:353-354`), but workspace-session parsing emits a separate call keyed only by session (`src/providers/kiro.ts:597-617`, `src/providers/kiro.ts:633-652`).

That means a workspace-session entry that references the same `executionId` can be counted once from the execution file and once again from the workspace-session file. The test at `tests/providers/kiro.test.ts:900-925` is the risky case: the assistant only says `"On it."`, the real output is in execution files, but the workspace-session still emits input tokens. Please either dedupe against referenced execution IDs or skip stub-only workspace-session entries when the execution file is present.

3. Actually scan both Linux Kiro roots.

The comment says to check both paths, but `getKiroAgentDir()` returns `~/.kiro-server/...` as soon as it exists and never scans `~/.config/Kiro/...` (`src/providers/kiro.ts:677-681`). Since `createKiroProvider()` passes one `agentDir` into discovery (`src/providers/kiro.ts:813-817`, `src/providers/kiro.ts:835-836`), this can still miss current sessions if `.kiro-server` exists but is stale or empty while `.config/Kiro` has data.

The targeted Kiro test file passes locally for me: `npm run test -- tests/providers/kiro.test.ts` reports 42 passing tests. Once the three issues above are fixed, I am good to merge this.

Enclavet:
Updated PR with fixes.

iamtoruk:
All three review items are addressed in your update, thanks. I pushed one maintainer commit (926dcdb) on top to close the remaining gaps so this can merge now:

- The cache test file now actually exercises the stale-cache path. It seeds a session-cache.json with a zero-turn Kiro entry and runs the full parseAllSessions() pipeline twice: at the current fingerprint the cache is honored (control), at the pre-fix fingerprint the section is discarded and the file re-parsed. The previous version compared a hardcoded string against a hash, which could not fail. It also caught a real trap while I wrote it: the env-isolation setup clears CODEBURN_CACHE_DIR before every test, so the seed has to re-assert it in beforeEach.
- Stub detection no longer depends only on the literal 'On it.' text. Any history item carrying an executionId is treated as execution-backed and its text is never counted, so a future stub-text change cannot silently reintroduce double-counting. The text check stays as a fallback for stubs whose executionId rides a separate item.
- Workspace-session timestamps resolve before the dedup key is consumed, and a failed stat now drops the call instead of stamping it with the current time.
- selectedModel is read through stringField instead of a cast, stat is imported statically, and the unused test imports are gone.

One note on the PR description: with the parser version registered, the manual cache deletion step is no longer needed. The fingerprint change invalidates the old section automatically on first run, and the regression test covers exactly that scenario. Worth editing the note so users do not clear caches for nothing.

Two follow-ups worth checking against your real data, neither blocking: whether mixed sessions exist where history holds both real inline replies and executionId refs (the user prompts of exec-backed turns would still be counted from both paths), and whether the base64 decode of workspace-session dir names should share one helper with resolveWorkspaceProject, which strips a trailing underscore instead of mapping underscores to padding. If you see either in the wild, a follow-up PR is very welcome.

Full suite is green locally (1504 tests) and tsc is clean. Merging once CI completes.

iamtoruk (approved):

PR #622, opened 2026-07-05
fix(opencode): merge legacy file sessions and SQLite DB instead of short-circuiting

### LINKED ISSUE #621: OpenCode provider ignores SQLite DB when legacy JSON files exist (Windows, v0.9.15)
## Summary

CodeBurn v0.9.15 on Windows only reads the **legacy file-based JSON sessions** from OpenCode and completely ignores the current **SQLite database** (`opencode.db`), even when both exist side-by-side. This causes the dashboard to show only a tiny fraction of actual usage (in my case: **$3.31 / 12 sessions** instead of the real **~$2,617 / 71 models**).

## Environment

- **OS:** Windows 11
- **Node:** v24.13.0 (`node:sqlite` available and working)
- **CodeBurn:** v0.9.15 (installed globally via `npm install -g codeburn`)
- **OpenCode:** current (sessions stored in SQLite `opencode.db`, 1.5 GB, modified today)

## Data layout on disk

```
~\.local\share\opencode\
├── opencode.db                          ← 1.5 GB SQLite, current (modified today 20:41)
├── opencode - Copy.db                   ← backup
├── opencode - Copy (2).db               ← backup
└── storage\
    ├── session\*.json                   ← 13 legacy JSON files (Jan 22 – Feb 13 2026, ~5 KB total)
    └── message\*.json                    ← 797 legacy message files (same period)
```

## Symptom

`codeburn report --provider opencode -p all` shows:
- **$3.31 cost, 549 calls, 12 sessions**
- Last data point: **2026-02-13**
- Only the 13 legacy JSON sessions are visible

The 1.5 GB SQLite DB (containing all usage from Feb through today, ~$2,617 across 71 models including gpt-5.5 at ~$1,000) is **completely ignored**.

For comparison, `tokscale` reading the same machine shows the correct numbers from the SQLite DB:
- **$2,616.89** across 71 models, last data point today (2026-07-05)
- Top model: gpt-5.5 at $1,000

## Root cause (found in bundled `dist/main.js`)

In `src/providers/opencode.ts`, `createOpenCodeProvider` → `discoverSessions()`:

```js
async discoverSessions() {
  const fileSessions = await discoverOpenCodeFileSessions(resolvedDataDir, "opencode");
  if (fileSessions.length > 0) return fileSessions;   // ← short-circuits, never falls back to SQLite
  return discoverSqliteSessions(sqliteConfig);
}
```

The comment above it says:
> "OpenCode 1.1+ stores sessions as file-based JSON; older builds used a SQLite DB. Prefer file-based when present, otherwise fall back to the DB so pre-migration installs keep reporting."

**This assumption is backwards for real-world OpenCode installs on Windows.** OpenCode migrated from file-based JSON (legacy, `storage/session/*.json`) to SQLite (`opencode.db`) at some point. After the migration, the legacy JSON files are **left behind** on disk but are no longer written to. The current data lives exclusively in `opencode.db`.

Because the discovery short-circuits on `fileSessions.length > 0`, any install that still has the old `storage/session/` directory (even with just a handful of stale files from January) will **never** read the SQLite DB. The SQLite fallback only fires when the legacy directory is completely empty/absent — which is not the case for users who upgraded OpenCode in place.

## Expected behavior

CodeBurn should read **both** sources and merge them (dedup by session ID), the way tokscale does (`opencode.db` **or/and** `storage/message/` legacy). At minimum, it should prefer the SQLite DB when it exists and is larger/newer than the legacy files, rather than short-circuiting on any legacy file presence.

## Suggested fix

Either:
1. **Merge both sources** (dedup by session ID), so legacy + SQLite are both counted; or
2. **Prefer SQLite when present** and only fall back to legacy files when no `opencode*.db` exists; or
3. At minimum, compare mtime/size and prefer the source that is actively being written to.

## Reproduction

1. Install OpenCode on Windows, use it for a while (creates `storage/session/*.json`)
2. Upgrade OpenCode in place (it migrates to `opencode.db` SQLite, leaves legacy files behind)
3. Continue using OpenCode (data now flows into `opencode.db` only)
4. `npm install -g codeburn` and run `codeburn report --provider opencode -p all`
5. Observe: only the stale legacy JSON sessions are shown; current SQLite data is missing

## Related

- #359 reported a similar "No usage stats for OpenCode on windows" issue for v0.9.9 (compound `:sessionId` path bug, fixed in #347). This is a **different** bug: the SQLite DB is not even attempted because the legacy file short-circuit fires first.

morpheus9393:
## Summary

Fixes #621

The OpenCode provider's `discoverSessions()` short-circuited on legacy file-based JSON sessions (`storage/session/*.json`) and never fell back to the SQLite DB (`opencode.db`) when any legacy file existed. This caused **migrated installs** — which keep stale legacy JSON files from before the OpenCode SQLite migration alongside the now-current SQLite DB — to report only a tiny fraction of actual usage, with the SQLite DB silently ignored.

## Root cause

```ts
async discoverSessions(): Promise<SessionSource[]> {
  const fileSessions = await discoverOpenCodeFileSessions(resolvedDataDir, 'opencode')
  if (fileSessions.length > 0) return fileSessions   // ← short-circuits
  return discoverSqliteSessions(sqliteConfig)
}
```

The comment said "Prefer file-based when present, otherwise fall back to the DB so pre-migration installs keep reporting." But the real-world migration direction is the opposite: OpenCode migrated **from** file-based JSON **to** SQLite. After an in-place upgrade, the legacy `storage/session/` directory is left behind with stale files, and all new data flows into `opencode.db`. The short-circuit fires on the stale legacy files and the SQLite DB is never read.

## Fix

Merge both sources instead of either/or:

```ts
async discoverSessions(): Promise<SessionSource[]> {
  const fileSessions = await discoverOpenCodeFileSessions(resolvedDataDir, 'opencode')
  const sqliteSessions = await discoverSqliteSessions(sqliteConfig)
  return [...fileSessions, ...sqliteSessions]
}
```

Per-message dedup is already handled in `createSessionParser` via `seenKeys` (keyed by `${provider}:${sessionId}:${messageId}`), so sessions that exist in both stores are not double-counted.

## Verification (Windows 11, Node v24.13.0)

Before fix (v0.9.15, `codeburn report --provider opencode -p all`):
- **$3.31** cost, 549 calls, 12 sessions
- Last data point: 2026-02-13 (stale legacy JSON only)
- Top model: missing (GPT-5.5 not visible)

After fix (patched dist, same command):
- **$3,182.91** cost, 40,338 calls, 470 sessions
- Last data point: 2026-07-05 (today)
- Top model: GPT-5.5 at $999.99 (matches tokscale's independent reading of the same SQLite DB)

Cross-checked against `tokscale` reading the same `opencode.db`: $2,616.89 across 71 models — consistent with the patched CodeBurn output (the small delta is explained by CodeBurn counting a few legacy JSON sessions that tokscale deduplicates differently).

## Why this also helps Mac users

The bug is not Windows-specific. Any OpenCode install that was upgraded in place (on any OS) and retains the `storage/session/` directory with even one legacy JSON file will hit the same short-circuit. The issue title mentions Windows because that's where it was reported, but the fix applies to all platforms.

## Checklist

- [x] Fix applied to `src/providers/opencode.ts` (not just the bundled dist)
- [x] Verified on Windows 11 with a real 1.5 GB `opencode.db` + 13 legacy JSON files
- [x] Cross-validated against tokscale's independent SQLite reader
- [x] No double-counting (seenKeys dedup confirmed)
- [x] Commit message references #621

### DIFF
--- a/src/providers/opencode.ts
+++ b/src/providers/opencode.ts
@@ -55,13 +55,16 @@ export function createOpenCodeProvider(dataDir?: string): Provider {
       return toolNameMap[rawTool] ?? rawTool
     },
 
-    // OpenCode 1.1+ stores sessions as file-based JSON; older builds used a
-    // SQLite DB. Prefer file-based when present, otherwise fall back to the DB
-    // so pre-migration installs keep reporting.
+    // OpenCode migrated from file-based JSON (storage/session/*.json) to a
+    // SQLite DB (opencode.db). After an in-place upgrade, legacy JSON files
+    // remain on disk while all new data flows into the SQLite DB. Merge both
+    // sources so migrated installs keep reporting legacy sessions AND pick up
+    // current SQLite data. Dedup is handled per-message in createSessionParser
+    // via seenKeys (keyed by `${provider}:${sessionId}:${messageId}`).
     async discoverSessions(): Promise<SessionSource[]> {
       const fileSessions = await discoverOpenCodeFileSessions(resolvedDataDir, 'opencode')
-      if (fileSessions.length > 0) return fileSessions
-      return discoverSqliteSessions(sqliteConfig)
+      const sqliteSessions = await discoverSqliteSessions(sqliteConfig)
+      return [...fileSessions, ...sqliteSessions]
     },
 
     createSessionParser(source: SessionSource, seenKeys: Set<string>): SessionParser {


ozymandiashh:
Nice catch, and the diagnosis is right. I traced the parser path to sanity-check the merge approach and it holds up:

- `discoverSqliteSessions` already returns `[]` gracefully when `opencode.db` is absent (the `isSqliteAvailable` guard plus the `readdir` catch and the `dbPaths.length === 0` early return), so calling it unconditionally does **not** regress pure file-based / pre-migration installs.
- `parserDedup` is created once per provider run in `parser.ts` and shared across every source, so the file pass populates the seen keys first and the SQLite pass skips the duplicates. The dedup you rely on works as described, and the updated header comment matches the real migration direction now.

Two things worth tightening before merge:

1. **Regression test for the both-stores case.** The existing discovery tests in `opencode.test.ts` are all SQLite-only, so `[...[], ...sqlite]` reduces to the old behavior and none of them actually exercise the bug this fixes. A fixture with legacy JSON **and** a SQLite DB present that asserts no double-count would lock this in; otherwise a future revert to the short-circuit stays green.

2. **ID stability across the migration.** The per-message dedup only holds if OpenCode preserves `sessionId`/`messageId` when it migrates JSON to SQLite. If the migration re-keys them, sessions present in both stores double-count. Your numbers suggest the practical impact is tiny (the time windows barely overlap), but if you can confirm the IDs are stable it'd be worth a one-line comment noting why the dedup is safe; if they aren't, the guarantee needs a second look.

Out of scope for this PR, but the header comment in `opencode-file-parser.ts` still describes file-based JSON as the *newer* 1.1+ format, which now contradicts the migration direction documented here. Might be worth reconciling in a follow-up.

ozymandiashh (approved):
Approving on correctness. I traced the parser path (graceful `[]` when `opencode.db` is absent, `parserDedup` shared across sources so the file pass dedups the SQLite pass) and CI is green. The both-stores regression test and the ID-stability note above remain worth a follow-up, but neither blocks the fix.

Issue #614, opened 2026-07-03
Design: tool-def deferral coverage (detect deferral gaps, restore/tune native deferral, proxy for non-native agents)


AVSRPA1KR:
Follow-up to the acting-layer epic (#602). This is the deferral piece called out in the epic's non-goals as a separate, evidence-gated track. Design issue first, then PR, same flow the epic used.

> **Scope update:** the proxy fallback for non-native agents (originally part 3) has been **split into its own design issue** — request-path (Class C) code warrants a separate review, and it now also covers a provider-agnostic gateway (Anthropic + OpenAI), which is beyond this issue's analyzer/config scope. This issue is now scoped to **detection (part 1)** and **native config restore/tune (part 2)**.

## Motivation

Tool-def deferral is no longer a missing capability on the happy path: Claude Code ships tool search **enabled by default** — MCP tool defs are deferred and discovered on demand, and the API mechanism behind it (`defer_loading` + a tool search tool) is GA on the Claude API, with the prefix untouched and prompt caching preserved by design (per current docs at https://code.claude.com/docs/en/mcp and https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool).

The real gap is coverage:

1. **Deferral silently turns off.** Claude Code disables tool search on Vertex AI and whenever `ANTHROPIC_BASE_URL` points to a non-first-party host (docs: "most proxies don't forward `tool_reference` blocks"). Users behind any gateway/proxy, or with a stale `ENABLE_TOOL_SEARCH=false`, are paying full tool-def cost on every turn with no signal that anything changed.
2. **Deferral config has tunable waste even when on.** `alwaysLoad: true` servers that are rarely invoked, and threshold mode (`auto:N`) left at defaults, both leave measurable prefix cost on the table.
3. **Non-native agents have no deferral at all.** Cursor, Cline, Aider and similar load every def every turn — addressed in the **separate proxy design issue**, not here.

None of this is routing. No model changes, no per-turn classification, no rewriting of anything the user wrote.

## Captured evidence (real usage, Opus 4.x)

Measured deferred vs non-deferred on the same workloads, in separate sessions each:

- 20-turn tool-heavy session: all defs loaded every turn **$1.31** vs deferred **$0.90** (~31% less, no model change).
- Two-turn "chat → one tool lookup" flow: **$0.96** non-deferred vs **$0.35** deferred (~64% less).
[...]

PR #630, opened 2026-07-08
feat(optimize): mcp-deferral-gaps finding family (part 1 of #614)

AVSRPA1KR:
Implements **part 1 (detection)** of the deferral-coverage design, #614.

Adds a read-only `mcp-deferral-gaps` finding family to `optimize` — three detectors, no file mutation anywhere in this PR:

- **`mcp-deferral-off`** — sessions carrying MCP tool-def overhead with zero `ToolSearch` invocations and no `deferred_tools_delta` inventory across the window. Cause attribution, each with its own message and fix: stale `ENABLE_TOOL_SEARCH=false` (settings env, any scope, or shell profile), non-first-party `ANTHROPIC_BASE_URL` (reported as *unknown proxy* — never assumes capability), Vertex AI, all observed Claude Code versions predating default-on tool search (v2.1.7), or a generic fallback.
- **`mcp-alwaysload-hygiene`** — `alwaysLoad`-pinned servers whose observed call rate is below 1 call / 5 sessions. Notes the 5s startup-blocking cost.
- **`mcp-defer-threshold`** — `ENABLE_TOOL_SEARCH=auto:N` overrides whose threshold the estimated def volume never reaches; recommends the tightest `auto:N` that would defer.

Detection only — no `FindingApply` payloads, no `act` changes, no file mutation. **Zero new output** for users without deferral gaps (deferral-active evidence suppresses the findings). 32 fixture-based unit tests; config readers take an injectable `homeDir` for hermetic tests.

Live-docs discrepancies noted in the commit body (changelog records default-on at v2.1.7 not first-ship; per-tool `anthropic/alwaysLoad` lives in server-served `_meta`, so server-level only).

Refs #614

### DIFF
--- a/src/optimize.ts
+++ b/src/optimize.ts
@@ -7,7 +7,7 @@ import { homedir } from 'os'
 import { readSessionLines, readSessionFileSync } from './fs-utils.js'
 import { discoverAllSessions } from './providers/index.js'
 import { parseJsonlLine, shouldSkipLine } from './parser.js'
-import type { DateRange, ProjectSummary } from './types.js'
+import type { DateRange, ProjectSummary, SessionSummary } from './types.js'
 import { formatCost } from './currency.js'
 import { formatTokens } from './format.js'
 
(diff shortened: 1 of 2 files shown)


iamtoruk:
Merged as part of #631, whose branch carried this PR's commits as its base per the stacking you described. The detection family, all 51 dedicated tests, and the finding text shipped exactly as designed here; the squash credits you as author. Closing this one so the pair has a single merge history. The empirical note from verification, since this PR asked the right question: across 120 real deferral-active sessions the deferred-tool inventory is emitted at session start regardless of usage, so the detector's no-signal heuristic holds with no minimum-usage floor needed. Thank you for the cleanest big first contribution this repo has seen, and looking forward to the part 3 design issue.

AVSRPA1KR:
Thanks @iamtoruk, will update [768](https://github.com/getagentseal/codeburn/issues/768) on the part 3 design.

PR #631, opened 2026-07-08
feat(act): defer-* plan kinds for native deferral config (part 2 of #614)

AVSRPA1KR:
Implements **part 2 (restore/tune native deferral)** of #614.

> **Stacked on #630.** The `defer-*` plans wire to the detection findings and can't compile without them, so this branch necessarily contains #630's detection commit as its base. **Review/merge #630 first**; this PR's actual contribution is the second commit (`feat(act): defer-* plan kinds…`). Once #630 lands, this retargets to a one-commit diff.

Wires the `mcp-deferral-gaps` findings to the existing `act` machinery with three new plan kinds, all through `ConfigDocs`/`runAction` (journaled, backed up, stale-guarded, dry-run previewable, `act undo` restores byte-identical):

- **`defer-enable`** — removes a stale `ENABLE_TOOL_SEARCH=false` from the settings env of the recorded scope. Refusal paths render as manual notes instead of plans: shell-profile lines (codeburn only appends marker blocks to shell files, never edits user lines), **unknown proxies** (setting the override blind makes requests fail on proxies that don't forward `tool_reference` — the note says verify first), Vertex, old versions.
- **`defer-alwaysload`** — strips `alwaysLoad: true` from named servers in the exact config files the finding recorded. Version-gated to Claude Code v2.1.121+ (below it / unparseable / probe failure all refuse). Preview notes the 5s startup block removed.
- **`defer-threshold`** — rewrites the `auto` override to the recommended `auto:N`, or deletes it when the default already defers.

Findings now carry `FindingApply` payloads; detector text is unchanged, so plain `optimize` renders byte-identically. Every plan links its finding id into the `ActionRecord` and states the change takes effect on the next session. 22 new tests (byte-identical apply/undo on odd-key-order JSON, unknown-proxy refusal, version gate, shell-profile refusal, drift guard).

Refs #614


iamtoruk (approved):
Cleared for 0.9.16 after deep verification. The one open question from the earlier review round, whether the detector false-positives on setups with deferral correctly on but low MCP usage, is settled empirically: across 120 real sessions on a deferral-active machine, the deferred-tool inventory is emitted at session start (typically line 3) regardless of usage, including 28 sessions that never called ToolSearch, and every session lacking the signal had zero MCP involvement. Negative control on a live deferral-on setup produces no finding; a synthetic gap fires with honest cause text. All three plan kinds were live-fired in a fully isolated home: each edit was surgical (sibling keys and file bytes preserved around the change), and act undo restored both configs byte-identical to their originals. Merged cleanly with current main (pushed to this branch), tsc clean, suite green apart from the known #681 items, and disabling the detector kills 19 of the 51 dedicated tests. The reserved proxy-verified path stays: it is documented, test-covered, and part 3 is an accepted upcoming track. Thanks for following the design-issue flow exactly and for the strongest first contribution this repo has had.

AVSRPA1KR:
Thanks @iamtoruk

PR #637, opened 2026-07-09
Add progress indicator + event-loop yielding for large cold-cache Claude scans

41fred:
## Summary
- On a large `~/.claude/projects` tree (hundreds of project dirs, e.g. a git-worktree-per-task workflow, plus months of session history), the CLI report-style commands (`overview`, `export`, `status`, etc.) go through `parseAllSessions` → `scanProjectDirs` with no progress indicator of any kind. A cold-cache run can take tens of seconds with zero output, which reads as a hang rather than as working. (The interactive `dashboard.tsx` TUI already has its own `loading` state for this — this gap is specific to the non-interactive CLI paths.)
- Adds a lightweight, TTY-gated `\r`-updated progress line during both the directory-discovery and file-parsing passes in `scanProjectDirs`, and yields to the event loop every 25 items in each pass so a long scan gives the OS scheduler more regular break points instead of running as one long block.
- Gated on `process.stderr.isTTY === true` so piped/captured output (`--no-color`, `export`, tests, any subprocess capturing stderr) is completely unaffected — verified by the full existing suite passing untouched.

## Verification
- Ran a true cold-cache scan against a real `~/.claude` tree with 934 project dirs / 18,182 session files (`rm ~/.cache/codeburn/session-cache.json` then `overview -p all` under a pty): 51s wall, 60% CPU (not pegged), progress visible throughout instead of silence. Warm-cache totals matched exactly before/after.
- `npx tsc --noEmit` clean.
- `npx vitest run`: all 121 test files / 1549 tests pass, unchanged.

## Test plan
- [x] Cold-cache run on a large real tree (934 dirs / 18k files) — progress renders, totals unchanged, CPU not pegged
- [x] Warm-cache run — no behavior change, no stray output
- [x] Piped/non-TTY output (`--no-color`, redirected to file) — no progress bytes leak into stdout/stderr
- [x] Full existing test suite passes unchanged
- [x] `tsc --noEmit` passes

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1530,6 +1530,8 @@ async function scanProjectDirs(
   const unchangedFiles: Array<{ filePath: string; dirName: string; cached: CachedFile }> = []
   const changedFiles: Array<{ filePath: string; info: FileInfo }> = []
 
+  const discoverProgress = createScanProgress('scanning claude project dirs', dirs.length)
+  let dirsDone = 0
   for (const { path: dirPath, name: dirName } of dirs) {
     const jsonlFiles = await collectJsonlFiles(dirPath)
     for (const filePath of jsonlFiles) {
@@ -1544,7 +1546,10 @@ async function scanProjectDirs(
         changedFiles.push({ filePath, info: { dirName, fp } })
       }
     }
+    dirsDone++
+    await discoverProgress.tick(dirsDone)
   }
+  discoverProgress.finish()
 
   // Pre-seed dedup set from cached (unchanged) files
   for (const { cached } of unchangedFiles) {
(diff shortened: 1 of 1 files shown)


iamtoruk:
Thanks @41fred — this shipped via #652, which preserves your commit and adds one hardening fix on top: the scan-progress line was gated only on stderr.isTTY, which is also true inside the interactive dashboard and `codeburn compare`, so it printed over the Ink frame and garbled the TUI. #652 gates it on an explicit 'interactive UI is live' flag instead, so plain CLI commands keep the progress line while the dashboard/compare stay clean. Your progress-indicator + event-loop-yield feature is now on main. Closing in favor of #652.

PR #651, opened 2026-07-09
Cold-scan progress indicator (hardened) — supersedes #637

iamtoruk:
Supersedes #637 (credit: @41fred). Their commit is preserved as-is; this adds one hardening commit on top.

## What #637 does
A cold-cache scan over a large `~/.claude/projects` tree runs 30-50s with no output, reading as a hang. #637 adds a `\r`-updated progress line during the scan (gated on isTTY) plus event-loop yielding every 25 items. Good, real problem, correct instinct.

## The bug it had
The progress line was gated only on `process.stderr.isTTY`, which is ALSO true when the interactive dashboard and `codeburn compare` render Ink to the same terminal. So the `\r` progress printed over their frame and garbled the TUI. I reproduced it under a PTY: running the cold dashboard showed `codeburn: scanning…` bytes in the dashboard's own output stream.

## The fix
Gate on an explicit "an interactive Ink UI is live" flag instead of isTTY alone. `renderDashboard` and `renderCompare` call `setInteractiveScanUI()` before they render, so every scan they trigger — including the dashboard's 30s auto-refresh and its `getPlanUsages → parseAllSessions` path — stays silent, while plain CLI commands (overview, export, status) keep the progress line. The event-loop yielding stays ungated (it's a responsiveness win independent of display).

## Verification (all under a PTY, cold cache, 60-dir tree, plan configured)
- Interactive dashboard (even with `--refresh 3` forcing rescans): **0** progress bytes
- `codeburn compare`: **0** progress bytes
- `codeburn overview` (plain CLI): progress **shows**
- Piped / non-TTY: **0** progress bytes
- Full suite: 1,572 tests pass. tsc clean. Warm-cache totals unchanged.

## Reviewed by Fable 5 (adversarial)
The review caught that my first approach (per-call-site opt-out) missed a transitive scan path via `getPlanUsages` and over-suppressed compare's pre-render progress. That led directly to this module-flag design, which covers every transitive path and is verified end-to-end. (The specific plan-usage corruption did not reproduce empirically even when forced — caching absorbs it — but the flag design defends against that whole class regardless.)

## Note
The unit test guards the gate; the end-to-end interactive-silence proof is the PTY runs above (documented here, since a live Ink render isn't reproducible in vitest).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "codeburn",
-  "version": "0.9.12",
+  "version": "0.9.15",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "codeburn",
-      "version": "0.9.12",
+      "version": "0.9.15",
       "license": "MIT",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/src/compare.tsx
+++ b/src/compare.tsx
@@ -4,7 +4,7 @@ import { render, Box, Text, useInput, useApp, useStdout } from 'ink'
 import type { ModelStats, ComparisonRow, CategoryComparison, WorkingStyleRow } from './compare-stats.js'
 import { aggregateModelStats, computeComparison, computeCategoryComparison, computeWorkingStyle, scanSelfCorrections } from './compare-stats.js'
 import { formatCost } from './format.js'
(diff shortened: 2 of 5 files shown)


iamtoruk:
Replaced by a version with the disallowed Claude co-author trailer stripped from the contributor's commit (repo CI blocks it). See the new PR.

PR #665, opened 2026-07-14
Kiro: IDE v2 session store support + credit-based pricing across all formats

Enclavet:
## Summary

Kiro's newly released 1.0 IDE (observed on 1.0.138; the 0.12.x builds still write the v1 globalStorage format) moves sessions to `~/.kiro/sessions/<workspace-hash>/sess_<id>/` — a self-contained `session.json` + `messages.jsonl` event log — instead of globalStorage. This PR adds a v2 parser and discovery, and — validated against a real upgraded machine — moves Kiro cost accounting from token estimation to the **real metered credits** Kiro records, across every format that carries them (CLI, v1 IDE executions, v2 IDE).

Token estimation wasn't just imprecise, it inverted model rankings. On a real machine, one week of Kiro usage:

| Model | Token-estimated | Metered credits | Error |
|---|---:|---:|---|
| Sonnet 4.6 | $339.82 | $161.52 | 2.1× overstated |
| Opus 4.6 | $11.15 | $130.57 | 11.7× understated |
| Fable 5 | $7.20 | $118.15 | 16× understated |

### New: IDE v2 store (`sess_*/messages.jsonl`)

- Parses the event-sourced turn stream (`user` / `turn_start` / `assistant` / `tool_call` / `tool_result` / `usage_summary` / `turn_end`) with per-execution dedup keys (`kiro-v2:<session>:<execId>`) and defensive flushes for out-of-order events and in-flight sessions.
- Prices turns from `usage_summary.promptTurnSummaries[].usage` credits at the public overage rate (`USD_PER_KIRO_CREDIT = $0.04`; individual plan is $20/mo for 1,000 credits) — same never-understate approach as the Codebuff provider.
- Counts `tool_result` content as input context (matching the CLI parser's `ToolResults` treatment) and takes the real `modelId` from `session.json`, so v2 sessions aren't mislabeled `kiro-auto`.
- v1 → v2 is a clean cutover (no migration/dual-write observed); dedup namespaces are disjoint, so nothing is counted twice.

### Fixed: credit pricing now actually reaches reports

- **`metering_usage` values are credits, not dollars.** The existing CLI parser summed them as USD, overstating CLI cost 25× at the real rate.
- **Kiro joins the `costUSD` pass-through allowlist** in `providerCallToCachedCall` (alongside mistral-vibe, devin, hermes, vercel-gateway, antigravity). Without this, the parser's credit-based cost was discarded at the cache boundary and re-derived from estimated tokens — silently undoing all credit pricing. Tradeoff (shared with all allowlisted providers, documented): `price-override` / `model-alias` don't affect Kiro dollar amounts.
- **v1 IDE executions are credit-priced too.** `usageSummary[].usage` (the v1 predecessor of v2's `promptTurnSummaries`) was already being iterated for `usedTools` but its credit values were ignored. Since v2 only ships in brand-new builds, v1 is the format nearly all Kiro IDE users are on today. Verified per-turn semantics on real data (values fluctuate; not cumulative counters).
- **Uniform fallback contract.** All three credit parsers gate on summed credits > 0 and fall back to token-estimated pricing with `costIsEstimated: true`. This also fixes two CLI edge cases: no-metering turns were priced at a frozen $0, and empty `metering_usage` arrays (observed on 75 of 10,105 real turn metadatas — meta written before metering lands) passed the old truthy-presence check and froze $0 marked as real.

### Fixed: v2 reasoning tokens double-counted

Reasoning was folded into `outputTokens` *and* reported in `reasoningTokens`, but aggregation sums the two fields. Now disjoint (matching codex/gemini); reasoning is still billed as output in the pricing fallback.

### Cache invalidation

Both persistence layers froze pre-fix dollar values and never self-invalidate (historical session files don't change):

- session cache: `CACHE_VERSION` 4 → 5
- daily rollup: `DAILY_CACHE_VERSION` 10 → 11 with `MIN_SUPPORTED_VERSION` raised (same pattern as the v10 cursor accounting bump)

One-time full re-parse; verified live that the corrected numbers land.

## Testing

- 20+ new tests: v2 parsing/discovery/dedup/routing, credit vs fallback pricing per format, `tool_result` turn scoping, empty-metering edge, reasoning disjointness, and a **mixed-format coexistence suite** (legacy `.chat` + v1 executions + workspace-sessions + CLI + v2 on one machine) asserting exact aggregate counts and no double counting.
- The cost pass-through test is mutation-verified: it runs end-to-end through `parseAllSessions` and fails with the token re-price ($0.000048 vs $0.10) if the allowlist entry is removed.
- Validated against a real machine: 9,995 CLI turns, 221 v1 executions, and 24 v2 turns now carry metered cost ($1,793 total); call counts unchanged before/after (no calls gained, lost, or double-counted).
- Full suite green; `tsc` clean; `tsup` bundle builds.

## Docs

`docs/providers/kiro.md` updated: v2 store layout, credit pricing sources per format, dedup namespaces, the parse-time cost-freezing contract (and its `price-override` implication), and the disjoint-store v1→v2 verification. A comment at `fingerprintFile` documents the companion-file blind spot (CLI credits / v2 modelId live next to, not in, the fingerprinted transcript).


iamtoruk (approved):
**Approve.** Reviewed the full diff and verified the key mechanisms against the source rather than trusting the description; CI is green (assess / check / semgrep).

### Verified independently
- **The allowlist is the correct lever.** `cachedCallToApiCall` uses `costUSD: call.costUSD ?? costUSD`, so it takes the cached pass-through value when present and otherwise re-derives from tokens. Adding `kiro` to the `providerCallToCachedCall` allowlist is exactly what stops the credit cost from being silently re-priced, and the mutation-verified test ($0.10 vs $0.000048 if the entry is removed) genuinely guards it.
- **v2 routing** is checked before the generic `.jsonl`/CLI branch — correct, since v2's `messages.jsonl` also ends in `.jsonl`. The `v2Root` derivation correctly skips the scan when only `agentDirOverride` is set, and the "trap" test confirms the `cli/` dir isn't descended into.
- **Reasoning disjointness is necessary, not cosmetic:** downstream cost sums `outputTokens + reasoningTokens` for non-Claude providers, so folding reasoning into output would double-count. Keeping them separate matches codex/gemini.
- `parseWorkspaceSession` is a behavior-preserving extraction of the previous inline block.

### On the `$0.04`/credit rate
Pricing every credit at the *overage* rate ($0.04) rather than the in-plan rate ($20 / 1000 = $0.02) overstates cost for users within their monthly allotment. I'm comfortable with it here for two reasons: it's a first-party change (Kiro sets its own credit pricing and writes the on-disk format, so both the rate and the format decode are authoritative), and the direction of the bias makes Kiro look *more* expensive in a provider-comparison tool — against the vendor's own interest — so the "never understate" convention reads as applied honestly. Consistent with the existing Codebuff treatment.

### Minor / non-blocking (all pre-existing)
- `costIsEstimated` is unified across all three parsers but isn't part of `CachedCall` and isn't read by any report/audit today, so through the cached path it's currently forward-looking metadata. The cost *values* are what flow, and those are correct.
- Credit-less turns whose model resolves to `kiro-auto` fall back to `$0` (`calculateCost` returns 0 for unknown models). v2/CLI/v1 all carry a real model, so this only touches legacy workspace-sessions — strictly better than before.

### Notably good
Mixed-format coexistence suite asserting an exact aggregate with disjoint dedup namespaces; the empty-`metering_usage`-array edge fix (turns that previously froze $0 as "real"); honest documentation of the companion-file fingerprint blind spot and the future v1→v2 migration guard; correct `CACHE_VERSION` and `DAILY_CACHE_VERSION` bumps with `MIN_SUPPORTED_VERSION` raised.

PR #672, opened 2026-07-15
feat(dashboard): make Daily Activity history directly scrollable

ihearttokyo:
## Summary

- keep full Daily Activity history directly in the main dashboard instead of a separate view or show/hide mode
- scroll one active day at a time with Up/Down, page with Page Up/Page Down or Space/Shift+Space, and jump with `g`/`G`
- keep the overview and other breakdowns scoped to the selected period while Daily Activity loads full history for the active provider
- document the in-place controls and add cursor boundary coverage

## Validation

- `npm run build`
- `npx tsc --noEmit`
- `npm test -- --run tests/dashboard.test.ts --maxWorkers=1 --testTimeout=20000` — 14 passed
- `npm test -- --run --maxWorkers=1 --testTimeout=20000` — 128 files passed, 2 opt-in infrastructure files skipped; 1,677 tests passed, 5 skipped
- live PTY test against the Homebrew-installed CodeBurn 0.9.15 runtime: Up/Down, Page Up/Page Down, and `g`/`G` all updated Daily Activity in place and `q` exited cleanly
- source-backed reconciliation: 21 genuine-format Codex rollout JSONL files, 21 active days, 42 calls, $2.5676 total; source SHA-256 `29364829f265a006cd2cbca965eec1714aea1fea94039b609453c7eeadd32f1c`
- screenshot verification matched every visible date, cost, and call count to CodeBurn's JSON report; selected 7-Day overview remained 17 calls / $1.3748

## Terminal evidence

Initial position (`Showing 1–14 of 21`):

Daily Activity initial position

After one Down Arrow (`Showing 2–15 of 21`):

Daily Activity after one Down Arrow

Screenshot SHA-256:

- `daily-activity-initial.png`: `e72e16591a5e0a2528a54e682327cbe01e7089e80c77c9291a80eb6cd15e3867`
- `daily-activity-arrow-down.png`: `5e69bce1aeffd1d00f037450115d78989ed3654072a2799dbd16e340609c8cf7`


ihearttokyo:
This is my first-ever PR, so apologies if the implementation is rough. The UX is the proposal: keep Daily Activity on the main dashboard and scroll it in place with ↑/↓. Please feel free to reimplement it rather than take my commit verbatim.

ozymandiashh:
Thanks for the work here — the UX direction and the evidence are unusually thorough, especially for a first PR. I found two issues that need addressing before merge:

1. **Network-provider correctness:** `loadDailyHistory()` calls `parseAllSessions(undefined, provider)`. The Vercel AI Gateway parser explicitly returns without yielding anything when `dateRange` is undefined (`src/providers/vercel-gateway.ts`), so its Daily Activity will be empty; on the **All** provider tab, Vercel usage will be silently omitted from the history. Please pass a concrete range and add coverage for the network-provider path.

2. **The default dashboard now starts an unbounded history parse:** `cli-date.ts` intentionally caps the existing **All / 6 Months** period because older Codex/Cursor/Claude histories can be very large. Passing `undefined` bypasses that bound for providers that do not cap themselves. On Today/7 Days, the refresh interval also starts this scan in parallel with the bounded `reloadData()` parse every 30 seconds. The current 21-day fixture does not exercise that scale/concurrency path. Please keep the default history bounded (the existing six-month range would be consistent), or make deeper history lazy/explicit, and avoid running two parser passes concurrently.

Local verification:

- `npm test -- --run tests/dashboard.test.ts --maxWorkers=1 --testTimeout=20000` — 14 passed
- `npm run build` — passed
- `git diff --check` — passed

I could not independently reproduce the PR's full-suite/typecheck claim from a clean `npm ci` because this checkout does not install several existing optional packages (`undici`, MCP SDK, `selfsigned`, etc.); those failures are not introduced by this diff.

PR #677, opened 2026-07-16
Extract shared token estimation helper, migrate all CHARS_PER_TOKEN sites

### LINKED ISSUE #644: Shared token-estimation module: CHARS_PER_TOKEN is independently defined in 7 files
## Problem

The content-length token heuristic is copy-pasted across the codebase: `CHARS_PER_TOKEN` is defined independently in src/context-tree.ts, src/context-budget.ts, src/providers/kiro.ts, src/providers/cursor.ts, src/providers/warp.ts, src/providers/codex.ts, and src/providers/cursor-agent.ts, with `Math.ceil(len / CHARS_PER_TOKEN)` re-derived at each site. If the heuristic is ever refined (different ratio for code vs prose, per-tokenizer), it will silently apply to some providers and not others, skewing exactly the cross-provider comparisons the product exists for.

## Evidence

- `grep -rln 'CHARS_PER_TOKEN = ' src --include='*.ts'` lists the 7 files above.

## Proposed fix

One exported `estimateTokensFromChars(chars: number): number` in a shared leaf module; delete the local constants. Mechanical change, no behavior difference today, and it gives estimated-token provenance a single home (pairs well with #639).

avs-io:
## Problem

`CHARS_PER_TOKEN` is defined independently in 7 files (`src/context-tree.ts`, `src/context-budget.ts`, and the kiro / cursor / warp / codex / cursor-agent providers), with `Math.ceil(len / CHARS_PER_TOKEN)` re-derived at each site. Any future refinement of the heuristic would silently apply to some providers and not others, skewing cross-provider comparisons. (#644)

## Fix

New leaf module `src/token-estimate.ts` exporting `CHARS_PER_TOKEN` and `estimateTokensFromChars(chars)`; all 7 local constants deleted and every call site migrated. One additional hardcoded `/ 4` site in `src/context-tree-codex.ts` (same heuristic, just never named) is migrated too.

Behavior-identical by construction: every site already used ratio 4 with `Math.ceil`, and the two sites with extra semantics keep them — `cursor-agent`'s non-positive guard is preserved, and `context-tree-codex`'s fractional decoded-size input goes through the same `ceil(chars / 4)`.

The module imports nothing from `src/`, so providers stay lazy-loadable.

## Tests

- New `tests/token-estimate.test.ts` pins the heuristic: ratio value, ceil rounding, zero/boundary inputs. The ratio can now only change in one place, guarded by one test.
- `npx tsc --noEmit` clean.
- Full suite: `npx vitest run` — 130 files passed, 1713 tests passed, 0 failed.

Closes #644

### DIFF
--- a/src/context-budget.ts
+++ b/src/context-budget.ts
@@ -4,8 +4,8 @@ import { join } from 'path'
 import { homedir } from 'os'
 
 import { readSessionFile } from './fs-utils.js'
+import { estimateTokensFromChars } from './token-estimate.js'
 
-const CHARS_PER_TOKEN = 4
 const SYSTEM_BASE_TOKENS = 10400
 const TOOL_TOKENS_OVERHEAD = 400
 const SKILL_FRONTMATTER_TOKENS = 80
@@ -20,7 +20,7 @@ export type ContextBudget = {
 }
 
 function estimateTokens(text: string): number {
-  return Math.ceil(text.length / CHARS_PER_TOKEN)
+  return estimateTokensFromChars(text.length)
 }
 
 async function readConfigFile(path: string): Promise<Record<string, unknown> | null> {
--- a/src/context-tree-codex.ts
+++ b/src/context-tree-codex.ts
@@ -4,6 +4,7 @@ import { basename, join } from 'path'
 import { homedir } from 'os'
 
 import { readSessionLines } from './fs-utils.js'
+import { estimateTokensFromChars } from './token-estimate.js'
 import {
   add,
   estimateTokens,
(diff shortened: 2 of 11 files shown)


iamtoruk (approved):
Verified in depth: no stray CHARS_PER_TOKEN or hardcoded /4 sites remain, the ratio pin tests fail when the constant changes, the cursor-agent non-positive guard is preserved, and a 30-day all-provider report on real data is byte-identical between main and this branch, so the behavior-identical claim holds empirically. The two local test failures during verification are pre-existing on main (app/electron/cli.test.ts expects a built dist/cli.js) and unrelated.

Issue #681, opened 2026-07-16
app/electron/cli.test.ts fails on a fresh clone: dev-resolution tests require a built dist/cli.js


iamtoruk:
## Symptom

On a fresh clone (or any checkout that has never run `npm run build`), `npx vitest run app/electron/cli.test.ts` fails 2 tests:

- resolveCodeburnPath (Vite development) > prefers the executable repo dist/cli.js when the Vite dev server is set
- resolveCodeburnPath (Vite development) > prefers the repo dev CLI over a persisted-path file (stale global) in dev

```
TypeError: .toMatch() expects to receive a string, but got object
  app/electron/cli.test.ts:47:35
  expect(resolveCodeburnPath()).toMatch(/dist\/cli\.js$/)
```

CI stays green because the pipeline builds before testing, so this only bites contributors running the suite locally. Reproduced on a clean checkout of main at 5869cad on 2026-07-16, and confirmed absent when dist/cli.js exists.

## Root cause

The Vite-development branch of `resolveCodeburnPath()` only returns the repo CLI path when `dist/cli.js` actually exists and is executable. Without a prior build it falls through and the function returns a non-string shape, so the assertion gets an object.

## Suggested fix

Have the two dev-resolution tests create the fixture they depend on: write a stub executable `dist/cli.js` into the temp repo root the test already controls (or chmod-touch one in a tmpdir and point resolution at it), rather than depending on the developer having built the real bundle. Skipping when the file is absent would also work but hides real resolution regressions, so a stub fixture is preferred.

Introduced in 916aa64 (Electron security + spawn efficiency batch, 2026-07-16).

PR #684, opened 2026-07-16
fix(copilot): bill CLI input and cache tokens from session.shutdown

### LINKED ISSUE #676: Copilot CLI provider under-reports cost — ignores  session.shutdown  token/cost data in  ~/.copilot/session-state/*/events.jsonl
Summary

The Copilot provider ( src/providers/copilot.ts ) under-reports cost and token usage for Copilot CLI sessions ( ~/.copilot/session-state/*/events.jsonl ). It only parses  assistant.message  events, which carry  outputTokens  only. It never reads the session's  session.shutdown  event, which already contains the full token breakdown (input, cache-read, cache-write, output) and the actual computed cost per model — data the CLI itself writes to disk.

The code even acknowledges the gap in a comment:

```
// Copilot JSONL only logs outputTokens; inputTokens are NOT available.
// Cost will be lower than actual API cost. This is the original
// behaviour — OTel data (below) replaces it when available.
const costUSD = calculateCost(currentModel, 0, outputTokens, 0, 0, 0)
```

The OTel/ agent-traces.db  fallback mentioned in the file only applies to VS Code Copilot Chat — it does not exist for standalone Copilot CLI usage, so CLI-only users get no compensating richer source at all. For CLI users, cost is silently and significantly underreported (in my session, cache-read tokens alone were 35k+, entirely excluded).

Evidence

Actual  session.shutdown  event from  ~/.copilot/session-state/<session-id>/events.jsonl  (produced by Copilot CLI 1.0.68):

`{"type":"session.shutdown","data":{"shutdownType":"routine","totalPremiumRequests":2,"totalNanoAiu":10001450000,"tokenDetails":{"input":{"tokenCount":4},"cache_read":{"tokenCount":35495},"cache_write":{"tokenCount":35783},"output":{"tokenCount":345}},"totalApiDurationMs":7495,"sessionStartTime":1784102040274,"eventsFileSizeBytes":98693,"codeChanges":{"linesAdded":0,"linesRemoved":0,"filesModified":[]},"modelMetrics":{"claude-sonnet-5":{"requests":{"count":2,"cost":2},"usage":{"inputTokens":71282,"outputTokens":345,"cacheReadTokens":35495,"cacheWriteTokens":35783,"reasoningTokens":31},"totalNanoAiu":10001450000,"tokenDetails":{"input":{"tokenCount":4},"cache_read":{"tokenCount":35495},"cache_write":{"tokenCount":35783},"output":{"tokenCount":345}}}},"currentModel":"claude-sonnet-5","currentTokens":27176,"systemTokens":11899,"conversationTokens":605,"toolDefinitionsTokens":14668}}`

Note  modelMetrics.<model>.usage  already contains  inputTokens ,  outputTokens ,  cacheReadTokens ,  cacheWriteTokens ,  reasoningTokens , plus  totalNanoAiu  (looks like a pre-computed cost/AIU unit) and  requests.cost . This is a per-session rollup written once at shutdown, in addition to the per-turn  assistant.message  events currently parsed.

Expected behavior

The JSONL parser ( createJsonlParser  in  src/providers/copilot.ts ) should also handle a  session.shutdown  event type and either:

1. Use  `modelMetrics.<model>.usage`  (inputTokens/outputTokens/cacheReadTokens/cacheWriteTokens) to emit a corrected/supplementary  ParsedProviderCall  per model with the real token breakdown, replacing the output-tokens-only estimate from  `assistant.message`  events, or
2. At minimum, use it to scale/correct the aggregate cost computed from  `assistant.message`  events for that session.

This event only appears once per session file (at shutdown) and is present even without any VS Code/OTel/JetBrains integration, so it's the most reliable source of accurate cost for CLI-only users.

Environment

• GitHub Copilot CLI version: 1.0.68 (and later — same schema observed across recent versions)
• Session files:  ~/.copilot/session-state/<uuid>/events.jsonl 
• OS: Windows (but the event schema itself is OS-independent)

iamtoruk:
## Problem

The Copilot CLI provider parsed only assistant.message events, which carry outputTokens alone. The session.shutdown event the CLI itself writes to ~/.copilot/session-state/*/events.jsonl carries the full per-model token breakdown (input, cache read, cache write, output, reasoning), and it was ignored, so every CLI session under-reported real spend. Closes #676.

## Fix

When a CLI session's events.jsonl contains session.shutdown with modelMetrics, emit one supplementary call per model carrying only the input and cache tokens the per-turn events lack (output stays owned by assistant.message, so nothing double-counts). Pure input is recovered as inputTokens minus cache reads and writes, since the CLI writes inputTokens cache-inclusive (verified arithmetically against the issue sample: 4 + 35495 + 35783 = 71282). requests.cost is a premium-request count, not USD, so cost is priced from the measured tokens. Gated on the CLI (non-transcript) format: VS Code, JetBrains and OTel paths are untouched. Sessions without a shutdown event (crashed or still running) keep exactly the old behavior.

A shutdown event with neither its own timestamp nor sessionStartTime falls back to the last stamped event in the session, so the supplementary call is never dropped by the date-range filters.

PROVIDER_PARSE_VERSIONS copilot is bumped so already-cached CLI sessions re-parse and pick up the new billing.

## Verification

- 6 new tests: shutdown-backed totals and cost, per-model attribution for multi-model sessions, no-shutdown sessions unchanged, transcript sessions ignore shutdown, dedup stability across re-parses, and the missing-timestamp fallback. Each verified by mutation: disabling the shutdown handling fails exactly the shutdown-dependent tests; reverting the timestamp fallback fails exactly its test.
- End to end through the real CLI with synthetic session-state fixtures: main bills 0.010 EUR and 0 percent cache hit for two sessions; this branch bills 0.152 EUR with the correct 49 percent cache split, including a session whose shutdown carries no timestamp.
- tsc clean; full suite green apart from the two pre-existing app/electron/cli.test.ts failures on unbuilt checkouts (#681).

## Honest limits

No real events.jsonl existed on the dev machine; the shutdown shape comes from the CLI 1.0.68 sample in #676, cross-checked arithmetically. If a future CLI version renames modelMetrics fields the supplementary call simply does not emit and behavior degrades to the current undercount, never to a crash or double-count.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1,3 +1,4 @@
+import { existsSync } from 'fs'
 import { lstat, readFile, readdir, stat } from 'fs/promises'
 import { basename, dirname, join, resolve, sep } from 'path'
 import { readSessionLines } from './fs-utils.js'
@@ -2040,7 +2041,19 @@ function getOrCreateProviderSection(cache: SessionCache, provider: string): Prov
   const envFp = computeEnvFingerprint(provider)
   const existing = cache.providers[provider]
   if (existing && existing.envFingerprint === envFp) return existing
-  const section = { envFingerprint: envFp, files: {} }
+  const section: ProviderSection = { envFingerprint: envFp, files: {} }
+  // A fingerprint change (env override or parse-version bump) must re-parse
+  // every present source, but for durable providers the cache is the ONLY
+  // remaining record of usage whose source rows were already pruned (OTel
+  // orphans). Discarding those with the section would permanently erase
+  // month-to-date history that cannot be re-derived, so carry forward exactly
+  // the entries whose source no longer exists; everything present on disk is
+  // dropped and re-parsed under the new fingerprint.
+  if (existing && DURABLE_PROVIDER_NAMES.has(provider)) {
+    for (const [path, file] of Object.entries(existing.files)) {
+      if (!existsSync(path)) section.files[path] = file
+    }
+  }
   cache.providers[provider] = section
   return section
 }
--- a/src/providers/copilot.ts
+++ b/src/providers/copilot.ts
@@ -738,6 +738,11 @@ function createJsonlParser(
         if (!currentModel) return // no toolCallIds to infer model from
       }
 
+      // Shutdown rollups may lack their own timestamp; remember the last
+      // stamped event so the supplementary call is never left with an empty
+      // timestamp, which the date-range filters silently drop.
+      let lastEventTimestamp = ''
+
       for (const line of lines) {
         let event: CopilotEvent
         try {
@@ -745,6 +750,7 @@ function createJsonlParser(
         } catch {
           continue
         }
+        if (typeof event.timestamp === 'string' && event.timestamp) lastEventTimestamp = event.timestamp
 
         if (event.type === 'session.start') {
           if (!isTranscript) {
@@ -787,7 +793,7 @@ function createJsonlParser(
           if (!isRecord(modelMetrics)) continue
 
           const shutdownTimestamp =
-            event.timestamp ?? timestampToISO(shutdownData.sessionStartTime)
+            (event.timestamp ?? '') || timestampToISO(shutdownData.sessionStartTime) || lastEventTimestamp
 
           for (const [model, metrics] of Object.entries(modelMetrics)) {
             if (!model || !isRecord(metrics)) continue
--- a/src/providers/copilot.ts
+++ b/src/providers/copilot.ts
@@ -195,12 +195,28 @@ type SubagentSelectedData = {
   tools?: string[]
 }
 
+// Per-model usage rollup the CLI writes into session.shutdown. inputTokens is
+// cache-INCLUSIVE (input + cache_read + cache_write); see the shutdown handler.
+type ShutdownModelUsage = {
+  inputTokens?: number
+  outputTokens?: number
+  cacheReadTokens?: number
+  cacheWriteTokens?: number
+  reasoningTokens?: number
+}
+
+type SessionShutdownData = {
+  modelMetrics?: Record<string, { usage?: ShutdownModelUsage }>
+  sessionStartTime?: number
+}
+
 type CopilotEvent =
   | { type: 'session.start'; data: SessionStartData; timestamp?: string }
   | { type: 'session.model_change'; data: ModelChangeData; timestamp?: string }
   | { type: 'user.message'; data: UserMessageData; timestamp?: string }
   | { type: 'assistant.message'; data: AssistantMessageData; timestamp?: string }
   | { type: 'subagent.selected'; data: SubagentSelectedData; timestamp?: string }
+  | { type: 'session.shutdown'; data: SessionShutdownData; timestamp?: string }
 
 type ChatJournalPathSegment = string | number
 type ChatSessionRequest = Record<string, unknown>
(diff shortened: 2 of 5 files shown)


ozymandiashh (changes requested):
The supplement design is the right call: input/cache on a separate call with output owned by the per-turn events means the durable union-merge works with you instead of against you, and scanned-while-active sessions converge to the correct total without any cache surgery. We stress-tested exactly this problem space today and this shape is the one that survives it. Two confirmed issues before merge, both small:

1. **The rollup's reasoningTokens get billed as output at query time.** Copilot is not in the cost-preserving set at `providerCallToCachedCall` (parser.ts:1882), so the parse-time costUSD you tested is discarded on cache write and recomputed by `cachedCallToApiCall`, where `outputForCost = outputTokens + reasoningTokens` for non-claude providers (parser.ts:1973-1975). The supplementary call carries `reasoningTokens` with `outputTokens: 0`, so those tokens are priced as output on every cached read, which is exactly what the PR states it avoids. The new tests read `ParsedProviderCall.costUSD` directly and never cross the recompute, so they stay green. Suggested fix: emit `reasoningTokens: 0` on the supplementary call (consistent with the per-turn and OTel paths), and route at least one test through the cache round trip.

2. **Repeated shutdown events: first wins, newer totals are dropped.** The handler emits inside the line loop and registers the dedup key immediately, so a second `session.shutdown` appended to the same file (resumed session) is skipped even though its cumulative totals supersede the first. In an append-only log the last shutdown is the authoritative one. Suggested fix: accumulate per-model rollups while scanning and emit once at end of file, last write wins.

Notes, non-blocking:
- The parse-version bump wipes the whole durable copilot section, including cached OTel rows whose spans may have been pruned from agent-traces.db. #688 deliberately avoids bumping copilot for exactly that reason, so the two PRs currently disagree on the same line of session-cache.ts (and will conflict textually). Worth deciding the migration story once for both.
- The transcript gate is content-based (session.start producer sniffing). A truncated or rotated file that lost its first line would classify as CLI; a discovery-level flag on the source (session-state dir vs transcript dir) would make the gate independent of file content.
- The timestamp fallback chain is nice; the all-three-absent case (no event timestamp, no sessionStartTime, no prior stamped event) still yields an empty timestamp and the date-range filters drop such turns silently. Cheap to pin with a test.

ozymandiashh:
Two upgrades to my review after deeper verification, one raising severity and one new:

**The parse-version bump note is not a nit, it is certain data loss.** On fingerprint mismatch `getOrCreateProviderSection` replaces the entire copilot section with empty files (parser.ts:2039-2046), and the durable orphan pass then has nothing to preserve. OTel conversations still in agent-traces.db re-parse and self-heal; conversations already pruned from the db exist only in the cache the bump just discarded, so their tokens vanish from month-to-date totals on upgrade day, unrecoverable. #688's own session-cache comment cites this verbatim as the reason copilot must never be bumped. Suggested mitigations, either works: scope the invalidation to jsonl sources (a per-source-type fingerprint or a jsonl-only sub-version), or carry OTel entries across the fingerprint change.

**The timestamp fallback order attributes the supplement days too early.** `(event.timestamp ?? '') || timestampToISO(sessionStartTime) || lastEventTimestamp` prefers session START over the last stamped event. For a shutdown with no timestamp on a multi-day session, the supplement lands at session start, so the date-range filters hide the input/cache cost from today/week views while the per-turn output stays visible. The last stamped event approximates when the rollup was written; it should come before sessionStartTime in the chain.

ozymandiashh:
One more, and it violates the PR's own stated safety property ("behavior degrades to the current undercount, never to a crash or double-count"):

`numberOrZero` returns 0 for any non-number (string-typed numbers, null, negative), so if a future CLI version ships `cacheReadTokens`/`cacheWriteTokens` as strings while `inputTokens` stays numeric, the subtraction silently stops applying and the full cache-inclusive `inputTokens` is billed at the pure-input rate. With the issue's real sample that turns a 4-token input into 71,282 input-rate tokens, an overcount, not an undercount. Cheap guard: emit the supplement only when all four usage fields are finite numbers, otherwise skip it (true degradation to the old behavior).

Issue #675, opened 2026-07-16
OpenCode does not work


chimon2000:
When running `npx codeburn overview --provider opencode -p week`

```bash
CodeBurn  Last 7 Days

No usage found for Last 7 Days.
```

ozymandiashh:
Thanks for the report. Two OpenCode parsing fixes landed on main after the latest npm release (0.9.15, published Jul 2), so `npx codeburn` does not include them yet: #622 (merge legacy file sessions and the SQLite DB instead of short-circuiting) and #620 (support custom data dir and DB prefix). There is a good chance your case is already covered.

To confirm, could you share: (1) OpenCode version, (2) OS, (3) where your OpenCode data lives, default `~/.local/share/opencode` or a custom dir, and (4) whether that folder holds a SQLite db, JSON session files, or both?

A release including both fixes is coming shortly. If it still reproduces there with the details above, we will dig further.

re-ava:
```
opencode -v
1.18.3
```

OS: Mac OS

Data location is `~/.local/share/opencode`. The folder holds a SQLite database: `opencode.db`

ozymandiashh:
Thanks, that confirms it. OpenCode 1.18.3 stores sessions in the SQLite database only (`opencode.db`), and codeburn 0.9.15 (what `npx codeburn` runs today) predates #622, which is exactly the fix for this: reading sessions from the SQLite db instead of short-circuiting on the legacy JSON store. Your setup (default `~/.local/share/opencode`, `opencode.db`) is discovered correctly by current main.

The fix will be in the next release. If you want to verify sooner: clone the repo, `npm install && npm run build`, then `node dist/cli.js overview --provider opencode -p week` should show your sessions.

iamtoruk:
Fixed by #622: discovery merges legacy JSON sessions and opencode.db instead of short-circuiting, shipped in v0.9.16 (current npm latest). Verified end to end in the 0.9.18 pre-release sweep: SQLite-only sessions are read and attributed. Reopen if 0.9.16+ still shows nothing.

PR #685, opened 2026-07-16
feat(doctor): per-provider detection diagnostics

### LINKED ISSUE #642: codeburn doctor: per-provider detection status so empty or wrong numbers are self-diagnosable
## Problem

The largest issue class in this tracker is "numbers are wrong or missing", and a big fraction of those are environment problems, not parser bugs: unsupported Node version making SQLite providers silently empty (#250), provider paths not found (five separate OpenCode-shows-nothing reports: #364 #360 #346 #341 #392), unpriced models at $0 (#629). Today none of this is self-diagnosable; users file a bug because the product fails silently.

## Proposed fix

A `codeburn doctor` command that prints, per provider:

- data path(s) probed and whether they exist
- whether parsing succeeded, and how many sessions/calls were found
- why a provider is empty when it is (path missing, unreadable, Node too old for zstd/sqlite, cache stale)
- unpriced models encountered, with the model-alias / price-override one-liner to fix them
- environment summary: Node version, platform, relevant env vars (CLAUDE_CONFIG_DIR etc.)

Exit non-zero when something is broken so it can be pasted into bug reports. Most of the probing logic already exists inside provider detection; doctor mainly needs to surface it instead of swallowing it.

## Why it matters

Every hour spent triaging a "shows zero" issue is an hour not spent on real parser bugs. Doctor converts that class into a self-serve check, and doubles as the first debugging step we ask reporters to run.

iamtoruk:
## Problem

When a provider shows zero, users have no way to see why codeburn found nothing, so it becomes a support issue (#675 and friends: wrong env override, tool not installed, empty custom dir, stale npm build). Closes #642.

## What it does

`codeburn doctor` (and `--provider <name>`, `--json`) prints one row per provider: sessions discovered, a bounded parse sample (up to 8 sources, 500 calls each, so huge histories stay fast), cached-file counts, and an honest verdict: OK (n sessions), NOTHING FOUND with the likely cause, or ERRORS. A Details section lists the exact probed paths with existence, any env override that is set next to the path it resolved to, and the provider parse version. Example diagnosis with a wrong override:

    Claude | 0 | - | 0 | NOTHING FOUND (override CLAUDE_CONFIG_DIR set; /nonexistent/projects does not exist)

## Design

- collect/render split: collectDoctorReport is pure and injectable (providers, cache, sample limit); the table and JSON renderers only format.
- No duplicated path logic: providers expose an optional probeRoots() (claude, codex, opencode implement it today) that reuses their existing resolution helpers; everything else reports from discovered sources plus env-override state.
- Every provider is wrapped in its own try/catch, so one thrower becomes an ERROR row instead of killing the report. Fully offline: the one network provider is discovered but never called. Strictly read-only: doctor never writes caches or config (verified by comparing cache mtimes across a run).

## Verification

- 12 unit tests over the collect logic with fixture dirs (found, empty, missing, override, error, render). Mutation-checked twice: breaking the exists-check fails 3 tests; blanking the override reporting fails exactly the override test.
- Real machine run shows real numbers (Claude 16, Codex 408 sessions, correct verdicts) and doctor --json parses.
- tsc clean; full suite green apart from the two pre-existing app/electron/cli.test.ts failures on unbuilt checkouts (#681).

## Notes

The Cached column shows 0 on a checkout whose CACHE_VERSION is ahead of the on-disk cache; that is faithful (this build would re-parse), not a bug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,6 +415,7 @@ Sync sends token counts, costs, models, and projects — never prompts or code. 
 
 | Command | What it does |
 |---------|--------------|
+| `codeburn doctor` | Per-provider detection status: paths probed, sessions found, parse health (`--json`, `--provider`) |
 | `codeburn audit` | Per provider-model token source table: where every number comes from |
 | `codeburn context` | What fills a session's context window: interactive browser (Claude Code and Codex) |
 | `codeburn context <id> --json` | The same context tree, scriptable |
@@ -559,6 +560,18 @@ codeburn report --to 2026-04-10                      # earliest data through thi
 ```
 
 Either flag alone is valid. Inverted or malformed dates exit with a clear error. In the TUI, the custom range sets the initial load only; pressing `1` through `5` switches back to predefined periods.
+
+### Diagnosing detection
+
+When a tool shows zero (or a number that looks wrong), `codeburn doctor` explains why. It runs fully offline and read-only, and never writes to caches or config.
+
+```bash
+codeburn doctor                     # every provider, human-readable table
+codeburn doctor --provider opencode # diagnose one provider
+codeburn doctor --json              # machine-readable, pipe to jq
+```
+
+For each provider it shows the exact directories or databases probed (with any env override such as `CLAUDE_CONFIG_DIR`, `CODEX_HOME`, or `OPENCODE_DATA_DIR` and whether the path exists), how many session files were found, how many of a bounded sample parsed cleanly, the cached file count, and a one-line verdict: `OK (n sessions)`, `NOTHING FOUND` with the likely cause (directory missing, override points at an empty dir, or the tool is not installed), or `ERRORS (n parse failures)`. A provider that throws is caught and reported as its own error row, never crashing the rest of the report.
 
 ### JSON Output
 
--- a/src/cursor-cache.ts
+++ b/src/cursor-cache.ts
@@ -79,6 +79,10 @@ export async function writeCachedResults(
   calls: ParsedProviderCall[],
   lookbackFloor: string,
 ): Promise<void> {
+  // Diagnostic contexts (codeburn doctor) sample-parse providers under a
+  // strictly read-only promise; this is the one parse path that writes to
+  // disk before its first yield, so it honors the suppression flag.
+  if (process.env['CODEBURN_SUPPRESS_CACHE_WRITES']) return
   const fp = await getDbFingerprint(dbPath)
   if (!fp) return
 
--- a/src/doctor.ts
+++ b/src/doctor.ts
@@ -1,0 +1,346 @@
+import { existsSync } from 'fs'
+import { dirname } from 'path'
+
+import { Chalk } from 'chalk'
+
+import { getAllProviders } from './providers/index.js'
+import type { Provider } from './providers/types.js'
+import {
+  PROVIDER_ENV_VARS,
+  PROVIDER_PARSE_VERSIONS,
+  loadCache,
+  type SessionCache,
+} from './session-cache.js'
+import { renderTable } from './text-table.js'
+
+// ── Types ──────────────────────────────────────────────────────────────
+
+export type DoctorProbePath = {
+  path: string
+  label: string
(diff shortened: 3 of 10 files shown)


ozymandiashh (changes requested):
The collect/render split is clean, the verdict logic is honest and override-aware, and the per-provider isolation holds (traced: a throw in probeRoots, discovery, or parse lands in the row's catch). Two issues against the PR's own headline claims:

1. **Doctor writes to disk on Cursor.** `cursor.ts:1016` runs `await writeCachedResults(dbPath, allCalls, timeFloor)` before the parser's first yield, and that lands in `~/.cache/codeburn/cursor-results.json` (writeFile + atomic rename). Any doctor run that samples a Cursor source with a cold or stale bubble cache performs a real cache write, so "strictly read-only: doctor never writes caches" does not hold. The mtime verification in the body likely passed because the dev machine had a warm cursor cache. Suggested fix: an injectable no-write option on the parse (or skip cursor parse sampling and report discovery-only for it), plus a test that fails if a doctor run mutates anything under the cache dir.

2. **"Bounded sampling" bounds yielded calls, not I/O.** The 8x500 cap breaks out of the async iterator, but several parsers (codex among them) read the entire file into memory and split all lines before the first yield, so a single multi-GB session file is fully loaded regardless of the cap. Worth either wording the claim precisely (bounds parse work, not file reads) or capping reads for the sampled path.

Notes, non-blocking: `probeRoots()` duplicates the resolution logic that discovery already owns for claude/codex/opencode, with nothing enforcing they stay in sync; a future discovery path added without updating probeRoots makes doctor confidently point at the wrong directories. And when both CLAUDE_CONFIG_DIRS and CLAUDE_CONFIG_DIR are set the one-line verdict names both without saying which one actually controls resolution (Details disambiguates, the verdict line does not).

ozymandiashh:
Follow-ups after a deeper verified pass; the first one changes the picture:

1. **The Parsed column is vacuous for Claude.** `claude.ts:325-329` returns an empty generator from `createSessionParser` (Claude parses through a different pipeline), so doctor's sample loop "parses" every Claude source instantly without reading a byte: a projects dir containing only garbage JSONL reports `Parsed 3/3, OK`. The flagship provider gets a rubber-stamped health check; only `cachedFailed` can ever reflect real Claude parse trouble. Worth either routing the Claude sample through its real parse path or rendering `-` for providers whose session parser is a stub.

2. **Cached shows 0 in exactly the state doctor exists to explain.** `loadCache` returns an empty cache on any `CACHE_VERSION` mismatch, so right after a codeburn upgrade every provider reports `Cached: 0` while a populated session-cache.json sits on disk. A diagnostic tool should distinguish "no cache" from "stale cache, will re-parse".

3. **"Fully offline" does not hold for Antigravity.** It is not flagged `network` (only vercel-gateway is), so doctor sample-parses it, and that parser probes a live local server (`detectServer` HTTP with 1s timeout, `rpc()` with 5s), stalling up to ~6s per sampled source against a hung server.

4. **No timeout anywhere, providers sequential**: a discovery that never returns (dead network mount in a readdir) hangs the whole report. "One thrower cannot kill the report" holds; one hanger still can.

5. Env-override faithfulness details: `CODEBURN_DESKTOP_SESSIONS_DIR` changes a probed root but is absent from `PROVIDER_ENV_VARS`, so it never shows as an override; both `CLAUDE_CONFIG_DIRS` and `CLAUDE_CONFIG_DIR` are named as in-effect when only the first controls resolution; and antigravity's listed `CODEBURN_CACHE_DIR` is an internal cache var displayed as if it affected detection. Paths sourced from codeburn's own config-file fallback (`claudeConfigDirs`) are shown with no attribution, so a stale config-file entry misdiagnoses as "tool likely not installed".

For clarity on my earlier read-only finding: codex's parse-time cache write is memory-only (flushed elsewhere), so codex is safe; cursor's `writeCachedResults` (cursor-cache.ts:77-109, mkdir + writeFile + rename inside the function) is the one real disk write doctor can trigger.

PR #688, opened 2026-07-16
fix(report): surface estimated costs distinctly

### LINKED ISSUE #639: costIsEstimated is dropped at the parser boundary; estimated costs render exactly like measured ones
## Problem

Providers that estimate tokens/cost set `costIsEstimated: true` (kiro, cursor, warp, copilot, grok, hermes, codex proxy path), but the field exists only on `ParsedProviderCall` (src/providers/types.ts:24). It is not carried onto `ParsedApiCall` or `SessionSummary` (src/types.ts), and src/parser.ts never reads it. Result: a Warp figure whose output tokens are synthesized from prompt length (src/providers/warp.ts:199-206) renders with the same authority as a Claude measured figure in the dashboard, overview, exports, and MCP tables. The only estimate disclaimer in the product is the optimize findings caption (src/dashboard.tsx:681), which does not cover the cost surfaces.

## Evidence

- `grep -rn costIsEstimated src/parser.ts src/types.ts` returns nothing; the flag dies at the provider layer.
- Setters: src/providers/kiro.ts:464, cursor.ts:768, warp.ts:407, copilot.ts:1467, grok.ts:199, codex.ts:477, hermes.ts:410.

## Proposed fix

Propagate an estimated-cost portion through `ParsedApiCall` -> `SessionSummary` -> aggregates (e.g. `estimatedCostUSD` alongside `costUSD`), then mark affected rows and totals in all four output layers (TUI, overview text, CSV/JSON export, MCP tables) with an est. tag or tilde. Related to the accuracy reports behind #574: users cannot currently tell which numbers deserve how much trust.

iamtoruk:
## Problem

Eight providers set costIsEstimated on their parsed calls (cursor, kiro, warp, copilot, grok, hermes, codewhale, and the codex proxy path), but the flag died at the parser boundary: ParsedApiCall, the session cache, and every aggregate had no field for it, so estimated money rendered exactly like measured money. Closes #639.

## Fix

Plumb the truth through, mirroring the savingsUSD/isLocalSavings pattern: call-level isEstimated, and an additive estimatedCostUSD amount on model breakdowns, sessions, projects, and periods. The amount, not a bare boolean, because mixed rows are the norm: on real data GPT-5.5 is 161 estimated out of 1359 total, which a boolean would render the same as a fully guessed row.

Surfaces: report and overview mark estimated figures with a tilde and one legend line; report --json, the menubar/web payload, and MCP get_usage carry the fields. Totals math is untouched, verified byte-identical against main on a 30-day real-data report (only the marker and legend differ).

Cache: CachedCall persists the flag; parse versions bumped for the providers that estimate. Copilot deliberately not bumped: it is a durable provider, so discarding its cache loses OTel rows whose sources may be pruned; its flag fills in as new data arrives.

Also fixes a real bug found during verification: the cross-provider project merge (same repo used with two tools) summed cost and calls but dropped merged-in estimated dollars, under-reporting exactly for multi-provider users. The merge is now an exported pure function with regression tests that fail if the summing line is removed. The same merge has a latent identical gap for totalSavingsUSD, left for a follow-up issue rather than widening this diff.

## Verification

- New tests: end-to-end estimated and measured propagation through parseAllSessions and the cache round trip, marker rendering, and the merge regression pair. Mutation-checked: dropping the cache plumb fails the e2e tests; deleting the merge summing line fails exactly the two merge tests.
- Real data: estimated providers render with the tilde, measured rows do not; totals identical to main.
- tsc clean; full suite green apart from the two pre-existing app/electron/cli.test.ts failures (#681).


ozymandiashh (changes requested):
The plumbing is disciplined: additive amounts, `?? 0` back-compat on every payload boundary, totals math untouched (verified), the cross-provider merge fix is real and regression-pinned, and the copilot cache exemption is the right call since its durable OTel path is measured. Two things before merge:

1. **cursor-agent estimates without the flag.** `cursor-agent.ts:452-454` prices tokens from character lengths (`estimateTokens`) exactly like its sibling, but never sets `costIsEstimated`, and its parse version is unbumped, while `cursor.ts:768` flags the identical method. A whole estimating provider renders as measured, which is the precise failure mode this PR exists to eliminate. Either add it to the covered set (flag + bump) or name the exclusion in the body.

2. **The field stops at model rows.** `estimatedCostUSD` exists on sessions and projects in the parser, but `buildPeriodData` drops it for category totals and the per-project payload rows, `projectsJson` in main.ts (lines ~330-336) omits it, and the overview's project/day/provider aggregates never track it, so the tilde can only ever appear on model rows. The body says the amount lands on "model breakdowns, sessions, projects, and periods" and that `report --json` carries the fields; today that is true for models only. Either surface it on the project/task rows too or narrow the body so the description matches the diff.

Nit: the body calls the savings merge gap latent, but the un-merged `totalSavingsUSD` reaches output at four sites today (main.ts:263, :329; usage-aggregator.ts:47, :386). Deferring it is fine and disclosed; "live, deferred" is the accurate label for the follow-up issue.

ozymandiashh:
Two additions after a deeper pass, one of them the biggest missing surface:

1. **`codeburn models` never sees the flag.** `aggregateModels()` recomputes per-model cost straight from calls and reads no `isEstimated` anywhere in `src/models-report.ts` (zero occurrences), so the one command dedicated entirely to per-model cost shows estimated-priced dollars as metered in all four formats (table, markdown, json, csv). Since the body's core claim is estimated amounts on model breakdowns, this surface belongs in the covered set.

2. **The new merge comment asserts a rule the code below it still violates.** The added comment says additive totals must ALL be summed in the cross-provider merge, and the block sums `totalCostUSD` and `totalEstimatedCostUSD` but still not `totalSavingsUSD`. Deferring the savings fix is fine and disclosed; the comment as worded documents the deferred bug as impossible. One line of comment wording, or sum it while there.

Smaller notes, take or leave: the day-aggregator's `buildPeriodDataFromDays` produces PeriodData without the field (harmless today only because its result is overwritten by `buildPeriodData` two lines later, a trap for whoever removes that overwrite), and the web dash React bundle plus the Swift menubar receive `estimatedCostUSD` in the payload but do not render the marker yet.

PR #689, opened 2026-07-16
test(electron): make dev-resolution tests self-sufficient on fresh clones

### LINKED ISSUE #681: app/electron/cli.test.ts fails on a fresh clone: dev-resolution tests require a built dist/cli.js

ozymandiashh:
Closes #681.

## Problem

On a fresh clone (no `npm run build`), the two Vite-development `resolveCodeburnPath` tests failed with `.toMatch() expects to receive a string, but got object`: the dev branch only returns the repo CLI path when `dist/cli.js` exists and is executable.

## Fix

Per the issue's preferred direction, the tests now create the fixture they depend on:

- `fakeRepoDevCli()` writes a stub executable `dist/cli.js` at the exact path the production `sourceDevBin` fallback resolves, and **snapshot-restores** any pre-existing real build artifact (content and mode restored byte-identically in `afterEach`; the fixture deletes only what it created, and `dist/` is kept when it was non-empty before).
- A fail-fast precondition asserts the parent-level `dist/cli.js` candidate (which resolution prefers under vitest's module layout) does not exist, naming the offending absolute path instead of failing later with a confusing `toBe` mismatch.
- No skip-when-absent: the tests run and pass with no build present, and still pin the preference order (repo dev CLI over a persisted stale global).

Production code untouched.

## Verification

- `npx vitest run app/electron/cli.test.ts` on a checkout with NO `dist/cli.js`: 19/19.
- Same suite with a pre-existing sentinel `dist/cli.js` (distinct bytes, mode 0711): 19/19, then the sentinel is restored byte-identically with its exact mode (`cmp` clean).
- `npx tsc --noEmit` clean.

### DIFF
--- a/app/electron/cli.test.ts
+++ b/app/electron/cli.test.ts
@@ -12,6 +12,7 @@ const originalPathDirs = process.env.CODEBURN_PATH_DIRS
 const originalPathFile = process.env.CODEBURN_CLI_PATH_FILE
 const originalViteUrl = process.env.VITE_DEV_SERVER_URL
 const originalBundled = process.env.CODEBURN_BUNDLED_CLI
+const originalDevRepoRoot = process.env.CODEBURN_DEV_REPO_ROOT
 
 /** Writes an executable node script and points CODEBURN_BIN at it. */
 function fakeBin(name: string, body: string): string {
@@ -19,6 +20,17 @@ function fakeBin(name: string, body: string): string {
   writeFileSync(p, `#!/usr/bin/env node\n${body}\n`, { mode: 0o755 })
   chmodSync(p, 0o755)
   process.env.CODEBURN_BIN = p
+  return p
+}
+
+/** Writes the repo CLI under this test's isolated dev-root override. */
+function fakeDevRepoCli(): string {
+  const repoRoot = join(dir, 'dev-repo')
+  const p = join(repoRoot, 'dist', 'cli.js')
+  mkdirSync(dirname(p), { recursive: true })
+  writeFileSync(p, '#!/usr/bin/env node\n', { mode: 0o755 })
+  chmodSync(p, 0o755)
+  process.env.CODEBURN_DEV_REPO_ROOT = repoRoot
   return p
 }
 
@@ -37,6 +49,8 @@ afterEach(() => {
   else process.env.VITE_DEV_SERVER_URL = originalViteUrl
   if (originalBundled === undefined) delete process.env.CODEBURN_BUNDLED_CLI
   else process.env.CODEBURN_BUNDLED_CLI = originalBundled
+  if (originalDevRepoRoot === undefined) delete process.env.CODEBURN_DEV_REPO_ROOT
+  else process.env.CODEBURN_DEV_REPO_ROOT = originalDevRepoRoot
   rmSync(dir, { recursive: true, force: true })
 })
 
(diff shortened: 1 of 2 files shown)


iamtoruk (changes requested):
The core fixture is sound and the assertions got stronger (exact-path toBe instead of toMatch, real statSync and X_OK checks still exercised), but three things before this can merge.

1. It is stale against main: there is a content conflict in app/electron/cli.test.ts, and main gained a third fresh-clone failure after this branched (the resolveTarget test where the dev repo CLI beats the bundled CLI). As-is, Closes #681 would close the issue with one of its three symptoms still failing. Rebase and cover that test too.

2. The stub writes the real repo dist/cli.js, which is the published bin and exactly what npm run dev spawns. The snapshot-restore in afterEach is careful, but a SIGKILL or worker crash between the write and the restore leaves a developer's real build replaced by a 20-byte no-op, silently. Since there is no CI test job, this suite runs precisely on developer machines where that hurts. The cleaner shape already has precedent in this file: CODEBURN_PATH_DIRS exists for tests and advanced setups. A matching dev-root override env var lets the whole fixture live in the per-test tmpdir and deletes the snapshot machinery, the precondition check, and the crash window.

3. Please drop the beforeAll repo-scanning guard that globs test files and regex-scans imports. It misses import-star and indirect uses, will spuriously fail the suite when someone adds a legitimate isolated resolver test, and likely self-detects on Windows path separators. It is the needless-defensiveness pattern we strip in review.

Happy to re-review quickly, the direction is right.

ozymandiashh:
Reworked per review, force-pushed as a fresh implementation on current main:

- **Env-var shape adopted**: `CODEBURN_DEV_REPO_ROOT` on the Vite-dev resolution branch (mirroring the `CODEBURN_PATH_DIRS` precedent). The whole fixture now lives in a per-test tmpdir; the real repo `dist/cli.js` is never written, so the crash-window concern is gone along with the snapshot-restore machinery and the precondition check. With the env unset, resolution is byte-identical to before (both candidates, same order), covered both ways.
- **All three fresh-clone failures covered** on current main, including the newer resolveTarget dev-repo-beats-bundled case; verified on a checkout with no `dist/cli.js` at any point (27/27).
- **beforeAll scanning guard dropped.**

iamtoruk:
Verified the current head end to end on a fresh clone and with a built dist. All three earlier review asks are addressed and the production change in app/electron/cli.ts is safe (gated on VITE_DEV_SERVER_URL, byte-identical resolution when the override is unset). One remaining issue: the new test at app/electron/cli.test.ts:88-96 asserts toBeNull() with the override unset, but the else branch resolves <repo>/dist/cli.js when it exists, so the suite fails 26/27 on any checkout that has run a build (fresh clone passes 27/27). The assertion is coupled to the absence of a gitignored build artifact and cannot be made environment-independent as written. Suggest dropping that test: the override-set cases at :58/:68/:143 already prove the new branch, and :98-105 covers the null path outside Vite dev. With that one change this is ready.

ozymandiashh:
Done in d5efe3b: dropped the environment-coupled test. The suite passes 26/26 both on a fresh clone and with a built dist/cli.js present (verified both locally). Branch coverage stays intact via the override-set cases at :58/:68 and the non-Vite null path at :98-105.

PR #690, opened 2026-07-16
Propagate costIsEstimated end-to-end; mark estimated costs in all output surfaces

### LINKED ISSUE #639: costIsEstimated is dropped at the parser boundary; estimated costs render exactly like measured ones

avs-io:
## Problem

Providers that estimate tokens/cost set `costIsEstimated: true` (kiro, cursor, warp, copilot, grok, hermes, codex proxy path), but the flag exists only on `ParsedProviderCall` and dies at the parser boundary. A Warp figure synthesized from prompt length renders with the same authority as a Claude measured figure in the dashboard, overview, exports, and MCP tables. (#639)

## Fix

- `estimatedCostUSD` propagated additively (always a subset of `costUSD`, never added into it) through parser → turn/category/skill/model breakdowns → `SessionSummary` → `ProjectSummary` → daily/period aggregates. Partial sessions (some calls measured, some estimated) are representable.
- Rendering: one convention on all text surfaces — `~$X` when the estimated portion > 0 (TUI dashboard, overview, MCP text tables). Machine formats get explicit numeric fields, not decorated strings: CSV `Estimated Cost` columns, JSON export field, MCP `structuredContent` totals + rows + zod schema, menubar JSON optional fields + `estimatedProviders` (Swift app decodes via explicit CodingKeys, so the new sibling keys are non-breaking).
- Cache correctness: session cache v5→6 and daily cache v11→12 with rationale comments — pre-existing cache entries from estimating providers carry no marker and must re-hydrate rather than silently reload as measured.

## Tests

- End-to-end invariant: a session built from an estimated provider call can never surface as fully measured — `tests/parser.test.ts` ("marks a session built from an estimated provider call downstream") plus per-surface assertions in overview/export/mcp-tables/menubar-json/daily-cache/day-aggregator tests.
- Red-before verified: with only this PR's test changes applied to main's source, exactly those 10 new assertions fail; green with the change.
- Session-cache round-trip test: estimated portion survives persist → clear → reload.
- `npx tsc --noEmit` clean; full suite 1717 passed / 0 failed.

Non-goals (follow-ups): the Swift menubar UI does not yet display the new fields; `web-dashboard` untouched.

Closes #639

### DIFF
--- a/src/daily-cache.ts
+++ b/src/daily-cache.ts
@@ -5,7 +5,14 @@ import { homedir } from 'os'
 import { join } from 'path'
 import type { DateRange, ProjectSummary } from './types.js'
 
-// Bumped to 11: kiro cost accounting changed (metered credits pass through
+// Bumped to 12: estimated-cost propagation (#639) added estimatedCostUSD to
+// day/model/category rollups. Days finalized at v11 were built before the
+// field existed, so estimated provider spend (cursor, warp, grok, copilot,
+// kiro fallback paths) would silently read as fully measured. Raising
+// MIN_SUPPORTED_VERSION forces the one-time re-hydration that backfills the
+// estimated portion.
+//
+// v11: kiro cost accounting changed (metered credits pass through
 // the session cache instead of being re-priced from estimated tokens), so
 // days finalized at v10 carry token-estimated kiro costs that were off by up
 // to 16× per model. Raising MIN_SUPPORTED_VERSION forces the one-time full
@@ -20,13 +27,14 @@ import type { DateRange, ProjectSummary } from './types.js'
 // that older binaries skipped. v8 added local-model savings to the daily
 // rollup; the `savingsConfigHash` field is invalidated separately when the
 // user changes their `localModelSavings` mapping.
-export const DAILY_CACHE_VERSION = 11
-const MIN_SUPPORTED_VERSION = 11
+export const DAILY_CACHE_VERSION = 12
+const MIN_SUPPORTED_VERSION = 12
 const DAILY_CACHE_FILENAME = 'daily-cache.json'
 
 export type DailyEntry = {
   date: string
   cost: number
+  estimatedCostUSD?: number
   savingsUSD: number
   calls: number
   sessions: number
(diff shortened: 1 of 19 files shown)


iamtoruk:
Thanks for the thorough pass on #639. This crossed in flight with #688, which merged minutes after this was filed and implements the same call-level flag, aggregate propagation, and tilde convention, so the two now conflict on essentially every file. Two things to flag from review before any rebase attempt: the global session-cache v5 to v6 and daily-cache MIN_SUPPORTED_VERSION bumps would discard copilot's durable section wholesale (loadCache returns emptyCache on version mismatch, and the orphan carry-forward can only rescue entries from a loaded cache), and the cross-provider project merge does not sum estimatedCostUSD, which is exactly the multi-provider under-report that is now test-pinned on main. That said, this covers real surfaces #688 missed: the CSV/JSON export columns, daily-cache and day-aggregator propagation, and the broader overview/dashboard/MCP row marking. We will port those onto the merged schema in a credited follow-up. Closing as superseded by #688.

avs-io:
Both flags confirmed on re-read: the version bumps would indeed drop copilot's durable section on load (emptyCache before the orphan pass can see it), and the project-merge gap is real — good that main now pins it. Happy to take the export/daily-cache/day-aggregator port onto the #688 schema if that saves you the follow-up; otherwise consider it yours.

PR #693, opened 2026-07-16
copilot: count session.shutdown token rollup for CLI sessions

### LINKED ISSUE #676: Copilot CLI provider under-reports cost — ignores  session.shutdown  token/cost data in  ~/.copilot/session-state/*/events.jsonl

avs-io:
## Problem

The Copilot JSONL parser handles only `session.start` / `session.model_change` / `subagent.selected` / `user.message` / `assistant.message`; `session.shutdown` falls through. Per-turn events carry only output tokens, and for standalone CLI sessions no other source (OTel/chatSessions are IDE-side) supplies input/cache/reasoning tokens — so cost is computed from output alone, under-reporting exactly as @fgapito reported. (#676)

## Fix

New `session.shutdown` branch (JSONL path only, skipped for transcripts): for each entry in the rollup's `modelMetrics`, yield one supplementary call per model with `outputTokens: 0` — per-turn events already carry output, so the rollup's output is never re-counted. The supplementary call carries what the per-turn events lack:

- non-cached input = `usage.inputTokens - cacheReadTokens - cacheWriteTokens` (clamped ≥ 0) — `usage.inputTokens` is cache-inclusive in the schema, so subtracting avoids triple-counting cache tokens in cost
- cache read/write and reasoning tokens (reasoning priced at the output rate via `calculateCost`; `requests.cost`/`totalNanoAiu` are Copilot-internal units and deliberately not interpreted as USD)

Dedup key `copilot:shutdown:${sessionId}:${model}` — stable and idempotent across re-parses, no collision with per-turn keys. Copilot parse version bumped (`otel-durable-v1` → `shutdown-v1`) so cached sessions re-parse once.

## Tests

- New `tests/providers/copilot-shutdown.test.ts` (written red-first: `expected +0 to be 35495` on cacheRead before the fix). It also pins `totals.output === 345`, not 690, guarding against double-counting, and covers the crashed-session (no shutdown event) path.
- Existing `tests/providers/copilot.test.ts` untouched and green.
- `npx tsc --noEmit` clean; full suite 1712 passed / 0 failed.

Fixes #676

### DIFF
--- a/src/providers/copilot.ts
+++ b/src/providers/copilot.ts
@@ -189,6 +189,18 @@ type AssistantMessageData = {
   toolRequests?: ToolRequest[]
 }
 
+type SessionShutdownData = {
+  modelMetrics: Record<string, {
+    usage: {
+      inputTokens: number
+      outputTokens: number
+      cacheReadTokens: number
+      cacheWriteTokens: number
+      reasoningTokens: number
+    }
+  }>
+}
+
 type SubagentSelectedData = {
   agentName: string
   agentDisplayName?: string
@@ -200,6 +212,7 @@ type CopilotEvent =
   | { type: 'session.model_change'; data: ModelChangeData; timestamp?: string }
   | { type: 'user.message'; data: UserMessageData; timestamp?: string }
   | { type: 'assistant.message'; data: AssistantMessageData; timestamp?: string }
+  | { type: 'session.shutdown'; data: SessionShutdownData; timestamp?: string }
   | { type: 'subagent.selected'; data: SubagentSelectedData; timestamp?: string }
 
 type ChatJournalPathSegment = string | number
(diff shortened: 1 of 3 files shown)


iamtoruk:
Thanks for the careful work here, especially the red-first test and the correct call to treat requests.cost as a Copilot-internal premium-request count rather than USD; that matches exactly what we found digging into the same rollup. This crossed in flight with #684, which merged shortly before this was filed with the same supplementary-call design, plus timestamp fallbacks for unstamped shutdown events and a durable-cache fix so the parse-version bump cannot erase pruned OTel history. Closing as superseded by #684, which also resolves #676. Appreciate you independently validating the approach.

PR #697, opened 2026-07-16
copilot: extract skill invocations so Skills & Agents populates

### LINKED ISSUE #654: Copilot JSONL parser misses skill usage — Skills & Agents section always empty for Copilot CLI sessions
I am a github copilot user and when running the report in the terminal with the command `codeburn report` the Skills and Agents section of the report is empty and I know there should be skills present.  I unleashed copilot on the issue to determine the root cause of the problem and this is what it came up with.  If this is not enough information or does not follow the format you are looking for I am happy to adjust the formatting to your expectations.  I remove the names of the skills as my employer may not want those in a github issue.

### Summary
`codeburn report`, `codeburn overview`, and `codeburn export` all show an empty **Skills & Agents** section for GitHub Copilot CLI users (sessions in `~/.copilot/session-state/`), even when skills are actively invoked during sessions. This affects v0.9.15 (latest).

### Repro

```bash
# Check a date where Copilot skills were used
codeburn overview --provider copilot --from 2026-07-09 --to 2026-07-09 --no-color
```

**Observed:** No "Skills" section in the overview output. The "By activity" table shows turns categorised as `Conversation` rather than `General`, and `codeburn export --format json` shows `skills: []` throughout.

```bash
codeburn export --format json --from 2026-07-09 --to 2026-07-09 --provider copilot --output /tmp/cb-out.json
```

```json
{
  "sessions": [
    { "Project": "...", "Turns": 148, "skillBreakdown": {} }
  ],
  "skills": []
}
```

### Evidence of skills in raw session logs

`~/.copilot/session-state/<session-id>/events.jsonl` contains `assistant.message` events with `toolRequests` entries where `name === "skill"` and the actual skill name is carried in `arguments.skill`:

```json
{
  "type": "assistant.message",
  "data": {
    "messageId": "6681c347-6454-4945-8492-cba70a6b14cc",
    "model": "claude-sonnet-4.6",
    "outputTokens": 111,
    "toolRequests": [
      {
        "toolCallId": "toolu_bdrk_01FrUw1ya3xi7MNK5aRK8LVc",
        "name": "skill",
        "arguments": { "skill": "{skill-name}" },
        "type": "function"
      }
    ]
  },
  "timestamp": "2026-07-09T16:18:12.463Z"
}
```

Multiple distinct skill names appear across sessions.

### Parser gap in `src/providers/copilot.ts`

**Gap 1 — `toolNameMap` is missing `skill`** (line 96–130)
The map normalises raw tool names to display names. `"skill"` has no entry, so `normalizeTool("skill")` returns the raw lowercase string `"skill"`. The classifier's `hasSkillTool()` checks `tools.some(t => t === "Skill")` (capital S), so i
t never matches. Turns that only call a skill fall through to `classifyConversation()` instead of being classified as `"general"`, and `subCategory` — which feeds `skillBreakdown` — is never set.

```typescript
// src/providers/copilot.ts  line 96
const toolNameMap: Record<string, string> = {
  // JSONL session-state tool names
  bash: 'Bash',
  read_file: 'Read',
  // ... no entry for 'skill' ← BUG
}
```

**Gap 2 — `skills[]` field is never populated in the JSONL parser yield** (~line 358)

`bashCommands` is extracted by inspecting `toolRequests` where `name` is in `BASH_TOOL_NAMES` and reading `arguments.command`. The same pattern is not applied for skill invocations: the actual skill name in `arguments.skill` is never read
and the `skills` field (defined as optional on `ParsedProviderCall`) is never set. Without it, `getAllSkills(turn)` always returns `[]` and `skillBreakdown` is never populated regardless of how many skills were invoked.

```typescript
// src/providers/copilot.ts  ~line 340 (createJsonlParser, assistant.message branch)
const bashCommands = toolRequests.flatMap((t) => {
  if (typeof t !== 'object' || t === null) return []
  const name = (t.name ?? t.toolName) ?? ''
  if (!BASH_TOOL_NAMES.has(name)) return []
  const cmd = t.arguments?.['command']
  return typeof cmd === 'string' ? extractBashCommands(cmd) : []
})

yield {
  // ...
  tools,
  bashCommands,
  // skills is never set here ← BUG
  subagentTypes: currentSubagentType ? [currentSubagentType] : undefined,
}

### Minimal patch suggestion

```diff
--- a/src/providers/copilot.ts
+++ b/src/providers/copilot.ts
@@ toolNameMap (line 97) @@
 const toolNameMap: Record<string, string> = {
   // JSONL session-state tool names
+  skill: 'Skill',
   bash: 'Bash',

@@ createJsonlParser, assistant.message yield (~line 350) @@
+          const skills = toolRequests.flatMap((t) => {
+            if (typeof t !== 'object' || t === null) return []
+            const name = (t.name ?? t.toolName) ?? ''
+            if (name !== 'skill') return []
+            const skillName = t.arguments?.['skill']
+            return typeof skillName === 'string' && skillName ? [skillName] : []
+          })
+
           yield {
             ...
             tools,
             bashCommands,
+            skills: skills.length > 0 ? skills : undefined,
             subagentTypes: ...
           }
```

Also bump the `copilot` entry in `PROVIDER_PARSE_VERSIONS` so previously-cached sessions are automatically re-parsed on next run.

### Expected
`codeburn overview` shows a populated **Skills** section listing skill names with turn counts and costs. `codeburn export --format json` includes non-empty `skills` entries when Copilot CLI sessions contain skill invocations.

### Actual
Skills & Agents section is always empty. Skill invocations are silently dropped during parse.

### Environment
- codeburn: `0.9.15` (latest)
- Node: `v22.18.0`
- Platform: macOS (Darwin)
- Copilot sessions: `~/.copilot/session-state/` (JSONL format, `sourceType: "jsonl"`)
- Skills are invoked via `toolRequests[].name === "skill"` with `arguments.skill = "<skill-name>"`

avs-io:
## Problem

The Copilot parser has no notion of skills: the tool-name map lacks a `skill` entry (the shared classifier matches exact `Skill`), and the JSONL `assistant.message` branch never reads `arguments.skill` from toolRequests into `skills[]`. Everything downstream (`getAllSkills` → `subCategory` → `skillBreakdown` → the Skills & Agents section) is permanently empty for Copilot CLI sessions, as @rfsmart-mmann reported. (#654)

## Fix

- `skill: 'Skill'` added to the copilot tool-name map.
- Skill names extracted from `toolRequests` where the tool is `skill`, reading `arguments.skill` (handles `name`/`toolName`, trims and drops empties — mirroring the claude parser's extraction); yielded as `skills` when non-empty.
- Subagent handling untouched (already worked via `subagent.selected`).
- Copilot parse version bumped (`cli-shutdown-cost-v1` → `cli-shutdown-cost-v1-skills`) so cached sessions re-parse once. Rebased on current main over #684's bump.

## Tests

- New `tests/providers/copilot-skills.test.ts` (written red-first: `expected [ 'skill' ] to include 'Skill'` and `expected undefined to deeply equal [ 'deploy-checklist' ]` before the fix).
- `npx tsc --noEmit` clean; full suite green except the three pre-existing `app/electron/cli.test.ts` failures already noted as environmental in the #677 review (expect a built `dist/cli.js`).

Fixes #654

### DIFF
--- a/src/providers/copilot.ts
+++ b/src/providers/copilot.ts
@@ -96,6 +96,7 @@ const modelDisplayNames: Record<string, string> = {
 const toolNameMap: Record<string, string> = {
   // JSONL session-state tool names
   bash: 'Bash',
+  skill: 'Skill',
   read_file: 'Read',
   write_file: 'Edit',
   edit_file: 'Edit',
@@ -873,6 +874,14 @@ function createJsonlParser(
             })
             .filter((t): t is string => t !== null)
 
+          const skills = toolRequests.flatMap((t) => {
+            if (typeof t !== 'object' || t === null) return []
+            const name = (t.name ?? t.toolName) ?? ''
+            if (name !== 'skill') return []
+            const skill = t.arguments?.['skill']
+            return typeof skill === 'string' && skill.trim().length > 0 ? [skill.trim()] : []
+          })
+
           // Extract base command names from bash-type tool requests, routing the
           // raw command through the shared extractBashCommands helper so chained
           // commands are normalised the same way as every other provider
@@ -904,6 +913,7 @@ function createJsonlParser(
             costUSD,
             tools,
             bashCommands,
+            skills: skills.length > 0 ? skills : undefined,
             subagentTypes: currentSubagentType ? [currentSubagentType] : undefined,
             timestamp: event.timestamp ?? '',
             speed: 'standard' as const,
--- a/src/session-cache.ts
+++ b/src/session-cache.ts
@@ -133,7 +133,7 @@ export const PROVIDER_PARSE_VERSIONS: Record<string, string> = {
   codex: 'mcp-attribution-v2-est-cost',
   cursor: 'composer-anchored-crediting-v1-est-cost',
   'cursor-agent': 'workspaceless-transcript-v1',
-  copilot: 'cli-shutdown-cost-v1',
+  copilot: 'cli-shutdown-cost-v1-skills',
   grok: 'estimated-cost-v1',
   hermes: 'reasoning-output-accounting-v1-est-cost',
   'lingtai-tui': 'token-ledger-registry-activity-v3',
--- a/tests/providers/copilot-skills.test.ts
+++ b/tests/providers/copilot-skills.test.ts
@@ -1,0 +1,100 @@
+// Regression test for issue #654: Copilot JSONL parser drops skill invocations,
+// leaving the Skills & Agents section empty for Copilot CLI sessions.
+//
+// Copilot CLI (~/.copilot/session-state/<id>/events.jsonl) records skill usage
+// as assistant.message toolRequests with name === 'skill' and the skill name in
+// arguments.skill. The parser must:
+//   1. normalise the tool name to 'Skill' (classifier's hasSkillTool matches
+//      exactly 'Skill', which makes the turn 'general'), and
+//   2. populate ParsedProviderCall.skills with the invoked skill name(s)
+//      (getAllSkills -> turn.subCategory -> skillBreakdown -> Skills report).
+import { describe, it, expect, beforeEach, afterEach } from 'vitest'
+import { mkdtemp, mkdir, writeFile, rm } from 'fs/promises'
+import { join } from 'path'
+import { tmpdir } from 'os'
+
+import { copilot } from '../../src/providers/copilot.js'
+import type { ParsedProviderCall } from '../../src/providers/types.js'
+
+let tmpDir: string
+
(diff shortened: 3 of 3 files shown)


iamtoruk (approved):
Verified locally since CI does not run tests: tsc clean, all copilot tests pass, full suite green apart from the pre-existing #681 electron failures, and dropping the skills yield fails exactly the new regression test. The parse-version bump is correct and safe post-#684 (durable orphans survive fingerprint changes now). Extraction mirrors the claude parser pattern and stays clear of the shutdown rollup handling. Shape of arguments.skill rests on the #654 report samples, which is the best available source.

PR #698, opened 2026-07-16
parser: dedupe the twin string/Buffer large-JSONL scanners

### LINKED ISSUE #643: parser: dedupe the twin string/Buffer large-JSONL scanners
## Problem

src/parser.ts contains two structurally identical hand-rolled JSON scanners: the string version (findJsonStringEnd and friends, src/parser.ts:152-478) and a Buffer twin of every function (findJsonStringEndBuffer and friends, src/parser.ts:483-822). Roughly 340 lines exist twice. Any scanner fix (escape handling, nesting off-by-one) must land in both; a fix applied to one branch produces divergent parsing between small-string and large-buffer inputs, which surfaces as a silent size-dependent miscount rather than an error.

## Evidence

- src/parser.ts:152 `function findJsonStringEnd(source: string, ...)`
- src/parser.ts:483 `function findJsonStringEndBuffer(source: Buffer, ...)`

## Proposed fix

Extract one scanner module generic over an indexable source (both string and Buffer support charCodeAt-style access; a tiny accessor closure keeps the hot path monomorphic). Delete the twin. Existing parser tests plus a fixture run through both entry paths guard the refactor.

avs-io:
## Problem

src/parser.ts carried two structurally identical hand-rolled JSON scanners — a string family and a Buffer twin of every function, ~340 lines duplicated. Any scanner fix had to land in both or parsing diverged between small-string and large-buffer inputs, surfacing as silent size-dependent miscounts. (#643)

## Fix

One scanner implementation generic over an indexable `JsonSource`. The hot loops stay monomorphic: low-level primitives (string-end/container-end/value-bounds scans) branch once per call on source type into type-specific loops; all higher-level logic (field lookup, capped string reads, usage/tool-block extraction) exists once. Net −67 lines with the parity test included; the eleven `*Buffer` twins of the higher-level family are gone.

Careful spots, called out for review:
- **Byte vs UTF-16 semantics**: capped reads count UTF-16 output units while Buffer offsets are bytes; the escape-scanner only yields ASCII-safe segment boundaries, so caps can't split a multi-byte character. The parity test covers raw multi-byte UTF-8 through the Buffer path, surrogate pairs at cap boundaries, escapes-before-cap + multibyte-after, and truncation mid-string / mid-escape / mid-UTF-8-sequence.
- **One deliberate unification**: large lines with leading whitespace used to parse on the Buffer path but return null on the string path. Both now parse (the permissive behavior). This is the only behavioral divergence between old and new, verified by differential execution of the old implementation against the new one across all the nasty-input cases.
- Rebased over the advisor-iterations and web-fetch usage additions; both are ported into the unified extraction and covered by their existing tests.

## Performance

Measured against the old implementation (same methodology as the differential check; ~300KB assistant lines, 50-iteration averages):

| Case | Old | New |
|---|---|---|
| string large line | 6.06 ms | 5.56 ms |
| buffer large line | 4.47 ms | 3.19 ms |
| string pathological cap | 4.70 ms | 4.66 ms |
| buffer pathological cap | 4.25 ms | 2.77 ms |

## Tests

- New `tests/parser-large-json-scanner-parity.test.ts` (13 cases): identical results asserted through both entry paths on escaped quotes/backslashes, unicode escapes, nested braces in strings, raw multi-byte UTF-8, emoji at cap boundaries, truncated lines, and whitespace variants.
- `npx tsc --noEmit` clean; full suite green except the three pre-existing `app/electron/cli.test.ts` environment failures noted in the #677 review.

Closes #643

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -255,12 +255,6 @@ function findJsonContainerEnd(source: JsonSource, start: number, open: number, c
     : findJsonContainerEndBuffer(source.raw, start, open, close, limit)
 }
 
-function findJsonValueBounds(source: JsonSource, start: number, limit = source.length): JsonValueBounds | null {
-  return typeof source.raw === 'string'
-    ? findJsonValueBoundsString(source.raw, start, limit)
-    : findJsonValueBoundsBuffer(source.raw, start, limit)
-}
-
 function findObjectFieldValue(source: JsonSource, objectStart: number, objectEnd: number, field: string): JsonValueBounds | null {
   return typeof source.raw === 'string'
     ? findObjectFieldValueString(source.raw, objectStart, objectEnd, field)
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -129,14 +129,14 @@ const LARGE_JSONL_LINE_BYTES = 32 * 1024
 
 export function parseJsonlLine(line: string | Buffer): JournalEntry | null {
   if (Buffer.isBuffer(line)) {
-    if (line.length > LARGE_JSONL_LINE_BYTES) return parseLargeJsonlBuffer(line)
+    if (line.length > LARGE_JSONL_LINE_BYTES) return parseLargeJsonl(line)
     try {
       return JSON.parse(line.toString('utf-8')) as JournalEntry
     } catch {
       return null
     }
   }
-  if (line.length > LARGE_JSONL_LINE_BYTES) return parseLargeJsonlLine(line)
+  if (line.length > LARGE_JSONL_LINE_BYTES) return parseLargeJsonl(line)
   try {
     return JSON.parse(line) as JournalEntry
   } catch {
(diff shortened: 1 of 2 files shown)


iamtoruk (approved):
Verified hands-on since CI does not run tests: tsc clean, full suite green apart from the pre-existing #681 failures, and a cold-cache month-scale report over real data (22k calls including many 32KB-plus lines and the advisor fixtures) is byte-identical between this branch and main. Mutation checks: corrupting the shared number reader is killed by the parity suite's absolute pins, and the advisor large-line tests pass on both scanner entry paths. One nit for a follow-up or a quick push: the generic findJsonValueBounds wrapper is dead code, defined with zero callers; the shared higher-level family calls findObjectFieldValue directly. Nice net reduction on the riskiest duplicated code in the repo.

avs-io:
Dead wrapper removed — `findJsonValueBounds` deleted, the String/Buffer variants stay as the direct callees. tsc clean, parser + parity suites green.

Issue #703, opened 2026-07-16
menubar: idle energy on AC still high after #648 — 30s cadence x ~10 CPU-s Node spawn per tick, no change-detection, userInitiated QoS (and #648 is unreleased)


avs-io:
## Problem

Follow-up to #647 / #648 with on-machine measurements from a real user machine running the menubar (0.9.15). Idle energy use is still substantial, and the merged #648 work — which is real and large — (a) has not shipped in any release (latest tag `mac-v0.9.15` predates the merge) and (b) leaves the dominant idle cost intact on AC power.

## Evidence (15-min idle window, popover closed, AC, M-series; 12 GB codex + 1.2 GB claude/opencode session corpus)

- The menubar spawned **15 `codeburn status --format menubar-json --provider all --period today --no-optimize` Node processes**, one every 59-63s. Each burned **7.6-16.7 CPU-s**, peaking at 112-360% CPU. Children alone: **142.8 CPU-s / 896s = 15.9% of a core, sustained indefinitely**.
- The Swift process itself is innocent: 0.32% avg. Activity Monitor attributes the children to the app.
- Direct measurement of one fetch: `today` = 9.9 CPU-s, 2.1 GB peak footprint, 93B instructions; `week` = 15.8 CPU-s. Cold start is not the cost: bare `node -e 0` = 0.03s, `codeburn --version` = 0.4s. `sample` of a live child shows V8 GC (ConcurrentMarking), UTF-8 decode, and JSON stringify — i.e. a full re-parse of today's sessions on every tick.
- Total: **~16.2% of a core sustained while idle**, nearly all of it in the per-tick subprocess re-parse.

## What #648 already covers (merged, unreleased)

Display-sleep gating, popover-closed payload trimming, battery/low-power backoff, scoped activity assertions, cadence picker. On battery this will land a big cut once released.

## Proposed fix (remainder, cheapest first)

1. **Ship a release containing #648** — users currently have none of it.
2. **Change-detection guard** (#648 ladder item 1, unimplemented): stat provider data roots before spawning; skip the tick when nothing changed since the last successful fetch. Most idle ticks are no-ops.
3. **AC backoff**: `RefreshCadence.interval` returns 30s whenever on AC with the popover closed (`RefreshCadence.swift:34`); back off to 2-5 min closed-popover regardless of power source (popover-open stays 30s).
[...]

PR #704, opened 2026-07-16
menubar: skip unchanged background refreshes, 120s AC idle cadence, utility QoS for tick spawns

avs-io:
## Problem

Items 2-4 of #703: even with #648's work, an idle menubar on AC still spawns a full Node re-parse every ~60s (measured ~10 CPU-s each, ~16% of a core sustained on a real user machine; details and raw numbers in #703). Items 1 (release) and 5 (CLI incremental parse) are intentionally not addressed here.

## Fix

1. **Change-detection guard** (`Data/UsageDataChangeGuard.swift`): before a background tick spawns the CLI, a shallow snapshot (root + first-level directory mtimes — deliberately not a recursive walk) of the provider data roots is compared against the snapshot taken at the last successful fetch; unchanged → skip the spawn. Interactive paths (popover open, Refresh Now) always bypass. The snapshot anchors only on fetch success, so failures can't mask changes. **Backstop:** unchanged-skips are honored for at most 30 minutes — the hand-maintained root list is a perf optimization, not a correctness contract; a provider missing from it degrades to a 30-minute refresh, never stale-forever.
2. **AC idle cadence** (`RefreshCadence.swift`): closed-popover Auto on AC moves from 30s to 120s (battery 150s / low-power 300s from #648 unchanged; popover-open stays 30s). Settings label updated.
3. **QoS** (`Security/CodeburnCLI.swift`, threaded through `DataClient`/`AppStore`): background tick spawns run `.utility` (E-cores); interactive fetches keep `.userInitiated`.

Combined with the guard, an idle machine goes from ~120 spawns/hour (post-#648 AC behavior) to at most 2/hour, and genuine refreshes land on efficiency cores.

## Tests

- `UsageDataChangeGuardTests` (swift-testing): fresh-skip, stale-no-skip, first-run, force-bypass, and the 30-minute backstop boundary.
- `RefreshCadenceTests` updated to pin the 120s closed-popover AC value.

## Verification

`swift build` clean (Swift 6.3.3 toolchain). Test targets need XCTest (Xcode/CI) to execute — same constraint as #702.

Part of #703 (items 2-4).

### DIFF
--- a/mac/Sources/CodeBurnMenubar/AppStore.swift
+++ b/mac/Sources/CodeBurnMenubar/AppStore.swift
@@ -523,19 +523,19 @@ final class AppStore {
         lastErrorByKey[key] = nil
         switchTask = Task {
             if scope == .combined {
-                async let local: Void = refresh(key: localKey, includeOptimize: false, force: false, showLoading: false)
-                async let combined: Void = refresh(key: key, includeOptimize: false, force: true, showLoading: true)
+                async let local = refresh(key: localKey, includeOptimize: false, force: false, showLoading: false)
+                async let combined = refresh(key: key, includeOptimize: false, force: true, showLoading: true)
                 if provider == .all {
                     _ = await (local, combined)
                 } else {
-                    async let all: Void = refreshQuietly(key: allKey, includeOptimize: false, force: false)
+                    async let all = refreshQuietly(key: allKey, includeOptimize: false, force: false)
                     _ = await (local, combined, all)
                 }
             } else if provider == .all {
                 await refresh(key: key, includeOptimize: false, force: true, showLoading: true)
             } else {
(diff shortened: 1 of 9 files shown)


iamtoruk (approved):
Verified locally: swift build green, 81 tests across 17 suites green, and both load-bearing behaviors are mutation-pinned; inverting the change comparison and removing the 30-minute backstop each fail the guard tests, including the dedicated backstop-expiry case. The design reads right: the shallow snapshot is a perf optimization with the backstop as the correctness contract, interactive refreshes always bypass, and anchoring only on fetch success means failures cannot mask changes. The Swift-side provider root list duplicating TS path logic is unavoidable cross-language and is exactly what the backstop bounds. The CPU reduction figures rest on the #703 measurements; the mechanism they depend on is what is verified here.

PR #709, opened 2026-07-16
fix(pricing): add kimi-k3 to the pricing fallback

ozymandiashh:
kimi-k3 usage renders $0: the model id is in neither the bundled LiteLLM snapshot nor `pricing-fallback.json` (verified across all 143 kimi/moonshot keys). Observed on real data: 14 kimi-k3 calls via the hermes provider priced at $0 that reprice to $18.29 with the correct rates.

Adds the official Moonshot rates to the fallback: $3.00/1M input, $15.00/1M output, $0.30/1M cache read (source: https://platform.kimi.ai/docs/pricing/chat-k3, figures verified against the live page), plus a `calculateCost` pin in tests.

Noted while verifying, not included: `kimi-k2.7-code` (bare id, without the `-highspeed` suffix that IS priced) also shows up unpriced on real data; left out pending an official figure.


iamtoruk:
Closing as superseded: kimi-k3 is already in the committed pricing fallback on main (landed with the 0.9.16 release regeneration, c58a705), and the Kimi Code provider that consumes it just merged in #750. This branch now conflicts on the regenerated file. Thanks for getting the pricing in motion — it's what made #750's cost computation work out of the box.

PR #710, opened 2026-07-16
feat(models): agent x model cross-breakdown (--by-agent)

### LINKED ISSUE #708: Feature request: agent × model cross-breakdown (e.g. codeburn agents --by-model)
## Problem

The dashboard shows "Skills & Agents" and "By Model" as separate panels, but there's no way to see which model each agent ran on. When agents are pinned to different models (common in orchestrator setups — e.g. planner on Opus, implementers on Sonnet, scouts on Haiku), the flat panels can't answer:

- Which agent is driving my Opus spend?
- Did an agent's model assignment change over time? (I found one of my agents had silently switched models — only visible by joining the two dimensions manually.)

## Current behavior

- `codeburn models --by-task` crosses model × task category, but not agent.
- `codeburn export` session records don't include `subagentTypes` or per-agent model, so the cross can't be reconstructed from the export either.

## Proposed behavior

One row per (agent, model), with the usual token/cost columns. Possible shapes:

- `codeburn agents --by-model`, or
- `codeburn models --by-agent` (mirroring the existing `--by-task` pattern)

Plus ideally exposing `subagentType` + `model` on exported per-record data so users can build custom crosses.

## Data is already there

For Claude Code, both fields are already parsed:

- `~/.claude/projects/<proj>/<session>/subagents/agent-*.meta.json` → `agentType`
- the paired `agent-*.jsonl` → `message.model` + `message.usage` per turn

They're just never joined in the report. Verified with a shell script over raw transcripts; sample output:

```
AGENT        MODEL              INPUT       OUTPUT
planner      claude-sonnet-x    5,567,464   49,166
reviewer     claude-opus-x      6,335,804   48,825
scout        claude-haiku-x     5,882,333   31,908
```

## Workaround

jq/awk aggregation over the raw subagent transcripts — works but Claude-Code-only and misses codeburn's pricing/cache-tier logic (raw input tokens conflate cache reads with fresh input, so they don't map to cost).

iamtoruk:
## Problem

The dashboard shows Skills and Agents and By Model as separate panels, so orchestrator setups with agents pinned to different models cannot see which agent drives which model's spend. Closes #708.

## Fix

codeburn models --by-agent: one row per (provider, model, agent), mirroring --by-task exactly across the flag, aggregation, and all four output formats (table, markdown, csv, json). The agent dimension comes from each subagent session's own agentType joined with its calls, which is the authoritative pairing; spawn markers on main sessions are deliberately not used. Ordinary main sessions and non-Claude providers bucket under main. --by-task and --by-agent are mutually exclusive with a clear error. The renderers gained a generic third-column concept rather than a copy of the task path.

Real output from a machine with a heavy orchestrator day (the exact question from the issue, answered):

    Claude  Opus 4.8   main             691.7M  446.51
                       general-purpose   61.3M   50.20
    Claude  Fable 5    main             203.2M  248.13
                       general-purpose   24.7M   55.21
    Claude  Haiku 4.5  general-purpose   68.8K    0.03

The issue's secondary ask, exposing subagentType on per-record exports, is left for a follow-up to keep this diff on one surface.

## Verification

- 8 new tests: the model x agent cross (one agent on two models, two agents on one model, main, non-Claude), group ordering, all four renderers, and the mutual-exclusion error. Mutation-checked twice: forcing every bucket to main fails exactly the two aggregation tests; mislabeling the third column fails exactly the two renderer pins.
- tsc clean; full suite green apart from the pre-existing #681 electron failures and the playwright-dependent sync e2e on this checkout.
- aggregateModels callers without options (usage-aggregator) are behaviorally unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,6 +447,7 @@ Sync sends token counts, costs, models, and projects — never prompts or code. 
 |---------|--------------|
 | `codeburn models` | Per-model token + cost table (last 30 days) |
 | `codeburn models --by-task` | Break each model into per-task-type rows |
+| `codeburn models --by-agent` | Break each model into per-agent rows (which agent drove which model's spend); Claude subagent transcripts only, other providers and main sessions bucket under "main" |
 | `codeburn models --top 10` | Only the 10 most expensive models |
 | `codeburn models --format markdown` | Emit a paste-friendly markdown table |
 | `codeburn models --task feature` | Filter to feature-development work |
--- a/src/main.ts
+++ b/src/main.ts
@@ -1895,19 +1895,24 @@ program
 
 program
   .command('models')
-  .description('Per-model token + cost table, optionally exploded by task type')
+  .description('Per-model token + cost table, optionally exploded by task type or agent')
   .option('-p, --period <period>', 'Analysis period: today, week, 30days, month, all', '30days')
   .option('--from <date>', 'Custom range start (YYYY-MM-DD)')
   .option('--to <date>', 'Custom range end (YYYY-MM-DD)')
   .option('--provider <provider>', 'Filter by provider (e.g. claude, codex, cursor)', 'all')
   .option('--task <category>', 'Filter to one task type (e.g. feature, debugging, refactoring)')
   .option('--by-task', 'One row per (provider, model, task) instead of one row per (provider, model)')
+  .option('--by-agent', 'One row per (provider, model, agent) instead of one row per (provider, model). Claude subagent transcripts only; other providers and main sessions bucket under "main"')
   .option('--top <n>', 'Show only the top N rows', (v: string) => parseInt(v, 10))
   .option('--min-cost <usd>', 'Hide rows below this cost threshold', (v: string) => parseFloat(v))
   .option('--no-totals', 'Suppress the footer totals row')
   .option('--format <format>', 'Output format: table, markdown, json, csv', 'table')
   .action(async (opts) => {
     assertProvider(opts.provider, 'models')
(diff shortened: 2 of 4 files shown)


ozymandiashh (commented):
Reviewed against a parallel implementation of the same issue that went through a three-model adversarial panel today, so these findings are battle-tested rather than speculative. The core design matches what we converged on independently: session-level attribution from the subagent transcript's own agentType, spawn markers deliberately unused (crossing per-call subagentTypes would bill the parent's Task-invocation tokens to the child agent). The generic third-column renderer is a nicer shape than mirroring the task path. Three findings:

1. **The bare `main` sentinel collides with a real agent named "main".** agentType is free-form text from the meta file, not an enum: a user-defined subagent literally named `main` merges silently into the main-session bucket and the attribution the feature exists for is corrupted for exactly that agent. A non-slug sentinel (we used `(main)` with parentheses) cannot collide; worth a test pinning that a subagent named `main` stays distinct.

2. **The inherited `--min-cost 0.01` default quietly hides sub-cent agent buckets.** At task granularity that default is invisible; at agent granularity sub-cent buckets are the norm (scouts, one-shot helpers), so agents disappear from the audit without a trace. One README sentence recommending `--min-cost 0` for a complete agent inventory saves the support question.

3. For the deferred export follow-up: when subagentType/model land on sessions.csv, append them AFTER all legacy columns — inserting mid-row shifts positions for by-index CSV consumers. We have the export half already built and panel-reviewed (records.csv with one row per call, sessions.csv with appended fields, formula-injection-safe); happy to open it as the follow-up PR so it composes with this one.

PR #712, opened 2026-07-16
feat(export): subagentType and model on per-record exports; --by-agent follow-ups

ozymandiashh:
Follow-ups to #710, as offered in its review. Three pieces, all small:

1. **Export (issue #708's second ask, deferred there)**: new `records.csv` with one row per priced call including `subagentType` and `model`; `sessions.csv` gains both fields appended after all legacy columns, so by-index CSV consumers keep their positions; JSON fields are additive; `escCsv` formula-injection escaping applies to the new values.

2. **Collision-safe main bucket**: `agentType` is free-form text from the transcript meta, so the bare `main` sentinel merges a real subagent literally named `main` into the main-session bucket. The sentinel becomes `(main)`; a test pins that an agent named `main` stays distinct across table, markdown, json and csv.

3. **README**: one sentence noting the inherited `--min-cost 0.01` default can silently hide sub-cent agent buckets (common at agent granularity) and that `--min-cost 0` shows the full inventory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,7 +447,7 @@ Sync sends token counts, costs, models, and projects — never prompts or code. 
 |---------|--------------|
 | `codeburn models` | Per-model token + cost table (last 30 days) |
 | `codeburn models --by-task` | Break each model into per-task-type rows |
-| `codeburn models --by-agent` | Break each model into per-agent rows (which agent drove which model's spend); Claude subagent transcripts only, other providers and main sessions bucket under "main" |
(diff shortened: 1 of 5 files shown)


iamtoruk (approved):
Verified locally: tsc clean, full suite green apart from the #681 electron failures, and both behaviors mutation-pinned; reverting the (main) sentinel to bare main fails exactly the collision test, and blanking subagentType in the record rows fails exactly the two export pins. Ran a real export over yesterday's data: records.csv carries subagentType and model on 631 real agent rows, and sessions.csv appends the new fields after every legacy column so positional consumers keep working, exactly as described. The (main) collision catch on #710 was a good one, thanks. This also closes out the export follow-up promised on #708.

PR #714, opened 2026-07-17
fix(yield): group attribution by canonical repo root, not path string

### LINKED ISSUE #713: yield: single-owner attribution double-credits commits across monorepo subdirs and worktrees (groups keyed by path string, not repo root)
## Problem

The single-owner commit attribution shipped in #692 enforces "each commit is credited to at most one session" per **launch-directory string**, not per repository. Two facts combine:

- `attributeCommits` groups sessions into repo groups keyed by the raw `project.projectPath` string, and `isGitRepo` uses `git rev-parse --is-inside-work-tree`, which is true for ANY subdirectory of a repo (src/yield.ts, grouping block).
- `getCommitsInRange` runs `git log --all --since --until` with no pathspec, so from any subdirectory it returns the identical repo-wide commit list.

So in a monorepo (sessions launched from `repo/packages/a` and `repo/packages/b`) or with git worktrees, the same underlying repository forms two independent groups, each pulls the full commit history, and the same commit is independently awarded once per group.

## Repro (verified on the merged code)

Two sessions in one temp repo, launched from two subdirectories, one commit inside both windows: the commit is credited to a session in EACH group — one commit, two `productive` sessions, `commitCount` totals 2. The second "owner" is not even `ambiguous`, since it is the sole candidate within its own group. Three scenario variants reproduced; every existing test stays green because each test uses a single repo path per group.

## Suggested fix

Key the repo groups by the canonical repository root — `git rev-parse --show-toplevel` (and `--git-common-dir` if worktrees should unify) — instead of the raw `projectPath` string. A regression test with two sessions launched from two subdirectories of one repo pins it.

Found during an independent review pass with adversarial verification; happy to provide the full repro scripts.

ozymandiashh:
Closes #713.

## Problem

The single-owner attribution from #692 enforces "one commit, one owner" per launch-directory string: repo groups are keyed by the raw `projectPath` while `getCommitsInRange` runs `git log --all` with no pathspec. Sessions launched from two subdirectories of one monorepo (or from linked worktrees) form independent groups that each pull the identical repo-wide history, and the same commit is credited once per group - one commit, two productive sessions, `commitCount` 2. Repro details in the issue; every pre-existing test stayed green because each used a single path per repo.

## Fix

- Repo groups key on the canonical root (`git rev-parse --show-toplevel`), resolved once per path and cached.
- Linked worktrees unify via `--git-common-dir`, both sides realpath-canonicalized before comparison (symlinks, platform case-insensitivity). If git or realpath fails, the session degrades to the existing cwd fallback; a comment documents the accepted residual for a main repo moved after `git worktree add` (git's own metadata is stale in that state).
- Non-git and missing paths keep the current behavior.

## Verification

- New regression: two sessions from two subdirectories of one temp repo, one commit inside both windows -> exactly one productive owner, the other lost-candidacy, `commitCount` 1. Pre-fix run on the base recorded both sessions productive.
- New linked-worktree regression, same shape. Both tests are mutation-checked: removing the canonical grouping (or the common-dir merge lookup) makes them fail with the exact #713 symptom.
- `npx tsc --noEmit` clean; yield suites 8/8.

### DIFF
--- a/src/yield.ts
+++ b/src/yield.ts
@@ -1,4 +1,6 @@
 import { execFileSync } from 'child_process'
+import { realpathSync } from 'node:fs'
+import { resolve } from 'node:path'
 import { parseAllSessions } from './parser.js'
 import type { DateRange, SessionSummary } from './types.js'
 
@@ -60,8 +62,31 @@ function runGit(args: string[], cwd: string): string | null {
   }
 }
 
-function isGitRepo(dir: string): boolean {
-  return runGit(['rev-parse', '--is-inside-work-tree'], dir) === 'true'
+type RepoResolution = {
+  readonly root: string
+  readonly commonDir: string
+}
+
+function canonicalizePath(path: string): string {
+  try {
+    return realpathSync(path)
+  } catch {
+    return path
+  }
+}
+
+function resolveRepo(dir: string): RepoResolution | null {
+  // Accepted residual: moving the main repo after worktree add leaves stale Git metadata; Git fails and sessions fall back to cwd.
+  const root = runGit(['rev-parse', '--show-toplevel'], dir)
+  if (!root) return null
+
+  const commonDir = runGit(['rev-parse', '--git-common-dir'], root)
+  return {
+    root,
+    // Linked worktrees have different top-levels but `git log --all` reads
+    // the same shared repository history, so they must share attribution.
+    commonDir: canonicalizePath(commonDir ? resolve(root, commonDir) : root),
+  }
 }
 
 function getMainBranch(cwd: string): string {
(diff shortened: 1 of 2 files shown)


iamtoruk:
Apologies — this collided with a maintainer-side fix (#719) that was developed in parallel straight off your issue writeup, before we saw this PR; entirely our process miss for not checking the issue's linked PRs first. #719 landed on the same design you proposed (canonical grouping, worktrees unified via realpath'd common-dir, separate repos guarded), so the bug itself is fixed on main.

Two details in yours look potentially additive over #719: the explicit platform case-insensitivity handling in the canonical comparison, and the documented residual for a main repo moved after `git worktree add`. If either isn't fully covered by #719's realpath canonicalization, a focused follow-up PR for the delta would be very welcome — and either way, thank you for both the sharpest bug report this repo has received and a complete fix to go with it.

ozymandiashh:
No apology needed, glad #719 landed on the same design. I checked whether the two details were covered: the residual comment was indeed missing, and the case-insensitivity one turned out to be a real gap — node's fs.realpathSync preserves the caller's casing (only realpathSync.native returns on-disk casing), so differently cased paths to the same repo still split into separate groups after #719. Both are in the focused follow-up you suggested: #751 (one-line native switch + guarded regression test + the residual comment). Closing this one as superseded.

Issue #645, opened 2026-07-09
Caches have no cross-process lock: concurrent CLI/menubar/MCP runs clobber each other's work


avs-io:
## Problem

Cache writes are correctly atomic (temp file, fsync, rename), but nothing coordinates across processes. daily-cache serializes only in-process via a promise chain (src/daily-cache.ts:184-190); session-cache, cursor-cache, and codex-cache do read-mutate-rename with no lockfile. CodeBurn deliberately runs as several concurrent processes: CLI, menubar refresh (every 30s), the guard statusline daemon, and the MCP server. Last-writer-wins means a full parse can overwrite a concurrent incremental append; nothing corrupts, but both processes re-parse next run. It is an invisible throughput tax that peaks exactly when the tool is used most, and it can make first-paint feel slow for no visible reason.

## Evidence

- src/daily-cache.ts:184-190 (in-process lock chain only)
- No O_EXCL/flock usage in src/session-cache.ts or src/daily-cache.ts; the only lockfile in the repo is src/act/journal.ts:56.
- Verified on main at 1a75484.

## Proposed fix

Reuse the act journal's lockfile pattern: acquire an O_EXCL lockfile with a stale-lock timeout around saveCache/saveDailyCache (and optionally around the expensive parse itself so concurrent processes share one result instead of racing). The orphaned-temp GC already in session-cache.ts is the model for stale-lock cleanup.

avs-io:
Design proposal before any code, since this touches every cache consumer. Feedback welcome — particularly on lock scope (§2 recommends locking the full parse+write transaction, not just writes) and the first-PR slice (§7, session-cache only).


## 1. Problem restatement

Cache publication is atomic, but cache computation is not coordinated across processes. `daily-cache` serializes only within one process, while the session, Cursor, and Codex caches can each load an old snapshot, compute independently, and replace a newer result; the file remains valid, but the losing work must be repeated on the next refresh. The macOS menubar and dashboard can refresh every 30 seconds, MCP calls aggregate through the same parser, and `guard install`/`guard refresh` invoke `parseAllSessions`; the guard statusline uses a separate per-session cache and is not part of this race.

## 2. Lock scope decision

| Cache | Write-only lock | Parse + write lock |
|---|---|---|
[...]

PR #743, opened 2026-07-18
cache: strict cross-process gate for the warm session-cache refresh transaction

avs-io:
Draft on purpose: per the commitment in #645, this stays a draft until you give the go-ahead — it's here so the code is reviewable whenever bandwidth exists, not to jump the queue.

Implements the **revised design** (third comment on #645): first slice, warm session-cache transaction only.

## What changed

- `hydrating.lock` is untouched — still cold-only, advisory, fail-open.
- New `src/cache-refresh-lock.ts`: a strict cross-process gate (`session-refresh.lock`) around the warm read→reconcile→parse→save transaction. Four explicit outcomes: `acquired`, `completed-by-other` (reload the now-fresh cache), `timed-out` / `unavailable` (serve the prior complete snapshot read-only — a correctness lock never fails open into mutation). Exclusive-create + random ownership token, 10s heartbeat, 90s staleness with a takeover-guard sidecar that never steals a heartbeating owner, in-process single-flight ahead of the filesystem lock, monotonic-clock wait deadlines, bounded Windows EPERM/EBUSY retries.
- Publication fence: `saveCache` re-verifies ownership under the takeover guard immediately before the payload rename and aborts publication if displaced — a suspended/clock-jumped owner can no longer overwrite its successor's work.
- Waiter budget is 30s, comfortably under the menubar's 45s subprocess timeout; a timed-out waiter serves the previous complete snapshot and never invalidates the holder.
- Read-only serve path threads through `parseAllSessions` so timeout/stale serving performs no cache mutation at all.

## How tested

- **Red-before**: `tests/cache-refresh-lock-process.test.ts` is a real child-process lost-update regression — two warm processes parse disjoint changed files; with the lock wiring stashed it fails on stock code (later save drops the other's update), with the wiring it passes. Verified both directions locally.
- Unit suite with injected clock/poll: every acquisition outcome, single-flight, heartbeat-prevents-takeover, exactly-one-takeover-winner, fence aborts a displaced save, timed-out waiter leaves the holder intact, token-scoped release.
- Full suite: 2162 passed with `dist` built (the `app/electron/cli.test.ts` fresh-clone failures are #681, pre-existing).

Follow-ups per the design: daily-cache cross-process transaction (lock order `daily → session`), then #703's incremental parse runs inside this gate (draft PR opening alongside, stacked on this branch).

### DIFF
--- a/src/cache-refresh-lock.ts
+++ b/src/cache-refresh-lock.ts
@@ -1,0 +1,331 @@
+import { randomBytes } from 'crypto'
+import { existsSync } from 'fs'
+import { mkdir, open, readFile, stat, unlink, utimes, writeFile } from 'fs/promises'
+import { homedir } from 'os'
+import { join } from 'path'
+
+const LOCK_FILE = 'session-refresh.lock'
+const TAKEOVER_FILE = `${LOCK_FILE}.takeover`
+const DEFAULT_HEARTBEAT_MS = 10_000
+const DEFAULT_STALE_MS = 90_000
+const DEFAULT_WAIT_MS = 30_000
+const DEFAULT_POLL_MS = 100
+const WINDOWS_RETRIES = 3
+
+type LockRecord = { pid: number; token: string; at: number }
+
+export type RefreshLockClock = {
+  monotonicNow: () => number
+  wallNow: () => number
+}
+
+export type RefreshLockOptions = {
+  cacheDir?: string
+  clock?: RefreshLockClock
+  heartbeatMs?: number
+  staleMs?: number
+  waitMs?: number
+  pollMs?: number
+  sleep?: (ms: number) => Promise<void>
+}
+
+export type RefreshLockHandle = {
+  token: string
+  release: () => Promise<void>
+  verifyStillOwner: () => Promise<boolean>
+}
+
+export type RefreshLockOutcome =
+  | { outcome: 'acquired'; handle: RefreshLockHandle }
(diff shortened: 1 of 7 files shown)


iamtoruk:
Full adversarial review before merge: spec-fit against the #645 revised design confirmed, all four outcomes exercised, and every concurrency attack survived without corruption or lost updates: SIGKILL mid-transaction, SIGSTOP past staleness with a resumed zombie fenced out of publishing over its successor, racing warm refreshes with exactly one parser, and a real 30s timeout serving the prior snapshot with provably zero cache writes. Both mutations (neutered heartbeat, disabled fence) are caught by tests. Real-data drive: two concurrent runs on a copy of a 109MB cache produced byte-identical totals.

Two amendments pushed: (1) the publication fence could race the owner's OWN heartbeat for the takeover guard and abort a legitimate save (reproduced 196/3000 at a 1ms heartbeat; fail-safe but it discards the parse this lock exists to protect) — owner-side guard operations now run through an in-process serializer, with a regression test at heartbeatMs 1 that fails against the unserialized code; (2) merged current main, resolving the readOnly gate against the incremental-append reconcile so readOnly never parses changed files while the append optimization is preserved, verified by both suites. The missing staleness marker on readOnly serves is accepted as bounded and self-healing for this release and tracked in the freshness-marker issue. Windows retry paths reviewed statically only; unverified on Windows.

PR #755, opened 2026-07-20
daily-cache: never lose history the sources can no longer re-derive (v14)

iamtoruk:
## Problem

Long-term usage history was silently destroyed, repeatedly. A user with heavy spend since March saw a 6-month total that was roughly half of reality, days like Apr 15 and Apr 19 rendered as 0.00 with 0 calls despite hundreds of dollars of tracked spend at the time, and an all-time number that never grew past ~15k.

Root cause chain:

1. Source session files are ephemeral. Claude Code deletes transcripts after ~30 days, so a cached day whose sources are gone exists nowhere except the daily cache.
2. The daily cache treated itself as disposable. Every invalidation (schema version bump, savings-config change, timezone change, incomplete-hydration retry) discarded all cached days and re-derived from surviving sources, truncating history to the source-retention window.
3. Five schema bumps landed between Jun 22 and Jul 16, each repeating the truncation.

Proof from a real machine: the v9 cache backup (Jul 2) held Claude spend for Apr 26 through Jun 1 that the current v13 cache had zeroed. The all-time plateau is the sliding-window equilibrium of ~30 days of Claude plus older Codex files that never get deleted.

## Fix

One invariant, enforced everywhere: an invalidation may re-derive a (day, provider) slice only if the parse actually produced data for it; everything else is carried forward.

- `ensureCacheHydrated` rebuilds by per-(day, provider) merge instead of wipe. Carried slices keep the accounting of the version that recorded them and are marked `carried` for provenance.
- Loading a missing or corrupt versioned cache file now adopts days from every older daily-cache file in the cache dir (legacy unversioned, older versioned, .bak copies, orphaned but parseable .tmp) as a union per (day, provider), higher schema version winning per pair. Version bumps are lossless from v14 on, and existing machines recover whatever their old files still hold on first run.
- Provider slices now store the full per-provider breakdown (tokens, models, categories, sessions) so future carry-forwards stay exact, not just cost-exact.
- A partial session parse can only fill gaps, never overwrite finalized baseline slices, so an interrupted hydration cannot freeze undercounts into history.

Hardening from adversarial review (three review rounds, five real findings fixed, final verdict SHIP):

- Opaque pre-v5 days (day totals without provider slices) merge all-or-nothing, closing a double-count path.
- Zero-data placeholder slices (a session starting before midnight with all turns after it) neither block nor lose carried data; session counts deduplicate by max.
- Adoption purges today/future entries, applies 730-day retention, and clamps a stale lastComputedDate so a purged day cannot be skipped forever.

## Verification

- New invariant test suite (`tests/daily-cache-carry-forward.test.ts`): every invalidation trigger run against an empty re-parse must preserve the baseline; corrections must still win for re-derivable days; adoption union priority; the real Apr-26 rescue scenario as a fixture. Full suite green (3653 tests).
- Mutation-verified: the invariant repros fail on the pre-fix code (days wiped) and pass on the new code.
- End-to-end on a copy of a real cache dir: the rebuilt v14 cache contains every (day, provider) pair present in any older cache file, zero missing, and restores ~165 USD of Apr-May spend that the July rebuilds had dropped, while leaving every pre-existing file byte-identical.

## Follow-ups (not in this PR)

- Aggregate-before-evict in the session cache (`parser.ts`) to close the residual case where partially expired sources re-derive a day slightly lower at invalidation time (~0.1% observed).
- Render days older than the oldest retained data as "no data" instead of 0.00 in the UI.
- `codeburn doctor` warning about Claude Code's `cleanupPeriodDays`.

### DIFF
--- a/src/daily-cache.ts
+++ b/src/daily-cache.ts
@@ -5,7 +5,15 @@ import { homedir } from 'os'
 import { join } from 'path'
 import type { DateRange, ProjectSummary } from './types.js'
 
-// Bumped to 14: NEVER-LOSE history. Session files are ephemeral (Claude Code
+// Bumped to 15: per-project daily rollups. Days and provider slices now carry
+// a `projects` breakdown (cost/calls/savings/sessions per project) so project
+// history outlives the session files, like models and categories already do.
+// This bump is the first to ride the v14 carry-forward: the old cache is
+// adopted losslessly and only days whose sources survive are re-derived (now
+// with projects); days already sourceless keep their totals and simply have
+// no project split.
+//
+// v14: NEVER-LOSE history. Session files are ephemeral (Claude Code
 // deletes transcripts after ~30 days), so a day that can no longer be re-derived
 // from sources exists ONLY in this cache. Every earlier version treated the
 // cache as disposable — schema bumps, savings-config changes, timezone changes
@@ -49,8 +57,8 @@ import type { DateRange, ProjectSummary } from './types.js'
 // that older binaries skipped. v8 added local-model savings to the daily
 // rollup; the `savingsConfigHash` field is invalidated separately when the
 // user changes their `localModelSavings` mapping.
-export const DAILY_CACHE_VERSION = 14
-const MIN_SUPPORTED_VERSION = 14
+export const DAILY_CACHE_VERSION = 15
+const MIN_SUPPORTED_VERSION = 15
 // Version-suffixed so different binaries each own a distinct file and never
 // clobber an incompatible schema. Bumping the version mints a fresh filename;
 // adoptOlderDailyCaches then unions days out of every previous file (including
@@ -70,6 +78,8 @@ export type ModelDayStats = {
 
 export type CategoryDayStats = { turns: number; cost: number; savingsUSD: number; editTurns: number; oneShotTurns: number }
 
+export type ProjectDayStats = { cost: number; calls: number; savingsUSD: number; sessions: number }
+
 export type ProviderDaySlice = {
   calls: number
   cost: number
@@ -87,6 +97,7 @@ export type ProviderDaySlice = {
   oneShotTurns?: number
   models?: Record<string, ModelDayStats>
   categories?: Record<string, CategoryDayStats>
+  projects?: Record<string, ProjectDayStats>
 }
 
 export type DailyEntry = {
@@ -104,6 +115,10 @@ export type DailyEntry = {
   models: Record<string, ModelDayStats>
   categories: Record<string, CategoryDayStats>
   providers: Record<string, ProviderDaySlice>
+  /// Per-project rollup (session-level project attribution). Absent on days
+  /// recorded before v15 — those days keep their totals but have no project
+  /// split, and nothing can reconstruct one once the sources are gone.
+  projects?: Record<string, ProjectDayStats>
   /// Present when some of this day's data was carried forward from an earlier
   /// cache generation instead of re-derived from session files (the files no
   /// longer exist). Carried values keep the accounting of the version that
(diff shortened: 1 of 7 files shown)


iamtoruk:
Added per-project daily rollups (v15, d2216d6): days and provider slices now carry a projects breakdown (cost/calls/savings/sessions per project), so project history survives source-file deletion like models and categories already do. First schema bump to ride the v14 carry-forward; verified lossless on a real cache dir (identical totals, zero lost pairs, project splits on all 36 derivable days). Two adversarial review rounds on the increment; findings fixed with regression tests, including a real prototype-pollution path via foreign cache keys named __proto__. Final verdict SHIP.

avs-io (changes requested):
There is one blocking integration gap in the final menubar payload path.

`buildPeriodDataFromDays(allDays, ...)` correctly builds cache-backed historical totals at `src/usage-aggregator.ts:275`, but the all-provider path immediately replaces that result at line 291 with `buildPeriodData(...scanProjects)`. `scanProjects` contains only sessions whose source files still survive. The new project rollups are similarly not surfaced: `buildPeriodDataFromDays()` never folds `DailyEntry.projects` into `PeriodData.projects`, and lines 411-431 rebuild `currentData.projects` solely from `scanProjects`.

I reproduced this on the exact PR head (`b2876145`) with an empty source directory and a v15 cache containing one carried $100 historical day/project:

- `history.daily` contained the $100 day (the new cache logic worked)
- `current.cost` was $0
- `current.calls` was 0
- `current.topProjects` was empty

This means the cache artifact is repaired, but the user-visible 6-month/all-time headline totals and By Project view described in the PR remain truncated once source files disappear.

Suggested implementation shape:

1. Keep the `buildPeriodDataFromDays(allDays, ...)` result as the authority for cache-backed totals instead of replacing it wholesale.
2. Merge only fields that genuinely require the surviving-session parse, such as `estimatedCostUSD`, `unpricedModels`, efficiency, and session-detail metadata.
3. Fold `DailyEntry.projects` across the selected cached days into `PeriodData.projects`, then merge today's live project rollups. Historical projects can omit `sessionDetails` when the source sessions no longer exist.
4. Add an end-to-end `buildMenubarPayloadForRange()` regression test that seeds a populated current-version daily cache, points discovery at an empty source directory, and asserts that headline cost/calls and `topProjects` match the carried cache data—not only `history.daily`.

The cache carry-forward implementation and focused tests otherwise look thoughtful. I ran the 70 affected daily-cache/day-aggregator tests, `tsc --noEmit`, and `git diff --check`; all passed. The missing final-payload coverage is the blocker.

iamtoruk:
Confirmed the finding and fixed it in 41209d9. You were right: line 291 replaced the cache-backed period data wholesale, so the headline, models table, and topProjects were all truncated to surviving sessions; my earlier end-to-end check asserted history.daily and the provider panel but never current.cost.

Fix follows your suggested shape: buildPeriodDataFromDays(allDays) is now the authority on the unscoped all-provider path; the scan merges in only estimatedCostUSD, unpricedModels, per-model estimated markers, sessionDetails, and project paths. Project totals fold DailyEntry.projects over the same day set as the headline (ProjectDayStats gained a stored path so carried-only projects keep a friendly display name). Sessions merge by max since start-day bucketing and active-day counting undercount differently; a --day filter regression surfaced exactly that and is covered.

Your regression test shape is implemented in tests/menubar-carried-headline.test.ts: seeded current-version cache with a carried day, empty HOME, asserts current.cost/calls/sessions, topModels, and topProjects against the carried data. Verified on a real cache dir: the 6-month headline now equals the history.daily sum to the cent, including the rescued April days. Known honest gap, documented in-code: days recorded before v15 have totals but no project split, so the projects list can sum below the headline.

ozymandiashh:
Pushed cba064f: closes the residual corrupt-input gaps in the v15 ingestion sanitization, found during an adversarial review of this PR.

What it covers:
- Day-level models/categories now sanitized at ingestion (a day with models:'bad' previously survived load and produced NaN in buildPeriodDataFromDays via per-character string iteration).
- Map keys shadowing Object.prototype (constructor, toString, ...) are dropped at ingestion in all sanitized maps. The merge path was already hasOwn-guarded, but modelTotals[name] ?? init in buildPeriodDataFromDays still resolved such keys to inherited prototype members and produced NaN. Deliberate tradeoff: such names are reclassified invalid at ingestion (a project literally named 'constructor' loses per-project attribution in carried days).
- Top-level lastComputedDate validated (YYYY-MM-DD string or null) - a non-string previously crashed the gap-start parse (.slice on a number).
- Reworded the stale savingsConfigHash comment (pre-v14 discard wording).

Verification: 3 regression tests, each verified to fail on the pre-fix code; relevant suites 73/73 green incl. the new menubar-carried-headline tests; tsc clean; an independent corrupt-cache battery (null slices, string providers maps, non-numeric fields, prototype keys, malformed lastComputedDate) passes on load + hydration + period aggregation.

Review findings NOT addressed here, worth separate issues:
1. TZ-change double-count: on tzChanged re-derive, a midnight-adjacent turn that shifts days is counted fresh on its new day while the old-TZ slice on the old day is carried - same calls counted twice within the source window.
2. Multi-provider turn stats: editTurns/oneShotTurns/category turns are duplicated per provider slice; a merge that re-derives one provider and carries the other double-counts those turn-level counts at day level (counts only, not cost).
3. Policy note for future semantic bumps: presence-based carry-forward cannot distinguish 'sources gone' from 'semantics moved the data' - a future bump that changes bucketing (like v12->v13) will double-count moved data within the source window unless that bump forces slice-drop for verified-source days.
4. +1 on your listed follow-ups (aggregate-before-evict is the top residual; partial-expiry re-derivation finalizing undercounts at invalidation was independently flagged by two reviewers).

PR #756, opened 2026-07-20
feat(insights): workflow intelligence (corrections, time-to-first-edit, file churn, coverage) + dash wire-up

iamtoruk:
Workflow-intelligence metrics for the CLI plus wiring three typed-but-never-rendered fields into the web dash.

## New metrics (all computed at report time from existing cached shapes; no parse-version bump, no daily-cache changes)

- **User corrections**: mirror of `scanSelfCorrections` pointed at user follow-ups ("that's not what I", "you missed", "revert that", "still broken"...). Conservative pattern list; bare "wrong"/"undo" excluded to avoid false positives in ordinary task text. Count + rate over prompt-bearing turns.
- **Median time-to-first-edit**: per session, first turn -> first edit-family call; no-edit sessions excluded, out-of-order timestamps clamped.
- **File-churn ranking**: edit-family tool calls' file paths aggregated to "top reworked files" (distinct sessions, then edits). Top 15 in optimize (project-relative paths), top 8 basename-only in the payload for privacy.
- **Pricing coverage**: 0-1 share of cost-bearing calls that were priced, alongside the existing unpricedModels.
- **Coaching notes**: up to 3 dry one-liners in optimize keyed on the strongest signals.
- **Payload (add-only)**: `current.workflow`, `current.topReworkedFiles`, `current.pricingCoverage`.

Deliberately NOT shipped: "sessions ending without an assistant reply" — the parser drops trailing user-only turns before they reach the cache, so a faithful version needs new parse-time capture + a global cold re-parse for one noisy count (tool-results ride user-role messages). The optional payload slot stays unset.

## Dash

`avgCostPerSession` column on Top projects, Top models list upgraded to a table exposing `savingsUSD`, new Model-efficiency table (`costPerEdit`, `oneShotRate`) — all three were in the data contract but rendered nowhere.

## Verification

CLI suite 2011 passed; 25 new tests (phrase guards, median edges, churn ties, coverage edges, note thresholds). Dash typecheck + build green. Real-machine run: corrections 8 (rate 0.004), median TTFE 5.8 min, coverage 99.99%, churn top = the files this repo's own recent work kept touching. No overlap with #755 (daily-cache untouched by design).

### DIFF
--- a/dash/src/App.tsx
+++ b/dash/src/App.tsx
@@ -83,9 +83,6 @@ function DeviceView({ payload, isRemote, unit }: { payload?: Payload; isRemote: 
   const toolBars: BarItem[] = c
     ? Object.entries(c.providers).filter(([, v]) => v > 0).sort((a, b) => b[1] - a[1]).map(([k, v]) => ({ name: k, value: v, display: usd(v) }))
     : []
-  const modelBars: BarItem[] = c
-    ? c.topModels.filter((m) => m.cost > 0).slice(0, 8).map((m) => ({ name: m.name, value: m.cost, display: usd(m.cost) }))
-    : []
   const activityBars: BarItem[] = c
     ? c.topActivities.filter((a) => a.cost > 0).map((a) => ({ name: a.name, value: a.cost, display: usd(a.cost) }))
     : []
@@ -136,7 +133,37 @@ function DeviceView({ payload, isRemote, unit }: { payload?: Payload; isRemote: 
           <BarList items={toolBars} total={c?.cost} />
         </Panel>
         <Panel title="Top models">
-          <BarList items={modelBars} total={c?.cost} />
+          <DataTable
+            columns={[
+              { key: 'name', label: 'Model' },
+              { key: 'cost', label: 'Cost', num: true },
+              { key: 'calls', label: 'Calls', num: true },
+              { key: 'savings', label: 'Savings', num: true },
+            ]}
+            rows={(c?.topModels ?? []).filter((m) => m.cost > 0).slice(0, 8).map((m) => ({
+              name: m.name,
+              cost: usd(m.cost),
+              calls: fmtNum(m.calls),
+              savings: usd(m.savingsUSD),
+            }))}
+          />
+        </Panel>
+      </div>
+
+      <div className="mb-3">
+        <Panel title="Model efficiency">
+          <DataTable
+            columns={[
+              { key: 'name', label: 'Model' },
+              { key: 'costPerEdit', label: 'Cost/edit', num: true },
+              { key: 'oneShot', label: 'One-shot', num: true },
+            ]}
+            rows={(c?.modelEfficiency ?? []).slice(0, 10).map((m) => ({
+              name: m.name,
+              costPerEdit: usd(m.costPerEdit),
+              oneShot: `${Math.round(m.oneShotRate * 100)}%`,
+            }))}
+          />
         </Panel>
       </div>
 
@@ -152,11 +179,13 @@ function DeviceView({ payload, isRemote, unit }: { payload?: Payload; isRemote: 
                 { key: 'name', label: 'Project' },
                 { key: 'cost', label: 'Cost', num: true },
                 { key: 'sessions', label: 'Sessions', num: true },
+                { key: 'avgCost', label: 'Avg/session', num: true },
               ]}
               rows={(c?.topProjects ?? []).slice(0, 10).map((p) => ({
                 name: p.name,
                 cost: usd(p.cost),
                 sessions: fmtNum(p.sessions),
+                avgCost: usd(p.avgCostPerSession),
               }))}
             />
           )}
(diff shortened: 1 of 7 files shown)


iamtoruk:
Reviewed against post-merge-train main. The feature is sound (median math, correction patterns, privacy claims, and add-only payload all verified, two behaviors mutation-tested), but the three new payload fields were silently dropped on the default all-provider path: the carry-forward headline (#755) serves totals from the durable day set and merges only listed scan-derived fields, and workflow/topReworkedFiles/pricingCoverage were not on the list. On real data they read 0/null/[] and pricingCoverage defaulted to a dishonest 1.0. Fixed in f8c9a6a (propagation plus a carried-headline regression test); verified on a cache copy: corrections 14, TTFE ~5.5 min, 8 reworked files, coverage 0.947, and the headline still equals the history sum. One optional nit left to author discretion: the revert pattern accepts bare 'revert the ...' so ordinary task requests match ('revert the migration'); the sibling undo pattern is tighter.

ozymandiashh (changes requested):
Ran an in-depth review of this PR (four independent reviewer passes plus my own; every finding below was verified on the code at the cited lines before inclusion). The new code is clean and well tested in isolation, but three issues are blocking.

**F1 (HIGH, merge semantics).** The branch is cut from a pre-#755 main. The merge with current main is textually clean, but on the default unscoped all-provider path `currentData` now comes from the daily-cache authority (`src/usage-aggregator.ts:296-320` on main) and only `estimatedCostUSD`, `unpricedModels`, `sessions` and the per-model estimated markers are grafted from the scan. The new `workflow`, `topReworkedFiles` and `pricingCoverage` fields are computed only inside `buildPeriodData`, so after the merge they are dropped on that path and the payload falls back to `corrections: 0`, `correctionRate: null`, `medianTimeToFirstEditMs: null`, `topReworkedFiles: []`, `pricingCoverage: 1`. The headline feature reports empty plus a false "100% priced" exactly on the default surface. CI stays green because no test asserts these fields on the cache-authority path. Fix: rebase onto current main and graft the three fields from `scanData` in the `isAllProviders` branch (they are derivable solely from surviving sessions, the same rationale the existing comment gives for `unpricedModels`), plus a test asserting them on that path.

**F5 (MED).** The correction scanner contradicts its own stated scope. The doc comment and the PR text scope it to user follow-ups, but `scanUserCorrections` counts every turn with a non-empty `userMessage`, including the session-opening prompt (`src/workflow-insights.ts:40-48`), and the `revert the...` / `undo the...` patterns (`:19-20`) match ordinary opening task requests such as "revert the migration we shipped". `corrections` and `correctionRate` overstate systematically, and the coaching note keyed on them fires for users who never corrected anything. The false-positive guard list in the tests happens to avoid exactly this case ("undo the migration when done" passes only because "migration" is not in the alternation). Fix: skip the first prompt-bearing turn per session and tighten those two patterns.

**F2 (MED).** `pricingCoverage: current.pricingCoverage ?? 1` (`src/menubar-json.ts:478`) turns "not computable" into "100% coverage". Make the payload field `number | null` and default to `null`.

**F6 (MED).** The `pricingCoverage` denominator counts calls of local/free models that `findUnpricedModels` deliberately excludes (`src/models.ts:724-735`). A user with 95% ollama calls and 5% unpriced cloud calls reads 0.95 coverage while 100% of their genuinely cost-bearing calls are unpriced. Use the same exclusions for the denominator.

**F3 (MED).** `sessionTimeToFirstEditMs` (`src/workflow-insights.ts:63-71`) skips an unparseable timestamp on the FIRST edit call and returns the time to the first parseable one, which measures "time to first parseable edit". Return `null` for the session instead, as already happens when `turns[0].timestamp` is invalid.

**F4 (MED).** File churn is keyed on the raw model-emitted path (`src/parser.ts:1196`) with no separator or relative-path normalization, while `relativizePath` and the payload basename hardcode `/` (`src/workflow-insights.ts:96-103`, `src/menubar-json.ts:418-420`). On Windows the privacy redaction is a no-op: the full absolute path, username included, ships in a payload whose own comment says it can leave the machine. Mixed absolute/relative emissions also split one file's churn across map entries. The parser already normalizes backslashes for project paths (`src/parser.ts:60-67`), so the same treatment fits here.

Non-blocking notes: `buildCoachingNotes` re-hardcodes the `editTurns >= 5` threshold that `worstOneShotCategory` already applies; the `still failing` pattern matches CI logs pasted into prompts; the three new payload fields ship with zero consumers in the repo today (disclosed as add-only, just noting it).

Also verified and fine, to save you the trip: `usd()` is null-safe, `findUnpricedModels` skips `<synthetic>`, `avgCostPerSession` exists on the payload, `modelEfficiency` nulls are filtered before the dash, and the deliberately excluded unanswered-sessions metric is fairly disclosed.

iamtoruk:
Full disposition of the review, finding by finding. F1 was fixed before merge in f8c9a6a (the three fields are grafted in the isAllProviders cache-authority branch, with a carried-headline regression test). F2 through F6 all verified valid against the merged code and fixed in #763: corrections now skip the session-opening prompt and the revert pattern matches demonstratives only (real-data corrections dropped from 14 to 5, all nine removed being exactly the opener/task-phrasing false positives you predicted); TTFE returns null on an unparseable first edit; churn normalizes backslashes so the Windows basename redaction actually redacts; the coverage denominator uses a new isExpectedFreeModel mirroring findUnpricedModels' exclusions; pricingCoverage is number | null with null for not-computable. The threshold duplication note is also fixed (shared constant). Remaining non-blocking notes (the 'still failing' CI-paste pattern, zero payload consumers) are acknowledged and left as-is. Thanks for the depth on this one; the Windows privacy catch in particular would have shipped.

PR #786, opened 2026-07-20
tui: Workflow panel, rotating coaching notes, titles in the context browser

iamtoruk:
Phase 1 for the terminal surfaces.

Workflow panel in the dashboard grid (paired with Claude Agent Types, existing responsive collapse): Corrections N% with count, First edit median, Rework top file with session count, Coverage N% priced. Reuses the workflow-insights functions over the already-parsed projects, no re-parse. Coverage computes null with no cost-bearing calls and the row drops entirely. Panel hides with no signal. Footer rotates one coaching note every 12s as a tip line, only when notes exist.

Context browser rows now lead with the captured title (untitled sessions read as untitled session), the 8-char id demoted into the dim metadata cluster with columns aligned; proven at 140 and 100 columns via pty drives on a real cache copy, with the real panel reading Corrections 0% (7) / First edit 6m / Rework sdk.py x15 / Coverage 95%.

18 new tests; typecheck green.

Wave-review note: surfaces currently differ on rounds-to-zero correction rates (app renders 0%, menubar hides, TUI renders 0% with count). Suggest settling one convention across the wave before merging, likely showing sub-1% as <1% (N).

Part of the phased surfacing wave; PRs only, no release.


ozymandiashh (commented):
Read through the panel and the tests, and traced the data path: `computeWorkflowPanelData` mirroring the report helpers (no re-parse), the coverage tally matching usage-aggregator's cost-bearing/unpriced shape (same `<synthetic>`/`isExpectedFreeModel` exclusions before `findUnpricedModels`, so the two figures agree by construction), and the null-coverage row dropping instead of rendering a hollow 100%. The `sessionPrimaryLabel` demotion of the id into the metadata cluster also reads much better at 100 columns. All solid.

Three things worth deciding before this lands, one of them cross-PR:

## 1. Merge-order collision with #863

#863 rewrites exactly this region of `dashboard.tsx` (427/177 churn): `DashboardContent` becomes an "eight panels retain source order" model with 1/2/3-column reflow and a 3/3/2 three-column arrangement, and `InteractiveDashboard` grows refresh/scroll state on the same lines this PR touches for `noteTick`. Whichever PR lands second has to re-place the Workflow panel as a ninth panel inside that ordering (3/3/3), re-anchor the tip line against the new scroll-offset handling, and re-resolve the `WorkflowInsights` pairing with `ClaudeAgentTypes`. Not an objection to either PR, but the two should be sequenced deliberately rather than left to whoever rebases last.

## 2. The rotation interval runs outside the dashboard view

`coachingNote` is computed only for `view === 'dashboard'`, but the `setInterval` at the `noteTick` effect keeps firing (and re-rendering via `setNoteTick`) in optimize/compare/day views where the tip line is not rendered. Cheap, but it is a spurious full-frame repaint every 12s in views that #863 is busy making repaint-free. Gating the effect on the view (or clearing the interval when the tip is hidden) closes it.

Related nit: `computeWorkflowPanelData` (inside `WorkflowInsights`) and `computeCoachingNotes` (in the parent) each run `scanUserCorrections` + `aggregateFileChurn` + `medianTimeToFirstEditMs` over the full project set, so every data refresh pays the scan twice. Computing once and deriving both would halve it.

## 3. The rounds-to-zero convention: +1 for `<1% (N)`

Your own real-data example makes the case: "Corrections 0% (7)" asserts zero corrections and then counts seven of them in the same cell. `<1% (N)` fixes the contradiction and keeps the count honest. I would also apply it to the menubar rather than keeping its hide-on-zero behavior, since "hidden" and "0%" are different claims and neither matches "<1% with a nonzero count".

One more edge for whichever convention wins: `formatCorrectionsValue` uses `Math.round`, so 0.4% renders as `0% (N)` today, while a genuine 0 with `userTurns > 0` renders `0% (0)`. Under `<1% (N)` the first becomes `<1% (N)` and the second stays `0% (0)`, which reads exactly right.

ozymandiashh (changes requested):
This head is not mergeable against current main and now needs a deliberate rebase after #863, which changed the same dashboard composition.

Beyond the real conflict in `src/dashboard.tsx`, please address:

- The 12-second interval continues rerendering in Optimize/Compare even though the coaching note is hidden there.
- `Math.round` can display the contradictory `0% (7)` for a nonzero rate below 0.5%; use a convention such as `<1% (N)`.
- The same expensive workflow scans are repeated independently for the panel and coaching notes.
- After #863, integrate Workflow explicitly into the new scrollable/responsive layout instead of resolving the conflict mechanically.

The branch tests (18/18) and typecheck pass on its old head, but they do not validate the now-required #863 integration. Please rebase, resolve these behaviors, and request a fresh review.

PR #787, opened 2026-07-20
models: resolve Fireworks-hosted models to friendly display names

### LINKED ISSUE #779: Fireworks-hosted models (accounts/fireworks/models/*) display as raw path/slug instead of friendly names
## Summary

Fireworks-hosted models (id form `accounts/fireworks/models/<slug>`) display as a
raw path/slug instead of a friendly name, unlike every other provider. In the
`codeburn` dashboard **By Model** panel they show as `fireworks/mode…` (truncated
raw path); in `codeburn models` they show as `glm-5p2`, `qwen3p7-plus`,
`kimi-k2p7-code`, `deepseek-v4-pro`, `deepseek-v4-flash`.

These arrive via the **OpenCode** provider (an `opencode`-format session log written
by a Fireworks-backed fleet driver).

## Environment

- codeburn 0.9.16 (Homebrew) — note the behavior is the same on current `main`.
- Provider: OpenCode; models hosted on Fireworks AI.
- macOS.

## Root cause

- `getShortModelName`'s path-style fallback returns the **last path segment**
  verbatim, and there is no `SHORT_NAMES` entry for these Fireworks slugs, so the
  bare slug leaks (`glm-5p2` rather than `GLM-5.2`).
- Separately, the dashboard **By Model** panel renders the cached `modelBreakdown`
  key directly (no `getShortModelName` pass), so an old/un-normalized cached key
  leaks the full raw path (`accounts/fireworks/models/…` → truncated `fireworks/mode`)
  until the session cache is rebuilt.

## Proposed fix

Display-only (pricing via `getModelCosts` is untouched, so no dollar amounts move):

1. `getShortModelName` path fallback: take the last path segment and **re-resolve
   it** through the normal name resolution, so a known slug earns a friendly name
   and an unmapped slug still falls through to the raw segment.
2. Add `SHORT_NAMES` entries for the Fireworks slugs (`glm-5p2` → `GLM-5.2`,
   `qwen3p7-plus` → `Qwen 3.7 Plus`, `kimi-k2p7-code` → `Kimi K2.7 Code`;
   `deepseek-v4-*` already resolve).
3. Route the dashboard **By Model** render through `getShortModelName` (matching
   `overview`'s "Top models") so already-cached raw keys normalize without a rebuild.

Verified locally against real data: `codeburn models` now shows GLM-5.2 / DeepSeek
v4 Pro / Kimi K2.7 Code / Qwen 3.7 Plus / DeepSeek v4 Flash with unchanged costs;
`tests/models.test.ts` extended (the existing last-segment fallback assertion is kept
for a still-unmapped id); full suite shows no new failures.

I have a branch ready and can open a PR if the approach looks right.

therickfactr:
## Summary

- Fireworks-hosted models (id form `accounts/fireworks/models/<slug>`, arriving via the OpenCode provider) were leaking the raw path/slug: `fireworks/mode…` in the dashboard **By Model** panel and `models/glm-5p2` in `codeburn models`. `getShortModelName`'s path fallback now takes the last segment and **re-resolves** it, so a known slug earns a friendly name (GLM-5.2, Qwen 3.7 Plus, Kimi K2.7 Code, DeepSeek v4 Pro/Flash) while an unmapped slug still falls through to the raw segment.
- **Display-only** — `getModelCosts` prices off the full id, so no dollar amounts move.
- The dashboard **By Model** panel now aggregates by the resolved display name (extracted to a pure `aggregateModelTotals`), so a mixed-vintage cache holding both the full `accounts/fireworks/models/<slug>` path (older sessions) and the bare slug (newer ones) **merges into one row** instead of rendering duplicates. This also aligns the panel with `modelEfficiency`, which is already keyed by `getShortModelName`.

Closes #779.

## Testing

- [x] `npm test` passes for the affected areas

Verified against real OpenCode/Fireworks sessions — `codeburn models` now shows
GLM-5.2 / DeepSeek v4 Pro / Kimi K2.7 Code / Qwen 3.7 Plus / DeepSeek v4 Flash with
**unchanged costs**. `tests/models.test.ts` extended (the last-segment fallback
assertion is kept for a still-unmapped id); new `tests/model-breakdown.test.ts`
covers the row-merge case. `tsc --noEmit` clean.

Note on the full suite: the only failures are two pre-existing flakes in
`app/electron/cli.test.ts` (`spawnSpecFor`) and `tests/cli-status-menubar.test.ts`
(menubar payload filter) that also fail on a clean `main` — unrelated to this change.


iamtoruk:
Reviewed in depth, and this is exactly the PR the issue thread hoped for. Security pass over every line: display-only, no network, environment, or process surface, sources and tests only, nothing touching CI. Correctness: the path-fallback re-resolution cannot recurse (segments carry no slashes) and the empty-segment edge is guarded; the new aggregateModelTotals merges mixed-vintage cache keys exactly as discussed, and your test covers the merge case; friendly names round-trip idempotently through the resolver so merged keys cannot drift; the dashboard now has zero direct getShortModelName calls, so no double resolution. Typecheck clean, your three touched suites green (163 tests) on the head, and the one full-suite failure was ours, not yours (a stale key-list assertion from an unrelated titles change, fixed separately in #789). Verdict: ready to merge. Thanks for the quality here, from the report to the doc comments.

therickfactr:
@iamtoruk, you are more than welcome. I don't like (or do) sloppy work unless there is no other option.

PR #799, opened 2026-07-23
kimicode: session discovery fixes + subscription quota in menubar

iamtoruk:
## What

**Kimi Code session visibility (CLI/dashboard)**
- Resolve `k3` / `k3-agent` / `k2d6-agent` model aliases to canonical names (`Kimi K3`, `Kimi K2.6`) so sessions stop showing raw aliases
- Discover sessions across all Kimi Code homes (CLI `~/.kimi-code` + desktop runtime), not just the first
- Accept `conv-*` / `ctitle-*` session directory naming alongside `session_*`
- Add the Kimi Code provider tab (with brand color) to the menubar

**Kimi Code subscription quota (menubar)**
- New `KimiSubscriptionService`: reads `~/.kimi-code/credentials/kimi-code.json` directly (no keychain copy) and polls `GET https://api.kimi.com/coding/v1/usages`, mirroring the existing Codex/Claude quota services
- Quota chip on the Kimi Code tab (Weekly + 5-hour windows), a Plan insight, and a Kimi tab in Settings with connect/disconnect
- Lenient decoding: string-or-number values, enum-style time units (`TIME_UNIT_MINUTE` 300 → 5-hour), `used` derived from `remaining` when absent, ISO-8601 or epoch reset times
- 429 backoff honoring `Retry-After`; expired tokens (15-min lifetime, CLI-owned refresh) surface as terminal with automatic recovery on the next tick after the CLI renews them
- Independent cadence anchor so Kimi-only setups respect the configured refresh interval instead of polling every payload tick
- Plan tab stays reachable for quota-capable providers even on days with no local usage

## Tests
- 8 new `KimiUsageParsingTests` fixtures, including one captured from the live API response
- Full Swift suite green (38/38); affected JS suites updated and passing

## Notes
- No OAuth self-refresh by design (CodexBar parity): the Kimi CLI owns token renewal
- Release menubar app untouched; verified live in the dev build

### DIFF
--- a/docs/providers/kimicode.md
+++ b/docs/providers/kimicode.md
@@ -8,13 +8,22 @@ MoonshotAI Kimi Code local session usage and tool activity.
 
 ## Where it reads from
 
-The provider reads `~/.kimi-code` by default and honors the Kimi Code CLI's `KIMI_CODE_HOME` environment variable. It scans:
+By default the provider scans every known Kimi Code runtime store:
 
 ```text
-$KIMI_CODE_HOME/sessions/wd_*/session_*/
+~/.kimi-code
+~/Library/Application Support/kimi-desktop/daimon-share/daimon/runtime/kimi-code/home   (Kimi desktop / IDE embedded runtime)
+```
+
+Setting `KIMI_CODE_HOME` (or passing a home override) narrows the scan to that single home. Inside each home it scans:
+
+```text
+$HOME/sessions/wd_*/<session-dir>/
 ├── state.json
 └── agents/<agent-id>/wire.jsonl
 ```
+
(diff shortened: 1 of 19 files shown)


iamtoruk:
Pushed a fix on top after an in-depth review of the reported flapping: Kimi tokens live ~15 min and only the CLI renews them, so .terminalFailure is the dominant steady state between CLI uses — and the Plan tab + tab-strip chip replaced a good last snapshot with the reconnect screen every cycle. Now terminal-with-data keeps the usage bars with a quiet 'login idle' caption (reconnect only when there is genuinely no data), the chip downgrades to stale instead of the red reconnect capsule, and any snapshot older than 10 minutes gets an 'as of <time>' stamp so a bar can never masquerade as current. Display decisions extracted into a pure KimiQuotaPresentation helper with 10 tests; Settings still explains the terminal reason. Full Swift suite green (155).

PR #800, opened 2026-07-23
fix(claude): discover Claude Desktop/Cowork sessions in Windows MSIX installs

### LINKED ISSUE #611: Feature Request: Support Claude Desktop / Cowork sessions on Windows
## Problem

CodeBurn currently supports Claude Desktop sessions on macOS (`~/Library/Application Support/Claude/local-agent-mode-sessions/`), but **not on Windows**, where the equivalent path is:

```
%APPDATA%\Claude\local-agent-mode-sessions\
```

Running `npx codeburn` on a Windows machine with active Cowork/Claude Desktop sessions shows zero usage — the provider doesn't scan the Windows data directory.

## Expected Behavior

CodeBurn should detect and parse Claude Desktop / Cowork session data on Windows, just as it does on macOS.

## Suggested Implementation

In the Claude Desktop provider, add the Windows path alongside the existing macOS one:

- **macOS**: `~/Library/Application Support/Claude/local-agent-mode-sessions/`
- **Windows**: `%APPDATA%\Claude\local-agent-mode-sessions\`
- **Linux** (if applicable): `~/.config/Claude/local-agent-mode-sessions/`

This is similar to how the Cursor provider already handles cross-platform paths.

## Environment

- **OS**: Windows 11
- **Node.js**: 20+
- **CodeBurn**: latest (via `npx codeburn`)
- **Claude Desktop**: active Cowork sessions present at `%APPDATA%\Claude\local-agent-mode-sessions\`

## Workaround Attempted

Setting `CLAUDE_CONFIG_DIR` to `%APPDATA%\Claude` — no effect, as the provider looks for `projects/` subfolder structure which doesn't match Cowork's `local-agent-mode-sessions/` layout.

ozymandiashh:
Fixes #611

## Problem

The Claude Desktop sessions resolver returned a single per-platform path. On Microsoft Store (MSIX) installs of Claude Desktop the session data is virtualized under:

```
%LOCALAPPDATA%\Packages\<Claude package>\LocalCache\Roaming\Claude\local-agent-mode-sessions
```

so codeburn showed zero Cowork usage. As @seb937 verified in #611 (comment), a directory junction workaround makes codeburn work but breaks Cowork's own file access inside active sessions, so the fix has to live in the resolver. This matters more now that #798 directs Windows installs to the Microsoft Store.

## Change

`getDesktopSessionsDir(): string` becomes `getDesktopSessionsDirs(): string[]` returning an ordered, deduped candidate list:

1. `CODEBURN_DESKTOP_SESSIONS_DIR` override (exactly that path, resolved, no probing)
2. macOS: `~/Library/Application Support/Claude/local-agent-mode-sessions`
3. Windows: classic `%APPDATA%\Claude\local-agent-mode-sessions` first, then MSIX candidates discovered by enumerating `%LOCALAPPDATA%\Packages` for directories named `Claude_*` or `*.Claude_*` (publisher-prefixed package family names), keeping only those whose sessions subpath exists, sorted lexicographically. Probing never throws; enumeration errors mean no MSIX candidates.
4. Linux: `~/.config/Claude/local-agent-mode-sessions`

All call sites scan every candidate: `isCoworkSession` classifies against all bases, `probeRoots` emits one `desktop` root per candidate, `discoverSessions` walks each base with its own `claude-desktop:<hash>` source id (project dirs deduped across bases), and the config selector treats desktop as present when any candidate exists.

Results are memoized per env-input tuple (platform, override, APPDATA, LOCALAPPDATA) so the parser's per-file classification never rescans the Packages directory.

Behavior on macOS, Linux, and classic Windows installs is unchanged. The package hash is never hardcoded.

## Testing

- 8 new resolver cases in `tests/providers/claude-config-dirs.test.ts` (30 total passing): MSIX discovery alongside classic, package-name filtering incl. near-miss rejections (`ClaudeXtra_foo`, `NotClaude_x`), missing `LOCALAPPDATA`/`Packages` robustness, override exact-return incl. unnormalized paths, non-win32 never probes Packages, missing sessions subpath excluded, deterministic ordering, resolved-duplicate dedupe, APPDATA fallback.
- `npx tsc --noEmit` clean; parser/aggregator consumer suites green (45/45); full CLI test tree green.

## Notes

- Known theoretical limitation: two MSIX packages whose sessions dirs are reparse points to the same underlying directory would not be deduped (`resolve()` does not follow junctions). Real Store packages use real directories, so this is not reachable in practice.
- Windows-native verification on an MSIX install would be appreciated - @seb937 if you can run this branch, `npx codeburn` should pick up your Cowork sessions with no junction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+- Claude Desktop and Cowork sessions are discovered for Windows Microsoft Store (MSIX) installs. (#611)
+
 ## 0.9.19 - 2026-07-20
 
 One version across every surface: CLI, macOS menubar, and the desktop app all ship as 0.9.19.
--- a/docs/providers/claude.md
+++ b/docs/providers/claude.md
@@ -12,10 +12,25 @@ Anthropic Claude Code CLI and Claude Desktop's local agent mode.
 |---|---|
 | Claude Code CLI | `$CLAUDE_CONFIG_DIR` if set, otherwise `~/.claude/projects/` |
 | Claude Desktop (macOS) | `~/Library/Application Support/Claude/local-agent-mode-sessions/` |
-| Claude Desktop (Windows) | `%APPDATA%/Claude/local-agent-mode-sessions/` |
+| Claude Desktop (Windows, classic) | `%APPDATA%/Claude/local-agent-mode-sessions/` |
+| Claude Desktop (Windows, MSIX) | `%LOCALAPPDATA%/Packages/<Claude package>/LocalCache/Roaming/Claude/local-agent-mode-sessions/` |
 | Claude Desktop (Linux) | `~/.config/Claude/local-agent-mode-sessions/` |
 
 For Desktop, `findDesktopProjectDirs` walks up to 8 levels deep looking for `projects/` subdirectories, skipping `node_modules` and `.git`.
+
+Desktop session roots are resolved in this order:
+
+1. A non-empty `CODEBURN_DESKTOP_SESSIONS_DIR` overrides discovery and is the
+   only returned root.
+2. macOS uses the single path shown above.
+3. Windows always includes the classic path first. It then scans
+   `%LOCALAPPDATA%/Packages` for package directories whose names start with
+   `Claude_` or contain `.Claude_`, sorted by package name, and includes only
+   packages whose full MSIX sessions path exists as a directory.
+4. Other platforms use the single Linux path shown above.
+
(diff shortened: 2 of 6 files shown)


ozymandiashh:
@iamtoruk this is ready to merge: real-world confirmation from @seb937 on a Windows 11 MSIX install (`Claude_pzs8sxrjxfjjc`) in https://github.com/getagentseal/codeburn/issues/611#issuecomment-5069814695 — all Cowork sessions detected, no junction needed, Cowork file access unaffected. Checks are green; it just needs a review approval.

Issue #404, opened 2026-05-27
No support of last version of kilocode


yoanhuret:
KiloCode ≥ 7.2.21 has migrated to a CLI backend that stores sessions in a SQLite database at:

C:\Users\HDDW7523\.local\share\kilo\kilo.db
The ui_messages.json file no longer exists for recent sessions. The latest tasks in globalStorage\kilocode.kilo-code\tasks\ date back to April 8, 2026 — this is the old architecture.

Codeburn 0.9.10 only supports the old format (ui_messages.json in globalStorage) for the KiloCode provider. It does not yet read ~/.local/share/kilo/kilo.db.

For comparison, Codeburn already supports OpenCode via SQLite (~/.local/share/opencode/opencode*.db) with the session, message, and share table schema. KiloCode uses the same scheme but in ~/.local/share/kilo/kilo.db.

iamtoruk:
Support added. KiloCode >= 7.2.21 now discovered from both the old VSCode globalStorage path (ui_messages.json) and the new SQLite backend at ~/.local/share/kilo/kilo.db. Same session/message/part schema as OpenCode so parsing, token extraction, tool normalization, and cost calculation all carry over.

Shipping in the next release.

yoanhuret:
Whaouh, so quick !
When is the fix will be available to setup ? Delivered ?
Is there a way to get it from source code ?

iamtoruk:
Already on main. I would be nice if you clone and check that out and let me know if that worked so I can publish in next version

yoanhuret:
I've just tested locally. It's OK with your fix !
Thanks

PR #801, opened 2026-07-24
fix(opencode): read session fallback model from the real schema

### LINKED ISSUE #769: opencode: SQLite token fallback selects model_id but the session column is model
Found during the 0.9.18 issue audit: tryQuerySessionTokens selects model_id from the session table (src/providers/sqlite-session-parser.ts:48) but the real OpenCode schema names the column model, so the fallback query throws against a real opencode.db. It only fires when per-message parsing yields zero tokens, which is why #675's fix works regardless, but the fallback is dead as written. One-line query fix plus a schema-accurate fixture test.

EuanTop:
## Summary

- query the real `session.model` column for SQLite token fallback
- decode the JSON model object into `providerID/id`
- invalidate stale OpenCode and KiloCode zero-result caches
- update the OpenCode fixture to match the real schema

## Context

#394 introduced the session-level fallback using `model_id`. #404 later introduced the shared OpenCode/KiloCode SQLite parser, and OpenCode was subsequently migrated onto it. Both current schemas store the session model as JSON in `model`.

The KiloCode parse-version bump only invalidates caches produced before this shared parser correction; normal caching remains enabled afterward.

Fixes #769.

## Testing

- targeted provider/cache tests: 113 passed
- `npm run build:cli`
- `npx tsc --noEmit`
- `git diff --check`

The full suite completed with 2449 passing tests, 5 skipped tests, and 2 existing unrelated failures in `tests/cache-refresh-lock.test.ts` and `tests/cli-durable-totals.test.ts`.


iamtoruk (approved):
Fixes #769. Clean, correct fix. The shared SQLite parser's session-level token fallback queried a `model_id` column that does not exist on the current OpenCode/KiloCode schema — the session model is stored as a JSON object in `model` — so the fallback silently produced no model. This now selects the real `model` column (cast to BLOB) and decodes the `{providerID, id}` JSON into `providerID/id`, returning undefined on a malformed/missing value.

The cache invalidation is right: `opencode` (which was absent from PROVIDER_PARSE_VERSIONS despite using the shared parser) gets a version key, and `kilo-code`'s is bumped, so stale caches parsed with the old column are re-derived once. Mutation-checked: the 'falls back to session-level tokens' test fails on the pre-fix parser and passes with it. On the merged tree (main was 68 ahead, clean): `tsc` clean, opencode/kilo-code/session-cache suites 110/110, full suite green apart from the two pre-existing `parser.test.ts` failures that reproduce on main. Good to merge.

PR #803, opened 2026-07-24
fix(mac): keep right-click status-item menu open and stable

### LINKED ISSUE #802: macOS menubar: right-click context menu flashes and/or scrolls away on mouse move
## Summary

On macOS (reproduced on **macOS 27 beta** with menubar **0.9.19**), right-clicking the CodeBurn status item either:

1. **Flashes** the context menu for a few ms then dismisses it, and/or
2. Opens the menu but **jumps/scrolls** on the first mouse move — the top "Today · $X · N calls" row slides away under a `^` chevron.

Settings / Refresh / Updates / About / Quit are effectively unreachable or hard to use.

Left-click → popover is unaffected.

## Environment

- macOS 27.0 beta (also relevant to the #472 macOS 27 right-click path)
- CodeBurn menubar `mac-v0.9.19` (includes #472)
- Apple Silicon

## Related history

| Change | What it did | Gap |
|--------|-------------|-----|
| #472 | Restored right-click on macOS 27 via global `rightMouseDown` monitor + `NSMenu.popUp` | Menu can **appear** but presentation timing/positioning is wrong |
| #509 | Polished menu; anchored `popUp` at `height + 6` to "clear the menu bar" | Scroll-on-hover still happens on 27; coordinate comment assumed non-flipped geometry |

There is **no open issue** tracking the flash or the jump after #472/#509.

### 1. Flash — present on `rightMouseDown`

#472's global monitor listens for **`rightMouseDown`**, then `DispatchQueue.main.async { popUp(...) }`.

`popUp` starts menu tracking immediately. The matching **`rightMouseUp`** is treated as an outside click → menu dismisses in milliseconds.

**Fix:** monitor **`rightMouseUp`** instead. Present only after the click completes.

- macOS ≤26: legacy button action already fires on `rightMouseUp`; global monitor also fires → existing **0.3s debounce** keeps a single present.
- macOS 27: legacy path still dead; global monitor on up is the only path (and no longer races mouse-up dismissal).

### 2. Jump — manual `NSMenu.popUp(at:in:)`

Even when the menu stays open, `popUp(positioning:at:in: button)` is a poor fit for status items:

- `NSStatusBarButton.isFlipped == true` (verified): `y = 0` is the **top** of the button, `y = height` is the **bottom**. #509's `height + 6` is the right *direction*, but manual popUp still tracks against a point while the cursor sits on the status item *above* the menu.
- First mouse move engages AppKit **scroll mode** → `^` chevron, first row ("Today · …") scrolls away.

**Fix:** stop using manual `popUp` for this menu. Use the status item's own menu slot so AppKit positions and tracks it:

```swift
menu.delegate = self
statusItem.menu = menu
button.performClick(nil)
// menuDidClose: statusItem.menu = nil  // restore left-click → popover action
```

#472 avoided `statusItem.menu + performClick` because assigning the menu from the right-mouse **action** never ran on 27. That is still true. We open from the **global rightMouseUp monitor** (or legacy action on ≤26), *then* attach `statusItem.menu` and `performClick`. At that point performClick is only "open the attached menu" and works on 27.

Clear `statusItem.menu` in `menuDidClose` so the next left-click hits the popover action again.

## Proposed change

Single file: `mac/Sources/CodeBurnMenubar/CodeBurnApp.swift`

1. Global monitor: `.rightMouseDown` → `.rightMouseUp`
2. `showContextMenu`: `statusItem.menu = menu; button.performClick(nil)` instead of `menu.popUp(...)`
3. Conform to `NSMenuDelegate`; `menuDidClose` clears `statusItem.menu`
4. Keep 0.3s debounce and legacy `rightMouseUp` action path for ≤26
5. Close an open popover before presenting the menu

## Testing

- [x] `swift build` / `mac/Scripts/package-app.sh` with full Xcode (SwiftUI macros)
- [x] Binary contains `mov w2, #0x10` (rightMouseUp mask) and `performClick`
- [x] **Manual on macOS 27:** right-click stays open; mouse move does not scroll; Today row stable; Settings/Quit work
- [x] **Manual on macOS 27:** left-click still opens popover after using the context menu
- [ ] **Smoke on macOS ≤26** (if available): right-click presents exactly once (debounce); no flash; no jump

## Workarounds (until released)

None in shipping 0.9.19. Local fix is built/installed as a dev menubar build from source.

## Notes for reviewers

- Do **not** "fix" the jump by switching to `y: 0` in view coords — button is flipped; that anchors into the menu bar and makes clipping worse.
- Screen-coordinate `popUp(in: nil)` is a reasonable fallback but still inferior to `statusItem.menu` for tracking.
- #472 author only verified on 27; ≤26 path relies on debounce — call that out in the PR the same way they did.

hyknerf:
## Summary

- Fixes the menubar right-click context menu on macOS 27 (and keeps ≤26 working): menu no longer **flashes** closed, and no longer **jumps/scrolls** the top "Today · …" row under a chevron on mouse move.
- **Flash:** global monitor presented on `rightMouseDown`; matching `rightMouseUp` dismissed the menu. Monitor is now `rightMouseUp`.
- **Jump:** manual `NSMenu.popUp(at:in:)` tracked poorly while the cursor sat on the status item. Open via `statusItem.menu` + `performClick` instead; clear the menu in `menuDidClose` so left-click still opens the popover.
- Closes #802.

## Context

#472 restored right-click on macOS 27 with a global monitor + `popUp`, but presentation on mouse-down caused the flash. #509 tried `height + 6` anchoring to avoid scroll; `NSStatusBarButton` is flipped and manual `popUp` still engaged scroll mode on mouse move.

#472 avoided `statusItem.menu` + `performClick` because the right-mouse *action* never runs on 27. We still open from the global monitor (or legacy action on ≤26); once `statusItem.menu` is set, `performClick` only means "open the attached menu" and works on 27.

## Test plan

- [x] `./mac/Scripts/package-app.sh dev` succeeds (Xcode 26.6 / Swift 6.3)
- [x] Built binary has global-monitor mask `0x10` (`rightMouseUp`) and references `performClick`
- [x] Installed dev build on macOS 27 beta; right-click opens menu (flash path fixed in earlier binary probe)
- [x] **Please verify on macOS 27:** right-click stays open; mouse move does not show `^` / scroll away Today row; Settings / Quit work
- [x] **Please verify on macOS 27:** after dismissing the menu, left-click still opens the popover
- [ ] **Smoke on macOS ≤26** if available: right-click presents exactly once (0.3s debounce); no flash; no jump

## Notes

- Single file: `mac/Sources/CodeBurnMenubar/CodeBurnApp.swift`
- No CLI / TS changes — `npm test` N/A for this diff
- ≤26 still relies on debounce if both legacy action and global monitor fire (same as #472)

### DIFF
--- a/mac/Sources/CodeBurnMenubar/CodeBurnApp.swift
+++ b/mac/Sources/CodeBurnMenubar/CodeBurnApp.swift
@@ -922,7 +922,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSM
         // Presenting on up (after the click completes) keeps it open. Harmless
         // on 15/26 too (the debounce in showContextMenu prevents a double-present
         // if the legacy path also fires).
-        rightClickMonitor = NSEvent.addGlobalMonitorForEvents(matching: [.rightMouseUp]) { [weak self] _ in
+        rightClickMonitor = NSEvent.addGlobalMonitorForEvents(matching: StatusItemContextMenuPolicy.presentEventMask) { [weak self] _ in
             guard let self,
                   let button = self.statusItem.button,
                   let window = button.window,
@@ -1139,9 +1139,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSM
     private func showContextMenu(from button: NSStatusBarButton) {
         // Debounce: on macOS <= 26 both the legacy action path and the global
         // monitor can fire for a single right-click. Present at most once per click.
-        let now = Date()
-        guard now.timeIntervalSince(lastContextMenuPresentedAt) > 0.3 else { return }
-        lastContextMenuPresentedAt = now
+        // Policy lives in StatusItemContextMenuPolicy so the gate is unit-tested (#802).
+        guard StatusItemContextMenuPolicy.acceptPresent(
+            now: Date(),
+            lastPresentedAt: &lastContextMenuPresentedAt
+        ) else { return }
 
         // Don't let an open popover steal the click / sit under the menu.
         if popover?.isShown == true {
--- a/mac/Sources/CodeBurnMenubar/CodeBurnApp.swift
+++ b/mac/Sources/CodeBurnMenubar/CodeBurnApp.swift
@@ -49,11 +49,14 @@ struct CodeBurnApp: App {
 }
 
 @MainActor
-final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate {
+final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSMenuDelegate {
     private var statusItem: NSStatusItem!
     private var popover: NSPopover!
     private var rightClickMonitor: Any?
     private var lastContextMenuPresentedAt: Date = .distantPast
+    /// Held only while the right-click menu is open. Cleared in menuDidClose so
+    /// left-click goes back to the popover action instead of re-showing the menu.
+    private var contextMenu: NSMenu?
     fileprivate let store = AppStore()
     let updateChecker = UpdateChecker()
     /// True while the displays are asleep. Refresh ticks skip spawning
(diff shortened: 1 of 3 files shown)


hyknerf:
Added unit tests for the extractable policy behind this fix:

- `StatusItemContextMenuPolicy` — event mask (`rightMouseUp`), presentation (`.statusItemMenu` not `.manualPopUp`), debounce gate
- `StatusItemContextMenuPolicyTests` — 7 cases, all green via `swift test --filter StatusItemContextMenuPolicyTests`

AppKit open/track/dismiss still needs a manual smoke (no good headless seam for `NSStatusItem` + real menu tracking). The tests lock the decisions that caused the regressions so they can’t silently flip back.

iamtoruk:
Merged — thank you for an exemplary contribution: researched issue first, surgical scope, the policy extraction with tests locking the anti-pattern, and clear archaeology of #472/#509. The fix ships with the next menubar release (mac-v0.9.20). Hope to see you around again.

PR #846, opened 2026-07-27
fix(omp): discover title-first session transcripts

### LINKED ISSUE #845: OMP title-slot metadata prevents session discovery
CodeBurn’s (v0.9.19) shared Pi/OMP JSONL discovery gate requires the first physical line of a transcript to be a `{"type":"session"}` record.

Oh My Pi can write a fixed-width title metadata slot before that header:

 ```json
   {"type":"title","v":1,"title":"…","updatedAt":"…","pad":"…"}
   {"type":"session","version":3,"id":"…","cwd":"…"}
 ```

 As a result, valid OMP transcripts were rejected during discovery and omitted from:

 ```sh
   codeburn sessions --provider omp
 ```

### Upstream context

 The earliest public upstream source found for this physical title-slot format is seemingly https://github.com/can1357/oh-my-pi/commit/0ce330ab7910c8fd0895fad1ceb9f06a902a74c9

 - Authored: 2026-06-27 13:13 UTC
 - Committed: 2026-06-28 05:27 UTC
 - Introduced SESSION_TITLE_SLOT_ENTRY_TYPE = "title" and fixed-width first-line title metadata.

 ### Expected behaviour

 Discovery skips blank lines, malformed records, and non-session metadata, then accepts the transcript when it finds a type: "session" record.

 ### Acceptance criteria

 - OMP title-first JSONL transcripts are discovered.
 - Blank lines before the session record do not prevent discovery.
 - Message-only files with no session record remain excluded.
 - Pi and OMP retain identical shared discovery behaviour.

jbspeakr:
## Summary

Fixes https://github.com/getagentseal/codeburn/issues/845 by changing shared Pi/OMP JSONL discovery to find the first `type: "session"` record rather than requiring it on the first physical line. 

OMP may store mutable session titles as a fixed-width type: "title" record before the JSONL session header. CodeBurn previously rejected these otherwise valid transcripts before parsing.

### Changes

 - Replaced first-line-only session detection with a JSONL scan for the first session record.
 - Skips blank lines, malformed JSON, and non-session metadata—including OMP title slots.
 - Preserves the existing readSessionFile() read-size and error behavior.
 - Adds a regression fixture with a title record and blank line before the session header.
 - Renames Pi/OMP negative-test wording: files are rejected only when they contain no session record.

## Testing

- [x] `npm test` passes

### For new providers only:


iamtoruk (approved):
Fixes #845. Clean, well-scoped fix: `readSessionEntry` now scans the JSONL for the first `type: "session"` record (skipping blank lines, malformed JSON, and non-session metadata like OMP's `type: "title"` slot) instead of requiring it on the first physical line. Since `readSessionFile` reads the whole file up to its size cap, the scan reaches the header regardless of how many title/blank lines precede it, and oversize files were already skipped before this change, so there is no read-size regression.

No regression for Pi or ordinary OMP transcripts (session on line 1 is found immediately); files with no session record are still rejected. Mutation-checked: the new "discovers title-first sessions" test fails on main and passes with the fix; the other discovery cases pass both ways. On the merged tree (main was 37 ahead, clean): `tsc` clean, omp/pi provider suites 31/31, full suite green apart from the two pre-existing `parser.test.ts` failures (and a cache-refresh-lock heartbeat test that flakes under parallel load and passes in isolation, untouched by this change). Good to merge.

PR #853, opened 2026-07-28
fix(parser): range-filter calls inside turns instead of dropping day-spanning turns

### LINKED ISSUE #852: Usage after local midnight vanishes when a turn spans days (turn-level range filter drops post-midnight calls)
**Summary:** In `parseProviderSources` (src/parser.ts), turns are range-filtered by `turn.calls[0]?.timestamp`. For long-running sessions whose turn spans midnight (common with autonomous Codex runs lasting many hours), all calls after midnight are dropped from the new day — `today` shows nothing until the turn ends, and daily history attributes the whole turn's cost to the turn-start day.

**Repro (synthetic rollout):**
```jsonl
{"timestamp":"2026-07-28T15:55:00.000Z","type":"session_meta","payload":{"session_id":"x","originator":"codex_cli_rs","cwd":"/tmp/t","model":"gpt-5.5"}}
{"timestamp":"2026-07-28T15:57:00.000Z","type":"response_item","payload":{"type":"message","role":"user","content":[{"type":"input_text","text":"work through midnight"}]}}
{"timestamp":"2026-07-28T15:58:00.000Z","type":"event_msg","payload":{"type":"token_count","info":{"total_token_usage":{"input_tokens":100000,"cached_input_tokens":50000,"cache_write_input_tokens":0,"output_tokens":2000,"reasoning_output_tokens":0,"total_tokens":102000},"last_token_usage":{"input_tokens":100000,"cached_input_tokens":50000,"cache_write_input_tokens":0,"output_tokens":2000,"reasoning_output_tokens":0,"total_tokens":102000}}}}
{"timestamp":"2026-07-28T16:10:00.000Z","type":"event_msg","payload":{"type":"token_count","info":{"total_token_usage":{"input_tokens":200000,"cached_input_tokens":100000,"cache_write_input_tokens":0,"output_tokens":4000,"reasoning_output_tokens":0,"total_tokens":204000},"last_token_usage":{"input_tokens":100000,"cached_input_tokens":50000,"cache_write_input_tokens":0,"output_tokens":2000,"reasoning_output_tokens":0,"total_tokens":102000}}}}
```
In timezone UTC+8: at local 2026-07-29 00:30, `codeburn today` reports "No usage data found", although the second `token_count` (2026-07-29 00:10 local) is squarely inside today. `export --from 2026-07-29` likewise returns 0 calls. Same-day control case (user message repeated before the second event) works fine, which isolates the cause to turn-level filtering.

Real-world case: a Codex session running autonomously for 30+ hours (one turn, last user message at 23:57 local) — all post-midnight usage (~150 calls) vanished from `today`.

**Suggested fix:** filter calls inside the turn instead of dropping the whole turn:
```ts
if (dateRange) {
  const inRangeCalls = turn.calls.filter((c) => {
    const ts = new Date(c.timestamp);
    return !Number.isNaN(ts.getTime()) && ts >= dateRange.start && ts <= dateRange.end;
  });
  if (inRangeCalls.length === 0) continue;
  turn = { ...turn, calls: inRangeCalls };
}
```
(applies to both the `servedSources` loop and the `durableSources` loop in parseProviderSources)

Optionally, in `aggregateProjectsIntoDays`, attribute per-call cost/token columns to `dateKey(call.timestamp)` rather than the turn day, so multi-day turns distribute cost across actual days in daily history. Turn-level stats (categories, one-shot) can stay turn-based.

I've verified both changes locally against the repro and a real 51MB rollout: post-midnight usage shows up in `today` correctly, monthly totals unchanged.

Happy to open a PR with this if you prefer.

KENSHI601:
Fixes #852.

**Bug:** `parseProviderSources` range-filtered turns by the turn's *first* call timestamp (`turn.calls[0]?.timestamp`). A long autonomous turn that starts before midnight (e.g. a 30h Codex run) was therefore dropped wholesale from the next day's view — `today` / single-day exports lost every post-midnight call in the turn.

**Fix:** filter calls inside the turn against the range instead. A turn is skipped only when none of its calls fall in range; otherwise it is kept with just the in-range calls, and downstream derivation (classified turn, project, sessionId, projectPath) uses the filtered turn. Applied to both identical filter blocks (served-sources loop and durable-orphans loop) in `src/parser.ts`.

**Tests:**
- New regression test in `tests/provider-turn-grouping.test.ts`: a synthetic Codex rollout whose single turn has `token_count` events straddling midnight; querying the next day's range now keeps the turn with only the post-midnight call. Verified the test fails without the fix and passes with it.
- `tsc --noEmit` and `tsup` build are green; parser-related suites (parser, session-cache, codex, date-range, export, durable-totals) pass. Full suite diffed against `main`: identical failure sets (pre-existing Windows-environment issues — missing jsdom, 5s-timeout flakes, `os.homedir()` fixtures), no new failures.


KENSHI601:
@iamtoruk Thanks for the thorough review on #857 — this update (7591e68) lands the conservation-correct version here, point by point:

1. **Conservation (blocker)** — `aggregateProjectsIntoDays` now buckets cost/savings/calls/tokens (and the model, project, and provider-slice rollups built from them) under each call's own day. Turn-level stats (categories, editTurns, oneShotTurns) stay turn-anchored, with the both-sides-of-a-split semantics documented in code. Your repro now reads today 1/$0.012, week/month 2/$0.018, with day-N + day-N+1 == whole range.
2. **Re-anchor** — sliced turns get `timestamp` = first surviving call in both `parseProviderSources` loops, adopting #857's split rule (`callsInRange` / `turnSlicedToRange` / `classifiedTurnSlicedToRange`, same null-on-empty contract).
3. **scanProjectDirs** — slices per call instead of whole-turn drop; `category`/`subCategory`/`retries`/`hasEdits` stay classified from the full turn, per avs-io's approach.
4. **Regression test** — added the straddling-turn `buildDurablePeriod` case to the durable-totals parity suite: day-N + day-N+1 == whole-range calls/cost/tokens, plus a today-only view case on the Claude path and surface-filter cases. Verified all three fail without the fix (including on this PR's previous revision) and pass with it.
5. **Surfaces** — `filterProjectsByDateRange` (dashboard) and `filterProjectsByDays` (menubar/history) now slice per call too, so every surface reports the same Today number; `buildJsonReport`'s dailyMap fallback follows the same per-call rule so it can't diverge from `durable.days`. `turnIsInDateRange`, orphaned by this change, is removed; `parseSessionFile` left untouched as agreed.

Verification: `tsc --noEmit` clean, `tsup` build ok, and the parser / day-aggregator / durable-totals / session-cache / codex / export suites are green — the durable-totals file is now also hermetic on machines with real provider data (USERPROFILE / KIMI_CODE_HOME / desktop-sessions redirects in beforeEach), which is what its 5s-timeout flakiness on this box actually was. Full-suite diff against the previous revision shows no new failures; remaining red is pre-existing environment noise (missing jsdom, homedir-based fixtures, load-sensitive timeouts).

iamtoruk (approved):
Reviewed on 6c411ad. This is the conservation-correct fix for #852, and I verified it end to end.

**Conservation (the hard part) holds.** Driven through the real `buildDurablePeriod` cache+period path with a turn straddling local midnight (call1 23:58 $0.006, call2 00:10 $0.012): today = 1 call/$0.012, week = month = 2 calls/$0.018, days split correctly (07-28: $0.006, 07-29: $0.012). No call is lost or double-counted. The per-call bucketing in `day-aggregator.ts` and `main.ts` (the layers a parse-only slice misses) is what makes the union exact.

**Cross-surface parity holds.** `turnIsInDateRange` is gone and `filterProjectsByDateRange` (dashboard) and `filterProjectsByDays` (menubar) now slice per-call, so headline / dashboard breakdown / menubar all agree (today 1, week 2). An adversarial pass also confirmed: cost/calls/tokens conserve across single-midnight, two-midnight, DST, and malformed-timestamp cases; the spawn-anchor change is a genuine fix (it recovers parent spend main was dropping); and the same-day common case is byte-identical to main.

**One thing I fixed and pushed (5a12dcc).** The provider path classified a sliced turn from only its in-range calls (`cachedTurnToClassified(slicedTurn)`) while the Claude path kept the full-turn classification, so a straddling turn's category/`hasEdits` differed by provider — contradicting this PR's own comment that those are whole-exchange judgments. I changed the two provider loops to classify the full turn then trim (mirroring the Claude path); cost/calls are unchanged, only the turn-level judgments are now provider-independent. Added a regression test (`provider-turn-grouping`, mutation-checked: it fails on the pre-fix classification and passes now). Also merged main to resolve the 7-commit lag (clean).

The `editTurns` sum-across-disjoint-reports inflation is documented/accepted and cost-safe (the persisted durable path parses the window once and doesn't inflate). `tsc` clean; full suite green apart from the two pre-existing `parser.test.ts` failures that reproduce on main. Merging this and closing #857 in its favor.

PR #857, opened 2026-07-29
parser: slice midnight-straddling turns instead of dropping them

avs-io:
> **Overlaps #853, which came first.** @KENSHI601 opened #853 on 2026-07-28 fixing this defect in the two `parseProviderSources` loops, and the diagnosis and repro here are theirs.
>
> This branch is a **superset** of #853: it carries the same two-loop fix plus the Claude Code path and the day-aggregator reconciliation described below. Either this lands and #853 closes as included, or #853 lands and the delta comes as a follow-up. Maintainer's choice — flagging the overlap rather than quietly duplicating it.

Relates to #852.

## The defect

Both `parseProviderSources` (Codex, OTel, network providers) and `scanProjectDirs` (Claude Code) range-filtered a whole turn by its first call's timestamp. A turn spanning local midnight — common for long autonomous sessions — had every call after the boundary silently dropped from the new day: `codeburn today` reported no usage until the turn ended, and daily history attributed the whole turn's cost to the turn-start day.

## Delta over #853

#853 fixes the two `parseProviderSources` loops. This branch additionally carries:

**1. The Claude Code path.** `scanProjectDirs` has the structurally identical whole-turn drop, filtering on `turn.assistantCalls[0].timestamp`. Claude Code is the highest-traffic provider in this tool, so the reported symptom persists for most users until this path is fixed too. Handled by `classifiedTurnSlicedToRange`, which slices `assistantCalls` and leaves `category`/`subCategory`/`retries`/`hasEdits` as classified from the complete turn — they are computed once ahead of the date filter, for `carriedBranch`/PR-ref tracking reasons already in the code, and are turn-level judgments rather than per-call sums.

**2. A `timestamp` re-anchor on a sliced turn.** `day-aggregator.ts` buckets by `turn.timestamp` rather than per-call timestamps — deliberately, per `ae18b1f`. Without re-anchoring the sliced turn to its first surviving call, the retained post-midnight calls are still filed under the pre-slice day, so they reappear in `parseAllSessions` output but remain invisible in `codeburn today`, which is the reported symptom. Verified against `getDateRange('today')` and `aggregateProjectsIntoDays` under a simulated UTC+8 clock.

## Fold-anchor safety

The PR-linked subagent fold anchor — a parent whose own turns fall entirely out of range is kept as a zero-cost anchor so an in-range child can still attribute — is unchanged. `carriedBranch`, `carriedPrRefs` and `spawnPrSets` are built from `cachedFile.turns`, the full pre-slice list, before the date filter runs. A slice yields a non-empty call list only when the turn genuinely has an in-range call, so a fully out-of-range turn still filters to nothing. The anchor test passes identically before and after the change, which is the correct obligation: the mechanism was not broken and must not break.

## Testing

Two regression files built from @KENSHI601's synthetic repro, made deterministic across machine timezones rather than dependent on the runner's local zone: `tests/parser-turn-midnight-range.test.ts` (provider path) and `tests/parser-claude-turn-midnight-range.test.ts` (Claude Code path).

Each covers: a turn spanning the boundary attributes post-boundary calls to the new day and pre-boundary calls to the prior day; a same-day control turn is unchanged; a turn entirely outside the range is still excluded; the unfiltered monthly/total aggregate does not double-count; and `aggregateProjectsIntoDays` with `getDateRange('today')` reconciles under a simulated UTC+8 clock. The Claude Code file adds the fold-anchor case.

All confirmed failing before the change. Full suite 2475 passed, `tsc --noEmit` clean.

## Out of scope

- The issue's optional per-call attribution in `aggregateProjectsIntoDays` (`dateKey(call.timestamp)` instead of the turn day) is deliberately not included. Once turns are sliced per range, a day-scoped query already sees only that day's calls. Reintroducing raw per-call bucketing on a wide unsliced parse — the common path for history and menubar rollups — would resurrect the reconciliation bug `ae18b1f` fixed, where a midnight-straddling turn split across two days and `history.daily` stopped reconciling to the live headline.
- `filterProjectsByDateRange` / `turnIsInDateRange`, the dashboard's in-memory re-narrowing over an already-wide parse, has the same shape but operates on turns already merged into `ProjectSummary`. Not covered here; a candidate follow-up.

## Credit

Root-cause diagnosis, the synthetic repro, and the fix shape are @KENSHI601's, in issue #852 and PR #853.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -2200,12 +2200,12 @@ async function scanProjectDirs(
     const spawnPrSets = cachedFile.prLinks?.length ? buildSpawnPrSets(cachedFile.turns) : {}
 
     if (dateRange) {
-      classifiedTurns = classifiedTurns.filter(turn => {
-        if (turn.assistantCalls.length === 0) return false
-        const firstCallTs = turn.assistantCalls[0]!.timestamp
-        if (!firstCallTs) return false
-        const ts = new Date(firstCallTs)
-        return ts >= dateRange.start && ts <= dateRange.end
+      // Slice rather than drop: a turn spanning local midnight would otherwise
+      // lose every call that lands in the requested day (issue #852). Only
+      // `assistantCalls`/`timestamp` are touched — see classifiedTurnSlicedToRange.
+      classifiedTurns = classifiedTurns.flatMap(turn => {
+        const sliced = classifiedTurnSlicedToRange(turn, dateRange)
+        return sliced ? [sliced] : []
       })
     }
 
@@ -2786,6 +2786,48 @@ export function createScanProgress(label: string, total: number) {
   }
 }
 
+// Shared by both turn-range slicers below: which of a turn's calls actually
+// fall inside dateRange. Returns null when none do (turn should be dropped
+// entirely, not kept with an empty call list).
+function callsInRange<T extends { timestamp: string }>(calls: T[], dateRange: DateRange): T[] | null {
+  const inRange = calls.filter(c => {
+    const ts = new Date(c.timestamp)
+    return !Number.isNaN(ts.getTime()) && ts >= dateRange.start && ts <= dateRange.end
+  })
+  return inRange.length > 0 ? inRange : null
+}
+
+// A turn can span local midnight (e.g. a long-running autonomous Codex
+// session): dropping the whole turn because its FIRST call falls outside
+// dateRange would discard every later call that lands in the requested day.
+// Instead, keep only the calls actually inside the range so each day gets the
+// calls that belong to it. Returns null when no call in the turn is in range.
+// `timestamp` is re-anchored to the first surviving call so day-aggregator's
+// turn-anchored bucketing (which keys off `timestamp`, not per-call
+// timestamps) buckets the slice under the day its retained calls actually
+// fall in, rather than the pre-slice turn's original (possibly prior-day)
+// start.
+function turnSlicedToRange(turn: CachedTurn, dateRange: DateRange): CachedTurn | null {
+  const inRangeCalls = callsInRange(turn.calls, dateRange)
+  if (!inRangeCalls) return null
+  if (inRangeCalls.length === turn.calls.length) return turn
+  return { ...turn, calls: inRangeCalls, timestamp: inRangeCalls[0]!.timestamp }
+}
+
+// Same slice, applied post-classification (scanProjectDirs classifies every
+// turn from its FULL call list up front, before date filtering — see the
+// carriedBranch/carriedPrRefs comments in scanProjectDirs — so this only
+// trims `assistantCalls` and re-anchors `timestamp`; `category`/`subCategory`/
+// `retries`/`hasEdits` stay exactly as classified from the complete turn.
+// Those are turn-level judgments about the whole exchange, not a per-call
+// sum, so they aren't recomputed from the partial call list.
+function classifiedTurnSlicedToRange(turn: ClassifiedTurn, dateRange: DateRange): ClassifiedTurn | null {
+  const inRangeCalls = callsInRange(turn.assistantCalls, dateRange)
+  if (!inRangeCalls) return null
+  if (inRangeCalls.length === turn.assistantCalls.length) return turn
+  return { ...turn, assistantCalls: inRangeCalls, timestamp: inRangeCalls[0]!.timestamp }
+}
+
 async function parseProviderSources(
   providerName: string,
   sources: SessionSource[],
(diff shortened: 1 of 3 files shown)


iamtoruk (changes requested):
The `today` fix is genuine, but as it stands this trades a visible bug for a silent one: **periods longer than a day lose usage.** Requesting changes on a proven conservation violation.

**The blocker, reproduced end-to-end through `buildDurablePeriod`** (the shared headline builder behind CLI/TUI/menubar): turn starting 07-27 23:57 with calls at 23:58 ($0.006) and 00:10 ($0.012). Truth for week/month = 2 calls / $0.018.

| period | main | this PR |
|---|---|---|
| today | 0 / $0 (the bug you're fixing) | 1 / $0.012 ✓ |
| week | 2 / $0.018 ✓ | **1 / $0.006** |
| month | 2 / $0.018 ✓ | **1 / $0.006** |

Mechanism: the backfill parse now slices the turn at the `backfillStart..yesterdayEnd` boundary, so the cached yesterday keeps only pre-midnight calls — but `aggregateProjectsIntoDays` stays turn-anchored, so in the period parse the whole turn buckets under yesterday and the today bucket gets zero. `unionDaysForPeriod` = cache (≤ yesterday) + live (today): the post-midnight calls are in neither. Per-call slicing and per-turn day bucketing are incompatible — the "out of scope" note on `aggregateProjectsIntoDays` is exactly where the gap opens. On a real month of data: 12 straddling turns, $8.98 silently gone.

**Second blocker — half-applied across surfaces**: `filterProjectsByDateRange` (dashboard Today/7-Days) and `filterProjectsByDays` (menubar/history) still whole-turn-drop, so the same data yields different Today numbers per surface — the exact divergence class the durable-totals parity suite exists to prevent (which has no midnight-straddling case; one should land with whichever PR carries this fix).

Also worth stating: turn-level aggregates (`editTurns`, categories, workflow) now report on BOTH sides of a split — a turn whose only Edit happened entirely on day N reports editTurns=1 on day N+1 too. If that's the intended semantics it needs documenting; summed-across-days turn counts inflate.

Given your own note that #853 should take precedence: agreed — suggest closing this in favor of landing the conservation-correct version once, there. Two things to carry over: (a) #853 currently does NOT re-anchor `timestamp` after slicing, so the two implementations diverge on the split rule — whichever survives should pin down one rule; (b) #853 doesn't touch `scanProjectDirs`, so its exposure at the cache-backfill boundary is structurally the same question and needs the same regression test (a straddling-turn `buildDurablePeriod` case asserting day-N + day-N+1 == whole-range for calls, cost, and tokens — I have a failing test ready to contribute).

Side-finding while auditing: `parseSessionFile` (~:1832-1846) is dead code carrying a fourth copy of the old filter and the historically-important "orphan assistant calls" comment — worth a separate cleanup so the next person doesn't have to reconcile four filter sites.

iamtoruk (changes requested):
Re-reviewed 1643aa9. The `today` fix is real and the re-anchor is a reasonable idea, but I have to keep this at changes-requested: reproduced end to end, the current commit does not fix the week/month conservation I flagged before, and it now regresses it below `main`. Requesting changes on a proven data-loss.

## Blocker 1 (still open, now a regression): week/month silently lose the post-midnight call

Driven through the real `buildDurablePeriod` path (the shared headline behind `codeburn today/week/month`, the TUI, and the menubar): daily cache on disk for the historical range, unioned with a live `parseAllSessions` for today. Faked clock at local `2026-07-29 00:30` (UTC+8). Fixture is the same straddling turn as last time: one turn, call1 `2026-07-28 23:58` ($0.006), call2 `2026-07-29 00:10` ($0.012). Truth: today = 1 call / $0.012; week/month = 2 calls / $0.018.

| surface | this PR | main | truth |
|---|---|---|---|
| today | 1 call / $0.012 (fixed) | 0 / $0 (the #852 bug) | 1 / $0.012 |
| week (7d) | **1 call / $0.006** | 2 / $0.018 | 2 / $0.018 |
| month | **1 call / $0.006** | 2 / $0.018 | 2 / $0.018 |

`main` counted week/month correctly; this PR drops call2 ($0.012) from both. It trades a visible bug (today under-reports) for a silent one (week/month under-report), on the surfaces most people read.

Why the re-anchor does not reach this: the re-anchor only moves the day when a turn is actually sliced, i.e. when part of it falls outside the range. A week/month **live** parse spans the whole turn, so nothing is sliced, and the intact turn buckets under `turn.timestamp`'s day (`day-aggregator.ts` buckets per turn, not per call). So `todayAllDays` is empty for this turn, and `unionDaysForPeriod` takes historical-from-cache (`2026-07-28` = call1 only) + today-from-live (empty) → call2 lands in neither. The day-rows the harness printed confirm it: this PR's week is `2026-07-28: 1c/$0.006`, with `2026-07-29` absent.

Scope: a turn straddling an **interior** week boundary (both days inside one cache parse) conserves on both this PR and main. The loss is specific to the **yesterday/today cache seam**, so it recurs for whichever turn straddled the last local midnight, in every period that includes today.

## Blocker 2 (still open): the parse path slices but the four post-parse filters do not

`turnIsInDateRange` (`parser.ts:3198`) still filters a whole turn by its first call, and it feeds both `filterProjectsByDateRange` (`:3533`, dashboard Today/7-Days and `codeburn export`) and `filterProjectsByDays` (`:3322`, menubar day view). Parsing the same straddling turn over a broad range and then filtering to today:

| path (Today) | this PR | main |
|---|---|---|
| sliced headline (`buildDurablePeriod`) | 1 / $0.012 | 0 / $0 |
| `filterProjectsByDateRange` (dashboard breakdown, export) | 0 / $0 | 0 / $0 |
| `filterProjectsByDays` (menubar) | 0 / $0 | 0 / $0 |

On this PR the header reads $0.012 while the dashboard project breakdown, `codeburn export`, and the menubar all read $0 for the same Today. On the dashboard specifically the week header ($0.006) also contradicts its own period breakdown ($0.018). `main` was internally consistent (all agreeing, even if wrong per #852); this PR makes surfaces disagree.

## Also (not blocking on their own, but part of the same root cause)

- **Turn-count double-count**: `classifiedTurnSlicedToRange` keeps `hasEdits`/`category` from the full turn, so a turn whose only Edit was in call1 reports `editTurns=1` on day N **and** day N+1. Summed across days that is 2 for one turn. The Codex/OTel path instead re-classifies from the sliced calls (`cachedTurnToClassified(slicedTurn)`), so the two providers now attribute edits/categories to slices by different rules.
- **Dead code**: `parseSessionFile` (`parser.ts:1790`) still carries a fourth copy of the old whole-turn filter (`:1838`). It has no callers and is unexported, so no runtime impact, but it re-introduces #852 the moment anyone rewires it. Worth deleting in whichever PR settles this.

## Root cause and suggested direction

Slicing the three parse loops is only one of the places the day rule lives. As long as `day-aggregator` buckets per turn, `unionDaysForPeriod` splits today-from-live vs historical-from-cache, and the four post-parse filters drop per turn, a sliced parse and the aggregate/filter/cache layers disagree at the yesterday/today seam. The fix needs one consistent per-call day rule across all of them, not a slice bolted onto turn-anchored bucketing.

Concretely, whichever version lands should carry a single regression test that asserts, through `buildDurablePeriod`, that for a straddling turn `day-N + day-(N+1) == whole-range` for calls, cost, and tokens (and, if turn counts are meant to be per-day, that they do not sum above the real turn count). This is the assertion that would have caught the table above; the current tests exercise `aggregateProjectsIntoDays`/`getDateRange` directly and never hit the cache+period path where the loss happens. I have that failing test ready to contribute.

As before, given the overlap with #853, my suggestion is to settle one conservation-correct rule once rather than iterate slices on top of turn bucketing. Happy to re-review quickly once week/month conserves and the surfaces agree.

iamtoruk:
Superseded by #853, which merged in b94d0ac. #853 fixes the same #852 defect but is conservation-correct end to end (week/month keep the post-midnight call — verified through `buildDurablePeriod`) and slices all four post-parse filter surfaces, so headline/dashboard/menubar agree. The two blockers on this branch (week/month data loss, cross-surface divergence) are resolved there, plus the provider-vs-Claude classification split. Closing in favor of #853 — thanks for the diagnosis and the repro here, which carried over.

PR #859, opened 2026-07-29
pi/omp: discover transcripts whose session record follows a title slot

avs-io:
> **This overlaps #846, which came first.** @jbspeakr — who reported #845 — opened #846 fixing the same defect. This branch was built independently from the issue before #846 surfaced in a PR sweep. **#846 should have precedence.**
>
> Left open as a draft only for the one difference described below. If that difference is not wanted, close this in favour of #846.

Relates to #845.

## The defect

The shared Pi/OMP JSONL discovery gate inspected only the first physical line of a transcript for a `type: "session"` record. Oh My Pi writes a fixed-width `type: "title"` metadata line ahead of that header, so valid OMP transcripts were excluded from `codeburn sessions --provider omp`. Both this branch and #846 scan past leading non-session lines to fix it, and both keep Pi and OMP on the identical shared path.

## Difference from #846

**Read cost.** #846 keeps `readSessionFile`, which reads the entire transcript into memory (up to the 128MB cap) and then iterates its lines. This branch streams through `readSessionLines` — the helper `codex.ts`, `droid.ts` and others already use — and stops after 20 leading lines.

The distinction matters most for the negative case. A message-only file with no session record is read in full under both the old code and #846; under a bounded scan it costs 20 lines. Session directories accumulate such files, and discovery walks all of them on every run.

Worth noting the pre-existing code only *looked* bounded: it read the whole file and then used line 0. So this is not a regression #846 introduces, it is an existing cost neither the old code nor #846 avoids.

One semantic change here: `readFirstEntry` now returns only session entries rather than whatever line 0 parsed to. The caller's `first.type !== 'session'` guard becomes redundant but is left in place rather than widening the diff.

## Acceptance criteria

- [x] OMP title-first JSONL transcripts are discovered
- [x] Blank lines before the session record do not prevent discovery
- [x] Message-only files with no session record remain excluded
- [x] Pi and OMP retain identical shared discovery behaviour

## Testing

Fixture tests in `tests/providers/pi.test.ts` reproducing a title-slot-first OMP transcript, confirmed failing before the change. Also covered: blank-line tolerance, malformed leading JSON handled without throwing, message-only exclusion, a session record placed beyond the scan bound (which must *not* be found, proving the scan is bounded rather than full-file), and a large message-only file completing well inside a generous time ceiling.

Full suite 2471 passed, `tsc --noEmit` clean.

## Credit

@jbspeakr reported #845 and traced the format to its upstream origin (`can1357/oh-my-pi@0ce330ab`, 2026-06-27), which is what made this straightforward to scope.


iamtoruk (approved):
Landing this for the one difference you flagged, now that #846 has merged the correctness fix. Rebased onto current main and reduced to exactly the perf delta: discovery streams through `readSessionLines` and stops after `MAX_HEADER_LINES_SCANNED` (20) leading lines, instead of #846's `readSessionFile` which reads each transcript in full (up to the 128MB cap) before looking at line 0. On the discovery hot path (walked for every file on every run), that turns a message-only file with no session record from a full read into a 20-line scan.

One reconciliation on top of your branch: the auto-merge reverted #846's `readFirstEntry` → `readSessionEntry` rename, so I re-applied it (and dropped the now-redundant `first.type !== 'session'` caller guard, since `readSessionEntry` only ever returns a session entry or null). Net effect on main is purely "bounded streamed scan," with #846's naming preserved.

Verified on the merged tree: `tsc` clean, omp + pi provider suites 37/37 (both #846's title-first omp case and your bounded-scan cases, including the "session record beyond the 20-line bound is not found" test that proves the scan is bounded rather than full-file), full suite green apart from the two pre-existing `parser.test.ts` failures. The 20-line bound is safe for the real OMP/Pi format (title slot + header at the top). Thanks for the analysis on the read-cost difference — it was the right call to keep it. Merging.

PR #860, opened 2026-07-29
post-0.9.18 polish: compare flags in the TUI, context id prefix, active-days label

### LINKED ISSUE #767: Post-0.9.18 polish: compare flags in TUI, context id asymmetry, active-days wording
Non-blocking UX nits logged during the pre-release functional sweep (all surfaces otherwise PASS):

- compare: the TUI picker ignores --model-a/--model-b (JSON path honors them), and explicit names require canonical ids (claude-opus-4-8) with no display-name resolution.
- context: the CLI accepts an 8-char session id prefix but /api/context/tree requires the full UUID; the 404 is clean but the asymmetry is confusing.
- report TUI: the Daily Activity panel counts active days from the bounded live scan (shows 'of 37') while the headline and daily series use the durable cache (63 days). Intentional per dashboard.tsx comments and each is internally consistent, but the two numbers sit on one screen; consider deriving the panel count from the same durable series.

avs-io:
Fixes #767 — three non-blocking UX nits from the pre-release sweep.

## 1. compare: the TUI ignored `--model-a`/`--model-b`

The default (TUI) format never passed the flags through to the interactive picker at all; only `--format json` honoured them. Both paths also required the canonical model id verbatim, with no way to pass the display name the picker itself shows.

Added `findModelStat` in `src/compare-stats.ts`: exact canonical-id match first, then a case-insensitive fallback through `getShortModelName`, the canonical-to-display map already used across the codebase, rather than a new alias table. `renderCompare` resolves the flags up front and seeds `CompareView`'s state so a valid pair goes straight to results. The JSON path's "both or neither" validation now applies to the TUI path too.

## 2. context: a session id prefix worked on the CLI but not through the API

`/api/context/tree` called `findClaudeSession`/`findCodexSession`, which already resolve an 8-char prefix exactly as `codeburn context <session>` does, and then additionally required the resolved id to equal the input verbatim — defeating its own resolver. Removed the redundant check so the endpoint trusts the same resolution the CLI relies on, keeping id resolution in one place.

The bundled frontend sends full ids from `/api/context/sessions`, so relaxing this cannot change its behaviour.

## 3. report TUI: two different active-day counts on one screen

**The count is correct by design and is not changed.** The Daily Activity panel scans a fixed six-month window independent of the selected period tab, so scrolling always works; the headline's day count comes from the durable cache scoped to the active period, which also counts days whose session files have since expired. These are two questions with two denominators, not one number computed two ways. Re-deriving the panel from the durable series would need a second `buildDurablePeriod` call scoped independently of the period tab — a behaviour and performance change, not polish.

What changed is the label: the panel footer now reads `of N days scanned` instead of a bare `of N`, so the two figures no longer read as a contradiction. The Optimize view's footer at the equivalent line was checked and left alone — it counts findings, not days, and has no headline to contradict.

**Adjacent finding, not part of item 3.** While investigating, the interactive dashboard's Overview headline turned out to have no explanation for durable-cache carry-forward, though `overview.ts`'s non-interactive report already has one. That wording is now shared as `carriedCostNote` and shown in the TUI as well. It explains carried *cost*, not the day count.

## Testing

Red-before-green for each change. New or extended: `findModelStat` cases in `compare-stats.test.ts`; display-name resolution and lone-flag rejection in `cli-emitters.test.ts`; full-id, prefix and no-match cases against a live server in `context-tree-api-prefix.test.ts`; `dailyActivityFooter` wording in `dashboard.test.ts`; `carriedCostNote` in `format-carried-cost-note.test.ts`.

Full suite 2476 passed, `tsc --noEmit` clean.

## Deliberately unchanged

`overview.ts` is untouched. Sharing `carriedCostNote` from there was tried and reverted: `overview.ts` has a local `formatCost` wrapper that adds thousands separators the shared helper does not replicate, so routing through it would have silently dropped that formatting on large carried-cost values.


iamtoruk (approved):
Reviewed on the merged tree (current main). All three fixes are scoped and verified:

- compare `--model-a/--model-b` now honored in the TUI with exact-then-display-name resolution (`findModelStat`); mutation-checked against the display-name test.
- `/api/context/tree` accepts an id prefix by reusing the CLI's own resolver; the 404 path is unchanged, and the mutation check confirms only the prefix case depended on the removed guard.
- Daily Activity footer labels the scanned-days denominator; pure display copy.

tsc clean; the five touched test files pass (76 tests); full suite 2600 passed with only the two standing parser.test.ts failures and one unrelated parallel-load flake (passes in isolation). Resolved the tests/dashboard.test.ts conflict as a union of the two new describe blocks. Good to merge.

PR #866, opened 2026-07-31
fix(dashboard): aggregate paired-device usage in the desktop Dashboard and menubar badge

### LINKED ISSUE #795: Desktop app dashboard shows only local device costs while paired devices are aggregated elsewhere
## Description

I have installed the Codeburn desktop app on two machines (a laptop and a workstation) and paired them under the same account.

The behaviour appears to be inconsistent across the different interfaces:

- The menu bar statistics show the combined usage/cost across both devices.
- The web GUI also reports the aggregated statistics correctly.
- In the desktop app, the Settings page correctly recognises that another device is paired.
- However, the Dashboard in the desktop app only displays the cost for the current device, rather than the combined total.

This makes the desktop dashboard inconsistent with both the menu bar and the web interface.

## Steps to reproduce
1. Install the desktop app on two devices.
1. Pair both devices to the same account.
1. Generate usage on both devices.
1. Open the desktop app on either device.
1. Compare:
    - the menu bar statistics,
    - the web GUI,
    - the desktop Dashboard.

## Expected behaviour

The desktop Dashboard should display the same aggregated usage/cost as the menu bar and the web GUI (or, if it is intentionally device-specific, this should be clearly indicated and documented).

## Actual behaviour

The Dashboard only reports the usage/cost of the current device, despite the app recognising the paired device in Settings.

## Environment
 - OS: OSX 26.5.2 and PopOS 24.04
 - Codeburn desktop app version: v0.9.19
 - Number of paired devices: 2 (laptop and workstation)

marcreynolds:
## Summary

Fixes #795. The desktop app's **Dashboard showed only the local device's
cost**, while the menu bar statistics and the web GUI reported usage
aggregated across paired devices. This makes the desktop Dashboard consistent
with those surfaces by adding a **Local / Combined scope** control, mirroring
the macOS menu bar's existing Scope setting.

Root cause: the desktop spawned `codeburn status --format menubar-json`
**without `--scope combined`**, so the CLI never ran its paired-device
aggregation and the renderer only ever had local data.

## What changed

**Desktop app**
- New **Scope: Local / Combined** setting in Settings › General (persists to
  `localStorage`, default `Local`), mirroring the menu bar.
- Combined passes `--scope combined` to the CLI, forces the provider filter to
  `all`, and clears the Claude-config scope (the CLI rejects `--scope combined`
  alongside a provider/project filter).
- The Overview hero shows the paired-device aggregate (cost / calls /
  sessions) with a **"Combined · N devices"** caption and a per-device
  breakdown when Combined is selected.

**Menu bar**
- The badge figure now reflects Combined scope instead of always showing the
  local total: it fetches the combined payload for the badge's period and
  renders the cross-device aggregate, falling back to local when no combined
  payload is available (cold cache, or an unreachable peer).

## Scope / limitations

- Only the hero KPIs aggregate; the detailed panels (daily chart, model table,
  etc.) remain local, because the combined payload carries totals only — this
  matches the menu bar's behavior.
- Aggregation is peer-to-peer over the LAN: a paired device contributes only
  while it is reachable. An unreachable peer degrades gracefully to the
  reachable subset.

## Testing

- **Desktop:** `npm run typecheck` and `npm run build:electron` clean; `npm
  test` → 445 passing (+9 new covering the scope argv, the Settings control,
  and the combined hero).
- **Menu bar:** `swift build` clean; new `AppStore` scope tests added (run
  under the Xcode toolchain: `swift test`).
- **End-to-end:** verified against the live CLI on a real two-device setup —
  Combined reported the aggregate ($163.54 month) vs local-only ($95.94),
  matching the menu bar and web GUI.


iamtoruk (approved):
Fixes #795. Verified end to end. Root cause was exactly as described: the desktop spawned `status --format menubar-json` without `--scope combined`, so the CLI never ran paired-device aggregation. The fix wires a Local/Combined scope control through `buildOverviewArgs` (combined passes `--scope combined` and drops the provider filter, which the CLI rejects alongside it; the renderer forces provider='all' in that mode so nothing is lost), and makes the menubar badge fetch the combined payload best-effort — success is gated only on the local fetch, so a slow or unreachable peer degrades gracefully to the local figure rather than failing the badge.

Local scope (the default) is untouched. On the merged tree (main was 29 ahead, clean merge): `tsc` clean, desktop `npm test` 462/462, `swift test` 128 tests + all XCTest suites at 0 failures, including the new scope-persistence, combined-total, and local-fallback cases. Merging as the base for #867.

PR #867, opened 2026-07-31
feat(menubar): mark badge when a paired device is unreachable in combined scope

marcreynolds:
## Summary

Follow-up to #866. Under Combined scope the menu bar badge falls back to the
**local** figure whenever a paired device doesn't report (asleep / off the
network) — which reads as a glitch ("it dropped back to local for a second").
This adds a small, dimmed **`reachable/total`** marker to the badge, plus a
matching tooltip, so a degraded total is legibly "a peer is unreachable"
rather than a number that mysteriously shrank.

Example badge: `🔥 $95.94 /mo · 1/2` — tooltip: *"CodeBurn month · 1 of 2
devices reporting"*. The marker clears the instant every paired device reports
again.

## What changed

- `AppStore.menubarBadgeDeviceShortfall` — `(reachable, total)` only when
  Combined scope is active **and** fewer paired devices reported than are
  paired; `nil` otherwise (all reported, single device, or Local scope).
- The badge appends the dimmed marker and sets the device-count tooltip when a
  shortfall is present.
- Tests for the shortfall condition (unreachable peer → marker; all reachable
  / Local scope → no marker).

## Stacked on #866

> [!NOTE]
> This branch is stacked on top of #866 (the combined-scope fix), since the
> indicator depends on the combined badge payload introduced there. Until #866
> merges, this PR's diff also includes #866's commit; once #866 lands on
> `main`, the diff reduces to just the indicator commit. **Merge #866 first.**

## Testing

- `swift build` clean; new `AppStore` shortfall tests added (run under the
  Xcode toolchain: `swift test`).


iamtoruk (approved):
Follow-up to #866 (now merged), and the diff reduces cleanly to just the indicator commit. `menubarBadgeDeviceShortfall` returns `(reachable, total)` only when combined scope is active and fewer paired devices reported than are paired — it reads `menubarBadgeCombined` (nil under local scope), so it is correctly silent on local scope and single-device setups, and clears the instant every peer reports again. That turns #866's graceful drop-to-local into a legible `reachable/total` marker instead of a number that mysteriously shrank.

On the merged tree: `swift test` 130 tests + all XCTest suites at 0 failures, including the two new shortfall cases (unreachable peer → marker; all reachable → no marker). Small, self-contained, and correct. Good to merge.

Issue #865, opened 2026-07-31
Overview total ignores --project/--exclude while every detail panel applies them


GodCC6:
## What happens

With `--project` / `--exclude`, the top **Overview** panel counts the filtered-out projects while every detail panel below it obeys the filter, so the two cannot be reconciled.

```
npx codeburn report -p week \
  --exclude Projects/ai-invest-copilot \
  --exclude writing/publishing/automation \
  --exclude Projects/novel-writing-publish \
  --exclude ai/invest/copilot
```

Overview reported **$7,962.65**; the **By Project** rows summed to **~$5,503.70** — a **~$2,458.95** gap, which is the excluded projects' spend almost exactly. **By Activity** and **By Model** both summed into the same $5,500–5,600 band, so the detail panels agreed with each other and only the headline was off.

Re-measured later with `--format json` (the rolling `-p week` window had moved on, so absolute figures differ):

| | Overview total | By Project sum | gap |
|---|---|---|---|
| filtered | $8,004.30 | $5,575.45 | **+$2,428.85** |
| unfiltered (control) | $8,095.19 | $8,166.12 | −$70.93 |

Calls are off by the same mechanism (84,458 vs. the filtered corpus's 56,492). The small negative gap in the unfiltered control is a separate, pre-existing effect — the durable path counts whole local days out of the cache while the live parse filters turns by exact timestamp, so they differ at a rolling range's boundary.

## Environment

- `codeburn --version`: **0.9.19**
- Providers involved: claude (dominant), codex, opencode
- Session history: `~/.claude/projects` 1.2G, `~/.codex/sessions` 2.3G, opencode 20M
- `daily-cache.v15.json`: 93 days, **15 of them with no per-project split**
- macOS, Node 25.9.0

Not a parsing bug, so no JSONL snippet — it reproduces from the cache alone.

## Root cause

`buildDurablePeriod()` in `src/usage-aggregator.ts` unions two sources for the headline:

1. **today** — `fp(await parseAllSessions(...))`, where `fp = filterProjectsByName(...)`: filtered.
2. **history** — `getDaysInRange(cache, ...)`, merged by `unionDaysForPeriod()`: **never project-filtered**.

[...]

PR #872, opened 2026-08-02
fix(parser): stop --provider filters from re-surfacing cached claude sessions

### LINKED ISSUE #871: `--provider <x>` leaks claude spend into every live-parse panel
# `--provider <x>` leaks claude spend into every live-parse panel

## What happens

`codeburn report --provider cursor` reports a correct cursor-only headline, but the panels underneath it carry claude spend. `By Model` lists Anthropic models under a cursor filter:

```
$ codeburn report -p all --provider cursor
│ CodeBurn  6 Months                                     │
│ €3.79 cost   132 calls   50 sessions                   │   <- cursor only, correct
...
│ By Model                                               │
│ ██████████ Fable 5       €1402.89  98.4%   2767        │   <- claude
│ ███░░░░░░░ Opus 5         €409.13  99.0%   1493        │   <- claude
│ ██░░░░░░░░ Opus 4.8       €224.34  94.4%    588        │   <- claude
│ ░░░░░░░░░░ Cursor (auto)   ~€2.40   0.0%     44        │
```

Same run via `--format json`:

| | value |
|---|---|
| `overview.cost` | 3.79 |
| `daily` summed | 3.79 |
| `projects` summed | **2038.99** |
| `models` summed | **2038.99** |
| `activities` summed | **2038.99** |

`--provider codex` produces byte-identical claude figures, so the leaked amount is independent of which provider you ask for.

The two leaked project buckets on my machine (`-Users-husamsoboh`, `-Users-husamsoboh-Downloads`, 11 sessions total) match exactly the 11 entries in my session cache that carry `prLinks`:

```
$ jq '[.providers.claude.files[] | select((.prLinks//[])|length>0)] | length' \
    ~/.cache/codeburn/session-cache.v7.json
11
```

## Root cause

Claude is scanned through `scanProjectDirs` rather than `parseProviderSources`, and that call has no provider-filter guard (`src/parser.ts:3739-3748`).

On a `--provider cursor` run, `discoverAllSessions` correctly returns no claude sources (`src/providers/index.ts:260-262`), so `claudeDirs` is empty. But `scanProjectDirs` still runs. Inside it, `allDiscoveredFiles` stays empty while `section` holds the entire cached claude history, so the orphan pass treats every cached file as no longer discovered (`src/parser.ts:1969-1976`):

```ts
for (const [filePath, cached] of Object.entries(section.files)) {
  if (allDiscoveredFiles.has(filePath)) continue
  if (!readOnly && !cached.prLinks?.length) continue   // only gate
  ...
  unchangedFiles.push({ filePath, dirName, cached })
}
```

That gate was written for genuinely deleted transcripts. It cannot distinguish "the file is gone" from "the file was not scanned because `--provider` excluded it this run".

- Write mode: every PR-bearing claude session leaks.
- Read-only mode (`readOnly = true`, taken whenever the cache refresh lock is unavailable, e.g. while the menubar app is running): the gate is bypassed entirely and every cached claude file leaks.

Every other provider already has this guard, a few dozen lines below in the same function (`src/parser.ts:3772-3775`):

```ts
// Skip if filtered to a different provider
if (providerFilter && providerFilter !== 'all' && providerFilter !== providerName) continue
```

Note that the paired eviction block does carry a `dirs.length > 0` guard (`src/parser.ts:2156`), so nothing is wrongly deleted from the cache. The asymmetry is only in the surfacing loop.

The headline escapes the bug because it comes from the provider-sliced daily cache (`sliceDayToProvider`), while the panels are built from the live parse. That is why the two cannot be reconciled, similar in shape to #865 but on the provider axis instead of the project axis.

## Affected surfaces

Anything that passes a specific non-claude provider into `parseAllSessions`, `buildDurablePeriod`, or `buildMenubarPayloadForRange`: `report`, `today`, `month`, `compare`, `audit`, `models`, `sessions`, `yield`, `spend`, `export`, `web`, and `status --format menubar-json`. `plan` usage tracking is the one with real consequences: a plan scoped to a non-claude provider counts leaked claude spend against that plan's budget.

`--provider all` and `--provider claude` are unaffected. `mcp` and the sharing paths are unaffected because they hardcode `provider: 'all'`.

## Environment

- codeburn 0.9.19, macOS 26.5.2, Node 22.23
- providers present: claude, cursor, codex, and others

I have a fix ready (a provider guard on the claude scan, plus regression coverage) and will open a PR against this issue unless you would rather take a different approach.

ozymandiashh:
Fixes #871

## The bug

`--provider <x>` filters the headline but not the live-parse panels, so `By Model` lists Anthropic models under `--provider cursor` while the total above it shows cursor alone.

```
overview.cost      3.79      <- correct, from the provider-sliced daily cache
projects summed    2038.99   <- leaked
models summed      2038.99   <- leaked
activities summed  2038.99   <- leaked
```

## Why

Claude is scanned through `scanProjectDirs` rather than `parseProviderSources`, and that call had no provider-filter guard. On a `--provider <other>` run `discoverAllSessions` correctly returns no claude sources, so `claudeDirs` is empty, but `scanProjectDirs` still ran. Its orphan pass reads the whole cached claude section and, with `allDiscoveredFiles` empty, treats every cached file as no longer discovered, re-injecting PR-bearing entries into the result. In read-only mode (cache refresh lock unavailable, which is the normal state while the menubar app is running) the `prLinks` gate is bypassed and every cached claude file is re-injected.

Every other provider already has this guard in the durable-orphan loop a few dozen lines below:

```ts
// Skip if filtered to a different provider
if (providerFilter && providerFilter !== 'all' && providerFilter !== providerName) continue
```

## The fix

One guard on the claude scan, mirroring the one above.

This is deliberately **not** a `claudeDirs.length > 0` check. When claude is in scope but every transcript has been pruned from disk, that orphan pass is exactly what keeps PR-attributed spend from vanishing, so gating on an empty dirs list would silently break the never-lose-history invariant. Gating on provider scope keeps both behaviours correct. The second test below locks that distinction in.

Nothing changes for `--provider all` or `--provider claude`. The paired eviction block already had a `dirs.length > 0` guard, so no cache entries were ever wrongly deleted and none are now.

## Verification against real data

`report -p all --provider cursor`, same machine, before and after:

| | before | after |
|---|---|---|
| `overview.cost` | 3.79 | 3.79 |
| `projects` summed | 2038.99 | 2.63 |
| `models` summed | 2038.99 | 2.63 |
| `models` listed | Fable 5, Opus 5, Opus 4.8, Cursor (auto), composer-2.5, grok-4.5, deepseek-reasoner | Cursor (auto), composer-2.5, grok-4.5, deepseek-reasoner |
| `projects` listed | -Users-husamsoboh, cursor, -Users-husamsoboh-Downloads | cursor |

`--provider all` before and after: identical model set, 489 projects in both, no rows dropped.

`--provider claude` after the fix: no cursor-only models present.

The residual `3.79` vs `2.63` delta is the pre-existing durable-versus-live difference, not part of this bug. It is one cached day (2026-07-26, €1.16) that the cache retains and the live parse can no longer reproduce because the Cursor SQLite store rotated. That is the never-lose-history invariant working as intended, and it is the same effect #865 documented in its unfiltered control.

## Tests

Two new cases in `tests/parser.test.ts`, using the existing `test-synthetic` provider harness:

1. `does not surface cached claude sessions when filtering to another provider` seeds a claude transcript carrying a `pr-link`, warms the cache with a `provider: all` parse, then parses again filtered to the synthetic provider and asserts the result contains only that provider. It is self-calibrating on cost, since the pricing engine re-derives cost from tokens. Without the fix it fails with `expected [ 'claude', 'test-synthetic' ] to deeply equal [ 'test-synthetic' ]`.
2. `still preserves a PR-bearing claude orphan when claude IS in scope` deletes the transcript and reparses unfiltered, asserting the total holds. This is the case a naive dirs-length guard would break.

Full suite: 2542 passing. Two failures in `tests/parser.test.ts` (`(a) copilot JSONL file-purge monotonic`, `(f) durable orphans survive a parse-version bump`) reproduce identically on unmodified `main` in my environment, both failing on their first parse with `expected +0 to be 200`, so they look environmental here rather than related to this change. `tsc --noEmit` is clean.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -3736,15 +3736,26 @@ async function runParse(
       ? { id: s.sourceId, label: s.sourceLabel, path: s.sourcePath, kind: s.sourceKind }
       : undefined,
   }))
+  // Claude is scanned through scanProjectDirs rather than parseProviderSources, so
+  // it needs the same provider-filter guard the durable-orphan loop below applies at
+  // its own level. Without it a --provider <other> run still enters scanProjectDirs
+  // with an empty dirs list, and the orphan pass there (which reads the whole cached
+  // claude section) treats every cached file as "no longer discovered" and re-injects
+  // it into the result. Note this is deliberately NOT a `claudeDirs.length > 0` check:
+  // when claude IS in scope but every transcript has been pruned from disk, that
+  // orphan pass is exactly what keeps PR-attributed spend from vanishing.
+  const claudeInScope = !providerFilter || providerFilter === 'all' || providerFilter === 'claude'
   if (claudeSources.length > 0) emitScanProgress({ kind: 'provider', provider: 'claude', state: 'start' })
   let claudeProjects: ProjectSummary[] = []
-  try {
-    claudeProjects = await scanProjectDirs(claudeDirs, seenMsgIds, diskCache, dateRange, saveProgress, readOnly)
-    if (claudeSources.length > 0) emitScanProgress({ kind: 'provider', provider: 'claude', state: 'done', files: claudeSources.length })
-  } catch (err) {
-    if (!isPermissionError(err)) throw err
-    process.stderr.write(`codeburn: skipped claude data (permission denied; grant Full Disk Access to include it)\n`)
-    emitScanProgress({ kind: 'provider', provider: 'claude', state: 'skipped' })
+  if (claudeInScope) {
+    try {
+      claudeProjects = await scanProjectDirs(claudeDirs, seenMsgIds, diskCache, dateRange, saveProgress, readOnly)
+      if (claudeSources.length > 0) emitScanProgress({ kind: 'provider', provider: 'claude', state: 'done', files: claudeSources.length })
+    } catch (err) {
+      if (!isPermissionError(err)) throw err
+      process.stderr.write(`codeburn: skipped claude data (permission denied; grant Full Disk Access to include it)\n`)
+      emitScanProgress({ kind: 'provider', provider: 'claude', state: 'skipped' })
+    }
   }
 
   const otherProjects: ProjectSummary[] = []
--- a/tests/parser.test.ts
+++ b/tests/parser.test.ts
@@ -488,3 +488,114 @@ describe('(g) skill attribution is independent of turn category', () => {
     expect(session!.skillBreakdown['telemetry-review']?.turns).toBe(1)
   })
 })
+
+// ═══════════════════════════════════════════════════════════════════════════
+// (h) Provider filter isolates claude: a --provider <other> run must not
+//     re-surface cached claude sessions through the orphan pass, while a run
+//     that DOES include claude still preserves PR-bearing orphans.
+// ═══════════════════════════════════════════════════════════════════════════
+describe('(h) provider filter excludes claude from the orphan pass', () => {
+  const SYNTH_SOURCE = (path: string): SessionSource[] =>
+    [{ path, project: 'synth-proj', provider: 'test-synthetic' }]
+
+  // The provider lives on each parsed call, not on SessionSummary.
+  const providersOf = (projects: Awaited<ReturnType<typeof parseAllSessions>>): Set<string> =>
+    new Set(projects
+      .flatMap(p => p.sessions)
+      .flatMap(s => s.turns)
+      .flatMap(t => t.assistantCalls)
+      .map(c => c.provider))
+
+  const SYNTH_CALL: ParsedProviderCall = {
+    provider: 'test-synthetic', model: 'gpt-4o',
+    inputTokens: 10, outputTokens: 5,
+    cacheCreationInputTokens: 0, cacheReadInputTokens: 0,
+    cachedInputTokens: 0, reasoningTokens: 0, webSearchRequests: 0,
+    costUSD: 0.25, tools: [], bashCommands: [],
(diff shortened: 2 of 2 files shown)


iamtoruk (approved):
Verified and clean. This fixes a real cross-panel money leak: under `--provider <other>`, `scanProjectDirs`'s orphan pass re-injected the whole cached Claude section into By Model / By Project / By Activity ($2,038 leaked vs $3.79 correct in your repro), worst in readOnly mode where the prLinks gate is bypassed.

The `claudeInScope` guard is the right fix — gated on provider scope, not `claudeDirs.length`, so it preserves the never-lose-history invariant (a PR-bearing claude orphan with pruned transcripts still surfaces under `--provider all`/`--provider claude`). Mutation-checked: the leak test fails without the guard and passes with it; the invariant test passes both ways.

An adversarial pass came back clean across all five vectors I care about here: the claude session cache is byte-identical after a filtered run (no corruption/eviction), `--provider claude` with pruned transcripts still preserves the orphan, no dangling progress event (claudeSources is non-empty only when claude is in scope), provider-name matching agrees across `assertProvider`/discovery/guard, and both readOnly states hold. Full suite green apart from the two pre-existing `parser.test.ts` failures that reproduce on main. Good to merge.

PR #880, opened 2026-08-03
fix(menubar): retry transient release-asset download failures

### LINKED ISSUE #876: menubar download fails with error 500
Running this to setup menu bar application:
`codeburn menubar`

Returns this:
```
Resolving CodeBurn Menubar v0.9.19...
Downloading CodeBurnMenubar-v0.9.19.zip...
Verifying checksum...

  Menubar install failed: Checksum download failed: HTTP 500
```

ozymandiashh:
Fixes #876

## What the user hit

```
Resolving CodeBurn Menubar v0.9.19...
Downloading CodeBurnMenubar-v0.9.19.zip...
Verifying checksum...

  Menubar install failed: Checksum download failed: HTTP 500
```

## Root cause, stated honestly

**The 500 is not ours.** I checked the release before touching any code:

- `gh release view mac-v0.9.19` reports both assets present and `state: uploaded` — zip 3,390,917 bytes, sidecar 94 bytes.
- Fetching the exact URLs the installer builds, 5 attempts each: `302` -> `200` every time, and the sidecar digest `3bc4b8d7...` matches what GitHub reports for the zip.
- `release-menubar.yml` uploads the zip and the `.sha256` together through `softprops/action-gh-release` with `fail_on_unmatched_files: true`, so a release cannot ship with the checksum missing.

So the asset was fine and is fine. The 500 came from the release-asset delivery path (github.com -> the Azure-backed signed URL), which occasionally 5xxs. That is outside this repo.

**What is ours** is that a single transient blip is fatal. `downloadToFile` and `verifyChecksum` each made exactly one attempt, with no retry anywhere in the file. `shouldFallbackToReleaseApi` treats only 404/410 as recoverable, so a 500 propagated straight out and aborted the install. And the message never named the URL it tried, so the user could not even retry it by hand or tell "GitHub hiccuped" from "this version was never published".

This PR does not claim to fix GitHub. It makes a known-transient failure survivable, and makes the surviving error actionable.

## The change

A single `fetchReleaseAsset` helper now wraps both asset downloads:

- **Retries only transient failures.** 5xx, plus network-level rejections (ECONNRESET / ETIMEDOUT / socket hang up). 3 attempts, 500ms then 1s backoff — at most ~1.5s added before a genuinely broken download reports back, which matters because `codeburn menubar` is interactive.
- **Does not retry 4xx.** 404/410 still fall through to the existing `shouldFallbackToReleaseApi` path completely unchanged, and `shouldFallbackToReleaseApi` itself is untouched, so the existing assertion that `shouldFallbackToReleaseApi(500) === false` still holds. I deliberately did not widen it: a 5xx does not prove the asset is missing, and retrying in place is the smaller, more predictable behavior change than rerouting to API discovery.
- **403/429 surface immediately** with the retry-after hint rather than being retried. A rate limit cannot clear inside a 1.5s window, so hammering it would only burn the budget.
- **Every thrown message now carries the requested URL.**
- A short line is printed when a retry happens, matching the existing `stageMenubarApp` console style, so the pause is not mysterious.

### The one thing worth reviewing closely

A naive retry here would be a security regression: re-downloading on a checksum **mismatch** would loop on a corrupted or tampered artifact until one attempt happened to pass. The digest comparison is deliberately outside the retry loop — retries cover transport only, and an integrity failure aborts on the first occurrence. There is a test pinning exactly that.

The fetch layer, sleep, logger and attempt budget are injectable so tests exercise retry without sleeping.

## Tests

9 new cases in `tests/menubar-installer.test.ts`:

```
✓ retries a transient 500 zip download and completes on the next attempt
✓ retries a transient 500 checksum download, the failure reported in the issue
✓ gives up on a persistent 500 and names the requested URL in the error
✓ does not retry a 404 and still routes to the missing-asset fallback
✓ does not retry a 429 and surfaces the retry-after hint instead
✓ retries a network-level failure and reports it with the URL when it persists
✓ recovers when a network-level failure clears on the next attempt
✓ fails a genuine checksum mismatch immediately instead of re-downloading
✓ honors an overridden attempt budget
```

`npx vitest run tests/menubar-installer.test.ts` -> **22 passed in 31ms**. The runtime confirms no real sleeping and no real network.

`npx tsc --noEmit` clean.

Full suite matches `main` on my machine: 14 failures in `tests/cache-refresh-lock.test.ts`, `tests/cli-status-menubar.test.ts` and `tests/parser.test.ts`, all of which reproduce identically on unmodified `b8c92bf` here, plus 26 `Cannot find package 'jsdom'` errors from the dash tests. None are touched by this change, which is confined to `src/menubar-installer.ts`.

## Scope

Two files, `src/menubar-installer.ts` and its test. No version bump, no unrelated changes, no behavior change on the happy path.

### DIFF
--- a/src/menubar-installer.ts
+++ b/src/menubar-installer.ts
@@ -37,6 +37,32 @@ type ProxyEnv = Partial<Record<'HTTPS_PROXY' | 'https_proxy' | 'HTTP_PROXY' | 'h
 type FetchOptions = Parameters<typeof undiciFetch>[1]
 type HeaderGetter = { get(name: string): string | null }
 
+/// Only the response surface the asset downloads actually touch, so tests can inject a
+/// plain object instead of constructing a full undici Response.
+type FetchLikeResponse = {
+  ok: boolean
+  status: number
+  headers: HeaderGetter
+  body: unknown
+  text(): Promise<string>
+}
+type FetchImpl = (url: string, options?: FetchOptions) => Promise<FetchLikeResponse>
+
+/// Release-asset delivery (github.com -> Azure blob) occasionally returns a transient 5xx or
+/// drops the socket. Three attempts with a short exponential backoff (0.5s, then 1s) rides out
+/// that class of blip while adding at most ~1.5s before a genuinely broken download reports
+/// back — `codeburn menubar` is interactive, so failing fast still matters.
+const ASSET_MAX_ATTEMPTS = 3
+const ASSET_BASE_DELAY_MS = 500
+
+export type AssetFetchOptions = {
+  fetchImpl?: FetchImpl
+  sleep?: (ms: number) => Promise<void>
+  log?: (message: string) => void
+  maxAttempts?: number
+  baseDelayMs?: number
+}
+
 class HttpStatusError extends Error {
   constructor(message: string, readonly status: number) {
     super(message)
@@ -135,7 +161,7 @@ export function formatGitHubReleaseLookupError(status: number, headers?: HeaderG
   return `${base}. ${details.join(' ')}`
 }
 
-function isMissingDirectAssetError(err: unknown): boolean {
+export function isMissingDirectAssetError(err: unknown): boolean {
   return err instanceof HttpStatusError && shouldFallbackToReleaseApi(err.status)
 }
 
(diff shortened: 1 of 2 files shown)


iamtoruk (approved):
Reviewed in depth on the merged tree (current main).

- Fixes the reported #876 failure: a transient 5xx on the checksum download is now ridden out (3 attempts, 500ms/1s backoff) instead of aborting the install.
- Security-critical property verified by mutation: the checksum-mismatch abort lives in verifyChecksum after fetchReleaseAsset returns, so a tampered artifact never re-downloads. Injecting a re-download-on-mismatch makes the "fails immediately" test fail (calls 1 to 2). Transient-classification and attempt-budget tests are likewise non-tautological.
- 404/410 still route to the release-API fallback (status preserved); 403/429 surface immediately with the retry-after hint.
- tsc clean; 22 menubar tests pass; full suite 2610 passed with only the two standing parser.test.ts failures (proven to reproduce on main, unrelated).
- Independent adversarial review: SHIP, no must-hold invariant broken.

A few edge findings from that review (mid-stream socket-drop retry, error-class/cause preservation, an exported-option guard, body-drain hygiene) are follow-ups; they will be applied in a separate hardening PR rather than held against this one. Good to merge.

Issue #50, opened 2026-04-16
Publish release with Codex Desktop originator fix


yjung199:
Hi, thanks for Codeburn.

I originally thought Codex Desktop usage detection was still broken on `main`, but after checking the repository I found that `main` already contains the fix.

Current `main` includes:
```
entry.payload.originator.toLowerCase().startsWith('codex')
```
and also has a test for accepting case-insensitive originator (Codex Desktop)

The issue I found is that the latest published/global version, codeburn@0.5.2, appears to have been released before this fix.

In 0.5.2, Codex Desktop usage reports as zero because Codex Desktop writes:

`"originator": "Codex Desktop"`
while the older released code only accepted lowercase `codex...` originators.

Before patching the installed 0.5.2 runtime, it showed
`{"currency":"USD","today":{"cost":0,"calls":0},"month":{"cost":0,"calls":0}}`

After applying the same case-insensitive check already present on main, it properly shows the usage:
`{"currency":"USD","today":{"cost":1.25,"calls":42},"month":{"cost":14.6,"calls":364}}`

**Request:**
Could you publish a new npm release, e.g. 0.5.3, that includes the current main fix for Codex Desktop session discovery?

**Expected behavior:**
Installing the latest released Codeburn package should detect Codex Desktop usage from ~/.codex/sessions/.

AgentSeal:
Just published 0.5.3 with the fix. `npm install -g codeburn@latest` or `npx codeburn@latest` will pick it up. Closing this -- let us know if Codex Desktop still shows zero after updating.

PR #881, opened 2026-08-03
fix(codex): validate rollouts structurally instead of by originator

### LINKED ISSUE #626: Codex provider skips JetBrains IntelliJ Codex ACP rollout sessions
## Problem

CodeBurn does not discover some Codex rollout files created through JetBrains IntelliJ AI Assistant / ACP.

The rollout files are valid Codex session files under `~/.codex/sessions/...`, but their first `session_meta` line uses this originator:

```json
"originator": "JetBrains.IntelliJ IDEA"
```

CodeBurn's Codex provider currently expects the originator to start with `codex`, so these JetBrains-created Codex ACP sessions are skipped during discovery.

## Real session example

Redacted path:

```text
~/.codex/sessions/2026/07/06/rollout-2026-07-06T16-07-01-019f3941-7d76-74d0-95eb-f0f0d70acdb2.jsonl
```

First line shape:

```json
{
  "type": "session_meta",
  "payload": {
    "originator": "JetBrains.IntelliJ IDEA",
    "cwd": "~/Projects/Tech-Ascension-Workspace",
    "session_id": "019f3941-7d76-74d0-95eb-f0f0d70acdb2"
  }
}
```

The same rollout also contains JetBrains ACP task context. In the real file, separate IntelliJ ACP tasks have their own `turn_id` values.

## Before fix

Running CodeBurn against the affected date missed the JetBrains-created Codex ACP rollout because the `session_meta.payload.originator` did not start with `codex`.

## Expected behavior

CodeBurn should treat JetBrains IntelliJ Codex ACP rollout files as Codex provider sessions when they are stored under the Codex session directory and have a valid `session_meta` record.

For JetBrains ACP tasks inside one rollout file, CodeBurn should split the tasks into separate sessions using a stable derived session id like:

```text
<codex-session-id>:jetbrains:<turn-id>
```

Example from my real cache after applying the fix locally:

```text
019f3941-7d76-74d0-95eb-f0f0d70acdb2:jetbrains:019f3941-cabb-7281-b95b-14e2f422a72b
019f3941-7d76-74d0-95eb-f0f0d70acdb2:jetbrains:019f394d-29db-7930-b072-e0fdbcae54dd
019f3941-7d76-74d0-95eb-f0f0d70acdb2:jetbrains:019f3950-65e7-7b41-9ac0-7d41f7ce8412
```

Normal Codex sessions should remain grouped under the original rollout session id.

## Local fix tested

I have a local patch that does this:

- Accepts `JetBrains...` originators for Codex rollout discovery.
- Detects JetBrains IntelliJ ACP task context from JetBrains `aia/agents` workspace roots.
- Uses `turn_id` to split JetBrains ACP tasks into separate CodeBurn sessions.
- Keeps normal Codex task records grouped under the original rollout session.
- Bumps CodeBurn cache fingerprints so stale bad parses are invalidated.
- Adds regression tests for discovery and parser behavior.

## Validation

Local Codex provider test suite:

```text
npm test -- --run tests/providers/codex.test.ts
```

Result:

```text
tests/providers/codex.test.ts (23 tests)
23 passed
```

Build:

```text
npm run build
```

Result: passed.

After installing the patched local build and clearing CodeBurn caches:

```text
rm -f ~/.cache/codeburn/codex-results.json ~/.cache/codeburn/session-cache.json
CODEBURN_TZ=America/Chicago codeburn report --provider codex --format json --day 2026-07-06
```

Result from real local data:

```json
{
  "sessions": 28,
  "calls": 160
}
```

## Environment

```text
OS: CachyOS Linux
Provider involved: Codex
JetBrains app: IntelliJ IDEA
Codex usage path: JetBrains IntelliJ AI Assistant / ACP running Codex
CodeBurn version tested locally: 0.9.15 patched
```

## Notes

I can open a PR with the local patch if this approach sounds acceptable.

### LINKED ISSUE #873: Codex discovery drops any rollout whose originator doesn't start with "codex"
## Problem

The Codex provider only counts rollouts whose `session_meta.payload.originator` passes:

```js
originator === "string" && entry.payload.originator.toLowerCase().startsWith("codex")
```

But `originator` is a free-form client identity, not a format marker. Any tool driving `codex app-server` (or embedding codex) writes structurally identical rollouts under `~/.codex/sessions/` with its *own* originator string — that's the field working as intended. The filter therefore silently excludes entire categories of usage from every frontend that isn't Codex itself.

This has now bitten at least three originators: `Codex Desktop` (exact-match era, fixed in #50 by relaxing to a prefix), `JetBrains.IntelliJ IDEA` (#626, open), and `t3code_desktop` ([T3Code](https://github.com/pingdotgg/t3code)), reported here. Each fix so far has admitted one more spelling; the next third-party frontend breaks it again.

## Concrete data

T3Code drives `codex app-server`; its rollouts begin:

```json
{"type": "session_meta", "payload": {"originator": "t3code_desktop", "source": "vscode", "cwd": "..."}}
```

In one week on my machine, 34 such rollouts were skipped while only 2 sessions (both `Codex Desktop`) were counted. The skipped sessions carry full `token_count` events and are not small: one top-level thread ended at 83.6M cumulative tokens, a spawned collab subagent at 23.4M, and 20 auto-review ("guardian") subagent rollouts totaled 9.0M. (Totals dominated by cached input — but that's the same basis CodeBurn prices from.)

Note this includes subagent-sourced rollouts (`"source": {"subagent": ...}`). Counting them does not double-count: I verified a parent thread's cumulative `token_count` excludes its child threads (the parent rollout's total matches the parent's own usage exactly).

## Expected behavior

Any structurally valid rollout under `~/.codex/sessions/` is counted, regardless of which frontend created it.

## Suggested fix

Stop using `originator` for discovery. A `session_meta` first line plus `token_count` events is already a strong structural signal that a file is a real codex rollout. Keep `originator` for what it's good at — labeling/attribution in reports — and this class of issue (this one, #626, and the next one) closes permanently.

Happy to provide a redacted sample rollout if useful.

ozymandiashh:
Fixes #873
Fixes #626

## The bug

Codex rollout discovery gated on `session_meta.payload.originator`:

```ts
const valid = entry.type === 'session_meta' &&
  typeof entry.payload?.originator === 'string' &&
  entry.payload.originator.toLowerCase().startsWith('codex')
```

But `originator` is a free-form **client identity**, not a format marker. Anything driving `codex app-server` writes structurally identical rollouts under `~/.codex/sessions/` with its own value, and those were dropped silently. This has now bitten three originators in a row: `Codex Desktop` (#50, fixed by relaxing exact-match to a prefix), `JetBrains.IntelliJ IDEA` (#626, still open), and `t3code_desktop` (#873). Each fix admitted one more spelling; the next frontend breaks it again.

Reproduced with three structurally identical rollouts differing only in `originator`:

```
$ ls sessions/2026/08/03/
rollout-...-aaaa.jsonl   originator: codex-tui
rollout-...-bbbb.jsonl   originator: t3code_desktop
rollout-...-cccc.jsonl   originator: JetBrains.IntelliJ IDEA

discovered: 1
 - rollout-...-aaaa.jsonl
```

## The fix

Validate **structurally**, never by string-matching a client name:

```ts
const payload: unknown = entry.payload
const valid = entry.type === 'session_meta' &&
  typeof payload === 'object' && payload !== null && !Array.isArray(payload)
```

That is signal enough. The walker only visits `rollout-*.jsonl` under the strict `YYYY/MM/DD` regex path or `archived_sessions/`, and `codex.ts` is the only provider that reads `~/.codex` — so **directory ownership**, not originator content, decides the provider. Foreign files (wrong entry type, missing or non-object payload, malformed JSON) are still rejected; the existing negative test still passes. No new originator allowlist, so this class of issue closes permanently rather than one spelling at a time.

**No double-counting.** Subagent rollouts are already handled by the existing `token_count` cumulative dedup and the fork-replay cutoff, not by the originator gate. The #873 reporter independently verified a parent thread's cumulative `token_count` excludes its child threads.

### 1. The daily cache would have made the fix invisible

`usage-aggregator` serves every day before today from `daily-cache`, `ensureCacheHydrated` only re-parses the gap after `lastComputedDate`, and retention is 3650 days. Newly-admitted rollouts were rejected *before they were ever parsed*, so nothing downstream can notice on its own. An upgrading user with a warm cache would have kept pre-fix history **forever**, while the session list disagreed with the headline:

| scenario | cost | calls | sessions |
|---|---|---|---|
| `main` @ b8c92bf, fresh cache | 4.55 | 1 | 1 |
| this branch **without** the bump, reusing main's warm cache | 4.55 | 1 | **2** |
| this branch, reusing that same warm cache | **18.2** | **2** | 2 |
| this branch, fresh cache (ground truth) | 18.2 | 2 | 2 |

Row 2 is what every existing user would have gotten: the session count moves because discovery now works, but cost and calls stay frozen — the fix looks broken and the undercount is permanent. Row 3 now matches ground truth.

So `DAILY_CACHE_VERSION` and `MIN_SUPPORTED_VERSION` go 15 -> 16, with a version-history entry in the established style. This rides the v14 carry-forward, so it is lossless: days whose sources are gone are preserved and marked `carried`, and there is a test pinning that.

`CODEX_CACHE_VERSION` and `PROVIDER_PARSE_VERSIONS.codex` deliberately do **not** move. Verified both statically and empirically: `codex-results.json` is only written after a source is discovered *and* parsed, so a file rejected at discovery never had an entry in either cache. After warming the fixture on `main`, both cache files contained exactly one key — the first-party rollout — with the third-party one absent. Bumping them would force a needless full re-parse of already-correct sessions.

### 2. A malformed `cwd` zeroed out the entire provider

`discoverSessionFile` did `meta.payload?.cwd ?? 'unknown'` and passed it to `sanitizeProject`, which calls `.replace()`. A rollout with a non-string `cwd` throws `TypeError: cwd.replace is not a function`; the throw escapes `discoverSessions()` and `safeDiscoverSessions` catches it into `[]` for **all** of Codex after one stderr line. Every Codex report reads zero.

This crash exists on `main`, but only for files already claiming a `codex*` originator. Relaxing the gate is exactly what makes it reachable from arbitrary third-party clients whose schema conformance is, by this PR's own argument, unverified — so it belongs here. Guarded on both the discovery path and the parse path (`sessionCwd` flows into `projectPath`/`workingDirectory`, and `src/parser.ts` calls `.trim()`/`.replace()`/`.toLowerCase()` on those).

## Tests

Nine new cases, all of which fail without the corresponding fix:

```
✓ accepts a third-party frontend originator (t3code_desktop)
✓ accepts the JetBrains plugin originator (issue #626)
✓ accepts a rollout with no originator field at all
✓ accepts an archived rollout from a third-party frontend
✓ still rejects foreign and malformed first lines regardless of originator
✓ survives a non-string cwd instead of zeroing out the whole provider
✓ does not leak a non-string cwd into projectPath/workingDirectory
✓ re-derives a warm v15 cache instead of serving its pre-fix rollups
✓ carries a v15 day forward when its sources can no longer re-derive it
```

Reverting only the two source files gives, verbatim:

```
× re-derives a warm v15 cache...            → expected +0 to be 1
× carries a v15 day forward...              → expected 4.55 to be 18.2
× survives a non-string cwd...              → cwd.replace is not a function
× does not leak a non-string cwd...         → cwd.replace is not a function
```

`npx vitest run tests/providers/codex.test.ts tests/codex-cache-invalidation.test.ts tests/daily-cache.test.ts tests/daily-cache-carry-forward.test.ts tests/daily-cache-degraded-completeness.test.ts tests/cache-persistence-coldstart.test.ts` -> **6 files, 111 tests passed**.

`npx tsc --noEmit` clean.

## Full suite, honestly

Branch and pristine `main` fail the same *count* on this machine but a rotating *cast*: the stable core is `parser.test.ts` (2), `cli-status-menubar.test.ts` (5), `cache-refresh-lock.test.ts` (10). The remainder are subprocess-spawning CLI suites timing out under parallel load — they rotate run to run and all pass in isolation (`cli-emitters`, `cli-json-daily`, `spend-flow`, `cli-status-menubar` -> 4 files, 27 tests passed). I re-ran `main` in the same session to confirm rather than assuming, since it did not match the baseline I started from.

## Docs

`docs/providers/codex.md` documented the old rule verbatim ("The first line must be a `session_meta` entry with `payload.originator` starting with `codex`") and now describes structural validation.

### DIFF
--- a/docs/providers/codex.md
+++ b/docs/providers/codex.md
@@ -4,7 +4,7 @@ OpenAI Codex CLI.
 
 - **Source:** `src/providers/codex.ts`
 - **Loading:** eager (`src/providers/index.ts:2`)
-- **Test:** `tests/providers/codex.test.ts` (374 lines)
+- **Test:** `tests/providers/codex.test.ts` (1075 lines)
 
 ## Where it reads from
 
@@ -24,7 +24,11 @@ The active-session discovery walk uses strict regex (`^\d{4}$`, `^\d{2}$`) on ea
 
 ## Storage format
 
-JSONL. The first line must be a `session_meta` entry with `payload.originator` starting with `codex` (case-insensitive). Files that fail this check are silently skipped.
+JSONL. Validation of the first line is **structural**: it must parse as JSON, have `type === "session_meta"`, and carry a `payload` that is a plain object (not missing, not a scalar, not an array). Files that fail this check are silently skipped.
+
+`payload.originator` is deliberately **not** part of the check. It is a free-form client identity string, not a format marker: Codex CLI writes `codex-tui` / `codex_exec` / `codex_cli_rs`, Codex Desktop writes `Codex Desktop`, and third-party frontends driving `codex app-server` write their own values (`t3code_desktop`, `JetBrains.IntelliJ IDEA`, ...) into structurally identical rollouts. Gating discovery on the spelling silently dropped those sessions from every report and required a new allowlist entry per client (issues #626, #873). Directory ownership decides the provider instead: `codex.ts` is the only provider that reads `~/.codex`, and the walk only visits `rollout-*.jsonl` under the strict `YYYY/MM/DD` path or `archived_sessions/`. `originator` is still parsed into the session meta entry, but nothing downstream reads it.
+
+Because admission no longer implies a known client, every payload field is treated as untrusted JSON. `payload.cwd` in particular is type-guarded before it reaches `sanitizeProject` (discovery) or `projectPath`/`workingDirectory` (parse): a non-string `cwd` falls back to the `unknown` project instead of throwing out of `discoverSessions`, which `safeDiscoverSessions` would have turned into an empty session list for the *entire* provider.
 
 The first line read is capped at 1 MB (`FIRST_LINE_READ_CAP`). Codex CLI 0.128+ embeds the full system prompt in `session_meta`, which can run 20-27 KB; the cap leaves headroom while bounding memory if a corrupt file has no newline.
 
--- a/src/daily-cache.ts
+++ b/src/daily-cache.ts
@@ -5,7 +5,17 @@ import { homedir } from 'os'
 import { join } from 'path'
 import type { DateRange, ProjectSummary } from './types.js'
 
-// Bumped to 15: per-project daily rollups. Days and provider slices now carry
+// Bumped to 16: Codex discovery is structural instead of originator-gated
+// (#873/#626), so rollouts written by third-party frontends driving
+// `codex app-server` ("t3code_desktop", "JetBrains.IntelliJ IDEA", ...) now
+// contribute usage that v15 rollups never contained. Those files were rejected
+// before they were ever parsed, so nothing downstream can notice on its own:
+// `usage-aggregator` serves every day before today from this cache, and
+// retention is ten years, so an upgrading user with a warm cache would keep the
+// pre-fix history forever while today's numbers silently disagreed with it.
+// Raising MIN_SUPPORTED_VERSION forces the one-time re-derivation.
+//
+// v15: per-project daily rollups. Days and provider slices now carry
 // a `projects` breakdown (cost/calls/savings/sessions per project) so project
 // history outlives the session files, like models and categories already do.
 // This bump is the first to ride the v14 carry-forward: the old cache is
@@ -57,8 +67,8 @@ import type { DateRange, ProjectSummary } from './types.js'
 // that older binaries skipped. v8 added local-model savings to the daily
 // rollup; the `savingsConfigHash` field is invalidated separately when the
 // user changes their `localModelSavings` mapping.
-export const DAILY_CACHE_VERSION = 15
-const MIN_SUPPORTED_VERSION = 15
+export const DAILY_CACHE_VERSION = 16
+const MIN_SUPPORTED_VERSION = 16
 // Version-suffixed so different binaries each own a distinct file and never
 // clobber an incompatible schema. Bumping the version mints a fresh filename;
 // adoptOlderDailyCaches then unions days out of every previous file (including
--- a/src/providers/codex.ts
+++ b/src/providers/codex.ts
@@ -185,12 +185,30 @@ async function readFirstLine(filePath: string): Promise<CodexEntry | null> {
   }
 }
 
+// Validation is STRUCTURAL, never string-matching on `payload.originator`.
+// `originator` is a free-form CLIENT IDENTITY string, not a format marker: any
+// tool driving `codex app-server` writes structurally identical rollouts under
+// ~/.codex/sessions with its own value ("codex-tui", "Codex Desktop",
+// "t3code_desktop", "JetBrains.IntelliJ IDEA", ...). Gating on the spelling
+// silently dropped every third-party frontend and needed a new allowlist entry
+// per client (issues #626, #873).
+//
+// A `session_meta` first line with a well-formed payload object is signal
+// enough: the walker only visits `rollout-*.jsonl` under the strict
+// YYYY/MM/DD path or `archived_sessions/`, and codex.ts is the only provider
+// that reads ~/.codex, so directory ownership — not originator content —
+// decides the provider. Genuinely foreign files (wrong entry type, missing or
+// non-object payload, malformed JSON) are still rejected.
 async function isValidCodexSession(filePath: string): Promise<{ valid: boolean; meta?: CodexEntry }> {
   const entry = await readFirstLine(filePath)
   if (!entry) return { valid: false }
+  // `entry` comes from an unchecked JSON.parse cast, so re-check the payload
+  // shape at runtime instead of trusting the declared type.
+  const payload: unknown = entry.payload
   const valid = entry.type === 'session_meta' &&
-    typeof entry.payload?.originator === 'string' &&
-    entry.payload.originator.toLowerCase().startsWith('codex')
+    typeof payload === 'object' &&
+    payload !== null &&
+    !Array.isArray(payload)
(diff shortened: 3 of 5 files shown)


iamtoruk (approved):
Reviewed in depth on the merged tree (current main), fixes #873 and #626.

Verification:
- Structural validation is a strict superset of the old originator gate, so no previously-counted session changes and third-party frontends (t3code_desktop, JetBrains, no-originator, archived) are now discovered. Mutation-checked: reverting to the originator gate fails the four admission tests.
- The two non-string-cwd guards prevent a malformed third-party rollout from throwing out of discovery and zeroing the whole codex provider. Mutation-checked: reverting both fails the "survives / does not leak a non-string cwd" tests.
- DAILY_CACHE_VERSION and MIN_SUPPORTED_VERSION 15->16 force the one-time re-derivation so a warm cache does not serve pre-fix rollups forever; it rides the v14 carry-forward, so v15 days whose sources are gone are preserved and marked carried (no data loss). Mutation-checked: reverting to 15 fails the re-derive and carry-forward tests.
- Not bumping CODEX_CACHE_VERSION is correct: a discovery-rejected file never had a codex-results.json entry, so newly-admitted rollouts parse fresh on the v16 re-derivation while first-party sessions are served unchanged.
- No double-count: admission is a superset and the seenKeys / basename / fork-replay dedup layers are untouched.
- tsc clean; 62 codex + daily-cache tests pass; full suite green apart from the two standing parser.test.ts failures (reproduce on main) and known parallel-load flakes that pass in isolation.

Independent adversarial review confirmed the money invariants (conservation, cache correctness, no data loss). One bounded, non-regressive follow-up: extend the same guard rationale to entry.timestamp (fork-cutoff Date math) and a non-string model, so a malformed third-party rollout is counted rather than silently reading zero for that one session. Tracking that as a small follow-up. Good to merge.

PR #883, opened 2026-08-03
feat(menubar): add preferred terminal setting with Terminal.app fallback

### LINKED ISSUE #877: Allow alternative terminal options (iTerm2) for TerminalLauncher in CodeBurnMenubar
On mac, I think it's safe to say that many developers prefer the fuller featured iTerm2 terminal app over Apple's Terminal.app. 

When using CodeBurnMenubar and clicking the "Full Report" button, it opens up the report using Terminal.app rather that the user's default terminal. 

This appears to be explicitly set in TerminalLauncher.swift at https://github.com/getagentseal/codeburn/blob/2de4d100bf746b0830b1d43c6118a91d853628b5/mac/Sources/CodeBurnMenubar/Security/TerminalLauncher.swift#L14

and 

https://github.com/getagentseal/codeburn/blob/2de4d100bf746b0830b1d43c6118a91d853628b5/mac/Sources/CodeBurnMenubar/Security/TerminalLauncher.swift#L55

Would it be possible to add a helper to detect if a default terminal is specified/preferred, and only fallback to the Apple Terminal.app if not set?

psuedo-code

```swift
import AppKit

struct TerminalDetector {
    /// Detects the Bundle Identifier of the default terminal handler.
    /// Falls back to Apple Terminal if no custom handler is registered or found.
    static func getPreferredTerminalBundleId() -> String {
        let shellScriptUTI = "public.unix-executable"
        let fallbackTerminal = "com.apple.Terminal"
        
        // Query LaunchServices for the default app assigned to open executable shell files
        if let defaultAppURL = NSWorkspace.shared.urlForApplication(toOpen: UTType(shellScriptUTI)),
           let bundle = Bundle(url: defaultAppURL),
           let bundleIdentifier = bundle.bundleIdentifier {
            return bundleIdentifier
        }
        
        return fallbackTerminal
    }
}
```

ozymandiashh:
Closes #877

Flagging up front, per `CONTRIBUTING.md`: this is a feature and **#877 has no maintainer response yet**. I've posted the design on the issue and I'm happy to rework or drop this if you'd prefer a different shape — the code is here so the discussion has something concrete to argue with, not to jump the queue.

## Two corrections to the issue's diagnosis

**1. Terminal.app isn't selected by bundle id**, so the `LSCopyDefaultRoleHandler`-style lookup doesn't slot in where you'd expect. `TerminalLauncher` picked by hardcoded filesystem path (`/System/Applications/Utilities/Terminal.app`, `/Applications/Utilities/Terminal.app`) checked with `fileExists`. There was no `NSWorkspace`/LaunchServices resolution anywhere.

**2. `urlForApplication(toOpen: UTType("public.unix-executable"))` isn't a reliable signal for "preferred interactive terminal."** It resolves the handler for executable *files* — what happens when you double-click a binary in Finder. Most people never set it, and on a stock machine it doesn't resolve to iTerm2 even for dedicated iTerm2 users. So this ships an **explicit setting** rather than auto-detection. I'd rather ask than guess wrong silently.

Also worth noting: a partial path already existed. If neither hardcoded Terminal.app path was present, the launcher fell back to a detached **headless** spawn (the doc comment names iTerm/Ghostty/Warp users). Alternate-terminal users got a silent background run with no window — arguably the worse half of the bug.

## The shape

- **A closed `PreferredTerminal` enum**, each case owning its app paths and its own script. This is deliberate: the `Security/` directory's invariant is that no user-influenced string ever reaches a shell or an AppleScript, so the `tell application` target stays a compile-time literal per case. A user-supplied bundle id interpolated into `osascript` would reopen exactly the injection surface this file exists to close. Unrecognised persisted values collapse to the default.
- **Settings → General → Terminal**, alongside the existing Currency/Metric pickers, same `@AppStorage` pattern.
- **Default unchanged** (Terminal.app), so nothing moves for existing users.
- **Graceful degradation**: chosen terminal not installed -> Terminal.app -> the existing headless spawn. Never silently nothing.

Only terminals with a real "run this in an interactive session and leave the window open" verb are listed. Ghostty, WezTerm, Warp, Alacritty and kitty expose no AppleScript equivalent — `-e <cmd>` tears the window down the moment the command exits, which would make Full Report flash and vanish. They stay out rather than shipping broken.

## Testing, on a real iTerm2

I installed iTerm2 3.6.11 specifically to verify this rather than shipping a recalled dialect. Two findings, both of which changed the patch:

**The launcher was failing silently.** `runInTerminal` ended in `try? process.run()` — fire and forget, no `terminationStatus` check. Any `osascript` failure meant no window, no error, no log, no fallback: the app just appears dead. The old code had the same shape, but Terminal.app's terminology is always registered so it never failed in practice; adding a second terminal is what makes it reachable. Now the exit status is checked and the fallback chain runs.

**`tell application "iTerm2"` does not work cold.** This is the one that would have shipped broken. Measured on this machine, iTerm2 not running:

```
$ osascript -e 'tell application "iTerm2" ... create window with default profile ...'
syntax error: Expected "," but found class name. (-2741)
EXIT=1
```

It only compiles while iTerm2 already happens to be running — which is exactly the state you're in when testing interactively, so the bug is easy to miss. `tell application "iTerm"` compiles cold, launches the app, and runs the command:

```
$ osascript -e 'tell application "iTerm" ... write text "echo COLD_OK > /tmp/proof" ...'
EXIT=0
$ cat /tmp/proof
COLD_OK
```

Command execution confirmed by writing from inside the session and reading the file back, not just by a zero exit code.

`swift build` succeeds. Note that **no CI workflow runs `swift test`** (`ci.yml` is Semgrep-on-`src/` only; `release-menubar.yml` only does `swift build`), so `mac/Tests/CodeBurnMenubarTests/TerminalLauncherTests.swift` is a local gate only — covering enum/path mapping, per-terminal script generation, fallback selection when an app is absent, and that argv validation still rejects unsafe tokens. The 7 XCTest suites in that target can't run without Xcode on this machine, which is equally true at baseline.

No `src/` changes; this is contained to `mac/` plus a README entry.

## Known trade-offs, not hidden

- A 30s `osascript` timeout is treated as success. If it hangs past 30s (e.g. a TCC/Automation prompt the user hasn't answered), no fallback fires. Falling back there would risk opening two windows once the user clicks Allow.
- If a terminal opens a window and runs the command but `osascript` still exits non-zero, the fallback chain could run it twice. I haven't been able to produce that in practice.
- `CodeBurnPreferredTerminal` is documented in the README next to the other `defaults write` keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,6 +306,14 @@ defaults write org.agentseal.codeburn-menubar CodeBurnMenubarRefreshSeconds -int
 
 Seconds between refreshes: `60`, `300`, or `900`; `0` is Manual and `-1` is Auto. Takes effect on the next refresh tick, no relaunch needed.
 
+**Preferred terminal** decides where Full Report and Optimize open. Set it in Settings → General → Terminal, or from Terminal:
+
+```bash
+defaults write org.agentseal.codeburn-menubar CodeBurnPreferredTerminal -string iterm2
+```
+
+Allowed values are `terminal` (macOS Terminal.app, the default) and `iterm2`. Anything else falls back to `terminal`. Only terminals that can script a command into a live window are offered; if the chosen app is missing or fails to accept the command, CodeBurn tries Terminal.app and then runs the command in the background, logging each step to Console.app. Takes effect on the next launch of a command, no relaunch needed.
+
 ### Linux (GNOME)
 
 Linux gets the same ambient view through a GNOME Shell extension (GNOME 45+): spend in the top panel, period switcher, compact mode, and daily budget alerts. It lives in [`gnome/`](gnome/):
--- a/mac/Sources/CodeBurnMenubar/Security/PreferredTerminal.swift
+++ b/mac/Sources/CodeBurnMenubar/Security/PreferredTerminal.swift
@@ -1,0 +1,124 @@
+import Foundation
+
+/// Closed set of terminal emulators CodeBurn knows how to drive (#877).
+///
+/// SECURITY: this type exists specifically so that a *user preference* can never become a
+/// free-form string inside an AppleScript. The preference is persisted as a raw value that is
+/// parsed back through `init(rawValue:)`; anything unrecognised collapses to `.default`. Every
+/// application name that reaches `osascript` is a hardcoded literal in `script(command:)` --
+/// never a stored or interpolated string. The only interpolated value is the command, which
+/// callers must have already validated token-by-token with `CodeburnCLI.isSafe`.
+///
+/// Only terminals with a real "run this in an interactive session and leave the window open"
+/// scripting verb are listed. Terminal.app has `do script`; iTerm2 has
+/// `write text` on a session. Ghostty, WezTerm, Warp, Alacritty and kitty expose no equivalent
+/// AppleScript verb -- launching them with `-e <cmd>` tears the window down the moment the
+/// command exits, which would make "Full Report" flash and vanish, so they intentionally stay
+/// out of this enum rather than shipping broken.
+enum PreferredTerminal: String, CaseIterable, Identifiable, Sendable {
+    case terminal
+    case iTerm2 = "iterm2"
+
+    var id: String { rawValue }
+
+    var label: String {
+        switch self {
+        case .terminal: return "Terminal (macOS default)"
+        case .iTerm2: return "iTerm2"
+        }
+    }
+
+    /// Bundle locations probed with a plain `fileExists` check, mirroring the pre-#877
+    /// behaviour.
+    ///
+    /// This is a simplicity/determinism choice, NOT a security boundary. A fixed list needs no
+    /// LaunchServices database state, gives the same answer on every machine, and keeps the
+    /// Settings "(not installed)" hint honest without a framework round-trip.
+    ///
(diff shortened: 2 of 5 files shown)


iamtoruk (approved):
Reviewed in depth on the merged tree, closes #877.

Security invariant (this lives in Security/) holds: the AppleScript `tell application` target is a compile-time literal per PreferredTerminal case, the only interpolated value (command) is gated by CodeburnCLI.isSafe's strict allowlist ^[A-Za-z0-9 ._/\-]+$ which rejects quotes, backslash, $, backticks and shell metacharacters, and runScript re-validates defensively. Unrecognised persisted values collapse to .default.

- swift build succeeds; 26 XCTests pass, including the security cases (metacharacters rejected before reaching AppleScript, a single unsafe token poisons the whole argv), the fallback chain, per-terminal script dialects, and persistence round-trip.
- Real Settings picker with honest "(not installed)" hints; default unchanged (Terminal.app) so existing users see no change.
- Fixes two real bugs the author verified on iTerm2 3.6.11: the silent try? process.run() launcher failure (now checks terminationStatus and falls back) and tell application "iTerm2" failing to compile cold (uses the "iTerm" bundle name).

Clean, well-scoped, security-conscious. Good to merge.

PR #893, opened 2026-08-03
models: name GPT-5.6, Grok 4.5 and ClinePass slugs instead of showing raw ids

therickfactr:
## Summary

- Adds `SHORT_NAMES` entries for model ids that **price correctly but had no display entry**, so the By Model panel rendered a raw slug next to properly named siblings: `gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`, `grok-4.5`, `qwen3.7-max`, `minimax-m3`, `mimo-v2.5-pro`.
- Display-only. No dollar amounts move — pricing already resolved for all of these (`mimo-v2.5-pro` is unpriced upstream both before and after; this only gives it a name).

Before / after, from a real local report:

```
qwen3.7-max      $0.562      ->  Qwen 3.7 Max     $0.562
minimax-m3       $0.154      ->  MiniMax M3       $0.154
mimo-v2.5-pro    $0.0090     ->  MiMo v2.5 Pro    $0.0090
grok-4.5        ~$10.82      ->  Grok 4.5        ~$10.82
gpt-5.6-sol      $280.56     ->  GPT-5.6 Sol      $280.56
gpt-5.6-terra    $122.98     ->  GPT-5.6 Terra    $122.98
```

## Notes on the non-obvious ones

**GPT-5.6 variants are listed individually, not as a bare `gpt-5.6`.** A base entry would swallow every future `gpt-5.6-*` via the prefix match and hide the variant behind a sibling's label — precisely what `getShortModelName`'s version-boundary rule exists to prevent. An unlisted variant still falls through to its raw id, and a test pins that.

**`grok-4.5` is the model, not the harness.** The Grok Build CLI reports the model it runs as `current_model_id`, so the id takes the model's own name. Ids that really are `grok-build*` keep the "Grok Build" label — also covered by a test, since that distinction is easy to regress.

**ClinePass needed no new prefix handling.** Its ids arrive as `cline-pass/<slug>`; `getShortModelName`'s path fallback already strips the prefix and re-resolves the bare slug, exactly as it does for `accounts/fireworks/models/<slug>`. Adding a dedicated prefix rule would have been dead code. Tests cover the `cline-pass/` path so the behavior stays intentional rather than incidental.

**MiniMax M3 is mapped under both spellings.** The gap-filled OpenRouter key is the lowercase slug while sessions report the capitalized name; `SHORT_NAMES` matching is case-sensitive, and the case-insensitive index covers pricing only.

## Testing

- [x] `npm test` passes

Every id above was taken from real local sessions, not from documentation — `gpt-5.6-luna` in particular only surfaced by grepping actual Codex transcripts (7 records), which is why it is included alongside the two obvious variants.

`npm test`: 20 failed / 2619 passed against a `main` baseline of 23 failed / 2612 passed on the same machine and commit base. No file fails here that does not also fail on `main`; the failures are the pre-existing `cache-refresh-lock` / `parser` durable-orphan / `app/electron/cli` set plus CLI-subprocess tests that flake under parallel load and pass in isolation. The 26 errors in both runs are `app/renderer` tests needing `jsdom` from `app/`'s own dependency tree.

Five new assertions in `tests/models.test.ts` cover the variant-collapse guard, the Grok Build distinction, the `cline-pass/` path fallback, and both MiniMax spellings.

### DIFF
--- a/src/models.ts
+++ b/src/models.ts
@@ -922,6 +922,27 @@ const SHORT_NAMES: Record<string, string> = {
   'glm-5p2': 'GLM-5.2',
   'qwen3p7-plus': 'Qwen 3.7 Plus',
   'kimi-k2p7-code': 'Kimi K2.7 Code',
+  // Ids that price correctly but had no display entry, so reports showed the
+  // raw slug. All display-only. The GPT-5.6 variants are listed individually
+  // rather than as a bare `gpt-5.6`: a base entry would swallow every future
+  // `gpt-5.6-*` via the prefix match and hide the variant, which is exactly
+  // what getShortModelName's version-boundary rule is there to prevent.
+  'gpt-5.6-sol': 'GPT-5.6 Sol',
+  'gpt-5.6-terra': 'GPT-5.6 Terra',
+  'gpt-5.6-luna': 'GPT-5.6 Luna',
+  // The Grok Build harness reports the model it runs (`grok-4.5`), so this is
+  // the model's own name; `grok-build*` ids still resolve to "Grok Build".
+  'grok-4.5': 'Grok 4.5',
+  // ClinePass routes models as `cline-pass/<slug>`; getShortModelName's path
+  // fallback strips the prefix and re-resolves the bare slug through this
+  // table, the same way it handles `accounts/fireworks/models/<slug>`.
+  'qwen3.7-max': 'Qwen 3.7 Max',
+  'mimo-v2.5-pro': 'MiMo v2.5 Pro',
+  // Both spellings occur in the wild: OpenRouter gap-filled keys are lowercase
+  // slugs while sessions report the capitalized name (see the case-insensitive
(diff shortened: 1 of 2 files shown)


ozymandiashh:
Maintainer note: the `check` failure was the `Co-Authored-By: Claude ...` trailer, which this repo blocks in commit messages. I amended the commit in place to drop that one line (diff, authorship and the rest of the message untouched) and force-pushed, so remember to `git pull --rebase` before any further local work on this branch. Review of the substance follows separately.

ozymandiashh (approved):
Verified locally (CI here does not run tests): 150/150 in models.test.ts, tsc clean. Display-only, each non-obvious mapping decision is argued in a comment and pinned by a test - the variant-collapse guard for gpt-5.6-*, the model-vs-harness split for grok-4.5, the path fallback for cline-pass/ slugs and the case-duplicate for MiniMax. Merging.

PR #895, opened 2026-08-03
fix(cline-cli): gate the rollup fallback on metrics presence, not emitted calls

### LINKED ISSUE #894: cline-cli: rollup fallback double-counts a session whose per-message calls were all deduped
Follow-up from the adversarial review of #891 (which shipped). Not a blocker: it over-reports and needs an abnormal on-disk layout, but it is a real money-path edge case worth closing.

## The bug

In `src/providers/cline-cli.ts`, the rollup fallback is gated on `emitted > 0`, but `emitted` is incremented only for calls that are actually yielded. The `seenKeys` dedup check runs before the increment:

```ts
if (seenKeys.has(deduplicationKey)) continue   // deduped: emitted NOT bumped
seenKeys.add(deduplicationKey)
...
emitted++
yield { ... }
...
if (emitted > 0) return
const rollup = parseMetrics(isRecord(metadata['usage']) ? metadata['usage'] : null)
```

So if a session has per-message metrics but every one of them is suppressed by the shared dedup, `emitted` stays 0 and the parser falls through to the `metadata.usage` rollup, emitting the session's cost a second time under `cline-cli:<id>:rollup`.

## Repro

Two session directories whose internal `session_id` collides (e.g. a copied session dir), same message ids, the second carrying a `metadata.usage` rollup:

```
sessions/aaa/aaa.json + aaa.messages.json   session_id "shared", msg_0 cost 0.01
sessions/bbb/bbb.json + bbb.messages.json   session_id "shared", msg_0 cost 0.01, metadata.usage rollup

reported: 2 calls, cline-cli:shared:msg_0 + cline-cli:shared:rollup, total $0.02
true    : one session, $0.01
```

## Fix

Gate the rollup on whether the session had any per-message metrics at all, set before the dedup check, not on the post-dedup `emitted` counter:

```ts
let hadMetrics = false
...
const metrics = parseMetrics(message['metrics'])
if (!metrics) continue
hadMetrics = true
...
if (seenKeys.has(deduplicationKey)) continue
...
// after the loop
if (hadMetrics) return   // instead of `if (emitted > 0) return`
```

A session whose messages were all deduped then correctly declines the rollup. Add a test with the duplicate-`session_id` layout above, mutation-checked against the current `emitted`-based gate.

ozymandiashh:
Closes #894 with exactly the shape the issue prescribes: `hadMetrics` set before the dedup check gates the rollup, and the now-consumerless `emitted` counter is gone.

Mutation check done as requested: the new duplicate-`session_id` test (rollup on the second-visited directory, which is the layout that distinguishes the two gates) fails against the old `emitted`-based gate with 1 failed / 33 passed, and passes post-fix at 34/34. The fixture gained a `sessionIdInMeta` override so a test can model a copied session directory without renaming it on disk.

`tsc --noEmit` clean, no new dependencies. Built with the same review flow as #891: verified locally since CI does not run the suite.

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/Users/husamsoboh/codeburn/node_modules
--- a/src/providers/cline-cli.ts
+++ b/src/providers/cline-cli.ts
@@ -265,12 +265,17 @@ function createParser(source: SessionSource, seenKeys: Set<string>): SessionPars
 
       const messages = isRecord(doc) && Array.isArray(doc['messages']) ? doc['messages'] : []
       const userMessage = firstUserMessage(messages)
-      let emitted = 0
+      let hadMetrics = false
 
       for (const [index, message] of messages.entries()) {
         if (!isRecord(message) || message['role'] !== 'assistant') continue
         const metrics = parseMetrics(message['metrics'])
         if (!metrics) continue
+
+        // Track BEFORE the dedup check: the rollup gate must see metrics-bearing
+        // messages even when every one is later suppressed by the shared dedup,
+        // or a duplicated session_id double-counts through the rollup - #894.
+        hadMetrics = true
 
         const modelInfo = isRecord(message['modelInfo']) ? message['modelInfo'] : {}
         const model = nonEmptyString(modelInfo['id']) ?? sessionModel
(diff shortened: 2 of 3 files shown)


iamtoruk:
Thanks for this, and sorry for the collision. I filed #894 and then landed the same fix in #896 (merged) a few minutes before this came in, so #894 is already closed with exactly this shape: `hadMetrics` set before the dedup check, the `emitted` counter removed, and the duplicate-`session_id` regression test mutation-checked against the old gate.

Closing as a duplicate, no reflection on the work, our two fixes are the same. One thing yours does nicer: the `sessionIdInMeta` fixture override to model a copied session directory without renaming it on disk is cleaner than the two-dirs-by-hand approach that landed. If you want to send that fixture improvement on its own, I would take it.

Issue #898, opened 2026-08-04
ci: actually run tsc and the vitest suite on pull requests


ozymandiashh:
## The gap

None of the required checks run the test suite or the typechecker. Today's `check` is the co-author-trailer gate, `appx` builds the Windows package, `semgrep` scans, `assess` is firstlook. So a PR can be CI-green with failing tests, and tonight that was nearly consequential twice:

- #893 and #891 landed with all checks green; their 150 and 57 tests only ran because the reviewer ran them locally before merging.
- #896's regression test (and the parallel #895) would have caught the #894 double-count in CI on the original #891 push, had anything executed vitest.

Review discipline covers this while the reviewer pool is two or three people who know to run the suite, but it does not scale past us, and it silently punishes outside contributors whose local environment differs.

## Proposal

A `test` job in ci.yml on `pull_request` + `push: main`:

```yaml
test:
  runs-on: ubuntu-latest
  steps:
    - uses: actions/checkout@v6
    - uses: actions/setup-node@v4
      with: { node-version: 22, cache: npm }
    - run: npm ci
    - run: npx tsc --noEmit
    - run: npx vitest run --exclude tests/cache-refresh-lock.test.ts --exclude tests/cache-refresh-lock-process.test.ts
    - run: npx vitest run tests/cache-refresh-lock.test.ts --pool=forks --poolOptions.forks.singleFork=true
```

Two knowns shape the shape:

1. **`tests/cache-refresh-lock.test.ts` is parallelism-sensitive**: several cases fail under a fully parallel run and pass in isolation (reproduced repeatedly on macOS/APFS on unmodified main; count varies run to run). Running that file serially in its own step keeps it enforced without making every PR roll dice. If it flakes even serially on ubuntu runners, quarantine it behind `continue-on-error: true` with an issue, rather than dropping it.
2. **The copilot fixture cases in `tests/parser.test.ts`** fail on some environments on their first parse (`expected +0 to be 200`). If the ubuntu runner reproduces that, the fixture needs fixing first; the job should land green, not aspirationally red.

Suggest landing it as non-required for a week to observe flake rate, then flipping it to required.

Happy to implement whichever variant you prefer; this wants a maintainer call on required-vs-observing and on the flake-quarantine policy before code.

iamtoruk:
[...]

PR #901, opened 2026-08-04
ci: run tsc and the vitest suite on pull requests

ozymandiashh:
Draft implementation of #898, opened as draft pending the required-vs-observing policy call.

What this run itself will teach us (the reason to open it now): the suite has never executed on an ubuntu runner, so this PR doubles as the experiment for the two environment-sensitive knowns. If the copilot fixture cases in `tests/parser.test.ts` fail here the way they fail on my macOS environment, the fixture needs fixing before this can go required, and we will see that in this PR rather than guess.

Flag syntax (`--exclude` glob, `--poolOptions.forks.singleFork`) verified locally against vitest 3.2.6 before pushing. Not adding it to required checks; that stays a maintainer call on #898.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,6 +30,9 @@ jobs:
       - name: Test suite (parallel)
         run: npx vitest run tests --exclude "tests/cache-refresh-lock*"
       # Single forked worker, so lock contention comes only from the child processes the
-      # tests spawn deliberately.
-      - name: Cache-lock suite (serial)
+      # tests spawn deliberately. Quarantined (reports, never gates): the process
+      # suite still races its own takeover window even serially on slow runners -
+      # tracked in #904; drop continue-on-error once that race is settled.
+      - name: Cache-lock suite (serial, quarantined)
+        continue-on-error: true
         run: npx vitest run tests/cache-refresh-lock.test.ts tests/cache-refresh-lock-corrupt-body.test.ts tests/cache-refresh-lock-process.test.ts --poolOptions.forks.singleFork=true
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,31 @@
+name: Tests
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22.13.0
+          cache: npm
+      - run: npm ci
+      - name: Typecheck
+        run: npx tsc --noEmit
+      # The cache-refresh-lock files exercise a cross-process file lock and are
+      # parallelism-sensitive (they fail under full worker pressure and pass serially -
+      # reproduced repeatedly on unmodified main), so they run in their own serial step
+      # below instead of making every PR roll dice.
+      - name: Test suite (parallel)
+        run: npx vitest run --exclude "tests/cache-refresh-lock*"
+      # Single forked worker, so lock contention comes only from the child processes the
+      # tests spawn deliberately.
+      - name: Cache-lock suite (serial)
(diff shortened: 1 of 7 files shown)


ozymandiashh:
The experiment ran, and the first ubuntu execution of this suite in the repo's history returned exactly the data the draft existed to collect. Six failing cases, five distinct causes, none of them "the workflow is wrong":

1. **`tests/parser.test.ts` - copilot file-purge + durable-orphans (2 cases):** fail on ubuntu identically to macOS (`expected +0 to be 200`, first parse of the copilot fixture). Now reproduced on two platforms, so this is a broken fixture on main, not an environment quirk. Needs a fixture fix before anything can go required.
2. **`tests/cli-durable-totals.test.ts` - provider-filter parity (1 case):** `expected 100 to be close to 101.08`. Same failure AVSRPA1KR reported reproducing on a clean checkout in the #862 thread. Real assertion drift on main; needs investigation (the 1.08 delta smells like a pricing-table or carried-day change the expectation never followed).
3. **`tests/parser-incremental-append.test.ts` - inode-change fallback (1 case):** `expected 8697244 not to be 8697244`. ext4 reuses the freed inode when a file is deleted and recreated immediately, so the "inode changed" premise does not hold on Linux. The test needs to force a genuinely different inode (create the replacement BEFORE unlinking the original, or assert on content/size instead).
4. **`tests/parser-proxy-pricing.test.ts` - case-insensitivity (1 case):** asserts macOS/Windows default-filesystem behavior and correctly fails on case-sensitive Linux. Needs a platform gate or a tmpdir-based case-sensitivity probe.
5. **`tests/cli-status-menubar.test.ts` - config-source filter (1 case):** plain 5s timeout on a 37s file on the shared runner. Timeout bump or file-level `testTimeout`.

Everything else passes on ubuntu, including both serial cache-lock steps - the quarantine held.

Suggested sequence: land the five test fixes (each is small and independent), then flip this from draft with the observing-first week from #898. I can take the five fixes as a follow-up branch on this PR or separately, whichever review shape you prefer.

ozymandiashh:
Run #7 is green: the job now enforces. The draft phase turned into a seven-run excavation of every latent environment sensitivity in the suite - none of them workflow bugs, all of them real test debt this job was proposed to surface:

1. noon-anchored fixture in cli-durable-totals (failed every before-noon run; same fix as 1596220)
2. copilot fixture with absolute 2026-05-01 dates that crossed the durable 90-day age-out on Jul 30 (2 tests)
3. inode-reuse on ext4 breaking the file-replaced premise (create-beside + rename now)
4. case-insensitivity asserted unconditionally; now asserts per-platform semantics
5. 5s default timeout vs real multi-parse CLI spawns (file-level 30s)
6. root vitest glob swallowing the app/ renderer suite whose jsdom lives in app/node_modules (scoped to tests/)
7. cache-refresh-lock-process racing its own takeover window even serially - quarantined behind continue-on-error, tracked in #904
8. context-tree teardown racing an in-flight cache write (rm retries)
9-10. the shared relative-time fixture escaping into yesterday during the first five minutes of UTC hours 0 and 1, at all four sites (midnight clamp)

Policy question from #898 stands: leave as reporting for a week, then flip required. The suite ran 2400+ tests green on the shared runner with the one quarantined step reporting separately.

iamtoruk (approved):
Landing this. It implements #898 and is exactly what would have caught the failures that had been sitting red on main (CI was Semgrep-only). The deterministic root-cause fixes here are better than the global retry I reached for in #914: fs.rm maxRetries for the ENOTEMPTY mid-rm window, longer timeouts for slow shared runners, UTC-midnight date clamps across five fixtures, and a serial (singleFork) cache-lock CI step instead of rolling dice.

Reconciled with #914 on the merged tree: kept this branch's equivalent date fixes for parser.test.ts and cli-durable-totals.test.ts, and dropped the global vitest retry #914 added now that the roots are fixed deterministically. Verified the way this workflow runs it: tsc clean, the parallel step (tests, excluding cache-locks) 2432 passed, the serial cache-lock step 26 passed, and the Tests workflow is green on ubuntu here.

Merging as an observing check. Promoting it to a required/blocking check is a branch-protection setting and stays the maintainer call per #898. Thanks for the thorough experiment and for fixing the flakes at the root.

PR #909, opened 2026-08-04
fix(kiro): estimate input tokens from the full prompt, not a 500-char slice

ozymandiashh:
Money-path accuracy fix from the systematic audit (find -> verify -> adversarial -> manual confirmation + mutation check).

`kiro.ts` `parseChatFile` estimated input tokens from `pendingUserMessage` - the last human turn sliced to 500 chars - while output correctly summed every bot char. So a multi-turn Kiro session, or any final prompt over 500 chars, undercounted input tokens and therefore `costUSD` severalfold. The asymmetry was the tell: output used a full accumulator, input used one truncated turn. The sibling modern-execution path in the same file already does the right thing with an `inputChars` accumulator; this makes `parseChatFile` match it, keeping the 500-char slice for the display `userMessage` only.

Mutation-checked: a 2400-char prompt reports 125 input tokens before the fix, 600 after (the real full-length estimate). `tsc --noEmit` clean, kiro suite green.

### DIFF
--- a/src/providers/kiro.ts
+++ b/src/providers/kiro.ts
@@ -203,12 +203,20 @@ function parseChatFile(data: KiroChatFile, sessionId: string, project: string, s
   if (modelId === 'auto' || !modelId) modelId = 'kiro-auto'
 
   let pendingUserMessage = ''
+  // Accumulate every human turn's full length for the input-token estimate,
+  // mirroring the modern-execution path (which sums inputChars). The prior
+  // code estimated input tokens from pendingUserMessage.length alone - the
+  // LAST human turn truncated to 500 chars - so a multi-turn session, or any
+  // final prompt over 500 chars, undercounted input tokens (and therefore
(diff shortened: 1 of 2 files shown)


iamtoruk (approved):
Reviewed on the merged tree. Correct money-path accuracy fix: parseChatFile estimated input tokens from pendingUserMessage (the last human turn truncated to 500 chars) while output summed every bot char, so a multi-turn session or any prompt over 500 chars undercounted input tokens and costUSD severalfold. The new inputChars accumulator sums every non-identity human turn's full length (the identity skip stays ahead of it), matching the modern-execution path, and the 500-char slice is kept for the display userMessage only. Mutation-checked: reverting to the slice makes a 2400-char prompt report the truncated estimate instead of 600 input tokens, and the test also pins userMessage.length at 500 so display is unchanged. tsc clean, kiro suite green. Good to merge.

PR #911, opened 2026-08-04
fix(context-budget): stop double-counting home skills and CLAUDE.md

ozymandiashh:
Context-budget accuracy fix from the systematic audit (find -> verify -> adversarial -> manual confirmation + mutation check).

When the project directory IS the home directory, `countSkills` built `[~/.claude/skills, <project>/.claude/skills]` - the same path twice - and counted every skill twice; `scanMemoryFiles` likewise read `~/.claude/CLAUDE.md` twice. The context-budget token estimate (skills + memory) was inflated for anyone running from their home dir. Both are now deduped by resolved path.

Mutation-checked: a single home skill counts 2 before the fix and 1 after; the second test confirms a genuinely distinct project skill still counts separately. `tsc --noEmit` clean.

### DIFF
--- a/src/context-budget.ts
+++ b/src/context-budget.ts
@@ -60,8 +60,13 @@ async function countMcpTools(projectPath?: string): Promise<number> {
 }
 
 async function countSkills(projectPath?: string): Promise<number> {
-  const dirs = [join(homedir(), '.claude', 'skills')]
-  if (projectPath) dirs.push(join(projectPath, '.claude', 'skills'))
+  // Dedupe by resolved path: when the project IS the home dir, the home and
+  // project skills dirs are the same directory, and counting both double-counts
+  // every skill (and inflates the context budget).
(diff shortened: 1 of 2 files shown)


iamtoruk (approved):
Reviewed on the merged tree. Correct accuracy fix: when the project dir is the home dir, countSkills built the same skills path twice and scanMemoryFiles read ~/.claude/CLAUDE.md twice, inflating the context-budget token estimate. Both now dedupe by path (path.join already normalizes trailing slashes and .. segments, so the identical strings collapse). Dedup fires only on identical paths, so it can only remove a double-count, never under-count a genuinely distinct project skill. Mutation-checked: reverting the skills dedup makes the one-home-skill case count 2 instead of 1; the sibling test confirms a distinct project skill still counts separately. tsc clean. Good to merge.

Issue #913, opened 2026-08-04
Support for Hermes Agent?


APZmedia:
Is it in the roadmap?

ozymandiashh:
Good news: Hermes Agent is already supported, and has been since v0.9.14 (June 2026), so it is not on the roadmap, it already shipped.

CodeBurn reads Hermes' local SQLite state directly:

- default profile: `~/.hermes/state.db` (or `$HERMES_HOME/state.db` if set)
- named profiles: `~/.hermes/profiles/<name>/state.db`

Hermes keeps durable token accounting at the session level, so CodeBurn reports one entry per Hermes session with input/output/cache/reasoning tokens. It uses the cost Hermes recorded when there is one, and falls back to CodeBurn's own pricing table otherwise. Tool calls (terminal, read_file, patch, ...) are normalized into CodeBurn's tool breakdowns too. Details: [docs/providers/hermes.md](https://github.com/getagentseal/codeburn/blob/main/docs/providers/hermes.md)

Any recent version works (current is v0.9.19). If you have Hermes installed and nothing shows up in `codeburn report`, run `codeburn doctor` and paste the output here and we will take a look.

Closing as already supported. Feel free to comment if anything does not behave as expected.

APZmedia:
Thank you! I'm seeing some issues though

Missing sessions (17 in DB vs 14 reported) Three sessions from today are not being counted. The hermes provider queries sessions WHERE input_tokens + output_tokens + cache_read_tokens + cache_write_tokens + reasoning_tokens > 0, so these sessions should match — but they're being dropped somewhere in the parser or dashboard aggregation.
 
Cost calculation is wrong by 3-4x
Even if all 17 sessions were counted, the LiteLLM-priced cost should be ~$5.52 for the token volumes observed. CodeBurn reports $1.65. 

Possible causes:
Model name resolution failure: Sessions store the name of the model. The getCanonicalName() strips the prefix to get the model, which does exist in LiteLLM. But the pricing lookup may be failing silently (returning 0) or resolving to a different/cheaper model.

Cache read tokens may not be priced: 28M cache read tokens = $1.22 of the expected cost. If these are being dropped or priced at $0, the remaining would still be higher suggesting the issue compounds.

[...]

Issue #890, opened 2026-08-03
Cost modes: reconcile displayed cost with the source's own reported cost (auto/calculate/display)


iamtoruk:
## Motivation

CodeBurn derives every cost from tokens times model pricing. But several sources write their own cost into the log, and where they do, we currently ignore it, so our headline can disagree with what the user sees in the tool itself.

- Claude Code writes a per-message `costUSD`; `claude.ts` never reads it and computes instead.
- crush, codewhale, openclaw, the sqlite-based sources, cline and session-message already read a reported cost, but only as a zero-fallback (used only when our computed value is 0).

There is no way for a user to say "trust the source's own number" or "always recompute", and no provenance in the output distinguishing a computed cost from a reported one. This adds an explicit cost mode, so the displayed number can match the source of truth when one exists.

## Semantics

A `--cost-mode` selector, mirrored in config and JSON:

- `calculate` - always compute from tokens times pricing (today's behaviour).
- `display` - prefer the source's own reported cost where the log carries one; a source with no reported cost is shown as a computed estimate, never a silent 0.
- `auto` (default) - use the reported cost when it is present AND the source is on the known-reliable list, otherwise compute. Conservative: numbers only move where we have a trustworthy reported cost.

## Non-goals (be honest)

- This does NOT fix the assumed-model divergence for sources that carry no cost field (Cursor computes against an assumed model and its data has no dollar figure to read). That is a separate discovery and pricing accuracy track.
- No new network calls. Reconciliation against a billing API is out of scope here (it overlaps #725 Part 5).

## Architecture

Cost is baked in at parse time and stored in the daily cache, so the mode has to be applied at read time, which means storing both numbers.

1. Add `reportedCostUsd?: number` to the parsed-call type, distinct from the computed `costUSD`. Populate it in the sources that read a reported cost (promote the existing zero-fallback reads to always capture it) and add Claude Code's per-message `costUSD`.
2. Carry both computed and reported cost through the daily rollup (`day-aggregator`), which is a `DAILY_CACHE_VERSION` bump (16 -> 17) riding the existing carry-forward so old days are preserved, not dropped. Same discipline as the recent structural-discovery bump.
3. One shared read-time selector applies the mode when a report is rendered, so the CLI, TUI, JSON and menubar can never disagree about which number they show.
[...]

PR #915, opened 2026-08-04
fix(parser): fold SQLite -wal siblings into source fingerprints

ozymandiashh:
## Symptoms (reported in #913)

A Hermes user with 17 token-bearing sessions in `~/.hermes/state.db` sees only 14 in reports; the 3 sessions started today are missing entirely.

## Root cause

Hermes (and Cursor, OpenCode, copilot OTel) keep their SQLite state DBs open in **WAL mode** for the life of the agent process. Committed writes park in `<db>-wal` until a checkpoint runs, so the main file's mtime can sit hours or days behind the newest committed data. Real-world stat from a live Hermes install (journal_mode=wal):

```
-rw------- 626180096 2026-08-04 03:21:24 state.db      <- 8h stale
-rw-------   7300672 2026-08-04 03:23:47 state.db-wal  <- newest commits
```

`fingerprintFile` only statted the main file, which broke two ways:

1. **Sessions vanish from reports.** The date-range mtime pre-filter in `parseProviderSources` (`fp.mtimeMs < dateRange.start` -> skip) reads the stale mtime as "nothing in range" and skips every source backed by that db. The today-parse (range start = local midnight) skips the whole db whenever the last checkpoint predates midnight, while the backfill union only keeps days through yesterday. Net effect: sessions started today after the last checkpoint appear nowhere. That is exactly the 17-in-DB / 14-reported split in #913.
2. **Stale cached data between checkpoints.** `reconcileFile` sees an unchanged fingerprint and serves cached turns, so sessions that grew since the last checkpoint report old totals until a checkpoint happens to run.

## Fix

Fold the `-wal` sibling into the fingerprint: newest mtime wins, sizes add. Both WAL growth and a checkpoint (db grows, wal truncates) move the fingerprint. Applied to the two virtual-suffix fallbacks (`#`, `:`) and to bare paths only when the extension is a SQLite one (`.db`/`.sqlite`/`.sqlite3`/`.vscdb`), so JSONL transcript fingerprints and their offset-based append detection are untouched. `-shm` is deliberately ignored: it mutates on reads and would churn the fingerprint without any data change.

## Testing

- 5 new unit tests for `fingerprintFile` (wal fold for `#` / `:` / bare-db forms, no-wal passthrough, non-SQLite paths never folded)
- 1 end-to-end regression: hermes session inside the range, main db utimes'd to before the range, fresh `-wal` sibling -> `parseAllSessions` must still report it
- Mutation check: with the src change stashed, exactly the 4 wal-fold tests fail; restored, 78/78 in the two touched files
- Full suite: 2702 passed / 0 failed (the two known-flaky files on this machine, cache-refresh-lock and menubar-json timeouts, passed on rerun; the 26 jsdom unhandled errors are a pre-existing local node_modules condition, present on clean main)

Fixes the missing-session half of #913. The cost-accuracy half is a separate discussion (recorded-cost precedence, #890 provenance) and is being handled on the issue.

### DIFF
--- a/src/session-cache.ts
+++ b/src/session-cache.ts
@@ -607,34 +607,65 @@ async function retryCacheFileMutation(operation: () => Promise<void>): Promise<b
 // append-only transcripts keep changing. Fixing this properly means
 // multi-file fingerprints per source.
 
+// SQLite database files by extension. Bare-db sources (copilot's
+// agent-traces.db) and the virtual-suffix bases below all match one of these.
+const SQLITE_DB_PATH = /\.(db|sqlite3?|vscdb)$/i
+
+/// Fingerprint a SQLite database file together with its `-wal` sibling.
+///
+/// A database in WAL mode parks committed writes in `<db>-wal`; the main
+/// file's stat only moves on checkpoint, and a long-lived writer connection
+/// (hermes, cursor and opencode keep their state DBs open for the life of
+/// the agent process) can defer checkpoints for hours or days. A fingerprint
+/// built from the main file alone then (a) carries an mtime older than the
+/// newest committed data, so the date-range mtime pre-filter in
+/// parseProviderSources skips the source and sessions committed after the
+/// last checkpoint never parse (issue #913: today's Hermes sessions missing
+/// from every report), and (b) does not change between checkpoints, so
+/// reconcileFile keeps serving stale cached turns for sessions that grew.
+/// Folding the WAL sibling in fixes both: the newest mtime wins, and the
+/// sizes add so both WAL growth and a checkpoint (db grows, wal truncates)
+/// move the fingerprint. `-shm` is deliberately ignored — it mutates on
+/// reads too and would churn the fingerprint without any data change.
+async function fingerprintSqliteFile(dbPath: string): Promise<FileFingerprint | null> {
+  try {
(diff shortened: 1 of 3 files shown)


iamtoruk (approved):
Reviewed in depth on the merged tree. Real data-completeness fix: SQLite state DBs kept open in WAL mode park committed writes in <db>-wal until a checkpoint, so the main file's mtime goes stale, the date-range mtime pre-filter skips the source, and today's sessions committed after the last checkpoint never parse (the 17-in-db / 14-reported split in #913). Folding the -wal sibling in (max mtime, sizes added) moves the fingerprint on both WAL growth and checkpoint; -shm correctly ignored. Gated to bare SQLite paths (.db/.sqlite/.sqlite3/.vscdb) and the #/: virtual-suffix forms, and for a non-SQLite base it degrades to exactly the old fingerprint, so JSONL append-detection is untouched. Mutation-checked: neutering the fold to main-only fails exactly the four wal-fold tests (the #/:/bare-SQLite cases and the Hermes checkpoint-stale e2e), while the no-wal and non-SQLite cases still pass. tsc clean; the two touched suites 78/78; full suite green (2438 parallel + 26 serial cache-locks, 0 failed). Good to merge.

PR #929, opened 2026-08-05
fix(parser): keep both halves of a midnight-straddling turn, and stop --provider leaking claude

ozymandiashh:
Ports upstream fixes for two defects this branch never received. Both are conservation bugs: work that happened is not where the report says it is.

## A turn spanning local midnight was filtered as a unit

The range and day filter keyed on the turn's **first** call, so every later call that landed in the requested day was discarded along with it. A long autonomous Codex run, or Claude work that crossed midnight, made `codeburn today` under-report until the turn ended — and multi-day totals attributed the whole turn to its start day.

Filters now slice **inside** the turn. Only the calls inside the requested window survive, and the turn's timestamp re-anchors to its first surviving call, so turn-anchored rollups (category, editTurns, oneShotTurns, the daily cache) land on the day the retained calls actually happened. Cost, calls, savings and tokens bucket under each call's own local day, so day N + day N+1 conserves the whole-range total.

A sliced turn is still **classified from its full call list** — category, hasEdits and retries describe the whole exchange, not the surviving slice. That matches the Claude path and is deliberate.

## The inverse leak, on provider-filtered runs

`--provider <other>` still entered the claude scan, whose orphan pass read the entire cached claude section, treated every cached PR-bearing transcript as no-longer-discovered, and re-injected it. **By Project and By Model listed Anthropic spend under `--provider cursor`** while the headline showed cursor alone.

Now guarded by an explicit in-scope check — deliberately *not* a directory-count check, so when claude is in scope but every transcript has been pruned, PR-attributed orphans still survive.

## Blast radius

Daily-history rows and the JSON daily fallback change shape for straddling days: call-derived values move to the call's own day. Session totals and the session-cache format are unchanged.

Measured on a reviewer's full real corpus, lifetime, all providers: **−51 calls and −$25.95**. Every difference is boundary-only in both directions — the old filter leaked post-window calls into a window (because the whole turn rode its first call in) and dropped in-window tails (because a turn anchored before the window was discarded whole). Nothing in the interior of a window moves. Window totals now conserve to the cent across a day split.

Desktop history bars will **visibly redistribute** on straddling days: spend that used to sit entirely on day N now shows the post-midnight share on day N+1. This is contract-safe — the JSON payload shape is unchanged and no app change is needed — but it is a visible number change on those days, not a no-op.

## Daily cache version

This PR bumps `DAILY_CACHE_VERSION` / `MIN_SUPPORTED_VERSION` to **24**, the next free number after #926's 23.

Per-call day slicing changes *which day* a historical straddling call lands on, so every rollup finalized by an earlier binary holds calls that now belong to the next day. Nothing downstream can notice on its own — `usage-aggregator` serves every day before today from this cache and retention is ten years — so an upgrading user would keep the stale whole-turn day N forever while day N+1 grew the sliced half. Raising `MIN_SUPPORTED_VERSION` forces the one-time re-derivation. It cannot ride on 23: a cache already written at 23 by a #926 binary would be adopted as current and complete, and the straddle re-derivation would never fire.

## Two behavior changes riding along, named explicitly

1. **PR-attribution freeze rule.** The `prRefsAtRangeStart` seed now freezes on the first turn that *contributes in-range calls* rather than the first turn whose unsliced first call is in range, and the frozen set is unioned into the session's `prLinks`. Without it, a range whose first content is a straddling turn's tail lost the pre-range PR row entirely. This **changes By-PR rows for sessions attributed by explicit reference** (`prAttributionSource: 'explicit-reference'`): spend that previously fell out of the By-PR table now appears under the PR that was active entering the range.
2. **Re-derive window widened 365 → 3650 days.** The forced re-derivation now walks the full retention window (`DAILY_CACHE_RETENTION_DAYS`) instead of the 365-day product backfill, so an old-but-still-sourced day is corrected too rather than keeping its pre-fix bucketing forever. One-time cost: the first run after upgrade parses every surviving source in the retention window, not just the last year. Days whose sources are gone are carried forward untouched, as before.

## One caveat, stated rather than left to be found

The multi-day all-provider **By Activity** rollup still derives today's slice from the unsliced range parse, so a straddling turn's category cost stays anchored on its start day and categories can sum below the headline on that surface. Upstream has a follow-up; it should land next.

### DIFF
--- a/packages/cli/src/daily-cache.ts
+++ b/packages/cli/src/daily-cache.ts
@@ -5,7 +5,31 @@ import { homedir } from 'os'
 import { join } from 'path'
 import type { DateRange, ProjectSummary } from './types.js'
 
-// Bumped to 15: per-project daily rollups. Days and provider slices now carry
+// Bumped to 17: range and day filters now slice a midnight-straddling turn
+// instead of filtering it as a unit (issue #852), so call-derived values
+// (cost, calls, savings, tokens) bucket under each call's own local day. A
+// v15 rollup finalized by the pre-fix binary holds the WHOLE turn on its
+// start day, and the new slicing then also puts the post-midnight half on
+// the next day — the same cost twice. Nothing downstream can notice on its
+// own: `usage-aggregator` serves every day before today from this cache, and
+// retention is ten years, so an upgrading user would keep the stale
+// whole-turn day N forever while day N+1 grew the sliced half. Raising
+// MIN_SUPPORTED_VERSION forces the one-time re-derivation: the new version
+// mints a fresh filename, adoption marks the result incomplete, and the next
+// hydration rebuilds every day within the retention window whose sources
+// survive under per-call bucketing — the re-derive parses the FULL retention
+// window, not the 365-day product backfill, so an old-but-still-sourced day
+// gets corrected too (days whose sources are gone are carried forward).
+//
+// v16 is SKIPPED: main already spent it on the codex structural-discovery
+// fix (eece4cf, #873/#626), which raised these same two constants to 16 with
+// a DIFFERENT meaning. Claiming 16 here too would make this binary load a
+// main-built v16 cache — which holds only the codex fix — as current and
+// complete, so the straddle re-derivation would never fire for anyone who
+// ever ran a main build. The next bump must take the next free number, not
+// the last one main used.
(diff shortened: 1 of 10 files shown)


ozymandiashh:

PR #930, opened 2026-08-05
fix(cli): restore pi/omp, cline and opencode session discovery

ozymandiashh:
Three discovery paths silently dropped sessions. Each fix carries a regression test that fails against the current code.

## pi/omp — invisible after an upstream format change

Discovery inspected only the **first physical line** of each transcript. When Oh My Pi started writing a `type: "title"` slot line ahead of the session header, every OMP session became invisible to codeburn.

Discovery now scans a bounded twenty leading lines for the session record, skipping blank and malformed lines rather than giving up on the first miss. The bound matters: without it a message-only or pathological file degrades into a full-file read.

## cline — only stable VS Code was scanned

Sessions written under **Code - Insiders** or **VSCodium** were never found. All variant roots are scanned now — matching what the roo-code and kilo-code siblings already do — plus Cline's home-data root, deduplicated by task id with the newest copy winning.

## opencode / kilo-code — a query against a column that does not exist

This is the worst of the three. The session-level fallback query selected `model_id`. Neither schema has that column; both store `model` as JSON text. **On any real database the query threw**, the fallback returned null, and the zero-yield session rollup never emitted a call — so usage for interrupted or user-only sessions was simply missing, with nothing surfaced to say so.

The query now reads the real column and reports `providerID/id`, which the display-name and pricing paths already canonicalize.

## Cache handling

Both parse versions are bumped so already-cached sessions re-parse once and pick up the working fallback. These are **per-provider fingerprints**, not the shared daily-cache constant — so unlike the sibling ports, they do not collide on the same line.

## One caveat

pi/omp discovery now uses the streaming reader with its 4 GB cap while parse still caps at 128 MB. A file between those sizes is listed at discovery and skipped at parse. No usage delta, but the asymmetry is real and worth knowing about.


ozymandiashh:

PR #931, opened 2026-08-05
fix(sync): key the device, span and trace digests

ozymandiashh:
The sync path derived three identifiers with bare SHA-256 and sent them to a configured endpoint.

## What was being sent

**`deriveDeviceId`** hashed `hostname:username`, truncated to 64 bits, commented *"pseudonymous, stable"*. An unkeyed digest of a host/username pair is not pseudonymous against anyone who can guess plausible values — hash the guess, compare, done.

**`deriveSpanId`** hashed the dedup key. For pi, zerostack, lingtai-tui and codebuff that key **embeds the raw absolute source path, home directory included**, because the bridge passes `source.path` straight through to the decoders. Guess a plausible home and project name and the same confirmation works.

## Why this is the project's own standard

Decision **D1** requires a caller-supplied HMAC key for fingerprints precisely so that digests of paths cannot be dictionary-attacked, and `packages/core/src/fingerprint.ts` throws on an empty key to enforce it. Every fingerprint that goes through core obeys this. The sync path bypassed the primitive and called `createHash` directly.

## The fix

All three ids are now HMAC-SHA256 under the **per-install privacy key** — the same key core's fingerprints use — with domain prefixes (`sync-span:`, `sync-trace:`, `sync-device:`) so one value appearing in two positions never yields the same digest. The derive functions **throw** on an empty key rather than degrading to unkeyed. The payload builder obtains the key itself, so the decode path — which runs with an empty key by design — never reaches them.

Determinism is preserved: the key is stable per install, so re-sends stay byte-identical and server-side dedup still holds.

## Scope, stated honestly

Sync is opt-in and needs an endpoint plus credentials. The digests are of identifiers, not prompts or file contents. This predates the extraction. **It is not an active leak of user content** — it is a weak construction in a project that already knows how to do this properly, on the one surface that skipped the primitive.

## Blast radius

Every id is re-keyed once at upgrade, so anything already pushed stops correlating with new sends and the backend sees a fresh device identity per install. Ids are stable afterwards unless the key file is lost, which re-keys again. The host-side sent-ledger keys off the raw dedup key and is unaffected, so re-push filtering keeps working. First sync push creates the key file if the optimize detectors have not already.

### DIFF
--- a/docs/sync/DEVELOPER.md
+++ b/docs/sync/DEVELOPER.md
@@ -105,12 +105,31 @@ Strict protobuf-JSON mapping of `ExportTraceServiceRequest`. lowerCamelCase fiel
 
 ### Span identity (deterministic)
 
+All ids are HMAC-SHA256 keyed by the per-install host privacy key (decision D1,
+see `packages/cli/src/privacy-key.ts`) with a role prefix — never a bare
+SHA-256, which would be confirmable by dictionary attack:
+
 ```
-span_id   = first 8 bytes of SHA-256(deduplicationKey) → hex (16 chars)
-trace_id  = first 16 bytes of SHA-256(sessionId)       → hex (32 chars)
+span_id   = first 8 bytes of HMAC-SHA256(privacyKey, "sync-span:" + deduplicationKey) → hex (16 chars)
+trace_id  = first 16 bytes of HMAC-SHA256(privacyKey, "sync-trace:" + sessionId)      → hex (32 chars)
 ```
 
-Re-sends are byte-identical. Server-side dedup is defense-in-depth.
+The key is generated once per install, persisted in the codeburn config dir,
+and never leaves the host, so re-sends are byte-identical on the same machine.
+Sync REQUIRES that persisted key: push aborts with an error if the config dir
+is unwritable (no per-process fallback key) or a key file exists but does not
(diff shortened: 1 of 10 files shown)


ozymandiashh:

ozymandiashh:
Cross-PR note from an independent review of #933: two decoders — `zerostack/decode.ts` and `lingtai-tui/decode.ts` — concatenate the **raw absolute source path** directly into `dedupKey`, and a reviewer produced a schema-valid envelope containing one. Since this PR keys span ids off `deduplicationKey`, the HMAC input for those providers is a real host path.

That does not weaken this change — keying is exactly what makes it safe to hash a path. But it means the two PRs are coupled: if #933 ends up fingerprinting `sourceRef` inside those decoders instead, the dedup key shape changes and the span ids derived here change with it. Whichever lands second should re-check the goldens.

iamtoruk:
Closing as absorbed: #932's first commit is patch-identical to this PR (verified via git patch-id), and its second commit additionally fixes this branch's race test (awaitValidKey + import.meta.url-anchored fixture path). Landing #932 lands all of this; review notes from here (CHANGELOG entry for the identity re-key + the two new hard-abort conditions) are being folded into #932.

PR #932, opened 2026-08-05
feat(sync): push git attribution spans, with the hardening that followed

ozymandiashh:
Ports three upstream commits this branch never received:

- `1bf7206` — push git attribution spans with `--attribution` (the feature)
- `ccee28a` — review hardening: cwd-fallback egress, Windows paths, PR-link validation
- `50c8251` — **close credential-leak paths**, session retraction, span/key/CLI hardening

`grep -ri attribution packages/cli/src/sync/` returned nothing on this branch; upstream's `push.ts` has several hits. The whole feature and its security follow-up were missing.

## Ported as an end state, not in sequence

Two and three revise one. Replaying them in order would have introduced the issues they fix and then removed them again — and anything missed on the third pass would have shipped a feature with a reopened security hole in the same change. That is the specific way this port could have gone wrong, so the credential-leak paths from `50c8251` are **enumerated and checked off individually** against the result rather than assumed to arrive with the feature.

## Scope

`packages/cli/src/sync/{cli,otlp,push}.ts`, `yield.ts`, the mock IdP fixture, docs, changelog, plus two new test files covering the feature and its CLI surface.

Nothing here touches the id derivations in `otlp.ts` — those are #931's subject. **No new unkeyed digest is introduced.**

## Verification

34 new attribution tests green, 93 across the seven sibling yield/sync suites, core 509, typecheck clean in both packages. The Playwright sync E2E self-skips without its three env vars, so it stayed inert.


ozymandiashh:

ozymandiashh:
**Merge-order blocker, from independent re-review.**

This PR adds a new attribution payload that sends `device_id`, span ids and trace ids — all still derived with bare SHA-256 on this branch — and introduces `stateHash` over user-derived state. #931 is the PR that keys those derivations, and it is **not an ancestor of this branch**.

So landing this alone ships new consumers of the unkeyed construction, and my earlier claim in the commit message that this is "not a D1 violation" is contested: one reviewer argued `stateHash` hides nothing because its inputs ride in cleartext in the same payload, another argued the derivations it feeds are exactly what D1 governs. The second reading is the safer one to act on.

Concrete requirement: **#931 must land before this**, and whoever merges should reconcile the one-argument call sites this PR adds against #931's keyed signatures. I would rather flag this than have it discovered at merge.

ozymandiashh:
**Correction to the note above, from the final cross-PR review.**

I wrote "land #931 first, or reconcile the attribution call sites into it." The *or* is wrong: `git merge-tree` reports **no conflict in either direction**, and the merged file does not compile regardless of order — #931 removes the `createHash` import that this PR's `stateHash` uses, and this PR's two call sites use the one-argument signatures #931 replaces. Rebasing replays the same call sites.

Reconciliation is **mandatory**, not an alternative. A reconciliation commit threading `privacyKey` into the attribution derivations has to land with this PR.

Worth stating why it matters beyond the build: if someone resolves it by loosening the signature instead, those two call sites are exactly the unkeyed span ids #931 exists to eliminate — reintroduced in brand-new code.

PR #933, opened 2026-08-05
fix(core): bound the model field, and stop the smuggling test certifying the hole

ozymandiashh:
Three contract defects in `@codeburn/core`'s published observation surface. 0.9.20 is already on npm, so an external consumer meets all three today.

**Framing first, because it changes what these are.** Nothing in this repo emits provider-derived observations: `packages/cli/src/optimize.ts` is the only file outside core that builds an envelope, and it hardcodes `dedupKey: 'host'` and `model: 'host'` to feed the detectors from host-side tool calls. So these are **latent API defects, not a live leak** — worth fixing before anything starts emitting, and before more consumers build on the schema.

## `model` was unbounded free text

`z.string().min(1)`, no max, no pattern — inside a schema whose docstring three lines above states that no field can hold free text. For vercel-gateway the model string arrives from a **fetched report**, so it is externally influenced.

Both `model` and `pricingModel` now use a bounded identifier pattern, following the existing `CanonicalToolName` convention rather than inventing a new shape. The character set was measured against real values across the provider decoders (`openai/gpt-4o`, `anthropic--claude-4.6-opus`, `us.anthropic.claude-3-5-sonnet-v2:0`, `@cf/meta/...`) and the cap sits comfortably above the longest slug in the pricing snapshot.

## The smuggling test asserted the leak survives

It planted a prompt in `model` and then asserted the prompt was **still there**, under the name *"contains the model secret (identifier-exemption convention)"*. A guard test that certifies the hole it exists to catch is worse than no test — it makes the gap look considered.

The assertion is inverted: a planted prompt is now rejected at validation, while a legitimate identifier-shaped slug still passes and is asserted explicitly so the exemption's real intent survives. **The sibling test in the vercel-gateway suite had the same shape and is fixed too** — this was a class, not a site.

## The contract described a fingerprint; every caller passed a raw path

`DecodeContext.sourceRef` was documented as *"an opaque fingerprint of the source"*. The bridge passes an absolute filesystem path, and zerostack, pi, lingtai-tui and codebuff fold it into dedup keys.

Changing the callers would break session identity that parity goldens certify — codebuff and pi parse structure out of that path — so the **contract now describes what is actually passed**, with an explicit policy for where the value may travel: into dedup keys and identity derivation, never into an envelope fingerprint field.

## One defect documented rather than fixed

The copilot JetBrains decode builds a dedup key from an unkeyed digest of assistant reply text. That path has no privacy key available, and routing it through core's fingerprint would break the bridge's documented key-free contract and move a golden — a maintainer decision, not a silent change. The constraint and the proposed fix are recorded at the site.

Core suite 509 green; no parity golden moved.

### DIFF
--- a/packages/cli/src/daily-cache.ts
+++ b/packages/cli/src/daily-cache.ts
@@ -5,7 +5,28 @@ import { homedir } from 'os'
 import { join } from 'path'
 import type { DateRange, ProjectSummary } from './types.js'
 
-// Bumped to 15: per-project daily rollups. Days and provider slices now carry
+// Bumped to 17: dedup-key hygiene (#931, this PR). The codebuff, zerostack,
+// pi/omp and grok decoders now thread a FINGERPRINT of the source path into
+// their dedup keys (and lingtai-tui normalizes the model component) instead of
+// the raw path / raw ledger text, because dedupKey ships on the observation
+// envelope. Unchanged files are served from the session cache, whose dedup
+// sets are seeded from the CACHED keys — so a warm cache built by the pre-fix
+// binary keeps the raw-path keys, the same records re-ingest under the new
+// key shape, totals go inconsistent, and the raw path stays on disk forever.
+// The per-provider parse versions (session-cache.ts PROVIDER_PARSE_VERSIONS)
+// force the session-cache re-parse that drops those keys; this bump forces the
+// daily cache to re-derive every day whose sources survive instead of serving
+// the pre-fix rollups. Raising MIN_SUPPORTED_VERSION to 17 makes a v16 file
+// load as an old-version file rather than the trusted current cache.
+//
+// v16 is SKIPPED: main already spent it on the codex structural-discovery fix
+// (eece4cf, #873/#626). A user who has ever run a main build owns a v16 cache
+// containing only the codex fix; claiming 16 here too would load that file as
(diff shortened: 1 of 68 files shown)


ozymandiashh:
Update after a sibling review: the copilot JetBrains unkeyed digest, which this PR documented rather than fixed, **is** fixed in #935. That review established the privacy key is actually available in `decodeCopilot`'s context, so it is now HMAC-keyed there and the bridge goldens were recomputed with the real decoder.

The constraint this PR recorded still holds and is worth keeping in view: the bridged CLI path passes an empty key, so full closure needs a real key threaded into rich decode. The note at the site in #935 says so.

Treat the "documented rather than fixed" paragraph above as superseded.

ozymandiashh:

iamtoruk:
Superseded by #1074 (merged as de536cf5), which lands this PR's work reconciled with #935: the ModelIdentifier bound and smuggling-test flip are in verbatim spirit, with one design change from review — normalization applies only at the observation boundary, never inside dedup keys, so the antigravity goldens stay byte-identical and no cache invalidation is owed. The sourceRef fingerprinting landed whole with its six parse-version bumps plus a privacy-key digest folded into the env fingerprint. Thanks @ozymandiashh — the CHANGELOG credits this PR.

PR #934, opened 2026-08-05
fix(kiro): estimate chat-file input tokens from every human turn, not a 500-char slice

ozymandiashh:
Ports upstream `6c4645a`.

`decodeKiroChatFile` estimated input tokens from `pendingUserMessage` — the **last** human turn, sliced to 500 characters — while output summed every bot character. So any multi-turn session, or any final prompt over 500 characters, under-reported input tokens and therefore `costUSD`, severalfold in the worst case.

It now accumulates every human turn's full character count, matching the modern-execution, ws-session, CLI-session and V2 arms which already did. The 500-char cap stays for the **display** `userMessage` only.

## Scope

The chat-file arm alone. The IDE-file dispatcher routes any record carrying a `chat` array plus metadata to `decodeKiroChatFile`, so this is chat-shaped Kiro IDE files — not every Kiro prompt. Identity preambles and tool content are excluded from the sum, so "full prompt" means the sum of human message characters rather than everything in the file.

One review-noted divergence from main: the identity-preamble guard uses `msg.content.trimStart().startsWith('<identity>')` instead of main's plain `startsWith('<identity>')`. This is directionally correct — it also strips a leading-whitespace identity preamble that main's exact-anchor check would otherwise sum into the input estimate — and it affects the displayed `userMessage` too, not just the token count. Reviewed and kept.

## The parity golden was defending the bug

`kiro-golden.test.ts`'s G2 case was titled *"A1 input tokens derived from truncated prompt"* — it pinned the buggy value, because the corpus was captured before upstream's fix. Fix the decoder and it goes red.

The pin moves from 125 to **750**, derived from the fixture rather than copied from the new output: the prompt is 3000 characters, the estimator is `ceil(chars / 4)`, so 3000/4 = 750 where 125 was 500/4. A comment marks 125 as a pre-fix value so nobody restores it. No other kiro golden encoded the same assumption — checked. For scale on real sessions rather than the fixture: review found a typical multi-turn kiro session going from ~125 input tokens under the old formula to ~1,937 under the corrected one — the old formula had a hard 125-token ceiling (500 chars / 4) no matter how long or how many turns the conversation had.

## Cache invalidation, and what this PR deliberately does not do

The user-visible part is cached history. Without invalidation, already-parsed sessions keep pre-fix tokens and cost forever while new ones use corrected maths — the report silently disagrees with itself.

The kiro **provider parse fingerprint** is bumped here, which discards the provider section and forces a re-parse of unchanged chat files. Verified independently: that check runs per provider section at scan time and has nothing to do with the daily cache.

The **daily rollup** constant is also bumped here, to **26** — the next free number after the landed sequence 23 (#926) / 24 (#873) / 25 (#930). This paragraph originally claimed the opposite (that this PR does not bump `DAILY_CACHE_VERSION` and that #926 carries it); that was wrong even at the time this PR was opened — the diff bumps the constant — and is now also stale, since #926 landed owning 23 and the base has since moved through 24 and 25 for unrelated fixes. The kiro chat-file fix changes historical kiro chat-file numbers, so cached daily rollups finalized before this fix need the same one-time re-derivation the session cache gets from the fingerprint bump above; riding on someone else's bump would leave that re-derivation to chance depending on merge order.

## Verification

Core 509, kiro 77, cache suites 67 — including a regression test that pins 125 under the old fingerprint and 750 under the new one.


ozymandiashh:
Converted to draft: this had two independent reviews but a third never landed. Re-running it; will mark ready once complete.

PR #935, opened 2026-08-05
test(core): make the privacy guards prove what they claim

ozymandiashh:
Four guards passed without demonstrating the property they exist for. Two turned out to be hiding live holes.

## What was wrong

**import-smoke stubbed seven modules.** It imports every public entry point with I/O replaced by throwing stubs, to show that importing core touches nothing. Global `fetch`, `node:os` (homedir), `node:process` and env, `node:sqlite`, tls, dgram, http2, worker_threads and `createRequire` — which bypasses ESM loader hooks entirely — were all open. Env was *emptied* rather than made to throw, so a read succeeded silently. It also only proved **import** purity: decoder bodies never ran, so a decoder that opened a file when called would have passed. Blocklist extended, env made to throw, and parser, decoder and detector bodies now execute under the same stubs.

**The schema string-field gate walked only part of the schema** — properties, items, definitions — so anything under `anyOf`, `oneOf`, `allOf`, `patternProperties` or a union type escaped both the enumeration and the bound check.

**The diagnostic sanitiser blocked path separators and had no live caller.**

**The userMessage gate greps one identifier**, so a decoder copying user text into a differently-named field was invisible, and eight allowlisted providers had no hostile-decode test.

## Two live holes closed, found by review

**`isolateRecords` let unkeyed detail through the public API.** Only *thrown* errors were protected; a caller returning a diagnostic could still emit an unkeyed fingerprint.

**The copilot JetBrains decode built a dedup key from an unkeyed SHA-256 of assistant reply text**, truncated to 48 bits — dictionary-attackable on short replies like "OK" or "Done." — and it was copied into the emitted envelope. Investigation showed the key **is** available in that decode context, so it is now HMAC-keyed. The bridged CLI path still passes an empty key, so full closure needs a real key threaded into rich decode; that constraint is documented at the site. The bridge goldens that pinned the unkeyed digests were recomputed with the real decoder rather than hand-edited.

## The near-miss worth recording

Wiring the sanitiser into decoders added a `detail` fingerprint where diagnostics had been bare — and the sanitiser's key was **optional**, degrading silently to unkeyed. The bridge decodes with an empty key, so on the real path that digest would have been unkeyed, over an error message, and Node embeds a fragment of the offending input in `JSON.parse` failures. **A hostile blob could have smuggled content out through the field added to prevent smuggling.**

The key is now required and throws like `fingerprint.ts` does. Decoders route through a helper that omits the field entirely when no key exists — a bare throw there would break poison-isolation and take a whole batch down with one malformed record. Four regression tests fail if the keyless path returns.

## Also

Two hostile tests planted secrets in fields their decoders never read, which proved only that ignored fields stay ignored while looking like coverage; they now plant in genuinely-read fields. The env probe uses an explicit sentinel instead of `HOME`, so it cannot false-red on a HOME-less runner. The 256-char bound's rationale is corrected — it is anti-free-text, not anti-credential, since a GitHub PAT is ~93 characters. Dead `byteOffset` surface removed.

556 core tests green, typecheck clean.

### DIFF
--- a/packages/cli/src/providers/bridge.ts
+++ b/packages/cli/src/providers/bridge.ts
@@ -83,7 +83,8 @@ export function createBridgedProvider<TRich>(spec: BridgedProviderSpec<TRich>): 
           if (records === null) return
           // The CLI holds the rich decode only; minimization / fingerprinting
           // happens on the sync path, so an empty privacy key is correct here
-          // (the rich decoder never consumes it), matching claude/codex.
+          // (the rich decoder treats it as "no detail": diagnostics carry the
+          // code + index only, never an unkeyed digest), matching claude/codex.
           const context: DecodeContext = { privacyKey: '', providerId: spec.name, sourceRef: source.path }
           const { calls } = spec.decode({ records, context, seenKeys })
           for (const rich of calls) {
--- a/packages/cli/src/session-cache.ts
+++ b/packages/cli/src/session-cache.ts
@@ -216,7 +216,18 @@ export const PROVIDER_PARSE_VERSIONS: Record<string, string> = {
   codex: 'mcp-attribution-v2-est-cost-rich-capture-v1-cross-provider-pr-v1',
   cursor: 'composer-anchored-crediting-v1-est-cost',
   'cursor-agent': 'workspaceless-transcript-v1',
-  copilot: 'cli-shutdown-cost-v1-skills',
+  // JetBrains dedup-key hardening (D1): the per-turn digest changed from an
+  // unkeyed sha256 to an HMAC keyed with the host's privacy key, so EVERY
+  // cached copilot dedup key's value changed, not just hostile ones. The
+  // session cache seeds its dedup sets from the CACHED keys, and copilot is
+  // the sole DURABLE_PROVIDER_NAMES member — its union-merge never deletes
+  // cached turns, it appends any turn whose keys are not already cached. A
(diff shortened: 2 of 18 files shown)


ozymandiashh:
Back to draft — a final cross-PR review found a critical defect this PR creates in combination with the rest of the batch.

Replacing `createHash` with `createHmac` changes the **value** of every copilot JetBrains dedup key, not just hostile ones. Copilot is the sole member of `DURABLE_PROVIDER_NAMES`, and the durable merge never deletes — it appends any turn whose keys are not already cached. The parse-version fingerprint is the only mechanism that drops old-shape keys.

So the first time a JetBrains transcript grows after this lands, the re-parse appends the session's entire history under new keys while the old copies remain, and the daily-cache bumps landing in the same batch then double-count it.

A sibling PR established exactly this rule and bumped a parse version for every provider whose key shape it changed. This one changed copilot's and did not. Fix in progress.

iamtoruk:
Superseded by #1074 (merged as de536cf5), which lands this PR whole — diagnostics keyed details, isolateRecords strip, import-smoke and architecture-gate strengthening, and the copilot JetBrains HMAC with its parse-version bump — reconciled with #933's key threading so the HMAC runs under the real per-install key, with goldens recomputed under a stubbed fixed key (machine-independent). Validation on the combined result also surfaced and fixed a copilot double-count on privacy-key rotation, now pinned by a permanent end-to-end test. Thanks @ozymandiashh.

PR #936, opened 2026-08-05
test(qwen): pin the orphaned-record dedup key shape

ozymandiashh:
The smallest item from the review pass, and the one that says the most about the extraction's test coverage.

The pre-extraction decoder built dedup keys with template interpolation, so a record missing `sessionId` or `uuid` produced the literal `qwen:undefined:<uuid>`. Phase 8.1 changed the spelling to `qwen::<uuid>` by coalescing both fields to an empty string.

**The intent was right** — a missing identifier should contribute nothing rather than a fake value. But the change shipped with nothing behind it: no test, no fixture, no golden pinned either spelling, and the CLI parity golden only exercises fully-formed records, where the two are byte-identical. **Nobody could tell the change from a defect.** That is what this PR fixes — not the shape, the absence of a contract.

The test pins the orphan spellings, asserts the collapsed key still dedups identical records, and fails against the pre-migration spelling (verified red: `qwen:undefined:a-orphan` against the pinned `qwen::a-orphan`). The decode-site comment records the reasoning so nobody "restores" the old one.

## Two caveats, worth knowing rather than discovering

- Records with an explicit `null` coalesce the same way as missing ones, so two previously-distinct keys now collapse into one.
- Keys persisted by pre-8.1 builds for orphaned records will not match the new spelling, so such a record can be counted once more across the upgrade. The qwen corpus is not in the frozen golden set, so no shipped fixture moves.

### DIFF
--- a/packages/core/src/providers/qwen/decode.ts
+++ b/packages/core/src/providers/qwen/decode.ts
@@ -128,6 +128,11 @@ export function decodeQwen({ records, seenKeys: liveSeen }: QwenDecodeInput): Qw
     const candidatesTokenCount = usage.candidatesTokenCount ?? 0
     if (promptTokenCount === 0 && candidatesTokenCount === 0) continue
 
+    // Deliberate shape: the pre-extraction decoder interpolated the raw fields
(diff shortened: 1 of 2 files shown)


ozymandiashh:

Issue #899, opened 2026-08-04
doctor: extend probeRoots() beyond 7/42 providers to kill the silent-zero class


ozymandiashh:
## Motivation

#874's real lesson was not the missing provider, it was the *silence*: Cline CLI usage was reported as $0.00 with no signal anywhere, not even `--verbose`, and the reporter had to reverse-engineer discovery to learn why. `probeRoots()` was added there so `codeburn doctor` can distinguish "tool not installed" from "override points at an empty directory" from "sessions found".

That capability currently covers 7 of the 42 providers. The other 35 keep the silent-zero failure mode: any of them whose data directory moves, whose env override is stale, or whose product update relocates storage will just quietly contribute nothing, and the user's headline will disagree with reality with no diagnostic anywhere.

## Proposal

Extend `probeRoots()` across the provider fleet, in tiers:

1. **Tier 1 - env-overridable local-dir providers** (the exact #874 shape): every provider with a `PROVIDER_ENV_VARS` entry should report its resolved root(s). The env override being wrong is the most common silent-zero cause and the cheapest to detect.
2. **Tier 2 - fixed-location providers**: report the well-known path and whether it exists. Still valuable: product updates move storage (the ChatGPT desktop schema churn in #234 is the extreme case of this class).
3. **Network providers**: out of scope; their failure modes are auth errors, which already surface.

Each implementation is mechanical (return the resolved discovery roots with a label), so this is a good campaign to run incrementally: a batch of 5-8 providers per PR, `doctor` output reviewed once at the start for format.

If the direction is agreed, I can start with Tier 1 (the `PROVIDER_ENV_VARS` set minus the 7 already covered) and keep PRs small.

therickfactr:
Reporter of #874 here — happy to take this if it is unclaimed.

Concrete before/after from that report, since it makes the tiering argument better than I can in the abstract. Before, with the CLI installed and 18 sessions on disk, every surface said the same thing:

```
$ codeburn export --format json --provider cline-cli --from ... --to ...
{"Cost (USD)": 0, "API Calls": 0, "Sessions": 0}
```

`--verbose` added nothing. There was no way to tell "not installed" from "installed, wrong root" without reading the provider source, which is exactly what I ended up doing.

After, the same machine:

```
│  Cline              │         0  │       -  │       0  │  NOTHING FOUND (tool likely not installed or no history yet)
│  Cline CLI          │        18  │    8/8+  │      18  │  OK (18 sessions)
[...]

PR #938, opened 2026-08-05
feat(doctor): probeRoots for six fixed-location providers (#899 Tier 2, batch 1)

therickfactr:
Tier 2 of #899, batch 1: `probeRoots()` for `cline`, `roo-code`, `kilo-code`, `grok`, `pi`/`omp` and `kimi`.

Standalone rather than stacked on #903 — no provider overlaps its twelve, and the shared-parser extraction below stands on its own. The Tier 2 tests are in their own file for the same reason; happy to fold them into the Tier 1 suite once #903 lands.

## Why these six first

`cline` is the provider whose silence motivated #874, and it still had no `probeRoots()` — the CLI half got one, the extension half never did. Its four roots (three VS Code variant globalStorage paths plus `~/.cline/data`) make the Tier 2 case concretely: four places to look, and no way to see which one CodeBurn found.

Real `doctor` output on a machine with the Cline CLI installed and the extension not:

```
│  Cline     │ 0 │ NOTHING FOUND (/Users/…/.cline/data exists but holds no sessions; no history yet)
│  KiloCode  │ 0 │ NOTHING FOUND (/Users/…/globalStorage/kilocode.kilo-code does not exist; tool likely not installed)
```

Both rows were an unexplained `0` before. Now one says *the root is there and empty* — the case where the user has the tool and something is wrong — and the other says *the tool is not installed*.

## One shared resolver instead of a mirror

`probeRoots()` is only worth having if it reports what discovery truly reads, so the resolution is shared rather than duplicated. `vscode-cline-parser.ts` gains `clineTaskRoots(extensionId, overrideDir)`, `discoverClineTasks` is rewritten to call it, and the three Cline-family providers use the same function for their probes.

That is the whole change to the shared parser: a pure extraction, no parsing touched. Roo Code, KiloCode and IBM Bob are unaffected — their suites plus the parser's own are green (44/44 across the six affected files).

This is worth spelling out because the first draft of this branch **mirrored** the logic in a local helper, and the mirror had already drifted before review: it detected "no override" with `=== undefined` while `discoverClineTasks` uses truthiness. An empty-string override therefore made `doctor` report `[""]` while discovery scanned the three default roots — a probe pointing where discovery never looked, stated authoritatively, which is precisely the failure this method exists to prevent. One resolver makes that class unrepresentable; there is a regression test pinning it regardless.

## Per-provider notes

- **kilo-code** reports both halves of its split discovery: the legacy task tree and the SQLite store, labelled separately.
- **pi and omp** are two providers in one module; each reports its own sessions dir, with a test that they cannot collide.
- **kimi** reports `<shareDir>/sessions`, not `shareDir`. Discovery walks the former; reporting the latter would point `doctor` at a directory that exists even when no sessions do — exactly the false reassurance this campaign removes.
- Candidate roots are reported **before** any existence filter, following your `lingtai-tui` call in #903: dropping non-existent dirs is right for discovery and wrong for `doctor`, whose job is to show where it looked.

## Testing

- [x] Tested locally against real data (not just unit tests)
- [x] `npm test` passes

Assertions pin **exact root sets** rather than substrings, so a wrong-but-similar path cannot pass. Each guard is mutation-checked — reverting the resolver to `=== undefined`, making cline's probe ignore its override, and making kimi report `shareDir` each fail the suite (3 failed / 5 passed with the mutations, 8/8 without).

`npm test`: 11 failed / 2699 passed. Every failure is pre-existing on `main` — 1 assertion failure and 10 timeouts, all in `cache-refresh-lock` and `app/electron/cli`. I ran `cache-refresh-lock` on clean `main` and on this branch and got identical results, so it is not introduced here; it looks like the lock family from #914 has not fully settled.

## Open question from #899

Still unanswered, and it shapes batches 2 and 3: for providers whose discovery globs (e.g. `~/.gemini/tmp/<project>/chats`), do you want the glob parent reported, or the resolved matches? Parent is cheap and stable; resolved matches are more useful but turn `doctor` into a directory walk. I have gone parent-only here, consistent with the cheap-existence-check shape, and will keep doing so unless you say otherwise.

### DIFF
--- a/src/providers/cline.ts
+++ b/src/providers/cline.ts
@@ -2,8 +2,8 @@ import { stat } from 'fs/promises'
 import { homedir } from 'os'
 import { basename, join } from 'path'
 
-import { discoverClineTasks, createClineParser, getVSCodeGlobalStoragePaths } from './vscode-cline-parser.js'
-import type { Provider, SessionSource, SessionParser } from './types.js'
+import { discoverClineTasks, createClineParser, clineTaskRoots } from './vscode-cline-parser.js'
+import type { ProbeRoot, Provider, SessionSource, SessionParser } from './types.js'
 
 const EXTENSION_ID = 'saoudrizwan.claude-dev'
 
@@ -38,6 +38,14 @@ async function dedupeTaskSources(sources: SessionSource[]): Promise<SessionSourc
 
 export function createClineProvider(overrideDirs?: string | string[]): Provider {
   const configuredDirs = normalizeOverrideDirs(overrideDirs)
+  // Cline may be installed in any VS Code variant (stable, Insiders, VSCodium),
+  // so every globalStorage root is scanned - same as the Roo Code and KiloCode
+  // siblings - plus Cline's own home-data root. Shared by discovery and
+  // probeRoots so doctor can never report a root discovery does not read.
+  const taskRoots = (): string[] => configuredDirs ?? [
+    ...clineTaskRoots(EXTENSION_ID),
+    getClineDataPath(),
+  ]
 
   return {
     name: 'cline',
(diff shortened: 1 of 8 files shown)


ozymandiashh (approved):
Reviewed against current main. The shared probe-root resolution preserves provider discovery semantics across Cline, Roo Code, KiloCode, Grok, Pi/OMP, and Kimi, with no cache, deduplication, privacy, or scope blocker found. Local validation: 89/89 affected tests, TypeScript, and diff check passed; all five remote checks are green.

PR #940, opened 2026-08-05
feat(cli): rehome cline-cli and codex throughput, with the schema sweep and decode tests it needs

ozymandiashh:
> Draft until independent review completes.

Upstream added two files under the old `src/` path that have **no counterpart on this branch**: the cline-cli provider (`448d470`, plus its follow-up `c642787`) and tool-excluded active throughput metrics for Codex (`a6bf81f`).

## Why this is not just a missing feature

Files added on one side only produce **no merge conflict**. When main is merged into this branch, git will happily create both at `src/…` — a path npm workspaces does not build and no registry imports. Green build, no conflict marker, and a whole provider quietly does not exist.

That is the failure mode this PR exists to prevent, and it is invisible by construction.

## What landed

**cline-cli** follows the phase-8 recipe: discovery and I/O host-side, pure record decode in core, registered alongside its neighbours, with the deduped-rollup follow-up folded in.

**codex-throughput** needed one repair the original did not. The mapper that writes provider calls into the session cache dropped `activeDurationMs`, `activeGeneratedTokens` and `toolWaitMs`, so the three fields never survived the round-trip — and the dashboard's Tok/s column could not have populated regardless of what was ported.

Worth being explicit that the consumer is real: the header, the `anyActiveTiming` gate, the calculation and the model-breakdown aggregation are all present on this branch. This is not a column ported into a void; the only broken link was the write hop.

The test goes end-to-end through `parseAllSessions` and asserts the exact shape the aggregation consumes. A mapper unit test would have passed while the column stayed empty — which is precisely how the gap survived the original port.

## Verification

Core 521. Seven targeted CLI files, 50 tests, covering the cache round-trip, throughput, the TPS surface, codex providers, rich capture and cache invalidation.

### DIFF
--- a/packages/cli/src/codex-cache.ts
+++ b/packages/cli/src/codex-cache.ts
@@ -26,7 +26,18 @@ import type { ParsedProviderCall } from './providers/types.js'
 //     decodes only the new bytes instead of re-streaming the whole file.
 // This is lossless: Codex rollout files are durable (never auto-deleted), so the
 // one-time re-derive on first run under v8 rebuilds byte-identical data.
-const CODEX_CACHE_VERSION = 8
+// v9: tool-excluded active timing — per-call activeDurationMs /
+// activeGeneratedTokens / toolWaitMs (a6bf81f). Cached calls lack the fields;
+// bump once so unchanged sessions re-decode and pick them up.
+// v10: the threaded decoder state now carries the task-timing window
+// (taskResultStart / taskGeneratedTokens / taskToolIntervals / taskStartedAt /
+// openToolStarts), so a task whose task_complete lands in an appended region
+// attributes the three timing fields to the calls emitted in the earlier
+// region — a parse that ended mid-task no longer strands them unattributed.
+// v9 states lack the window; bump once so unchanged sessions re-decode with it.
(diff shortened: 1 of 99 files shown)


ozymandiashh:
**This PR is now stacked, and absorbs #928.** Three commits:

1. the rehome itself
2. **#933's commit, carried underneath** — the new `cline-cli` provider calls `normalizeModelIdentifier`, which arrives with #933. Without it this branch does not typecheck, so the two are genuinely coupled rather than merely ordered. Cross-review flagged exactly this as a gap that fell between the two PRs and was therefore nobody's.
3. **the codex regression tests from #928**, which is closed in its favour. Review established #928's source fixes were already here in full — but its *tests* were not, so this branch was restoring behaviour that nothing proved stayed restored. Merging them one file into three duplicate helper definitions was the alternative.

Green as it stands: core 545, targeted CLI 69. Previously it was red on its own and would have stayed red until #933 landed — which is not a PR anyone should be asked to merge.

ozymandiashh:
Closing for now — drifted behind main with conflicts, and the slot is needed under the five-open-PR cap. The branch stays on the fork for a future rebase.

Issue #912, opened 2026-08-04
Systematic code audit: 6 verified fixes (PRs), plus confirmed + candidate findings


ozymandiashh:
Ran a systematic multi-agent audit of the codebase tonight and want to record the results so nothing is lost, and so the remaining real candidates get a second set of eyes.

## Method
Every app source line was reviewed in overlapping chunks by a bug-hunting pass, then each finding went through: a skeptical verifier (refute-by-default, given the real code window), then a 2-lens adversarial pass (reproducibility + caller-contract, survive only if neither refutes), then **my own manual verification with a mutation-checked test** on anything I turned into a fix. Funnel: 553 raw -> 270 triaged -> 121 verify-confirmed -> 103 adversarial-unanimous survivors.

**Key caveat for reviewers:** the flash swarm kept known-false findings through all three automated stages (e.g. it insisted a `SELECT DISTINCT ... ORDER BY <non-selected col>` crashes SQLite - it does not; I confirmed empirically). Automated confirmation is NOT sufficient here; each survivor needs manual verification before it becomes a fix. That is exactly why the six fixes below each ship a mutation-checked test.

## Fixed (PRs, each mutation-checked)
- #908 hardening: malformed request URL crashes the LAN share-server; invalid `ts` RangeError aborts a Cline-family session parse; a null entry in the remote LiteLLM JSON aborts the whole pricing load.
- #909 kiro: input tokens estimated from a 500-char slice of the last turn while output summed everything (cost undercount).
- #910 optimize: result-cache key collided on (project count, api-call sum), serving stale findings.
- #911 context-budget: skills and `~/.claude/CLAUDE.md` double-counted when the project dir is the home dir.

## Confirmed real, not yet fixed
- `models.ts` model-alias table: `claude-4-sonnet-thinking` aliases to `claude-sonnet-4-5` (its non-thinking sibling maps to `claude-sonnet-4`). Zero money impact today because the two price identically, but it will misprice if they ever diverge. Worth a one-line correction.

## Need a repro before I would touch them (plausible, subtle)
- `parser.ts` ~3805: a provider-filtered cold run may set the whole-cache completeness marker with out-of-scope providers unscanned.
- `providers/antigravity.ts` ~1098: the count-1 run skip interacting with the monotonic-delta snapshot logic.
- 2 Swift money-path items in the menubar credential/cache path (I can't run the Swift suite here).

[...]

PR #982, opened 2026-08-12
fix(models): keep claude-4-sonnet-thinking in the Sonnet 4 family

avs-io:
## Problem

Cursor's unversioned `claude-4-sonnet-thinking` slug was aliased to `claude-sonnet-4-5`, even though it identifies the thinking variant of Sonnet 4.

This is easy to miss today because Sonnet 4 and Sonnet 4.5 currently have identical rates. The incorrect identity nevertheless surfaces the wrong model name and could become a pricing error if those rates diverge.

## Change

- map `claude-4-sonnet-thinking` to `claude-sonnet-4`
- update the Cursor alias regression table
- add an independent display-name assertion that pins the model identity even while the two models share pricing

## User impact

Cursor sessions using this slug now appear under Sonnet 4 rather than Sonnet 4.5 and remain attached to the correct pricing family.

## Validation

- `npx vitest run tests/models.test.ts` — 154 passed
- `npm test` — 2,617 passed, 5 skipped
- `npm run test:locks` — 26 passed
- `npx tsc --noEmit`
- `npm run build:cli`
- `npm --prefix dash run build`
- runtime probe resolves the slug to `Sonnet 4`

Addresses the confirmed model-alias finding in #912.

### DIFF
--- a/src/models.ts
+++ b/src/models.ts
@@ -311,7 +311,7 @@ const BUILTIN_ALIASES: Record<string, string> = {
   // reports that quote literal slugs (e.g. forum.cursor.com/t/154933).
   'claude-4-sonnet':                'claude-sonnet-4',
   'claude-4-sonnet-1m':             'claude-sonnet-4',
-  'claude-4-sonnet-thinking':       'claude-sonnet-4-5',
+  'claude-4-sonnet-thinking':       'claude-sonnet-4',
   'claude-4.5-sonnet':              'claude-sonnet-4-5',
   'claude-4.5-sonnet-thinking':     'claude-sonnet-4-5',
   'claude-4.6-sonnet':              'claude-sonnet-4-6',
--- a/tests/models.test.ts
+++ b/tests/models.test.ts
@@ -505,7 +505,7 @@ describe('Cursor model variants resolve to pricing', () => {
     // Sonnet family
     ['claude-4-sonnet', 'claude-sonnet-4'],
     ['claude-4-sonnet-1m', 'claude-sonnet-4'],
-    ['claude-4-sonnet-thinking', 'claude-sonnet-4-5'],
+    ['claude-4-sonnet-thinking', 'claude-sonnet-4'],
     ['claude-4.5-sonnet', 'claude-sonnet-4-5'],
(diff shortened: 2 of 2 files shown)


ozymandiashh (approved):
Correct, and the repo corroborates it without needing to consult anything external: `src/providers/cursor.ts:33` and `src/providers/cursor-agent.ts:58` already render this slug as "Sonnet 4 (Thinking)", with `claude-4.5-sonnet-thinking` as a separate "Sonnet 4.5 (Thinking)" entry — pinned by `tests/providers/cursor.test.ts:31`. The provider display maps and the alias table disagreed, and this makes them agree.

It also removes the last inconsistent row in that table. Every dot-versioned slug carries its version through (`claude-4.5-sonnet-thinking` → `claude-sonnet-4-5`, `claude-4.6-sonnet-thinking` → `claude-sonnet-4-6`), and the unversioned `claude-4-sonnet` / `claude-4-sonnet-1m` already resolved to `claude-sonnet-4`. This row was the only one where a suffix quietly promoted a model to a different minor version — same bug class as `claude-4-opus`, which came from the same original commit and was corrected in #290.

The impact is more than cosmetic: `src/parser.ts` keys `modelBreakdown` off `getShortModelName`, so a Cursor user running both Sonnet 4 Thinking and Sonnet 4.5 had the two collapsed into one "Sonnet 4.5" row in the overview, menubar and sessions breakdowns.

Good call adding the standalone display-name assertion. The parametrized pricing row can't catch this on its own while the two models share rates in the snapshot, so that extra assertion is the only thing actually pinning the identity — and the comment says so, which is what makes it maintainable.

One pre-existing gap this doesn't touch, just noting it so it isn't a surprise: already-finalized days in the daily cache stay bucketed under the old name. `getDailyCacheConfigHash` is composed only of the local-savings and price-override hashes, and `getModelAliasesConfigHash` deliberately excludes the builtin table, so nothing invalidates history when `BUILTIN_ALIASES` changes. Today's numbers are right immediately; older days re-bucket only when their cache is otherwise re-derived. Not a reason to hold this.

Issue #969, opened 2026-08-11
Unpriced-models warning is truncated at every terminal width, hiding both the model list and the fix command


therickfactr:
## Problem

The unpriced-models warning in `codeburn report` is truncated below usefulness at every terminal width, so neither the affected models nor the fix command can be read from the output.

## Evidence

At an 80-column-ish default:

```
│ ! 3 models unpriced at $0, fix: codeburn model-alias (auto-ge… │
```

Widening the terminal does not help — the warning lives in a fixed-width panel in the report layout, so at `COLUMNS=400` it still clips:

```
$ COLUMNS=400 codeburn report | grep -i unpriced
│ ! 3 models unpriced at $0, fix: codeburn model-alias (auto-genius, big-pickle, .… │
```

The third model name is never reachable, and the fix command is cut before its arguments — `codeburn model-alias` alone is not runnable.

## Why it matters

The warning's whole purpose (per #638) is to make unpriced usage actionable in the product itself rather than behind `CODEBURN_VERBOSE=1`. As rendered, it tells the user a problem exists but withholds both which models are affected and the command to address it. Finding the third name meant reading the daily cache directly, which is the opposite of the intent.

The full renderer produces useful text — it is only the panel that clips it:

```
Unpriced   3 models at $0: auto-genius (35.3M tok), big-pickle (33.9K tok), …
           Fix: codeburn model-alias "<model>" <known-model>
```

## Suggested direction

Any of these would resolve it; listing options rather than assuming a preference:

1. Wrap the warning across lines inside the panel instead of clipping, so the model list and the fix line both survive.
2. Keep the panel line as a short pointer (`! 3 models unpriced at $0 — see codeburn models --unpriced`) and put the detail behind a command that can print at full width.
3. Let the warning render full-width outside the fixed panel grid, as the `Totals` block does.

Option 2 has the side benefit of giving a stable way to list every unpriced model, not just the first three — useful when the count exceeds the `+N more` cutoff.

Environment: codeburn 0.9.20 (npm), macOS 26.6.1 (25G76) arm64, Node v26.5.0.

adamkwhite:
Adding a related data point, since it strengthens option 2 in particular.

[...]

PR #985, opened 2026-08-12
Add unpriced filter to models report

kocaemre:
## Why
Related to #969.

The TUI warning for unpriced models points users at `codeburn model-alias`, but the detailed model list is still hard to reach/copy from the dashboard. `codeburn models` is already the plain-stdout escape hatch, so this adds a focused filter for the affected rows.

## What
- Adds `codeburn models --unpriced`.
- Keeps zero-cost unpriced rows visible by defaulting that mode's min-cost threshold to `0`.
- Reuses the existing `findUnpricedModels` logic so local/free/aliased/overridden models are treated the same as the dashboard warning.
- Covers the CLI path with a fixture containing one priced model and one unpriced model.

## Validation
- `npx vitest run tests/models-report.test.ts`
- `npm run build:cli`
- Manual fixture run:
  - `node dist/cli.js models --unpriced --from 2026-05-09 --to 2026-05-09 --provider claude --format json`

## Note
I also tried `npm test -- tests/models-report.test.ts`, but the script ignores the positional file argument and ran the broad suite. That broad run failed in unrelated cache/history CLI tests due the local environment picking up existing usage data; the targeted suite above passes.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -2075,6 +2075,7 @@ program
   .option('--by-agent', 'One row per (provider, model, agent) instead of one row per (provider, model). Claude subagent transcripts only; other providers and main sessions bucket under "main"')
   .option('--top <n>', 'Show only the top N rows', (v: string) => parseInt(v, 10))
   .option('--min-cost <usd>', 'Hide rows below this cost threshold', (v: string) => parseFloat(v))
+  .option('--unpriced', 'Show only models with usage that currently price at $0')
   .option('--no-totals', 'Suppress the footer totals row')
   .option('--format <format>', 'Output format: table, markdown, json, csv', 'table')
   .action(async (opts) => {
@@ -2099,13 +2100,21 @@ program
     }
 
     const projects = await parseAllSessions(range, opts.provider)
-    const rows = await aggregateModels(projects, {
+    let rows = await aggregateModels(projects, {
       byTask: !!opts.byTask,
       byAgent: !!opts.byAgent,
       taskFilter: opts.task,
       topN: typeof opts.top === 'number' && Number.isFinite(opts.top) ? opts.top : undefined,
-      minCost: typeof opts.minCost === 'number' && Number.isFinite(opts.minCost) ? opts.minCost : 0.01,
+      minCost: typeof opts.minCost === 'number' && Number.isFinite(opts.minCost) ? opts.minCost : (opts.unpriced ? 0 : 0.01),
     })
+    if (opts.unpriced) {
+      rows = rows.filter(row => findUnpricedModels([{
+        model: row.model,
+        calls: row.calls,
+        cost: row.costUSD,
+        tokens: row.totalTokens,
+      }]).length > 0)
+    }
 
     const fmt = (opts.format ?? 'table').toLowerCase()
     if (rows.length === 0 && (fmt === 'table' || fmt === 'markdown')) {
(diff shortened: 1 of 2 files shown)


ozymandiashh (changes requested):
Nice, focused change — and thank you for reusing `findUnpricedModels` rather than re-deriving the predicate. That keeps the alias / price-override / local-savings exclusions in one place, so the filter and the dashboard warning can never drift apart. Passing `row.model` rather than the display name is also the right call; it's a better pricing key than what `dashboard.tsx` and `overview.ts` currently hand the same helper.

Two things before this can go in.

**1. `--unpriced` combined with `--top` returns nothing.**

`--top` is applied inside `aggregateModels` (`src/models-report.ts:234-236`), after the sort by `costUSD + savingsUSD` descending. Unpriced rows are $0 cost by definition, and rows carrying a local-savings baseline are excluded by `findUnpricedModels`, so they're $0 savings too — which puts them last in that sort, every time. The new filter in `src/main.ts` runs afterwards, so `--top` has already sliced away exactly the rows `--unpriced` exists to show:

```
codeburn models --unpriced --format json              # -> the unpriced row
codeburn models --unpriced --top 2 --format json      # -> []
codeburn models --unpriced --top 2                    # -> "No model usage found for the selected period."
```

That last line is the part I'd like to avoid shipping: a user with unpriced models is told they have none. Both flags are documented on `models`, so it's a combination people will land on.

The straightforward fix is to stop letting `aggregateModels` do the slicing when `--unpriced` is set — pass `topN: undefined`, then slice to `opts.top` after the filter. Worth a test pinning `--unpriced --top 1` against a fixture where a priced model outranks the unpriced one, since that's the case that's silently wrong today.

**2. The flag needs a line in the README table and the changelog.**

Every other `models` flag is listed in the **Models** table in `README.md`, and user-visible flags land in `## Unreleased` in `CHANGELOG.md`. A filter nobody can discover doesn't do much for #969, which is the point of adding it.

On your note about `npm test -- tests/models-report.test.ts` running the broad suite: that's a real rough edge, not you. The script doesn't forward the positional path. `npx vitest run tests/models-report.test.ts` is the working form, and it's what CONTRIBUTING should probably say.

Everything else looks good to me — the end-to-end spawn test with an isolated `HOME`/`CODEBURN_CACHE_DIR` and a fixed `TZ` is the right shape, and defaulting that mode's min-cost to 0 is the correct call rather than a special case. Ping me when the `--top` interaction is handled and I'll take another pass.

ozymandiashh (approved):
Taking the two items myself rather than holding this — the feature is right and the remaining work is small enough that it shouldn't sit in your queue.

To be clear about what's going in: `--unpriced` on its own works correctly, and reusing `findUnpricedModels` rather than re-deriving the predicate is the part that matters long-term, since the alias / price-override / local-savings exclusions stay in one place and the filter heals on its own when pricing data arrives. Passing `row.model` instead of the display name is also strictly better than what `dashboard.tsx` and `overview.ts` currently hand the same helper.

Merging, and I'll follow up with:

- the `--top` interaction — `aggregateModels` slices before the filter runs, so `--unpriced --top N` returns nothing whenever N is smaller than the number of priced rows
- the README **Models** table row and the changelog entry

Also worth fixing separately: `npm test -- <path>` not forwarding the positional argument, which is what sent you into the broad suite. That's a real rough edge in the repo, not something you did wrong.

Thanks for a clean, well-scoped first contribution.

PR #999, opened 2026-08-14
fix(status): serve menubar-json status from a disk-persisted snapshot to eliminate per-poll re-parse latency

dgabehar:
## Summary

`codeburn status --format menubar-json` took 25-90+ seconds per call
regardless of cache state, causing the polling CodeBurnMenubar.app to
stall/stop refreshing. Root cause (confirmed via live profiling): the
menubar spawns a fresh CLI process per poll, so every call re-parses the
full session-cache blob and re-runs the full aggregation pipeline with
zero cross-process reuse — none of the existing in-process caches
(parser.ts's TTL/burst maps, session-cache.ts's cacheMemo) ever survive
between polls.

This adds a disk-persisted status snapshot keyed by a cheap corpus
fingerprint (stat-only pass over discovered sources — dev/ino/mtime/size,
no session-cache.json parse, no transcript content read), with a
settle-window debounce (CODEBURN_STATUS_SNAPSHOT_SETTLE_MS, default
2000ms, capped 60000ms) so rapid-fire source writes coalesce into one
recompute instead of one per poll tick.

## Related

#971 covers the same general problem class (expensive redundant
per-invocation recompute) from a different angle — a resident
`codeburn serve` daemon sometimes being bypassed on cold start. This PR
is orthogonal to that: the menubar's `status --format menubar-json` path
has no resident-process story at all today (a fresh CLI process every
poll, always), so the fix here is a disk-persisted snapshot rather than
daemon-sharing. Happy to reconcile approaches if useful once #972 lands.

## Changes

- `src/session-cache.ts` — disk-persisted status snapshot
  (`loadStatusSnapshot`/`saveStatusSnapshot`) keyed by corpus fingerprint
  + resolved query; atomic temp+rename write at 0600; schema-version
  guard against a payload-shape drift across binary upgrades; orphaned
  temp-file sweep extended to cover the new file.
- `src/parser.ts` — `computeCorpusFingerprint`, a stat-only pass over
  every discovered source (expanding Claude's directory-shaped sources
  to their actual `.jsonl` files first, since a directory's own mtime
  doesn't move when an existing file inside it is rewritten in place).
- `src/main.ts` — wires the `status --format menubar-json` path to check
  the snapshot before parsing/aggregating, and to persist a fresh one
  after a real recompute; a single shared options object now feeds both
  the cache key and the `buildMenubarPayloadForRange` call so they can't
  drift apart.
- `tests/cli-status-menubar.test.ts` — new coverage: snapshot-hit
  byte-identical output, debounce-defer during the settle window,
  reflect-after-settle.
- `SPEC-perf-cache-fix.md` — design write-up (root-cause investigation,
  why the debounce lives in the snapshot layer rather than
  `reconcileFile`, and why a first attempt at the latter was reverted).

## Test plan

- `npm test` — full suite passes, including the new
  `tests/cli-status-menubar.test.ts` coverage.
- `npx tsc --noEmit` clean.
- Verified live against a real ~386MB session-cache corpus: cold call
  ~18-23s (unchanged — that's the underlying parse+aggregate cost, not
  addressed by this fix), settled repeat call ~2s, byte-identical
  payload.
- Manually installed via `npm link` and run against the real
  CodeBurnMenubar.app.

## Checklist

- [x] Code follows project conventions and coding standards
- [x] Unit tests written and passing
- [x] No new secrets, credentials, or API keys committed
- [ ] Release notes updated (`docs/release-notes.md` + `README.md`)
- [x] `npm test` green


ozymandiashh (changes requested):
Thanks for the profiling work here — the numbers are real and the problem you're describing was real. A few things have moved under this PR since you opened it, though, and there are some correctness issues in the snapshot itself that I don't think we can ship. Taking the structural point first, because it changes what the rest of the work should look like.

## The premise has changed: #972 landed

When you opened this, `status --format menubar-json` genuinely had no cross-process story — a fresh CLI per poll, every time. That's no longer true on `main`. #972 introduced a resident `serve --stdio` child that both desktop clients route `status` through, with a process spawn only as a fallback, plus watcher-based invalidation keyed on config content, Claude discovery roots and watched session inputs.

That matters for this design specifically, not just as a rebase chore. A disk snapshot underneath serve's memo is a second, weaker cache with its own invalidation rules, and the weaker one wins whenever it hits first. You anticipated this in the PR description and offered to reconcile once #972 landed — it has, so I'd take you up on that before going further on the snapshot itself.

## Correctness issues in the snapshot

These would apply to any version of this approach, so they're worth having on record regardless of what shape the reconciliation takes.

**1. `queryKey` omits the pricing config that the parse cache key deliberately includes.** `cacheKey` in `src/parser.ts:3485` folds in `getProxyPathsConfigHash()`, `getModelAliasesConfigHash()`, `getPriceOverridesConfigHash()` and `getLocalModelSavingsConfigHash()`, with a comment stating the reason: a memoized parse must never "present costs priced under aliases/overrides/savings the user has since changed." Your `queryKey` carries range, provider, project/exclude, optimize/timeline and the Claude config source — none of the four. Edit a price override, and the snapshot keeps serving the old costs until something unrelated moves the corpus fingerprint.

**2. Network providers drop out of the fingerprint entirely.** `computeCorpusFingerprint` only records a source when `fingerprintFile(path)` returns non-null. `vercel-gateway`'s source path is the synthetic `'vercel-ai-gateway:report'` (`src/providers/vercel-gateway.ts:138`) with no file behind it, so it contributes nothing to the hash. The parser handles this deliberately — network providers get a synthetic fingerprint so they re-fetch every run — and the snapshot skips the run that would do the fetching.

**3. Directory-shaped non-Claude sources repeat the blind spot you fixed for Claude.** The leaf-file expansion is gated on `source.provider === 'claude'`. `mistral-vibe` also emits a directory as its source path (`src/providers/mistral-vibe.ts:421`) and its parser reads `join(source.path, 'messages.jsonl')`. Statting the directory misses in-place appends to that file — exactly the failure your own spec documents for the Claude case.

**4. A knowingly-degraded read-only parse gets persisted as authoritative.** When the cross-process refresh lock is contended, `parseAllSessions` falls into a read-only parse that deliberately serves cached turns for files whose fingerprint has already moved, and records that by setting `readOnlyServedStale` (`src/parser.ts:3059-3065`). `isSessionHydrationComplete()` exists precisely so callers can tell. The save path doesn't consult it, so that payload gets written to disk keyed on the current, already-advanced fingerprint — and now the stale answer is the cached answer.

**5. The settle window keys on corpus-wide newest mtime.** `newestMtimeMs` is the max across every discovered source, not the mtime of whatever changed. On a machine where any file in the corpus is touched more often than the window, every fingerprint mismatch is deferred and no fresh snapshot is persisted, so the stale payload is served indefinitely. The spec's claim that the window "only ever delays picking one up" by "at most the window" doesn't hold in that case.

## Two smaller things

**The merge commit `a90de16` carries a `Co-Authored-By: Claude` trailer.** `CONTRIBUTING.md` forbids it and `.github/workflows/block-claude-coauthor.yml` rejects it — you haven't seen it fail because workflow runs need approval on a first PR, so the gate hasn't run yet. It'll need a rebase to strip.

**`SPEC-perf-cache-fix.md` lands at the repository root.** Design write-ups live in `docs/design/`; the root holds only the standard files. It also opens by citing `PERF-DEFECT-FINDINGS.md`, which isn't in the tree.

None of this is wasted work — the fingerprint sweep and the cross-process reasoning are the useful parts, and points 1-5 are things any caching layer here has to get right. What I'd suggest is rebasing onto `main`, looking at what serve's invalidation already covers, and coming back with what's still missing after that. Happy to talk through the shape before you spend more time on it.

dgabehar:
Iter-1 response to @ozymandiashh (CHANGES_REQUESTED, 2026-08-18)

**Tally:** 8 closed, 0 open, 0 deferred (plus 1 bonus fix found while addressing the others)

| Item | Disposition | Where |
|---|---|---|
| Structural: #972 reconciliation | Kept the disk snapshot, scoped to `serve`'s CLI-spawn-fallback path (a fresh process has no in-process memo to hit, which is exactly the fallback case `serve.ts` documents). Brought its invalidation to parity with — and in some respects stronger than — `serve`'s watcher-based memo via the fixes below. Full reasoning in `docs/design/perf-cache-fix.md` | `docs/design/perf-cache-fix.md` |
| 1. queryKey omits pricing config | Added the same 4 hash getters `cacheKey` uses (proxyPaths/modelAliases/priceOverrides/localModelSavings) | `src/main.ts` |
| 2. Network providers drop out of fingerprint | Synthetic now-stamped entry per network source, mirroring `parseAllSessions`' own treatment | `src/parser.ts` `computeCorpusFingerprint` |
| 3. Directory-shaped non-Claude sources | Generic recursive expansion for any directory-shaped source, not gated on provider name | `src/parser.ts` (new `collectFilesRecursive`) |
| 4. Degraded read-only parse persisted as authoritative | Save now gated on `isSessionHydrationComplete()` | `src/main.ts` |
| 5. Settle window keyed on corpus-wide mtime | Anchored on wall-clock time since the fingerprint first mismatched instead | `src/session-cache.ts` `loadStatusSnapshot` |
| Hygiene: AI co-author trailer | Rebuilt the merge commit with the trailer removed (same tree, same parents); verified clean across the full branch history | commit history |
| Hygiene: spec doc at repo root | Moved to `docs/design/perf-cache-fix.md`; removed the dangling reference to the uncommitted findings file | same file |

Also found and fixed a 6th instance of the same defect class as #1 while chasing a test failure: the snapshot's query key also omitted the active display currency, so switching currencies kept serving stale numbers. Same fix shape as #1, same file.

Full test suite green (`tsc --noEmit` clean; general suite and the lock-contention suite both pass). One heads-up unrelated to any of the above: `main` has moved substantially since this branch's last sync (merge-base is still `da3d903`), so `mergeable` currently reads `CONFLICTING` — pre-existing, not introduced by this push. Happy to do a further rebase pass once you've had a look at the shape of this one.

ozymandiashh (changes requested):
Re-reviewing the current head, two correctness blockers remain in the persisted status snapshot. First, the snapshot key covers the session corpus/query but not the live LiteLLM pricing cache, bundled pricing generation, or parser/pricing semantic version. With unchanged sessions, a repricing or pricing fix can therefore keep serving old fully rendered costs indefinitely. Second, the default optimized status is cached even though optimize results depend on mutable project/config/prompt/hook state that is outside the fingerprint, so findings can remain stale with no session change. The new regression uses `--no-optimize` and does not exercise that path. Please add explicit semantic/pricing invalidation and either fingerprint the optimize inputs or exclude optimized output from the persistent snapshot, with regressions for both transitions.

dgabehar:
Iter-2 response to @ozymandiashh (CHANGES_REQUESTED, 2026-08-21)

**Tally:** 2 closed, 0 open, 0 deferred

| Item | Disposition | Where |
|---|---|---|
| Pricing/version staleness | `queryKey` now includes a new `getPricingGenerationKey()`, distinct from the four pricing-CONFIG hashes already there: the live LiteLLM cache's freshness (a timestamp tracked whenever the on-disk cache is read or refreshed), a content digest of the bundled `litellm-snapshot.json`/`pricing-fallback.json`, and the existing `CACHE_SCHEMA_VERSION` pricing-behavior version (reused — it already gated the on-disk LiteLLM cache, just hadn't reached this snapshot). Any of the three changing forces an immediate recompute regardless of corpus fingerprint. | `src/models.ts`, `src/main.ts` |
| Optimize-path caching | Optimize findings depend on project/config/prompt/hook state (`~/.claude` settings at user and project scope, `CLAUDE.md`, defined skills/agents/commands, MCP config) with no single enumerable fingerprint. Rather than fingerprint that surface, the optimize path (the default; `--no-optimize` unaffected) now never reads or writes the persisted snapshot — it always recomputes fresh. Computing optimize findings already requires the same full parse the snapshot exists to let a cache hit skip, so this costs nothing extra versus fingerprinting those inputs while removing the whole invalidation surface. | `src/main.ts` |

Added regressions for both: one that mutates the on-disk LiteLLM cache (new price, new timestamp) against an unchanged session corpus and asserts the second call reflects the new price instead of the cached one; one that adds a never-invoked custom agent definition between two default (optimize-enabled) calls and asserts both the finding count changes and the status snapshot file is never written.

Also picked up a second upstream sync while addressing this (`main` had moved another 201 commits) — two more disjoint conflicts (an import-list union in `src/main.ts`, and two independent new code blocks landing at the same insertion point in `src/parser.ts`), both resolved mechanically with no logic changes.

Full test suite green: `tsc --noEmit` and `build:cli` clean, general suite passing (3284/3302, the 13 residual failures are all `Test timed out` in files this change never touches — confirmed by diff — matching known contention noise on the CI-adjacent box this ran on), and the lock-contention suite verified clean by direct diff (this change touches neither `session-cache.ts`'s lock logic nor `cache-refresh-lock.ts` at all) plus two from-scratch 26/26 runs on this exact code earlier in this branch's history.

ozymandiashh:
Follow-up pushed in `e4f7615` after synchronizing the branch with current `main` (`d8d89f7`).

Reproductions against the previous head (`46f2ddd`):

- The claimed CAS was not atomic. In 50 genuine same-query `Promise.all` races, the older observation won the final rename **40/50** times.
- When mismatch bookkeeping could not be written (read-only cache dir), the first poll served stale and a second poll after the settle window still served stale because every process restarted the grace clock.
- Deleting the newest source lowered max mtime from 3000 to 2000; the later recompute was then refused and could not publish.
- With an unchanged corpus, adding a model to `flatRateModels` still served the cached unpriced warning because that config was absent from the query key.
- Reversing two same-basename `claudeConfigDirs` in `config.json` kept the old `profile 1` / `profile 2` path mapping because ordered source topology was invisible to the snapshot.

What changed:

- Status writers now use the existing hardened cross-process lock protocol with a deterministic per-query lock. Guard + write + owner fence + rename happen in one critical section; separate query keys remain independent. Writes are synced, temp files are cleaned, and a failed bookkeeping write returns a cache miss instead of stale data.
- Competing saves are ordered by a high-resolution observation timestamp captured before corpus discovery, not by non-monotonic max mtime.
- Snapshot envelope v3 carries a semantic key (`package + renderer + daily-cache version`). The query key now includes flat-rate config and ordered Claude source topology; the corpus fingerprint also includes explicit source metadata.
- Added a real two-process writer harness (20 rounds), the 50-round same-process race, read-only failure, deletion, semantic-version, flat-rate, and Claude-topology regressions.

Validation on the pushed head:

- `npm test`: **244 files passed, 3,306 tests passed, 5 skipped**
- `npm run test:locks`: **34/34 passed**, including the new child-process status writer test
- `npx tsc --noEmit`: passed
- `npm run build:cli`: passed
- Two independent adversarial reviews found no remaining blocker.

Please re-review the updated head. I have not merged while the existing `CHANGES_REQUESTED` review is still active.

ozymandiashh (approved):
Re-reviewed `e4f7615` after the `main` reconciliation and follow-up fixes. The previously requested snapshot correctness blockers are addressed, the added race/config reproductions pass, the full local suite is green, and all official GitHub checks on this head pass. Approving the updated implementation.

iamtoruk:
Post-merge review record (this PR was merged while held for founder review; full adversarial review was run after the fact - process now formalized via branch protection on main, so this situation resolves structurally rather than personally).

**Verdict: KEEP WITH FIXES - the merge stands.** The invalidation surface here is genuinely comprehensive (DAILY_CACHE_VERSION, package version, pricing generation, five config hashes, currency, provider env fingerprints, ordered topology - all rejected at load with regressions), the torn-write handling is correct, and the lock-path generalization actually improves SIGTERM-window cleanup over the old single-path code. Thank you @dgabehar for the feature and @ozymandiashh for three rounds of review plus the hardening commit.

**One blocker survived review, now fixed in e0a2b413:** the snapshot save gate at src/main.ts read the module-level hydration global after buildMenubarPayloadForRange had already run further parses (including its own 365-day history re-parse), at exactly the read point the aggregator's own comments call unsafe. A poll that went read-only under a held refresh lock could persist its under-reported payload once a later in-process parse reset the global - pinning degraded dollars to disk, served as authoritative until the corpus next changed, with no self-heal on an idle machine. The gate now requires the payload's own captured markers (stale !== true, no hydration block); a regression holds the refresh lock and asserts no snapshot file is written for the degraded pass (verified to fail with the old gate).

Minor follow-ups worth an issue each, none urgent: desktop app never hits this path (its fetches omit --no-optimize, so useSnapshot is false - the win currently lands on the menubar poll only); snapshot files are never garbage-collected (one permanent file per distinct query); the file path carries no binary version so mixed-version CLIs thrash each other's snapshots (correctness-safe, perf-only); network-provider users get a silent forced miss every poll.

PR #1001, opened 2026-08-16
feat: add DeepSeek Harness (dsh) provider

MiloMMIN:
## Summary

- Adds a provider for [DeepSeek Harness](https://www.npmjs.com/package/@deepseek-ai/dsh) (`dsh`): reads sessions from `$DSH_HOME/sessions/<encoded-cwd>/session-<uuid>/session.jsonl.zstd` (default `~/.dsh`, plus the uncompressed `session.jsonl` variant).
- DSH session logs are **concatenations of independent zstd frames** (one per appended event batch), which `node:zlib`'s one-shot API cannot decode as a whole. The provider ports the frame-boundary scan (`scanZstdFrames`) from the official `@deepseek-ai/dsh-session-persistence-jsonl` package and decompresses frame by frame using only public APIs. zstd requires Node >= 22.15; older runtimes get a stderr notice and DSH data is skipped (same pattern as the Zed provider).
- Usage accounting mirrors `dsh-token-meter`'s projection semantics: an `assistant/message` usage report is the final value for its `(turn, step)` and **replaces** the earlier `assistant/chunk` sample (adjacent reports of the same API call), never double counting. Model attribution comes from the most recent `request/header` config; reasoning tokens are billed at the output rate (same as Hermes/Gemini). One parsed call per `(turn, step)`, dedup key `dsh:<sessionId>:<turn>:<step>`.

## Testing

- [x] `npm test` passes — the new suite (`tests/providers/dsh.test.ts`, 22 tests) plus registry/env/probe-roots/session-cache suites are green. Full `tests/providers` on this Windows machine shows 45 failures that exist identically on the unmodified baseline (home-dir-related: claude-config-dirs, copilot, devin, ...); none are dsh-related.
- [x] `npm run build` succeeds — `build:cli` (tsup) and `tsc --noEmit` are clean. Note: the full `build` script's first step (`bundle-litellm`, re-fetching the upstream pricing snapshot) fails on this machine with `ECONNRESET` due to local network restrictions; that fetch is unrelated to this change, which does not touch pricing data.

### For new providers only:

- [x] `npm run dev -- models --provider dsh` shows correct model names and pricing

`report --provider dsh -p all --format json` against a real `~/.dsh` (25 sessions):

```
overview: cost $3.54, calls 1027, sessions 25, cache hit 96.5%
tokens:   input 6,380,935 / output 1,096,378 / cacheRead 174,656,064 / cacheWrite 0
models:   DeepSeek v4 Pro    501 calls  $2.8229
          DeepSeek v4 Flash  525 calls  $0.7158
          grok-4.6             1 call   $0.0000
projects: Desktop/share/script, Documents/WK, ... (grouped by session cwd)
```

Cross-check: token totals summed independently from the raw session logs (frame-by-frame decode + the replace-on-final usage rule) match the report exactly (sessions=25, calls=1027, input/output/cacheRead/cacheWrite identical).

Dashboard (`npm run dev`) with DSH included in the `all` view:

```
│ By Model                                                  │
│ DeepSeek v4 Pro    $2.82  97.0%  501  100.0%              │
│ DeepSeek v4 Flash ~$0.934 96.2%  571  100.0%              │
│ By Project                                                │
│ Desktop/share/script  $4.51  $0.215  21                   │
```

Provider cycling (`p`) detects `dsh` and the TUI shows it as `DeepSeek Harness`; `codeburn doctor` probes `$DSH_HOME/sessions`.


xianjianlf2:
I opened a stacked fix PR against the contributor branch: https://github.com/MiloMMIN/codeburn/pull/1

It addresses the zstd compatibility review feedback:
- zstd-dependent DSH tests now skip when `zlib.zstdCompressSync` is unavailable, so Node 22.13 CI can still run the plain JSONL coverage
- DSH discovery/header probing now emits the Node >= 22.15 zstd notice instead of silently skipping `.zstd` sessions
- the notice is deduped to avoid stderr spam during a scan

Validation run locally:
- `npm ci`
- `npx vitest run tests/providers/dsh.test.ts`
- `npx tsc --noEmit`
- `git diff --check`

ozymandiashh (changes requested):
This is strong provider work. Scanning zstd frame boundaries structurally rather than trusting `node:zlib` to handle a concatenation, bucketing usage per `(turn, step)` with the `assistant/message` report *replacing* the `assistant/chunk` sample rather than summing, and skipping `seq < seedLength` on a forked session so a parent's replayed events aren't billed twice — those are the three things this format could most easily get wrong, and all three are right. The 34-test suite including the upstream snapshot is well above the usual bar for a new provider.

Two things in the decode path I'd want fixed before this ships, both around `readZstdLines`.

## A valid log that decompresses past 128 MiB is discarded whole and reported as corrupt

`readSessionFile`'s size guard rejects a *compressed* file over `MAX_SESSION_FILE_BYTES`, and then `readZstdLines` applies that same 128 MiB figure as a budget on the *decompressed* output. zstd routinely gets around 10x on JSONL, so the effective ceiling for a DSH log is roughly 13 MB on disk — a size an ordinary long agentic session reaches without being remarkable.

What happens then is the part that concerns me more than the threshold itself. The overflow throws, the caller catches it alongside genuine structural corruption, and the user gets:

```
codeburn: skipped corrupt DSH session log <path>
```

So a perfectly intact log is dropped in full — every call in it, not a truncated tail — and the user is told their data is corrupt. They have no way to tell that apart from a real problem, and nothing suggests a remedy. Two separate things worth splitting: the budget should be sized against decompressed content rather than inheriting the compressed constant, and a size overflow should not share an error path with "this file is structurally broken."

## Records straddling a frame boundary are silently lost

Each frame's plaintext is `split('\n')` independently and the pieces are never rejoined, so the decoder is only correct if every frame happens to end exactly on a newline. If DSH ever flushes a batch mid-record, that record arrives as two fragments, both fail `JSON.parse`, and both are skipped — silently, because unparsable lines are the expected case for a torn tail.

In practice this probably holds today, since DSH appends whole batches. But it's an assumption about a third-party writer's flush behaviour that nothing in the code states or enforces, and the failure is invisible when it breaks. Carrying a remainder across frames and only treating the final leftover as a possible torn tail costs very little and removes the assumption. If it's straddling the header line specifically, the session is lost entirely.

## Smaller notes

`mapToolName` does `toolNameMap[raw] ?? raw` on an object literal, so a tool named `constructor`, `toString`, `valueOf` or `hasOwnProperty` resolves up the prototype chain and pushes a function into `tools: string[]`. `raw` comes from parsed session JSON. A `Map`, or a `hasOwn` check, closes it — and it's the same class of thing the Semgrep bracket-assign rule exists to catch.

The `[22.13.0, 22]` matrix on `tests.yml` is well justified given the 22.15 zstd gate, and the check-name change it causes (`test` becomes `test (22.13.0)` / `test (22)`) is safe here — the `Main Rule` ruleset requires an approving review but pins no status-check contexts, so nothing breaks. Worth a conscious yes/no on doubling CI minutes for every PR in the repo, though; gating the second entry on paths touching the zstd path would get most of the value.

One thing to flag on the current head: there are no check runs on this PR at all, so none of the suite has actually run in CI for the DSH code. Worth getting green before merge rather than after.

PR #1003, opened 2026-08-16
fix(optimize): scope transcript-derived findings to the selected provider

### LINKED ISSUE #1002: optimize --provider <non-claude> reports findings computed from Claude sessions
`codeburn optimize --provider <non-claude>` prints a header scoped to the selected provider, but the findings underneath are computed from Claude Code transcripts. The two halves of the report describe different providers, and the Claude-derived numbers read as the selected provider's.

## Environment

- `codeburn --version`: 0.9.20
- macOS 15 (Apple Silicon), Node 24.19.0
- Providers with data: Claude, Codex, Copilot, Antigravity, Gemini, Warp
- History size: `~/.claude/projects` 1.0G, `~/.codex/sessions` 2.9G

## Repro

Run the same command three ways:

```bash
codeburn optimize                    # all providers
codeburn optimize --provider codex
codeburn optimize --provider claude
```

Header line vs. the "Claude edits more than it reads" finding:

| run | header sessions | header calls | finding: reads | finding: edits |
|---|---|---|---|---|
| (no flag) | 2,403 | 133,623 | 16,341 | 10,461 |
| `--provider codex` | 347 | 75,929 | 16,368 | 10,510 |
| `--provider claude` | 2,046 | 57,580 | 16,369 | 10,512 |

The header filters correctly. The read/edit counts do not move — they are the same Claude-derived numbers in all three runs. (The small drift is the rolling 30-day window advancing between runs, which were minutes apart; note the `--provider codex` figure is *higher* than the unfiltered one, so it cannot be a subset of it.)

Codex accounts for 57% of calls but only 14% of sessions here, so if the finding were actually scoped, the three rows could not agree.

## Cause

`scanSessions()` in `src/optimize.ts` hardcodes the provider:

```ts
async function scanSessions(dateRange?: DateRange): Promise<ScanData> {
  const sources = await discoverAllSessions('claude')
```

`--provider` never reaches it. `runOptimize` receives the already-filtered `projects` (so the header is right) but `scanAndDetect` calls `scanSessions(dateRange)` with no provider, so every detector fed by `toolCalls` / `projectCwds` / `apiCalls` / `userMessages` reports Claude data:

`read-edit-ratio`, `junk-reads`, `duplicate-reads`, `unused-mcp`, `mcp-deferral-*`, `bloated-claude-md`, `bash-bloat`, `unused-skills`, `unused-agents`, `unused-commands`.

## Expected

Under a `--provider` that excludes Claude, those findings should not appear at all. Note that simply passing the filter down and letting the scan return empty is not sufficient — an empty scan is an absence of measurement, not a measurement of absence, and the ghost-item detectors read emptiness as "never invoked" and report *every* skill, agent and command as unused. On this machine that turns 19 reported unused skills into 38.

The `projects`-derived findings (MCP tool coverage, capability reliability, low-worth sessions, context bloat, session outliers, model recommendations) already filter correctly and should keep running — `--provider codex` legitimately surfaces Codex MCP servers in the retry-correlation finding today.

I have a fix and a regression test ready; happy to open a PR if the approach above sounds right.

AndrewDongminYoo:
## Summary

- `scanSessions()` hardcoded `discoverAllSessions('claude')`, so every finding it feeds was computed from Claude transcripts regardless of `--provider`, while the header came from the already-filtered `projects`. `optimize --provider codex` printed a Claude read/edit ratio, Claude ghost skills and Claude MCP findings under a Codex header.
- Skip the scan when the filter excludes Claude, and skip the detectors it feeds rather than handing them an empty scan — emptiness reads as "never invoked", so an empty scan reports every skill, agent and command as unused (19 → 38 on my machine).
- `projects`-derived findings (MCP tool coverage, capability reliability, low-worth sessions, context bloat, outliers, model recommendations) already filter correctly and still run. `cacheKey` now carries the provider, since the provider decides whether the scan runs at all.

Fixes #1002

## Testing

- [x] `npm test` passes

`npm test`: 199 files, 2,631 passed, 5 skipped, 0 failed. Two new cases in `tests/optimize-fs.test.ts` cover the positive (`--provider claude` still reports `read-edit-ratio`) and negative (`--provider codex` reports neither `read-edit-ratio` nor `unused-skills`) paths. I verified the negative case fails without the fix:

```log
AssertionError: expected [ 'read-edit-ratio', …(2) ] to not include 'read-edit-ratio'
 ❯ tests/optimize-fs.test.ts:413
```

Against real data (30-day window, Claude + Codex + Copilot + Antigravity + Gemini + Warp on disk):

```log
BEFORE
  --provider codex    347 sessions   75,929 calls   Health: F (20/100, 14 issues)
                      ─── 1. Claude edits more than it reads (16,368 reads / 10,510 edits)
                      ─── 10. 19 skills you never use

AFTER
  --provider codex    347 sessions   75,929 calls   Health: D (48/100, 4 issues)
                      ─── 1. 6 MCP servers correlate with retry-heavy edits
                      ─── 2. 79 possibly low-worth expensive sessions
                      ─── 3. 62 context-heavy sessions
                      ─── 4. 2 high-cost session outliers

  --provider claude   2,051 sessions   57,812 calls   Health: F (20/100, 14 issues)   unchanged
  (no flag)           2,409 sessions  134,068 calls   Health: F (20/100, 14 issues)   unchanged
```

The four findings that survive under `--provider codex` are genuinely Codex-derived; the Claude-only ones are gone. The default and `--provider claude` paths are byte-for-byte unchanged in finding count and health grade.

### DIFF
--- a/src/act/optimize-apply.ts
+++ b/src/act/optimize-apply.ts
@@ -13,6 +13,10 @@ export type ApplyOptions = {
   yes?: boolean
   dryRun?: boolean
   only?: string
+  // Mirrors `optimize --provider`. The scan below only reads Claude
+  // transcripts, and this path does not just report findings, it plans and
+  // applies them - a Codex-scoped run must never offer to edit ~/.claude.
+  provider?: string
   actionsDir?: string
   ctx?: PlanContext
   // Test seams: crafted findings skip the session scan; streams default to
@@ -103,7 +107,7 @@ export async function runOptimizeApply(
   let costRate = opts.costRate ?? 0
   if (!findings) {
     errout.write(chalk.dim('  Analyzing your sessions...\n'))
-    const scanned = await scanAndDetect(projects, dateRange)
+    const scanned = await scanAndDetect(projects, dateRange, opts.provider)
     findings = scanned.findings
     costRate = scanned.costRate
   }
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -1460,14 +1460,14 @@ export function InteractiveDashboard({ initialProjects, initialDailyHistoryProje
     const generation = reloadGenerationRef.current
     setOptimizeLoading(true)
     try {
-      const result = await scanAndDetect(projects, currentRange())
+      const result = await scanAndDetect(projects, currentRange(), activeProvider)
       if (reloadGenerationRef.current === generation) setOptimizeResult(result)
     } catch (error) {
       console.error(error)
     } finally {
       if (reloadGenerationRef.current === generation) setOptimizeLoading(false)
     }
-  }, [optimizeAvailable, projects, currentRange, optimizeLoading, optimizeResult])
+  }, [optimizeAvailable, projects, currentRange, optimizeLoading, optimizeResult, activeProvider])
 
   useEffect(() => {
     const refreshIntervalMs = getRefreshIntervalMs(refreshSeconds ?? 0)
--- a/src/main.ts
+++ b/src/main.ts
@@ -1849,9 +1849,9 @@ program
         appliedHeader = buildOptimizeAppliedHeader(applied) ?? undefined
         previouslyApplied = applied.appliedByFinding
       } catch { /* the header is optional; never block the findings */ }
-      await runOptimize(projects, label, range, { format, appliedHeader, previouslyApplied })
+      await runOptimize(projects, label, range, { format, appliedHeader, previouslyApplied, provider: opts.provider })
     } else {
-      await runOptimize(projects, label, range, { format })
+      await runOptimize(projects, label, range, { format, provider: opts.provider })
     }
   })
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -1831,7 +1831,7 @@ program
     const projects = await parseAllSessions(range, opts.provider)
     if (opts.apply) {
       const { runOptimizeApply } = await import('./act/optimize-apply.js')
-      await runOptimizeApply(projects, range, { yes: opts.yes, dryRun: opts.dryRun, only: opts.only })
+      await runOptimizeApply(projects, range, { yes: opts.yes, dryRun: opts.dryRun, only: opts.only, provider: opts.provider })
       return
     }
     assertFormat(format, ['text', 'json'], 'optimize')
--- a/src/optimize.ts
+++ b/src/optimize.ts
@@ -551,7 +551,18 @@ export async function scanJsonlFile(
   return { calls, cwds, apiCalls, userMessages }
 }
 
-async function scanSessions(dateRange?: DateRange): Promise<ScanData> {
+// The session scan reads Claude Code transcripts only, so a `--provider` that
+// excludes Claude leaves nothing for it to do. Callers must also skip the
+// detectors it feeds (see `claudeOnly` in scanAndDetect) — the empty scan
+// returned here is an absence of measurement, not a measurement of absence.
+export function providerCoversClaude(provider?: string): boolean {
+  return !provider || provider === 'all' || provider === 'claude'
+}
+
+async function scanSessions(dateRange?: DateRange, provider?: string): Promise<ScanData> {
+  if (!providerCoversClaude(provider)) {
+    return { toolCalls: [], projectCwds: new Set(), apiCalls: [], userMessages: [] }
+  }
   const sources = await discoverAllSessions('claude')
   const allCalls: ToolCall[] = []
   const allCwds = new Set<string>()
@@ -2977,7 +2988,7 @@ export function computeInputCostRate(projects: ProjectSummary[]): number {
 type CacheEntry = { data: OptimizeResult; ts: number }
 const resultCache = new Map<string, CacheEntry>()
 
-export function cacheKey(projects: ProjectSummary[], dateRange: DateRange | undefined): string {
+export function cacheKey(projects: ProjectSummary[], dateRange: DateRange | undefined, provider?: string): string {
   const dr = dateRange ? `${dateRange.start.getTime()}-${dateRange.end.getTime()}` : 'all'
   // Fingerprint enough of the dataset that two materially different inputs
   // cannot collide onto one cached OptimizeResult. Project count + api-call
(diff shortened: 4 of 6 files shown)


ozymandiashh (changes requested):
The diagnosis is right and the fix is the right shape. Two things in particular I want to call out as good calls, because they're the parts that are easy to get wrong:

Skipping the detectors rather than handing them an empty scan. An empty `ScanData` is an absence of measurement, not a measurement of absence, and `detectGhostSkills` and friends can't tell the difference — feeding them nothing would have turned "we didn't look" into "every skill you own is unused." Your comment says exactly that, which is the kind of note that stops someone helpfully "simplifying" it back in six months.

Putting the provider in `cacheKey`. The provider decides whether the scan runs at all, so two filters sharing a project fingerprint genuinely must not share a result. Easy to miss, and it would have produced a very confusing intermittent bug.

I also checked whether a comma-separated `--provider claude,codex` could slip past `providerCoversClaude` — it can't, `assertProvider` only accepts `all` or a single name, so the three-way check is complete.

## What I'd like added before merge

`scanAndDetect` has three callers that carry a provider filter, and this threads it through one of them.

**1. `optimize --apply` — same command, same flag.** `src/main.ts:1832-1835` branches to `runOptimizeApply(projects, range, { yes, dryRun, only })` before ever reaching the code you changed, and `src/act/optimize-apply.ts:106` calls `scanAndDetect(projects, dateRange)` with no provider. So `codeburn optimize --provider codex --apply` still computes Claude findings — and this path doesn't just print them, it plans and applies them. If anything it's the more important half: `--provider codex` offering to edit `~/.claude` config is worse than merely reporting a Codex-labelled read/edit ratio.

**2. `usage-aggregator` — the menubar, desktop and web surfaces.** `AggregateOpts.provider` exists and is applied at `src/usage-aggregator.ts:397` and `:525`, but `:938` calls `scanAndDetect(scanProjects, scanRange)` without it. Same issue #1002 describes, on the GUI surfaces.

`src/dashboard.tsx:1463` looks genuinely fine to leave as-is, since the TUI has no provider filter to honour — but worth a quick confirmation from you rather than my say-so.

Both fixes are the same one-argument change you've already made, and `providerCoversClaude` is exported, so the pieces are in place. A test on the apply path would be worth having, since that's the one with real side effects.

The test cleanup in `afterEach` is a nice touch — restoring the shared fake home so later suites still see an empty `~/.claude` is the sort of thing that usually only gets noticed after it breaks someone else's suite.

Ping me when those two are threaded through and I'll get this merged.

ozymandiashh:
Correction on my last comment: I said `src/dashboard.tsx:1463` looked fine to leave alone because the TUI has no provider filter to honour. That's wrong — it has one. `src/dashboard.tsx:1691` builds its projects with `parseAllSessions(range, provider)`, and the view supports switching provider interactively (`reloadData(pending.period, pending.provider, ...)` at `:1445`). So the optimize panel shows a provider-scoped header next to Claude-derived findings, and the mismatch moves as the user tabs between providers while the findings stay put.

That makes it three call sites, not two: `optimize --apply`, `usage-aggregator`, and the dashboard. Sorry for the extra round trip.

One thing worth flagging from a closer read, because it's a point in the PR's favour rather than against it: `detectMcpAlwaysLoadHygiene` is the one detector in the skipped set that could already emit a finding under a non-Claude filter. It calls `loadMcpConfigs`, which reads `~/.claude/settings*.json` regardless of `projectCwds`, and unlike `detectMcpDeferralOff` it isn't internally Claude-scoped — so under `--provider codex` it was charging Claude's `alwaysLoad` pins against Codex session counts. Skipping it fixes a second scoping bug that nobody reported. Might be worth a line in the PR description, since it's a real behaviour change beyond what #1002 describes.

AndrewDongminYoo:
Both threaded, plus a third — `dashboard.tsx` does have a provider filter, so I fixed it too rather than leaving it. Details below since you asked for a confirmation rather than taking it on my say-so.

## `optimize --apply`

Agreed this was the important half. `ApplyOptions` gains `provider`, `main.ts` passes it at the branch, and `scanAndDetect` receives it.

Worth recording why it matters concretely: `unused-skills` is appliable, and `buildArchive` moves directories out of `~/.claude/skills`. Neutralising just the one-argument change and running the new test reproduces it:

```
× plans no applies from Claude findings when scoped to another provider
  → expected '\n  Appliable config-class fixes:\n\n…' to contain 'No appliable config-class fixes'
```

So before this commit, `optimize --provider codex --apply` really did offer to archive Claude skills. Against my own data, after:

```
$ codeburn optimize --provider codex --apply --dry-run
  No appliable config-class fixes for this period.
```

Unfiltered and `--provider claude` still print `Appliable config-class fixes:` — unchanged.

## `usage-aggregator`

One-argument change at `:938`. `AggregateOpts.provider` was already honoured for the usage half at `:397` and `:525`, so this just makes the optimize half agree with it.

## `dashboard.tsx` — this one is not fine as-is

The TUI does have a provider filter to honour:

- `:1301` — `const [activeProvider, setActiveProvider] = useState(initialProvider)`
- `:1547` — `if (input === 'p' && multipleProviders …)` cycles `['all', ...detectedProviders]` and calls `reloadData(period, next, dayDate)`
- `:1528` — `if (input === 'o' && view === 'dashboard' && optimizeAvailable) { void loadOptimizeResult() }`
- `loadOptimizeResult` called `scanAndDetect(projects, currentRange())` with no provider

So `p` to Codex then `o` showed Claude findings — issue #1002 on a third surface. It now passes `activeProvider`, which also joins the `useCallback` deps so the cycled value isn't captured stale.

I checked whether switching provider needs the memo invalidated separately: it doesn't. `reloadData` already calls `setOptimizeResult(null)` on the non-background path (`:1401`), and the `p` handler goes through it, so the next `o` re-scans.

## One thing I hit that you may want to know about

`scanAndDetect` memoises for 60s on `(provider, range, project fingerprint)` and not on any on-disk state. It bit the new tests — two cases differing only in what was in the fake home served each other's results — so the fixtures now vary the fingerprint with a `seed`, commented in place.

In production the same shape means `optimize` followed by `optimize --apply` inside a minute replans off the first scan. That is usually what you want, but it goes stale if config changed in between. Pre-existing and out of scope here; flagging rather than touching it.

## Verification

- `npm test`: 205 files, 2717 passed, 5 skipped, 0 failed
- `npx tsc --noEmit` clean, `npm run build` succeeds
- Negative control run per call site: reverting each one-argument change fails the corresponding assertion
- Rebased onto current `main` (34 commits) — clean, no conflicts

Thanks for the careful read on the empty-scan point; that was the part I got wrong on the first attempt and only caught by running it.

AndrewDongminYoo:
@ozymandiashh both are threaded through and pushed — details in the comment above.

One thing needs your eyes rather than just a re-read: I did **not** leave `dashboard.tsx` as-is. The TUI does carry a provider filter (`activeProvider` at `:1301`, cycled by `p` at `:1547`, and `o` at `:1528` opens optimize off that same state), so `p` to Codex then `o` showed Claude findings. That is a deviation from what you specified, so please sanity-check it — the reasoning and line references are in the previous comment.

ozymandiashh (approved):
All three threaded, and thanks for going and checking `dashboard.tsx` rather than taking my word that it was fine — it wasn't, and you found the stale-capture problem in the `useCallback` deps that I would not have.

Verified on `fa2505c` here:

- `tests/optimize-fs.test.ts` + `optimize.test.ts` + `optimize-apply.test.ts` — 151 passed, `tsc --noEmit` clean
- negative control on the apply path: reverting the one-argument change at `src/act/optimize-apply.ts` fails `plans no applies from Claude findings when scoped to another provider`, so the appliable-archive path is genuinely pinned. That was the one I most wanted covered, given it can move `~/.claude/skills` directories
- all six checks green

The `usage-aggregator` call site is the one without a direct assertion — the existing `buildMenubarPayloadForRange` suites all pass `optimize: false`, so reverting it stays green. Not asking you to add one: it's a single argument, symmetric with the two that are covered, and the type checker carries it. Noting it so it's on the record rather than assumed.

The memoisation note is a good catch and worth keeping visible — `optimize` followed by `optimize --apply` inside the 60s window replanning off the first scan is fine right up until config changed in between, and nothing in the key would notice. Agreed it's pre-existing and out of scope here; worth its own issue if you feel like filing one.

Merging. Closes #1002.

Issue #1011, opened 2026-08-17
Flatpak install method via FlatPark


jing2uo:
Hi @iamtoruk — thanks for CodeBurn, it is genuinely useful, and the provider coverage is impressive.

I maintain [FlatPark](https://github.com/flatpark/flatpark), a small hub that offers Flatpak installs for apps that aren't on Flathub, and I've packaged the CodeBurn desktop app there. I'm opening this issue because there was no existing thread for it — and because I saw #970 and the `app/flathub/` manifest in your tree, so this is clearly a path you already wanted.

It uses Flatpak **extra-data**: the package downloads your **official `codeburn-desktop_<ver>_amd64.deb`** from GitHub Releases at install time, unmodified, and a resolver tracks new `desktop-v*` tags automatically. I don't rehost, patch or rebuild anything — the bytes users run are yours. Same shape as your own manifest otherwise: Electron base app, launched through `zypak-wrapper`.

```
flatpak remote-add --if-not-exists flatpark https://dl.flatpark.org/flatpark.flatpakrepo
flatpak install flatpark org.agentseal.CodeBurn
```

App page: https://flatpark.org/apps/org.agentseal.CodeBurn

Two packaging notes, in case either is useful to you:

- **The read-only enumeration from #978 translates straight across.** Instead of `--filesystem=home`, the FlatPark package grants each provider directory separately and read-only (`--filesystem=~/.claude:ro`, `~/.codex:ro`, `~/.config/Cursor/User/globalStorage:ro`, and so on — 18 of them), which is the Flatpak spelling of the `personal-files` plug you settled on for the snap. Your snap list is the better reference of the two, since you enumerated all 25; if you'd like, I'll align ours with it.
- **Two paths need the app's own env knobs once the grant is narrowed.** Flatpak redirects the XDG variables but not `$HOME` itself, so `$HOME/.cache/codeburn` lands on the sandbox tmpfs and every launch re-fetches the LiteLLM table and re-parses cold; the wrapper sets `CODEBURN_CACHE_DIR` into the per-app cache instead. Likewise `OPENCODE_DATA_DIR`, because OpenCode resolves its store through `XDG_DATA_HOME`, which Flatpak points inside the sandbox. Both are set only if unset, so a user can override them. (`~/.config/codeburn` is granted directly, so settings stay shared with a natively installed CLI.)

[...]

PR #1013, opened 2026-08-18
fix(flatpak): make the home grant read-only

ozymandiashh:
> **Blocked on #1017.** An adversarial pass found that `home:ro` breaks every WAL-mode SQLite provider: a read-only SQLite open still has to create `-shm`/`-wal` next to the database, so on a read-only parent it fails with `attempt to write a readonly database` — and both discovery sites swallow that silently. That covers cursor, cursor-agent, opencode, goose, warp, kilo-code, zerostack and the copilot agent-traces DB, whenever the tool in question was last shut down cleanly. #1017 fixes the open; this rebases on it. Details in the comment below.

Reported in #1011 by the FlatPark maintainer while packaging CodeBurn downstream.

## The problem

`app/flathub/org.agentseal.CodeBurn.yaml` asked for `--filesystem=home` — read-write access to the user's entire home directory, for an app that only ever reads. The snap was narrowed to a read-only `personal-files` plug in #978; the Flatpak manifest never got the same treatment, so the two packages have been shipping very different permission stories.

## Why this isn't the snap list

My first cut at this mirrored the snap's 35-entry read list as individual `:ro` grants. That was wrong, and it's worth writing down why, because it's the obvious thing to reach for.

**It cannot cover project directories.** Every entry on that list is a dot-directory. A user's repository lives at an arbitrary path like `/home/u/code/myrepo`, which nothing on the list matches. `resolveRepoIdentity` shells out to `git` with `cwd` set to the project path (`src/yield.ts:59`); inside the sandbox that directory doesn't exist, `runGit` returns null, and `codeburn yield` ends up attributing nothing and classifying essentially all spend as abandoned. `CLAUDE.md` discovery and per-project MCP config fail the same way.

**The list isn't a superset of what the providers actually read.** Cross-checking every `homedir()`-rooted path in `src/providers/` against it turns up `~/.kimi-code` (`kimicode.ts:82`), `~/.clawdbot/agents`, `~/.moltbot`, `~/.moldbot` (`openclaw.ts:66-68`), `~/.local/share/devin/cli` (`devin.ts:161`) and the multi-account Claude roots — none of them granted. A missing grant fails as `ENOENT`, which is indistinguishable from "that tool isn't installed", so those providers would report zero sessions with no error and no doctor warning.

The snap gets away with its list because it *also* holds the generic `home` plug. There's no Flatpak equivalent that quietly does the same job.

## What this does instead

`--filesystem=home:ro`. Reading stays unrestricted, which is what CodeBurn actually needs; writing is gone, which is the property worth having and the whole point of the report.

Two writable exceptions, both declared:

- `~/.config/codeburn:create` — everything CodeBurn writes resolves under `dirname(getConfigFilePath())`: `config.json`, the act journal and its lock, act backups, the sharing store, sync credentials, mcp redact state.
- `xdg-download:create` — `codeburn export` needs a real destination. Under a purely read-only home it would `mkdir -p` into the sandbox tmpfs, write eleven files, and return success, with the caller's existence guard never tripping.

The wrapper keeps the two env knobs from the original report: `CODEBURN_CACHE_DIR` at the persistent XDG cache (`$HOME` is not redirected inside the sandbox, so the default lands on tmpfs and every launch re-fetches the LiteLLM table and cold-parses), and `OPENCODE_DATA_DIR` back at the host store that `XDG_DATA_HOME` no longer resolves to. Both fill in only when the user hasn't set them to a non-empty value — `${VAR:+x}` rather than `${VAR+x}`, because `getCodeburnCacheDir` (`override?.trim() ? …`) and `getDataDir` (`if (override)`) both treat empty as unset. `XDG_CACHE_HOME` carries its own `:-` default: unset, it expanded to the literal `/codeburn`.

## The test

`tests/flatpak-grants.test.ts` pins the read-only property rather than a path list — a list test would only detect divergence between two hand-maintained copies of the same incomplete idea, and could never catch a provider that's in neither.

It fails if a writable whole-home grant reappears, if `home:ro` goes missing, or if a writable grant is added without an entry in its `ALLOWED_WRITABLE` map explaining why. Mutation-checked all three:

```
reintroduce --filesystem=home     -> 3 failed: "grants write access to the whole home: home … use home:ro"
add --filesystem=~/.ssh:create    -> 1 failed: "Undeclared writable Flatpak grant(s): ~/.ssh"
drop the XDG_CACHE_HOME default   -> 1 failed
restored                          -> 4 passed
```

## Verification

- `npx vitest run tests/flatpak-grants.test.ts` — 4 passed
- `npx tsc --noEmit` — clean
- `yaml.safe_load` parses the manifest; the wrapper commands extracted from it pass `sh -n`

**What I could not verify:** this is a macOS machine with no `flatpak` or `flatpak-builder`, so I have not built the package, inspected the resulting sandbox mounts, or run the wrapper under `zypak-wrapper`. Two things in particular deserve a check on a real Linux build before this ships: that a `:create` grant on a subpath of `home:ro` is honoured as writable (Flatpak documents most-specific-match, but I'd rather see it), and that the portal-backed directory picker still yields a usable export destination outside `xdg-download`.

The release `url`/`sha256` pair is deliberately untouched.

## Follow-up, not in this PR

**Six providers resolve through XDG variables, which Flatpak redirects into the sandbox.** This is pre-existing — XDG redirection is a property of the runtime, not of the filesystem grants, so it was equally broken under `--filesystem=home` — but the analysis is worth recording since this PR fixes one instance of it (OpenCode) and someone will hit the rest:

| Provider | Resolves through | Usable env override | Host default |
|---|---|---|---|
| `opencode` | `XDG_DATA_HOME` | `OPENCODE_DATA_DIR` (exact dir) | fixed here |
| `crush` | `XDG_DATA_HOME` | `CRUSH_GLOBAL_DATA` (exact dir) | `~/.local/share/crush` |
| `zerostack` | `XDG_DATA_HOME` | `ZS_DATA_DIR` (parent of `sessions/`) | `~/.local/share/zerostack` |
| `goose` | `XDG_DATA_HOME` | `GOOSE_PATH_ROOT` — **cannot express it** | `~/.local/share/goose/sessions/sessions.db` |
| `kilo-code` | `XDG_DATA_HOME` | none | `~/.local/share/…` |
| `copilot` | `XDG_CONFIG_HOME` | none for that path | `~/.config/github-copilot` |
| `ibm-bob` | `XDG_CONFIG_HOME` | none | `~/.config/…` |

`GOOSE_PATH_ROOT` is the interesting one: it resolves to `<root>/data/sessions/sessions.db` while the XDG path is `<base>/goose/sessions/sessions.db`, so no value of it reproduces the host layout.

I deliberately did not extend the wrapper here. Fixing three of six in a grant PR would be a confusing partial, and the other three need an upstream env knob that does not exist yet. Resetting `XDG_DATA_HOME` / `XDG_CONFIG_HOME` wholesale in the wrapper would cover all six but also moves Electron's own Chromium profile onto a now read-only host path, which I have no way to test from here. Worth its own issue.

## Snap follow-up, also not in this PR

`~/.local/share/opencode`, `~/.kimi-code`, `~/.clawdbot/agents`, `~/.moltbot`, `~/.moldbot`, `~/.local/share/devin/cli` and `~/.config/manicode-dev` / `-staging` are all absent from the snap's `personal-files` list, and snap's `home` interface exposes non-hidden files only — so a store under a dot-directory is outside it. The snap may be missing those providers entirely. Separate package, separate store review, so it wants its own change rather than riding along here.


ozymandiashh:
Holding this as a draft. An adversarial pass over my own branch found that `home:ro` breaks a large fraction of providers, so both versions of this PR have now been wrong for different reasons.

**A read-only SQLite open still writes.** `openDatabase` uses `new DatabaseSync(path, { readOnly: true })` (`src/sqlite.ts:123`), but `SQLITE_OPEN_READONLY` is not sufficient for a WAL-mode database: SQLite must create `<db>-shm` (and an empty `-wal`) in the database's own directory unless they already exist. Reproduced on this machine against a real WAL database with the same Node and the same call:

```
=== writable dir (today, --filesystem=home) ===
READ OK
  side effect: state.vscdb-shm, state.vscdb-wal created

=== read-only dir (this PR, home:ro) ===
READ FAILED: Error ERR_SQLITE_ERROR attempt to write a readonly database
```

It is conditional on `-shm` being absent, which is precisely the state after the tool exits cleanly. So the symptom would be "Cursor spend disappears whenever Cursor is not running" — intermittent and undiagnosable. Both discovery sites swallow the error with a bare `catch { continue }` (`src/providers/sqlite-session-parser.ts:312`, `src/providers/cursor.ts:190`), so the provider reports zero sessions with nothing on stderr. That is exactly the failure mode this PR's own README argues against.

It covers Cursor, cursor-agent, opencode, goose, warp, kilo-code, zerostack and the copilot agent-traces DB.

Two things follow from this that are bigger than the manifest:

1. **The snap is probably affected too.** #978 narrowed it to a read-only `personal-files` plug, which is the same read-only bind mount. Worth checking on a real snap install before this is treated as Flatpak-specific.
2. **The "CodeBurn only ever reads" claim is not true today.** A read-only open creates two files in the user's provider directory. That is worth knowing regardless of packaging.

So the fix belongs in the code, not the manifest: `openDatabase` needs to survive a non-writable parent — an `immutable=1` URI where safe, or copying the database plus its `-wal` into the cache directory and opening the copy — and the discovery sites need to stop swallowing `SQLITE_READONLY` silently. I am doing that as its own change, since it stands on its own merits; this PR will rebase on it.

iamtoruk:
Reviewed together with #1017 (merged locally, only CHANGELOG conflicts; grants test 4/4; manifest parses; final grants `home:ro`, `~/.config/codeburn:create`, `xdg-download:create`).

Verified: the cache needs no extra grant — Flatpak points `XDG_CACHE_HOME` at `~/.var/app/<id>/cache` (always writable) and the wrapper's `CODEBURN_CACHE_DIR` follows it; `src/config.ts` resolves the config dir from `homedir()`, so `~/.config/codeburn:create` matches what the code writes; an unwritable cache degrades to a notice + totals still render.

Holding it until:
1. **#1017 lands with the CANTOPEN fix** (a read-only parent whose DB has a stale `-wal` and no `-shm` fails with `SQLITE_CANTOPEN`, not `READONLY`, so the current fallback never triggers and every WAL provider still silently reports zero — exactly the state `home:ro` produces after an unclean shutdown). Being fixed on #1017 now.
2. One line in README + CHANGELOG saying what `home:ro` gives up: `optimize --apply` / act plans (writes to `~/.claude.json`, `~/.claude/settings.json`, project `CLAUDE.md`, shell rc) and the antigravity statusline install become EROFS in the Flatpak — acceptable for a sandboxed analytics build, but it must be stated.
3. The manifest comment's rationale: with a home grant `$HOME` is not redirected, so `~/.cache` is the real host dir, not tmpfs — under `home:ro` it's simply read-only. Right fix, wrong reason.
4. One real Linux Flatpak build confirming the two things neither of us can verify from macOS: that `:create` on a subpath of `home:ro` is honoured, and that `XDG_CACHE_HOME` lands writable.

Worth noting for the README once it lands: under `home:ro` the #1017 copy fallback becomes the *normal* path for every SQLite provider after a clean shutdown (a 1.9 GB Cursor `state.vscdb` copied into the app cache on each change) — #1017 is getting an `immutable=1` shortcut for the no-`-wal` case, which makes that free.

iamtoruk:
#1017 is merged (with the CANTOPEN fix, immutable in-place open when there is no WAL on Node ≥ 22.15, notice on unwritable cache, Windows-safe copies, eviction). Remaining before this can land: the README/CHANGELOG line on what `home:ro` gives up (`optimize --apply`/act writes, antigravity statusline install), the manifest comment's tmpfs rationale, and one real Linux Flatpak build confirming `:create` on a `home:ro` subpath and a writable `XDG_CACHE_HOME`. Rebase onto main when ready.

ozymandiashh:
Closing for now — drifted behind main with conflicts, and the slot is needed under the five-open-PR cap. The branch stays on the fork for a future rebase.

Issue #997, opened 2026-08-14
Chart legend shows session UUIDs instead of session titles


ghorbanies:
**Version:** 0.9.20 (npm, global) · **OS:** Windows 11 · **Provider:** claude

## What happens

In `codeburn web` → **Usage**, when "Hourly buckets" is grouped by **Sessions**, the legend
renders each series as the project path plus a truncated session UUID:

```
D:/repo · a1b2c3…7f01 (claude)   D:/repo · d4e5f6…7f02 (claude)
D:/repo · 0a1b2c…7f03 (claude)   D:/repo · 3d4e5f…7f04 (claude)
```

Every series has the same project prefix, so the only distinguishing part is a hex fragment.
There is no way to tell which line is which piece of work without leaving the chart.

## Why this is awkward

This is the normal case for a monorepo. A single repository holding several applications,
worked on in multiple concurrent sessions, produces a set of series that all share the same
project path — so the legend shows half a dozen entries that are visually indistinguishable.

Per-application cost attribution is the main reason to open the dashboard, and it is the one
thing the chart cannot currently answer.

## The data already exists

The **Context** tab already displays session titles next to the short id, e.g.:

```
a1b2c3d4   Refactor billing module          10.5MB
d4e5f678   Add reporting screen             17.4MB
```

`codeburn sessions --format json` also returns a `title` field per session. So the mapping is
already available — it just isn't used in the chart legend.

## Suggested change

**1. Prefer `title` in the session legend, falling back to the short id when a session has no
title.** Something like `Refactor billing module (claude)` instead of
`D:/repo · a1b2c3…7f01 (claude)`. Keeping the short id as a suffix or tooltip would preserve
the current disambiguation.

**2. (Secondary, larger) Make legend entries clickable to filter the view to a single
session** — isolating one series and having the summary tiles below recompute for it. Today
narrowing to one session means dropping to the CLI.

Item 1 alone would resolve the main problem; item 2 is a convenience on top.

Happy to test a change against a multi-session monorepo setup if that is useful.

PR #1014, opened 2026-08-18
fix(dash): lead the session legend with the session title

ozymandiashh:
Reported in #997 by a user on a monorepo.

## The problem

Grouping the hourly chart by session labelled every series `<project> · <short-id> (<provider>)`. In a monorepo that's the normal case: several applications in one repository, worked on in concurrent sessions, so every series carries the same project prefix and the only thing separating them is a truncated hex fragment.

```
D:/repo · a1b2c3…7f01 (claude)   D:/repo · d4e5f6…7f02 (claude)
D:/repo · 0a1b2c…7f03 (claude)   D:/repo · 3d4e5f…7f04 (claude)
```

Per-application cost attribution is the main reason to open that chart, and it was the one question the chart couldn't answer.

## The fix

The data was already there. `SessionSummary.title` is parsed from the transcript (`src/types.ts:264`) and the Context tab already renders it beside the short id. `buildGranularHistory` just wasn't using it.

The legend now prefers the title and keeps the short id as the suffix:

- with a title: `Refactor billing module · a1b2c3…7f01 (claude)`
- without one: unchanged — `app/web · a1b2c3…7f01 (claude)`

Keeping the id matters for the reporter's own scenario: two concurrent sessions in a monorepo can easily carry the same title, and the suffix is what keeps those two series apart. There's a test for exactly that.

Titles come from parsed transcripts, so they get the same treatment model names get in `src/models.ts`: ANSI stripped, control characters flattened to spaces rather than deleted (so a transcript line break can't join two words), whitespace collapsed, and a length cap. The parser's own cap is 200 characters, which would swamp both the `max-w-40` legend entry and the tooltip.

The label is also built once per session rather than once per API call. It was already being re-`set` on every call, but sanitising a title on every call in a large corpus is real work for no benefit, and every input to the label is constant for a given series key.

## Second pass: what an adversarial review of this branch turned up

I ran a hostile pass over my own change before asking anyone to look at it. Four things came back, all now fixed in the second commit.

**Two series could render byte-identical labels.** `sessionKey` is provider/projectPath/sessionId, so the same session id under two project paths is two separate series — a session resumed in a different cwd, or two Claude config directories holding the same project slug. The old label led with the project, so they stayed apart; leading with the title removed the only thing separating them, and the legend showed two identical entries.

Labels are now built once at the end from per-key inputs rather than inline. Identical base labels get the short project label appended; the residual case (two worktrees whose last two path segments match) falls back to full path plus full session id. The ordinary monorepo case still omits the project entirely, which is the whole point of the issue.

**The label depended on cache state.** The hoist carried a comment claiming every input was constant for a given `sessionKey`. That's false — `title` and `project` are not in the key, so two `SessionSummary` objects can share a key with different titles, and the `has()` guard meant whichever arrived first won. This reproduces through the real parser: two config dirs with the same project slug and session id, one carrying a title, and `scanProjectDirs` merges them into one project. Title candidates are now collected per key and a valid one preferred over none, deterministically.

**The cap sliced UTF-16 code units.** A title with an astral character straddling the boundary left a lone high surrogate in the payload. It counts code points now.

**A pre-existing injection in the dashboard bootstrap.** `serveIndexHtml` builds the bootstrap with `html.replace(string, string)`, so `$`-substitution patterns in the replacement string are interpreted. A payload value containing `` $` `` or `$'` expands to the raw document around the match — which contains `</script>` — so the `<` escaping right above it does not stop the tag from closing early.

To be clear about attribution: this is **not introduced by this branch**. `git diff upstream/main...HEAD -- src/web-dashboard.ts` was empty before the fix, and device names, project names and model names already reached that sink. Session titles only widen the input surface, since AI-generated free text is far likelier to contain those sequences than a filesystem path. It's one line and adjacent to what this branch touches, so it's fixed here rather than filed: the injection is factored into `injectDashboardBootstrap` and uses a replacer function. Happy to split it into its own PR if you'd rather it not ride along.

## Third pass

A further adversarial pass found the cap was sized against the wrong number, and turned up two more.

**The 80-code-point cap was ~2.5x what the legend renders.** `max-w-40` is 160px and the legend renders at `text-[10px]`, which shows roughly 32 characters. Everything past that is clipped — and that is where the short session id, the provider and every collision-tier suffix lived. Two sessions in one repository whose titles share a 32-character prefix rendered as the *same* legend entry, which is worse than `main` and is precisely the scenario #997 describes.

The label now leads with the disambiguator, so it is always inside the visible width:

```
a1b2c3…7f01 (claude) · Refactor billing module
```

and both the legend and the tooltip carry `title=` so the full label is reachable on hover. I tried the alternative — keeping the title in front and shortening the id to a 6-character prefix — and reverted it: session ids are not all UUIDs, and slug-shaped ids like `session-titled-123456` / `session-absent-123457` collapse onto the same prefix, so the first-6-plus-last-4 form is doing real work.

**`injectDashboardBootstrap` was not safe by construction.** Extracting the helper fixed the `$`-substitution problem but left the security-critical `<` escaping at the call site 94 lines away, and the test called the helper with raw `JSON.stringify` output — so deleting that escape left every test green while the served page became injectable through any project, device or model name. Nothing in `tests/` asserted that escaping at all. It has moved inside the helper, with a test that pushes `</script>` through a payload value.

**`preferredSessionTitle` picked alphabetically, not most recently.** `src/types.ts:264` documents `title` as the *last* ai-title entry, so when one session id yields two summaries the legend could show the superseded one. It now takes the greatest `lastTimestamp`, keeping the alphabetical order only as a tie-break. Entries are ordered by key before the collision tiers run, so the same corpus cannot emit a different label set depending on input order.

## Privacy boundary

Worth stating explicitly, since this puts transcript-derived text into a payload field: it doesn't cross any boundary that project paths weren't already crossing. Both consumers strip session detail wholesale — `redactProjectNames` empties `sessionSeries` and every point's `sessions` array unless the caller explicitly opts in (`tests/mcp-redact.test.ts` pins this), and `sanitizeForSharing` does the same unconditionally for the LAN share path. I ran both suites plus `menubar-json` to confirm.

## Scope

Only item 1 from the issue. The secondary request — clickable legend entries that filter the whole view to one session, with the summary tiles recomputing — is a much larger change and the reporter says item 1 alone resolves the main problem, so I left it.

No CSS change to `UsageChart.tsx`. Long labels truncate at `max-w-40` exactly as long project labels always have, and the tooltip shows more; the difference is that titles diverge at the *start*, which is what makes them readable under truncation where identical project prefixes were not. Happy to widen the legend if you'd rather.

## Verification

- `npx vitest run tests/granular-history.test.ts` — 16 passed
- adjacent suites — `web-dashboard`, `mcp-redact`, `sharing/*`, `menubar-json` — 86 passed across 11 files
- `npx tsc --noEmit` — clean
- mutation-checked, each independently: reverting the title preference, disabling the collision disambiguation, and capping on UTF-16 units instead of code points each fail exactly their own test; restoring returns 15/15. Restoring the string replacement in the bootstrap fails the `$`-pattern test with the raw document prefix visible inside the serialized value.

New cases cover the title path, the no-title regression pin, two sessions sharing a title, ANSI/control-character sanitisation, and the length cap.


ozymandiashh:
Holding as a draft. A second adversarial pass found three things, one of which makes this a regression on the exact case the issue is about.

**The 80-code-point cap is about 2.5x what the legend actually renders.** `max-w-40` is 160px and the legend renders at `text-[10px]`, which fits roughly 32 characters. Everything past that is clipped — and everything past that is where ` · <shortId> (<provider>)` and all of the collision-tier suffixes live. So two sessions in one repo whose AI-generated titles share a 32-character prefix now render as the *same* legend entry, where on `main` they were distinguishable because the label led with the short project plus short session id.

That is the monorepo scenario from #997, made worse rather than better. Fixing by sizing the title cap to what the box shows so the id survives, and adding `title={item.label}` to the legend and tooltip spans so the full label is reachable on hover.

**`injectDashboardBootstrap` is not safe by construction.** I extracted the helper to fix the `$`-pattern substitution, but the security-critical `.replace(/</g, '<')` stayed at the call site 94 lines away. The new test calls the helper with raw `JSON.stringify` output and asserts only the `$`-half, so deleting the escape at `web-dashboard.ts:188` leaves every test green while the served page becomes injectable through any project, device or model name. Nothing in `tests/` exercises `serveIndexHtml` or asserts that escaping at all — grepping for `u003c` across the suite returns nothing. Moving the escape inside the helper and adding a test that pushes `</script>` through a project name.

**`preferredSessionTitle` picks alphabetically, not most recently.** `src/types.ts:264` documents `title` as the *last* ai-title entry, so when one session id yields two summaries the legend can show the superseded one. Deterministic but wrong; switching to the greatest `lastTimestamp` with the alphabetical sort kept only as a tie-break.

Issue #998, opened 2026-08-14
Grok provider undercounts tokens by ~98% on cache-read: parser reconstructs from a heuristic instead of reading the authoritative turn_completed.usage event


saulcanina:
## Summary

The Grok provider in codeburn does not read the `turn_completed` event's `usage` field that Grok CLI itself writes to the session file — the authoritative token count (`inputTokens`, `outputTokens`, `cachedReadTokens`, `cacheCreationTokens`). Instead it reconstructs an estimate from `params._meta.totalTokens`, a running counter present on other event types (mostly `tool_call`), using a threshold heuristic to detect new "segments". On a real, cache-heavy session this reconstruction captures roughly **1.4% of the real cache-read volume**, and about **6% of total real tokens** for the day.

This is not a regression — the same logic (and the same order-of-magnitude gap) is present in already-consolidated data from two prior days, just at a scale where it went unnoticed. It became visible today because of an unusually large, cache-heavy Grok run.

## Environment

- codeburn `0.9.20` (npm `latest` as of 2026-08-10; installed via Homebrew, same bundle)
- macOS (Darwin), arm64
- Provider: **grok** (Grok CLI 1.0.3, model `grok-4.6`)
- Session store: default `~/.grok/sessions` (no `GROK_HOME` override)
- Session volume: 42 sessions with `summary.json` dated the same day, sizes 442 B–2.98 MB (well under the 128 MB `readSessionFile` cap). 35 of 42 produce non-zero reconstructed tokens under the current heuristic; 7 are legitimately empty/no-op sessions (~499 B, correctly discarded by the `input === 0 && output === 0` early-return).

## Expected vs. actual

**Expected**: token/cost totals for a Grok session should reflect what Grok CLI itself reports as consumed for that session (the `turn_completed.usage` object), the same way other providers use their tool's own authoritative usage report where one exists.

**Actual**: codeburn ignores `turn_completed.usage` entirely for Grok and instead derives input/output/cache-read by watching `_meta.totalTokens` climb across unrelated event types and guessing turn boundaries from a `total < prevTotal * 0.5` reset heuristic.

## Evidence

Single-session example (one of several parallel subagent sessions from the same run), comparing the real `turn_completed.usage` event in that session's `updates.jsonl` against what codeburn's parser reconstructs from the same file:

| | Real (`turn_completed.usage`) | codeburn (reconstructed) | Captured |
|---|---:|---:|---:|
[...]

PR #1015, opened 2026-08-18
fix(grok): read the CLI's own completed-turn usage instead of estimating it

ozymandiashh:
Reported in #998 with an unusually complete write-up, including a verbatim redacted `tool_call` → `turn_completed` pair from a real session.

## The problem

Grok CLI writes a `turn_completed` update at the end of each turn, carrying its own usage object — `inputTokens`, `outputTokens`, `cachedReadTokens`, `cacheCreationTokens`, `reasoningTokens` — into the same `updates.jsonl` the provider already reads line by line. We never branched on it.

Instead `parseUpdates` reconstructed an estimate from `params._meta.totalTokens`, a running context-size counter riding on unrelated events, using `total < prevTotal * 0.5` to guess turn boundaries. On the reporter's cache-heavy session that captured roughly **1.4% of the real cache-read volume** and about **6% of the day's tokens**, while over-counting output about fivefold. `cacheCreationInputTokens` and `reasoningTokens` were hardcoded to `0`.

## The fix

Branch on `update.sessionUpdate === 'turn_completed'` and read `update.usage`, keyed by the record's `prompt_id` (snake_case there, camelCase `_meta.promptId` on the streaming events) with last-write-wins so a re-emitted record cannot double count.

Two decompositions are load-bearing, both derivable from the reported numbers rather than assumed:

- `totalTokens === inputTokens + outputTokens` exactly (12,851,663 + 36,633 = 12,888,296), and `cachedReadTokens` is smaller than `inputTokens` without appearing separately in the total ⇒ **cache read and cache creation are subsets of input**. Fresh input is `input − cacheRead − cacheCreation`, clamped **per record** so one inconsistent record cannot eat another's fresh input.
- `reasoningTokens` (29,077) is smaller than `outputTokens` (36,633) and likewise absent from the total ⇒ **reasoning is a subset of output**.

That second one is where this gets interesting, because Grok's convention is the inverse of the repo's.

## The repo contract runs the other way

`ParsedProviderCall.reasoningTokens` is *exclusive* of `outputTokens` everywhere — codex, hermes, qwen, kiro, cursor-agent and lingtai-tui all price `output + reasoning`, and `tests/providers/kiro.test.ts:1247` states it outright: "downstream aggregation sums outputTokens + reasoningTokens, so reasoning must NOT be folded into outputTokens here."

So the split happens in the provider: reasoning is clamped to the reported output, `outputTokens` is emitted without it, and the downstream sum reconstructs exactly what Grok reported. `src/parser.ts` is untouched.

The clamp is not decoration. Every usage field is validated independently and nothing cross-checked reasoning against output, so a record with `reasoning > output` produced a negative `outputTokens` and left the pipeline pricing reasoning instead of output — the exact double-bill this change exists to remove.

## What is deliberately not in scope

Multi-model attribution. `usage.modelUsage` is a per-model breakdown, and pricing each bucket at its own rate is tempting — I tried it and reverted it. `chooseAuthoritativeModel` exists to avoid a truthful-but-$0 row when `modelUsage` names an id this checkout cannot price (`grok-4.6-build`, which is the id in the reporter's real record), and per-model pricing loses that fallback: the reporter's own session went from $1.20 to near zero the moment a second, priced id appeared alongside it. It also discarded the verified top-level totals in favour of a `modelUsage` sum nothing had checked against them.

So: one call per session from the top-level totals, with `modelUsage` used only to pick a priced attribution id. A session that used two models is priced at one rate — a bounded, documented inaccuracy, and better than silently pricing most of a session at $0.

## Sessions the record does not cover

Older Grok CLI versions do not emit `turn_completed`, so the existing heuristic remains the fallback, unchanged and still pinned by its original test.

The decision is taken from the **deduplicated** records rather than latched per line, so a superseded or all-zero record cannot flip a session off the heuristic and then drop it at the totals gate. And a session only partly covered — a turn written by an older build, or a run that died before writing its record — keeps `costIsEstimated: true` rather than presenting a partially covered session as fully provider-measured.

`usage.costUsdTicks` is deliberately not read. The scale of a "tick" is not documented anywhere I can verify, and guessing it would fabricate spend.

## Cache versions

The grok parse version moves, and `DAILY_CACHE_VERSION` moves with `MIN_SUPPORTED_VERSION`. The session-cache bump alone would leave existing users with the old accounting for all history, because the daily cache serves every day before today and retains ten years. Moving the two in lockstep is what keeps it lossless — the filename is version-suffixed, so the old file stays on disk and is adopted for days no source can still re-derive. There is a test that seeds a v17 cache with no sources present and asserts the days survive.

## One fix beyond Grok

`detectContextBloat` divided by `session.totalOutputTokens` alone. Since reasoning is stored *beside* output for every reasoning-bearing provider, the detector was seeing a fraction of the generated tokens and inventing findings: a session whose provider-reported ratio is 20:1 — below the 25:1 threshold — was reported as 133:1, "high impact", with 710K tokens of claimed savings. It now uses the same `output + reasoning` sum the reports use, which fixes codex, kiro, hermes, qwen and cursor-agent as well.

## Verification

- 119 passed across `tests/providers/grok.test.ts`, `tests/grok-parser-pipeline.test.ts`, `tests/daily-cache-grok-rederivation.test.ts` and `tests/optimize.test.ts`
- `npx tsc --noEmit` clean

The pipeline tests are the ones that matter. Two earlier rounds of this change passed a full provider-level suite while genuinely broken, because those tests assert `call.costUSD` — a value the pipeline discards for providers that are not in the `providerCallToCachedCall` allowlist. Everything load-bearing is now asserted through `parseAllSessions` on both a cold and a warm parse.

Mutation-checked, each independently:

| Mutation | Result |
|---|---|
| drop the `reasoning ≤ output` clamp | 1 failed — `expected -150 to be +0` |
| skip the cache-exclusive decomposition | 5 failed |
| first-write-wins on a duplicate `prompt_id` | 1 failed |
| revert the `detectContextBloat` denominator | 1 failed |
| restored | 119 passed |

## What I could not verify

**I have no Grok CLI install and no real session data on this machine.** Everything is built against the event shape in #998 and fixtures derived from it.

@saulcanina — you offered to sanity-check a candidate fix against your own data, and that is exactly the gap. Three things I would most like confirmed: that the numbers this produces match the `turn_completed.usage` values you quoted; that sessions predating the record still parse through the fallback unchanged; and whether any session emits two `turn_completed` records for one `prompt_id` with genuinely different usage, since the dedup assumes last-wins.


ozymandiashh:
Holding this as a draft: an adversarial pass on my own branch found that the reasoning-not-double-billed claim doesn't survive the pipeline. `grok` isn't in the costUSD persistence allowlist at `src/parser.ts:2490`, so the cost this provider computes is discarded and `cachedCallToApiCall` (`:2624-2631`) re-prices non-Claude providers with `outputTokens + reasoningTokens`. That was harmless while grok hardcoded `reasoningTokens: 0`; populating it from the authoritative record makes every Grok cost inflate by exactly the reasoning tokens.

Also reworking: a multi-model session is currently priced entirely at whichever `modelUsage` key comes first, an all-zero `turn_completed` record can flip a session onto the authoritative path and drop it, and the cache-exclusive clamp runs on the session sum rather than per record. Will push a corrected version with pipeline-level tests rather than provider-level ones — the latter is exactly why the first three got through.

saulcanina:
Reporter here. **I have not run this branch**, so I cannot yet do the check you asked for first: comparing what it produces against the values I quoted. What I could do is test the assumptions the patch rests on against my whole corpus rather than the single session in #998, and I can address your other two questions — one with data, one only by showing why my corpus cannot answer it.

Everything below describes the JSON on disk. Where a statement would be about your code's behaviour, I say so instead of guessing.

**Corpus**: 271 session files, 398 `turn_completed` records, spanning 20 Jul to 17 Aug 2026. 397 carry a `usage` object. The install here reports `grok 1.0.4`, but that does not tell you which version wrote the older records, and the schema note below suggests it was not all one version.

## Your two questions

> whether any session emits two `turn_completed` records for one `prompt_id` with genuinely different usage, since the dedup assumes last-wins

**Within a single session file: zero repeated `prompt_id`, in any of the 271.** Last-write-wins is never invoked there, so on this corpus it cannot be wrong.

Across files it does repeat, which may be outside the scope you meant but affects the same code. Two `prompt_id` values appear in more than one session file, 5 records in total. The three files involved hold 1, 2 and 3 records, and each file's sequence is an exact prefix of the next: same `prompt_id`, same `stop_reason`, and `usage` payloads that are byte-identical as raw JSON substrings. The session ids differ and the copied events carry fresh timestamps, so it is not a file copy. All five were written within about half an hour of each other on one day.

I can't tell you from the files *why* it happened — resume, fork, replay and import are all compatible, and nothing in the data names the mechanism. The consequence is what matters: if the dedup key is scoped per session and session totals are then summed, those turns are counted twice and three times, 1,954,010 tokens here. The PR says a re-emitted record "cannot double count", which holds if the key is global, so it may be worth stating the scope explicitly. For these two ids a global last-write-wins is safe, since the payloads are identical.

> that sessions predating the record still parse through the fallback unchanged

**I cannot test this: my corpus has no session that carries token data without a `turn_completed` record.** 87 of the 271 have no such record, but each is a single-line file holding one `hook_execution` / `session_start` update and nothing else. I checked all 87 exhaustively, walking every JSON object rather than sampling: `_meta.totalTokens` does not appear anywhere in them.

So those files do reach the fallback, they just carry nothing for it to work with. What is missing from my corpus is a token-bearing legacy session, and that is the case your fallback test would need. It would have to come from an older CLI than any that wrote these files.

## The decompositions hold

Zero violations across the 397 records with usage: `totalTokens === inputTokens + outputTokens`; `cachedReadTokens + (cacheCreationTokens ?? 0) <= inputTokens`; `reasoningTokens <= outputTokens`. The `?? 0` is load-bearing there, for the reason in the next section.

## Three schema facts I did not expect

**1. `cacheCreationTokens` is not always present.** The key appears in 204 records and is absent from 193. Where present it is **always 0** — not once non-zero in a month. Every absent case falls on or before 27 Jul; every record from 1 Aug onward has the key. There is a gap in my data between those dates, so I can only say the key appeared somewhere in it. Both shapes emit `turn_completed`. In this corpus `cacheCreationTokens ?? 0` is therefore always zero, and the cache-exclusive subtraction reduces to `input − cacheRead`.

**2. 13 of the 397 usage objects set `usageIsIncomplete: true`,** and they are exactly the 13 that omit `costUsdTicks`. The two sets are identical, which suggests the flag may be about the missing cost figure rather than about the token counters — every token field in those records is present, finite, and satisfies all three invariants above. I can't settle that from the files, and since this patch deliberately ignores cost it may not matter to you at all. Two things in case it does: excluding those records would drop 17,975,442 tokens, 3.0% of my corpus, and no later record repeats their `prompt_id` to replace them; and `stop_reason` is not a usable substitute for the flag, though the two are correlated (the flag appears on 5 of 24 `cancelled` records and on 8 of 373 `end_turn` ones).

**3. One record has no `usage` key at all.** Its entire update is `{"sessionUpdate": "turn_completed", "prompt_id": ..., "stop_reason": "end_turn"}`. There is no all-zero usage object anywhere in my corpus; absent is the degraded shape that actually occurs. It sits in a file with 11 valid records.

## `costUsdTicks` decomposes exactly on 78% of the records, and I think that changes the calculus

You wrote that the scale of a tick isn't documented anywhere you can verify and that guessing would fabricate spend. That's fair, but the field turns out to be much better behaved than a guess. Testing candidate coefficient vectors against the 384 records that carry it, with integer arithmetic and no fitting:

- `grok-4.5-build`: **188 of 257** records satisfy `ticks = 20000·fresh + 3000·cacheRead + 60000·output` **exactly**. A further 21 satisfy exactly twice that expression. All remaining 48 fall strictly between the two values.
- `grok-4.6-build`, earlier records: **28 of 76** exact on `20000 / 5000 / 60000`, 12 exact at double, 31 strictly between, 5 below.
- `grok-4.6-build` on the most recent day: **51 of 51 exact**, on `3400 / 850 / 10200` — precisely 0.17× the same vector.

**300 of 384 records, 78%, land exactly on one of those vectors**, and 379 of 384 are either exact or bracketed by the base/double pair. A pooled regression across the whole set mixes those regimes and returns a nonsense negative output coefficient, which is why the structure only shows up once the records are split by model and by period.

I'm not claiming these are official prices — my data documents neither the scale nor the reason for the doubled regime. But two things follow that seem relevant to the patch. First, a single scale factor of 1e-10 USD per tick turns the base vector into $2.00, $0.30 and $6.00 per million tokens, which are round numbers rather than the artefacts a wrong scale produces. Second, and more usefully: the rate visibly **changed** partway through my corpus, by a clean factor. A locally maintained price table silently goes stale when that happens, whereas the provider's own field tracks it. That seems worth more than the risk you were guarding against.

---

If it would help, I can run the branch against this corpus. It would have to be an isolated clone pointed at its own cache directory: the PR says it raises the parse version, and I'd rather not have the branch rewrite the cache the installed copy reads. If you want that, tell me the entry point that parses a session and what output you want compared, and I'll post the token figures side by side against the values in #998. One constraint on my end: these sessions are client work under confidentiality, so I can post derived counts like the ones above, but not session files, project paths, or anything from their contents.

If it's useful I can also put together a fixture from these logs: 271 session files carrying the 398 `turn_completed` updates with their `usage` objects, ids replaced by sequential ones so the cross-file repeats stay visible, and timestamps rebased so that only the intervals within a session survive. It is numbers plus protocol constants — every value in the usage payloads is an integer or a boolean, and the only strings anywhere are constants the CLI itself writes (the update types, the stop reasons, the RPC method name, the two model names) plus the ids I would be replacing — so it can sit in the repo as test data rather than in a comment. Every number in it is real and unmodified; only the ids and the timing are synthetic, because these are client sessions. One limit before you lean on it: all 397 usage objects carry exactly one `modelUsage` key, and that per-model block duplicates the top-level fields to the digit, so the multi-model pricing path you mentioned reworking cannot be exercised against this data.

Issue #968, opened 2026-08-11
Unpriced warning fires on subscription-billed models where $0 is correct, and its suggested model-alias fix would fabricate spend


therickfactr:
## Problem

The unpriced-models warning from #638 fires on models whose cost genuinely **is** zero, and its suggested fix would make the totals wrong if followed.

#638 anticipated exactly this ambiguity — *"a new model id shipping ahead of our pricing table … undercounts real spend and looks identical to a genuinely free model."* In practice the genuinely-free case is common: models billed under a flat-rate subscription have no per-token cost to compute.

## Evidence

On my machine the warning reports these at $0:

| model | tokens | calls | why it is $0 |
|---|---|---|---|
| `auto-genius` | 35.3M | 898 | ClinePass codename — opaque routed model, flat-rate pass |
| `warp` | 17.7M | 449 | Warp's built-in agent, subscription |
| `codex-auto-review` | 7.2M | 940 | runs under the ChatGPT subscription |
| `grok-composer-2.5-fast` | 1.9M | 10 | Grok subscription |
| `big-pickle` | 33.9K | 4 | ClinePass codename |

None of these has a per-token price to discover. The marginal cost of the call is zero because the spend already happened as a subscription fee.

## Why the suggested fix is harmful here

The warning advises:

```
Fix: codeburn model-alias "<model>" <known-model>
```

Applied to any of the above, that maps a flat-rate model onto another model's per-token rate and **invents spend that was never billed**. `auto-genius` alone is 35.3M tokens — aliasing it to a mid-tier model would fabricate hundreds of dollars, in the one number the tool exists to report accurately. For ClinePass specifically the underlying model is deliberately opaque and can rotate, so there is no correct alias to choose even in principle.

So today the user has two options and both are wrong: leave it and get warned indefinitely about correct data, or alias it and corrupt the totals.

## Suggested direction

A third state alongside "priced" and "unpriced": **flat-rate / subscription-billed**, meaning $0 is the right answer and no warning is due.

There is already precedent for this shape in the codebase — local models are special-cased out of the warning via `looksLikeLocalModel()` / `getLocalSavingsBaseline()`, and surfaced separately as a "Savings (local models)" line. Subscription-billed models seem to want the same treatment: excluded from the unpriced list, and ideally attributed to their plan rather than shown as free.

[...]

PR #1016, opened 2026-08-18
fix(models): let --unpriced survive --top, and document the flag

ozymandiashh:
Follow-up to #985, which added `codeburn models --unpriced` for #969. Two things I said I'd take rather than leave in a first-time contributor's queue.

## `--unpriced --top N` returned nothing

`--top` is applied inside `aggregateModels` (`src/models-report.ts:234-236`), after the sort by `costUSD + savingsUSD` descending and before the unpriced filter in `src/main.ts` ever runs. Unpriced rows are $0 cost by definition, and `findUnpricedModels` excludes anything carrying a local-savings baseline, so they're $0 savings too — which puts them last in that sort, every time. Any `--top N` smaller than the number of priced models therefore sliced away exactly the rows the flag exists to surface.

In table form the user got:

```
$ codeburn models --unpriced --top 2
No model usage found for the selected period.
```

Which is wrong twice: they do have unpriced models, and nothing hints that the two flags fought. Both are documented `models` flags, so it's a combination people land on.

The slice now runs after the filter when `--unpriced` is set. `aggregateModels` is handed `topN: undefined` in that mode and the rows are sliced afterwards.

## Docs

The flag is now in the README **Models** table alongside the others, with a changelog entry under Added crediting @kocaemre, plus one under Fixed for the `--top` interaction.

## Second pass: the fix was half-right

An adversarial pass over this branch found that filtering before the slice is necessary but not sufficient.

Every unpriced row is $0 on **both** cost and savings, so they all tie under `aggregateModels`' sort key, and `Array#sort` is stable. The surviving order was Map insertion order — the order each model's first assistant call appears in the transcript. So `--unpriced --top N` returned the N that happened to appear earliest, and a model holding almost all of the unpriced volume was dropped if it showed up late:

```
fixture, emitted small-first: zz-unpriced-small (1.1k), zz-unpriced-largest (9.1k), zz-unpriced-middle (5.1k)

before: --unpriced --top 2  ->  ['zz-unpriced-small', 'zz-unpriced-largest']
after:  --unpriced --top 2  ->  ['zz-unpriced-largest', 'zz-unpriced-middle']
```

`findUnpricedModels` already sorts by tokens descending, then calls, then model name — and that is exactly the order the dashboard warning renders. It is now called once over the whole row set, its output becomes a rank index, and rows are ranked before the slice. That also makes the README row honest: the CLI and the warning now agree on which N, which they did not before. In `--by-task` / `--by-agent` several rows share one model, so they share that model's rank and N still counts rows.

My original test could not have caught this. Its fixture held one unpriced model, so `--top 2` never truncated anything — deleting the `slice` line entirely left the suite green.

## Verification

- `npx vitest run tests/models-report.test.ts` — 32 passed
- `npx tsc --noEmit` — clean
- new test `keeps unpriced rows when --unpriced is combined with --top` uses a fixture where two priced models (Opus, Sonnet) both outrank the unpriced row, so `--top 2` would consume both slots
- mutation-checked three ways: restoring `topN` unconditionally fails with `expected [] to deeply equal [...]`; removing the rank sort fails with `expected [ 'zz-unpriced-small', …(1) ] to deeply equal [ 'zz-unpriced-largest', …(1) ]`; deleting the slice fails with `to have a length of 2 but got 3`. Each returns 32/32 when reverted.

## Still open on #969

The issue's original complaint was that the *TUI warning* is truncated at every terminal width, hiding both the model list and the fix command. This makes the escape hatch discoverable and correct, but it doesn't touch the warning itself. #968 is also still open and relevant: the warning fires on subscription-billed models where $0 is genuinely correct, and `--unpriced` inherits that same false positive since it reuses `findUnpricedModels` — deliberately, so the two can't drift, but it does mean fixing #968 fixes both at once.


ozymandiashh:
Holding as a draft: an adversarial pass found the fix is half-right. Filtering before slicing is necessary but not sufficient — every unpriced row is $0 on both cost and savings, so they are all *tied* under `aggregateModels`' sort key, and `Array#sort` is stable. The surviving order is Map insertion order, i.e. the order each model's first assistant call appears in the transcript. So `--unpriced --top N` returns the N that happen to appear earliest, and a model holding almost all the unpriced volume gets dropped if it shows up late.

My test missed it because the fixture has one unpriced model, so `--top 2` never truncates anything — deleting the slice line entirely leaves the suite green.

Fixing by ranking on the order `findUnpricedModels` already produces (tokens descending) before slicing, which also makes the CLI and the dashboard warning agree on which N — the README sentence about this being the copyable form of the warning is not true as written until that lands.

PR #1042, opened 2026-08-19
fix(models): alias bare MiMo 2.5 ids to the LiteLLM Xiaomi rows

avs-io:
## Problem

Hermes / Xiaomi token-plan sessions store `mimo-v2.5-pro`. CodeBurn showed `$0` even though LiteLLM already has `xiaomi/mimo-v2.5-pro`.

## Root cause

Pricing looks up the session id as written. The snapshot row is namespaced. `mimo-v2-flash` already had this alias; `mimo-v2.5-pro` did not. A naive alias then made `getShortModelName('cline-pass/mimo-v2.5-pro')` recurse forever (strip → alias back to `xiaomi/…` → last segment → strip).

## Change

- Alias `mimo-v2.5-pro` / `mimo-v2.5` to the existing `xiaomi/…` snapshot rows.
- Resolve display names on the stripped leaf before following a pricing alias.
- `kimi-k3` is unchanged: there is still no invented rate.

## User impact

MiMo 2.5 Pro sessions get the snapshot list price instead of `$0`. Name stays "MiMo v2.5 Pro".

## Preservation

- No new LiteLLM numbers.
- No Kimi rate invented.
- Other aliases unchanged.

## Testing

- `npx vitest run tests/models.test.ts`: 156 passed, including `mimo-v2.5-pro` / `MiMo-v2.5-Pro` through `xiaomi/mimo-v2.5-pro` and `cline-pass/mimo-v2.5-pro` short name.

### DIFF
--- a/src/models.ts
+++ b/src/models.ts
@@ -297,6 +297,10 @@ const BUILTIN_ALIASES: Record<string, string> = {
   'k3-agent':                      'kimi-k3',
   'k2d6-agent':                    'kimi-k2p6',
   'mimo-v2-flash':                 'xiaomi/mimo-v2-flash',
+  // Hermes / Xiaomi token-plan sessions store the bare id. LiteLLM's row is
+  // namespaced. Same class as mimo-v2-flash above — do not invent a rate.
+  'mimo-v2.5-pro':                 'xiaomi/mimo-v2.5-pro',
+  'mimo-v2.5':                     'xiaomi/mimo-v2.5',
   'kat-coder-pro-v1':              'kwaipilot/kat-coder-pro',
   // Cursor emits dot-version tier-last names plus tier/reasoning suffixes
   // that LiteLLM does not index (`-high`, `-low`, `-medium`, `-thinking`,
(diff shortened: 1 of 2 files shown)


avs-io:
Closing: this was a one-id alias on a class that already exists (mimo-v2-flash → xiaomi/…). Not enough product surface to keep a PR open. Branch kept if we want the cycle-safe getShortModelName later.

iamtoruk:
Reviewed in depth; it also fixes a live `RangeError` stack overflow on `main` for `mimo-v2-flash` display (pre-existing alias cycle). Maintainer edits are disabled on this PR so the review commits couldn't be pushed here — landing via #1048 with your three commits carried verbatim plus the crash-id regression tests, the two display names, and a CHANGELOG entry. Style note (not a defect): the rewrite is ~3× larger than needed — `lookupShortName` + a 3-line termination guard in the path-segment fallback passes all the PR's tests without the `seen` set / `userAliases` block; happy to share the diff if you want to slim it in a follow-up.

PR #1043, opened 2026-08-19
fix(models): cite unpriced session ids or leave them $0

avs-io:
## Problem

Session ids `kimi-k3` and `mimo-v2.5-pro` (and the Cline Pass / Hermes spellings) had no LiteLLM row under that exact string. `kimi-k3` only priced via last-resort gap-fill. Bare `mimo-v2.5-pro` stayed $0 even though LiteLLM already has `xiaomi/mimo-v2.5-pro`. Aliasing the bare id to the vendor-prefixed row made `getShortModelName` recurse forever (`mimo-v2.5-pro` → `xiaomi/mimo-v2.5-pro` → last segment → same alias).

## Root cause

Unpriced-without-a-row was treated as a missing screenshot string. The class is: a session id with no table row stays $0 unless there is an explicit cited alias or override. Display names are keyed by the bare id, so a bare→`vendor/model` alias must not re-enter path fallback.

## Change

- Pin `kimi-k3` to Moonshot official list price ($3 / $15 / $0.30 per 1M). Cite: https://platform.kimi.ai/docs/pricing/chat-k3. Do not inherit K2.
- Explicit aliases `mimo-v2.5-pro` → `xiaomi/mimo-v2.5-pro` and `mimo-v2.5` → `xiaomi/mimo-v2.5` (existing LiteLLM rows). Same pattern as `mimo-v2-flash`.
- `getShortModelName` looks up the pre-alias session id and refuses to recurse when the last path segment aliases back to `vendor/model`.

## User impact

Kimi K3 and MiMo v2.5 / v2.5 Pro session ids that previously showed $0 now price. Unknown next-gen ids (`kimi-k4`, `mimo-v2.6-pro`) stay unpriced.

## Preservation

- No invented family fallback (`glm-5.x`, `mimo-v2.5-*`).
- LiteLLM's `xiaomi/mimo-v2.5-pro` rates are used as-is; this PR does not invent a second Xiaomi official override.
- Fireworks `accounts/fireworks/models/<slug>` naming still peels the last segment.

## Testing

- `npx vitest run tests/models.test.ts tests/models-hoist.test.ts tests/menubar-json.test.ts tests/providers/kimicode.test.ts tests/model-breakdown.test.ts`: 204 passed.
- Sabotage: removing the `mimo-v2.5-pro` alias fails the new test (bare id returns null).

### DIFF
--- a/src/models.ts
+++ b/src/models.ts
@@ -51,6 +51,11 @@ const BUILTIN_PRICE_OVERRIDES: Record<string, SnapshotEntry> = {
   'composer-2': [0.5e-6, 2.5e-6, 0.5e-6, 0.2e-6],
   'composer-1.5': [3.5e-6, 17.5e-6, 3.5e-6, 0.35e-6],
   'composer-1': [1.25e-6, 10e-6, 1.25e-6, 0.125e-6],
+  // Moonshot official K3 list price. LiteLLM has no kimi-k3 row; do not
+  // inherit K2. https://platform.kimi.ai/docs/pricing/chat-k3
+  // $3.00 input / $15.00 output / $0.30 cache-read per 1M. No published
+  // cache-write (buildCosts uses the 1.25x input heuristic).
+  'kimi-k3': [3e-6, 15e-6, null, 0.3e-6],
 }
 
 // Assemble a ModelCosts, applying the cache-cost heuristics (write = 1.25x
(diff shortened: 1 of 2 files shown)


avs-io:
Overlaps #1042 (same mimo alias + short-name cycle). Keeping #1042.

Issue #868, opened 2026-08-01
Menubar icon never renders on macOS 26.5.1 (25F80) with 0.9.19 — #146 recurrence, no RenderBox fault


twodxme:
### Summary

Menubar icon never renders on macOS 26.5.1 (Build 25F80) with CodeBurnMenubar 0.9.19. The app launches and keeps running, the CLI contract is healthy, the status-item scene is created — but nothing ever draws. This looks like a recurrence of #146 on a newer macOS build, with one notable difference from the related CodexBar report (steipete/CodexBar#805): **no RenderBox precondition failure appears in the log**, so the #147 fix path ("start as .regular, activate, simple image first") may not cover whatever changed in 26.5.x.

### Environment

- CodeBurnMenubar: **0.9.19** (fresh `codeburn menubar --force` install; version verified via PlistBuddy)
- macOS: **26.5.1** / Build **25F80**
- Hardware: MacBook Pro (M3 Max)
- Install location: `/Users/tim/Applications/CodeBurnMenubar.app`
- No menubar manager (no Bartender/Ice/HiddenBar); ample free menubar space (notch ruled out)

### Actual behavior

- Process launches and stays alive; no crash.
- No menu bar icon, ever.
- Light/Dark appearance toggle has no effect (rules out the appearance-specific render failure seen in CodexBar#805).

### CLI contract is healthy

```
$ time codeburn status --format menubar-json | head -c 400
{"generated":"2026-08-01T05:03:14.207Z","current":{"label":"Today (2026-07-31)","cost":0,...}
codeburn status --format menubar-json  0.56s user 0.18s system 123% cpu 0.606 total
```

### Diagnostics (unified log)

Status-item scene **is created**:

```
CodeBurnMenubar: (AppKit) [com.apple.AppKit:StatusBar] 0 create scene 5DAF1172-A4D9-4468-A834-365D847350B3
CodeBurnMenubar: (BoardServices) [com.apple.BoardServices:XPCErrors] [C:1] Alloc com.apple.controlcenter.statusitems
```

Control Center scene requests complete, but the NSStatusItemView aux scene loops on reconnect:

```
[com.apple.FrontBoard:Common] Requesting scene <FBSScene: ...; com.apple.controlcenter:3B7E320E-...-Aux[1]-NSStatusItemView> from com.apple.controlcenter.statusitems
[...]

PR #1049, opened 2026-08-19
fix(optimize): scope remediation copy to --provider

### LINKED ISSUE #1044: optimize --provider codex still emits Claude-specific remediation wording after #1002
### Problem

Running `optimize` with the Codex provider still produces Claude-specific remediation text.

This is a follow-up to #1002 rather than the same issue. Provider filtering appears to be working correctly on current `main`; the remaining problem is that cross-provider findings still emit Claude-specific remediation wording.

For example:

```text
-- Ask Claude in the current session --
Ask Claude to audit the retry-heavy capability before changing config:
```

and:

```text
-- One-time session opener (do NOT add to CLAUDE.md) --
Paste at the start of your NEXT expensive thread (one-time, do not add to CLAUDE.md):
```

This appears when the report is explicitly scoped to Codex.

### Reproduction

Environment:

- Windows / PowerShell
- Node v22.14.0
- npm 10.9.2
- CodeBurn current `main`

Command:

```powershell
npx tsx src/cli.ts optimize --provider codex -p lifetime
```

The report is correctly restricted to Codex sessions, but the coaching instructions still refer to Claude and `CLAUDE.md`.

### What I found

I did a quick source trace before reporting this.

`runOptimize()` passes `opts.provider` into `scanAndDetect()`, and `scanAndDetect()` correctly uses the provider to suppress Claude-only detectors.

However, the provider is not passed through to the text rendering path (`renderOptimize()` / `renderFinding()` / `renderActionHeader()`).

There also appear to be some cross-provider findings that contain Claude-specific wording directly in their `fix.label`, including the retry-heavy capability finding and the session-level findings.

Because `buildOptimizeJsonReport()` exposes `f.fix` directly, this appears to affect structured output as well as the terminal rendering.

This seems related to the remediation destinations introduced in #277 / #281, which were originally phrased around Claude.

### Possible direction

I think either of these could work:

1. Make remediation wording for cross-provider findings provider-neutral.
2. Pass provider context through the remediation/rendering layer and generate provider-specific wording.

I'm happy to implement the fix if this is something you'd like changed. I wanted to check which approach fits the project before writing code.

avs-io:
## Problem

`optimize --provider codex` still told the user to ask Claude and to keep one-time openers out of `CLAUDE.md`. Provider filtering from #1002 was already correct; the remaining class was Claude-hardcoded remediation copy on findings that still run for every provider, plus the CLI/TUI destination headers and the JSON `fix` payload.

Closes #1044.

## Root cause

`scanAndDetect` skips Claude-only detectors when `--provider` is not Claude, but the cross-provider detectors (`retry-heavy-capabilities`, `low-worth-sessions`, `context-heavy-sessions`, `cost-outliers`, `mcp-project-scope`) baked `Ask Claude` / `CLAUDE.md` into `fix.label`. `renderActionHeader` and the TUI `actionDestinationHeader` always said Claude regardless of provider. `buildOptimizeJsonReport` exposes `f.fix` as-is.

## Change

One `optimizeRemediationCopy(provider)` table:

- unset / `all` / `claude` → Claude + `CLAUDE.md` (shipped copy)
- `codex` → Codex + `AGENTS.md` (the only other instruction file CodeBurn already names)
- every other provider → display name + generic `project instructions` (no invented filename)

Cross-provider detectors, CLI headers, TUI headers, and therefore JSON all use that table. Claude-only detectors are unchanged.

## User impact

`codeburn optimize --provider codex` now says Ask Codex / do not add to AGENTS.md. Claude-scoped and unscoped runs are unchanged.

## Preservation

`PasteDestination` values (`claude-md`, `session-opener`, `prompt`) are unchanged. Apply plans still target CLAUDE.md only for Claude-only findings. No new LiteLLM rates. No Keychain / Buzz / #868 / #916 / #998 work.

## Testing

- 5 new tests in `tests/optimize.test.ts` for the copy table, detector labels + JSON, and render headers
- Sabotage: forcing Claude-only copy fails those 4 non-default cases
- `npx vitest run tests/optimize.test.ts tests/optimize-fs.test.ts tests/mcp-coverage.test.ts tests/dashboard.test.ts` — 262 passed


NeoLorenzo:
Tested on Windows / PowerShell against the real Codex history that originally reproduced #1044.

optimize --provider codex -p lifetime now correctly renders Ask Codex and uses AGENTS.md for the session-opener guidance.

I also ran the targeted test suites. The Optimize/MCP suites pass; dashboard.test.ts has 3 Windows-specific shortProject failures. I checked the exact base commit (c9ad778) and reproduced the same 3 failures there, so they are pre-existing and unrelated to this PR.

Thanks for picking this up.

iamtoruk:
Reviewed in depth with real-data runs. The scoping mechanics are right — the two CLAUDE.md-destination findings are `claudeOnly` so `--apply` can never offer a Claude write under another provider, unscoped output is byte-identical, AGENTS.md is a real Codex file, and the display-name drift test is a genuine guard. Two changes requested:

1. **Blocker — the new empty-state copy claims a scan that didn't run.** Under `--provider codex` the output now reads "CodeBurn optimize scans your **Codex** sessions and config for token waste: junk directory reads, duplicate file reads, unused agents/skills/MCP servers, bloated **AGENTS.md**…" with Health A (100/100) — but `scanSessions` returns empty for non-Claude providers and every listed detector is `claudeOnly`-disabled. Nothing was scanned. Main's wording was confusing here; this makes it false. For non-Claude providers the empty state should say the session-scan detectors don't cover that provider yet.
2. **Cut the dead TUI plumbing:** `dashboard.tsx:1514` gates the optimize view to `all|claude`, so the non-Claude branch of `optimizeTuiPasteHeader` and the 5 threaded `provider` props can never execute (~33 lines). Revert `dashboard.tsx` until the TUI actually opens to other providers.

Adjacent note (out of scope but worth a follow-up): `mcp-low-coverage`/`mcp-project-scope` are not `claudeOnly` and still carry Claude-config `apply` destinations + claude.ai connector copy under `--provider codex` — the PR scopes copy, not destinations, so "codex users aren't told to do Claude things" isn't fully delivered yet.

Issue #638, opened 2026-07-09
Unpriced models silently contribute $0 to totals; warning is off by default


avs-io:
## Problem

When a model has no pricing data, `calculateCost` returns 0 (src/models.ts:722) and the only warning is gated behind `CODEBURN_VERBOSE=1` (src/models.ts:694). By default every call on an unpriced model adds $0 to all totals, breakdowns, exports, and MCP responses with no indication anywhere. A new model id shipping ahead of our pricing table (the #629 scenario, 27 models at $0) undercounts real spend and looks identical to a genuinely free model.

## Evidence

- src/models.ts:694: `if (process.env['CODEBURN_VERBOSE'] !== '1') return false`
- src/models.ts:708-722: unknown model returns cost 0
- Verified on main at 1a75484.

## Proposed fix

Track unpriced token volume during aggregation and surface it in the product itself, independent of the stderr warning:

- overview/report/dashboard: a line like `N calls / M tokens unpriced (model X). Map it: codeburn model-alias ...`
- JSON output and MCP: an `unpriced` field
- At minimum, emit the stderr warning once per run by default instead of only under CODEBURN_VERBOSE.

This turns silent undercounting into a visible, self-serve fixable state and should deflect a chunk of the wrong-numbers issue class.

PR #1050, opened 2026-08-19
fix(models): treat subscription SKUs as honestly $0

### LINKED ISSUE #968: Unpriced warning fires on subscription-billed models where $0 is correct, and its suggested model-alias fix would fabricate spend

avs-io:
## Problem

The unpriced-models warning from #638 fires on subscription / flat-rate product SKUs whose cost genuinely is $0. The suggested `model-alias` fix maps those ids onto another model's per-token rate and invents spend.

The first cut classified the wrong ids: it treated metered `codex-auto-review` as $0, missed `kimi-for-coding-highspeed` (the SKU #968 was filed around), and certified a synthetic Warp id (`warp`) that no provider emits.

Closes #968.

## Root cause

`findUnpricedModels` / `isExpectedFreeModel` only treated local-looking models, `model-savings` mappings, and exact zero-rate user overrides as "expected free". Subscription product SKUs look like missing LiteLLM rows. The built-in classifier then used screenshot strings instead of the ids providers actually write, and `model-flat-rate --remove` could not unsilence a wrong built-in.

## Change

Third state alongside priced and unpriced: **flat-rate / subscription-billed**.

- Built-in classifier for real product SKUs: ClinePass `auto-genius`, Warp `auto` / `warp-auto-*`, `kimi-for-coding-highspeed`, `grok-composer-*`, plus Warp Auto / Grok Composer display aggregation keys.
- Dropped `codex-auto-review` (metered; openai/codex#32224; priced by #1056) and unsourced `big-pickle`.
- User hatch: `codeburn model-flat-rate <model>` (`--list` / `--remove`). `--remove` opts out of a built-in so a false positive can warn again without a release.
- Shared `unpricedModelHint()`: alias only if billed per token; otherwise `model-flat-rate`.
- Parse-worker snapshot carries user marks and opt-outs. Daily-cache hash includes them (one-time re-derive on upgrade).
- Existing `warp-auto-*` aliases stay priced; those calls remain in the coverage denominator.

## User impact

Known subscription SKUs drop out of the unpriced warning. Remaining unpriced rows still get an alias path, plus an honest hatch when $0 is correct. A wrong built-in can be opted out with `--remove`.

## Preservation

No new LiteLLM rates. No removal of `warp-auto-*` aliases. No parser / billing-provenance rewrite. No Keychain / Buzz / #868 / #916 / #998 work. No second Extra High.

## Testing

- Focused: models (SKU class, Warp `auto` vs synthetic `warp`, kimi-for-coding-highspeed, auto-review unsilenced, built-in opt-out + snapshot + hash) and `model-flat-rate` CLI (user mark round-trip, built-in `--remove` / re-add)
- Sibling: `warp-auto-efficient` stays `isExpectedFreeModel=false` because it still has a billable alias


iamtoruk:
Reviewed in depth with real-data runs. The third-state concept and the `unpricedModelHint()` consolidation are genuinely good — but on the corpus the issue was filed around, the classifier gets 2 of 3 SKUs wrong and silences one real metered surface:

1. **Blocker — drop `codex-auto-review`** (leaf entry **and** the `/^codex auto review$/i` display regex): OpenAI's tracker says auto review consumes normal model usage (openai/codex#32224) — it's metered; #1056 prices it. Keeping the entry here hides $390+ of real spend and contradicts a sibling PR.
2. **High — the case #968 was filed around is not fixed:** `kimi-for-coding-highspeed` (3.4M tok, a real subscription SKU on a real corpus) is absent from the classifier and still fires the warning. Add it.
3. **High — `warp` claimed covered but isn't:** the observed Warp id in real data is `auto`, not `warp`; the test at `tests/models.test.ts:1029` uses a synthetic `{model:'warp'}` no provider emits, certifying a case that doesn't occur. Match what Warp actually writes.
4. **High — no user escape from a wrong built-in:** `isBuiltInFlatRateModel` is unconditional; `model-flat-rate --remove` only touches config entries. A false positive permanently suppresses a real unpriced warning until a release. Give `--remove` power over built-ins (or a `--force-priced` list) — the failure mode is silent under-reporting of money.
5. **Medium:** `getDailyCacheConfigHash()` losing its early return invalidates every user's stored hash → one-time full daily re-derivation for all installs. Fine, but undeclared — say it in the CHANGELOG.
6. **Medium:** `big-pickle` is classified as a "ClinePass codename" but appears under provider `opencode` in real data — an unsourced assertion of exactly the kind this PR exists to prevent. Cite or drop.
7. Smaller: display-string matching (`/^warp auto\b/i`) inside a pricing predicate is brittle (and #1053 changes those strings — land this before #1053); `parser.ts:3251` serve-memo key omits the flat-rate hash (benign, deserves a comment); stray unbalanced-paren comment edit in `shouldWarnAboutUnknownModel`.

Suggested landing order across the trio: **#1056 (with its version fix) → this one (rebased per above) → #1053 (re-keyed)**.

ozymandiashh (commented):
Code review is otherwise positive: the effective auto-merged code passed 239 targeted tests and typecheck, and the opt-out/flat-rate behavior matched the stated cases. I am not approving or merging this head because it currently conflicts with main in `CHANGELOG.md`. Please rebase and preserve both changelog entries; I can do the final live merge check on the resolved head.

Issue #1029, opened 2026-08-18
models: grok-4.5-build and grok-4.5 collapse to one 'Grok 4.5' display name, showing duplicate rows


iamtoruk:
Surfaced during the #1015 review on a real corpus. `codeburn models` / `codeburn audit` show two visually identical `Grok 4.5` rows (e.g. 487 calls and 50 calls) because the distinct ids `grok-4.5-build` and `grok-4.5` map to the same display name in the grok provider's `modelDisplayName`.

Expected: either the rows merge under one display name (if they are the same billable model — check pricing), or the display name distinguishes them (`Grok 4.5 (build)`), consistent with how other providers handle variant ids. Pre-existing; not introduced by #1015.

PR #1053, opened 2026-08-19
fix(models): resolve and merge raw ids in models report

### LINKED ISSUE #967: codeburn models shows raw ids (gpt-5.6-*, accounts/fireworks/models/*) and duplicate rows that report resolves correctly
## Problem

On 0.9.20, `codeburn models` and `codeburn report` disagree about the same models. `report` resolves friendly names; `models` shows raw ids for some families and also emits duplicate rows.

## Evidence (0.9.20, npm)

Same three models, same run:

| `codeburn models` | `codeburn report` |
|---|---|
| `gpt-5.6-sol` | GPT-5.6 Sol |
| `gpt-5.6-terra` | GPT-5.6 Terra |
| `accounts/fireworks/models/kimi-k2p6` | Kimi K2.6 |

`report` also resolves `gpt-5.6-luna` → GPT-5.6 Luna, which `models` does not surface under a friendly name either.

This is not a missing-names problem — the names clearly exist, since `report` produces them from the same data in the same binary.

It is also not "`models` deliberately shows raw ids". The same table mixes both conventions: `GLM-5.2`, `Kimi K2.7 Code`, `DeepSeek v4 Pro`, `Qwen 3.7 Plus` and `Grok 4.5` all render friendly, while the `gpt-5.6-*` family and the full `accounts/fireworks/models/*` path form stay raw. So the resolution path runs in `models`, but misses those two shapes.

## Second symptom, same view: duplicate rows

`codeburn models` lists `Cline CLI | GLM-5.2` **twice**:

```
│ Cline CLI  │ GLM-5.2   │ Testing (75%)      │ 867.3K │ $0.246 │
│ Cline CLI  │ GLM-5.2   │ Conversation (43%) │  15.7K │ $0.019 │
```

Two distinct raw ids collapse to one display name and are not merged, so one model is reported as two, splitting its tokens and cost across rows. Reproducible with:

```bash
codeburn models | awk -F'│' 'NF>3 {gsub(/^ +| +$/,"",$2); gsub(/^ +| +$/,"",$3); print $2" | "$3}' | sort | uniq -d
```

That row-merge already exists — `aggregateModelTotals()`, keyed by display name, added in #787 for the By Model panel. It looks like `models` does not go through it.

## Why it matters

`codeburn models` is the natural place to look up an id before running `codeburn model-alias`, and it is the view where a raw id is most likely to be copied into a command or a report. Two rows for one model also understates that model's real share when scanning the list by eye.

## Suggested direction

Route the `models` command through the same name resolution and the same `aggregateModelTotals()` merge that `report` uses, rather than adding more entries to the names table — the entries are already there.

Happy to send the PR if you'd like it; wanted to check the approach first per CONTRIBUTING.

Environment: codeburn 0.9.20 (npm), macOS 26.6.1 (25G76) arm64, Node v26.5.0.

avs-io:
## Problem

`codeburn models` shows raw ids (`gpt-5.6-sol`, `accounts/fireworks/models/kimi-k2p6`) that `codeburn report` already resolves, and splits one model into two rows when two raw ids are the same SKU.

Closes #967.

## Root cause

`models` asked each provider for a label, then bucketed by raw id. Providers with a local table miss fell back to the raw id. A later merge keyed on the display string, so live label collisions (GPT-5 / GPT-5.3 Codex / Kimi K2 Thinking / Opus 4.6) would sum distinct SKUs into one row.

## Change

- Provider-first resolution stays. Local labels (Cursor estimated suffix, provider overrides) still win.
- If a provider echoes the raw id, fall back to the global short-name table.
- `cursor-agent` and `vercel-gateway` consult that table after their local transform (suffix / vendor peel).
- Merge by provider + alias-resolved canonical id. Display names stay cosmetic.
- Keep the first-seen raw id (`models --format json` is not a lex-smallest rewrite).
- Mixed rated/unrated Codex credit merges partial-sum the rated buckets and set `creditsIncomplete`.

## User impact

`codeburn models` matches `report` for gpt-5.6-* and Fireworks path-form ids. Path-form `glm-5p2` merges with the bare slug. Distinct SKUs that only share a label stay two rows.

## Preservation

No new LiteLLM rates. No desktop / By Model rewrite (`aggregateModelTotals` already merges). No Keychain / Buzz / #868 / #916 / #998 work. Audit still buckets by raw id (it recomputes rates per id). No second Extra High.

## Testing

- 40 models-report tests + `resolveCanonicalModelId` unit cases (230 passing in the focused files; 3 pre-existing DeepSeek snapshot fails)
- Maintainer cases: provider-table miss (`gpt-5.6-sol` → `GPT-5.6 Sol (est.)`), path-form (`accounts/fireworks/models/kimi-k2p6` → `Kimi K2.6`), canonical merge of fireworks `glm-5p2` + `glm-5p2`, display-only collisions stay split, first-seen raw id, mixed credit partial-sum
- Same display name across providers stays two rows


iamtoruk:
Reviewed in depth on a rebased branch with real-data runs. The display-resolution half is right and verified ($-identical on a 31-row real corpus, `gpt-5.6-sol` → "GPT-5.6 Sol" etc.). Requesting changes on the merge half:

**Blocker — the merge key is a display string.** `bucketKey(provider, modelDisplayName, …)` merges rows whose *labels* collide, and there are live same-provider collisions today: "GPT-5" ← `gpt-5` | `gpt-5-fast` | `gpt-5.2-low`; "GPT-5.3 Codex" ← `gpt-5.3-codex` | `gpt-5.1-codex-high` | `warp-auto-efficient`; "Kimi K2 Thinking" ← `kimi-k2-thinking` | `kimi-code` | `kimi-for-coding`; "Opus 4.6" ← `claude-opus-4.6` | `warp-auto-powerful`. Those are separate SKUs (subscription vs API, fast tier vs base) whose spend would sum into one row. `grok-4.5-build` stays split only because #1029 hand-renamed its label — this converts every future display-name collision from a cosmetic bug into a money bug. Key the merge on the alias-resolved canonical id; keep display cosmetic.

Also:
- **High:** `if (bucket.model < existing.model) existing.model = bucket.model` makes `models --format json` emit the lexicographically-smallest raw id after a merge — an undocumented payload-semantics change (README promises exact ids for `--unpriced`).
- **Medium:** merging a credits-rated bucket with an unrated one nulls `credits`, which `usage-aggregator` sums into the menubar `codexCredits` — a merge can zero a real credit total. Partial sum + incomplete flag instead.
- Cleared: the provider-file changes are display-only (no cache/dedup key impact — correctly no parse-version bump). Pre-existing note: `BUILTIN_ALIASES` resolves before `SHORT_NAMES`, so the `'gpt-5.2-low'` display entry is dead — that mechanism is what makes the collisions above live.

Suggest splitting: land `fallbackRawModelDisplayName` now; re-key the merge + drop the `row.model` mutation in round two. Sequencing note: this should land **after** #1050 (its display-name regexes match `getShortModelName` output this PR changes).

ozymandiashh (changes requested):
The original canonical-ID merge work looks sound, but this head cannot be merged safely against current main. Its conflicted report paths still compute billable output as `output + reasoning`; current main deliberately uses `billableOutputTokens` because Codex output already includes reasoning. Taking the branch side would reintroduce double counting (for example 500k output + 500k reasoning becomes 1M instead of 500k). Please rebase, retain the current `billableOutputTokens` pricing semantics in both audit and models reports, and rerun the report/model suites on the resolved tree.

Issue #1051, opened 2026-08-19
copilot: rollup dedup key uses a per-file occurrence counter, so two journals for one session collide on the same leg index


iamtoruk:
Found while landing #946 (review B-4 fixture). The rollup dedup key `copilot:<sid>:shutdown:<model>:<n>` uses a **per-file** occurrence counter `n`. When a session has two journal files (e.g. resumed into a second `events.jsonl`) with the same model, the second file's leg 1 collides with the first file's leg 1 and is deduped away before the reconciliation sweep — one leg silently dropped. Pre-existing (not introduced by #946; #946's residual keys were changed to be timestamp-derived precisely to avoid this class). Fix direction: key rollups by the leg's shutdown timestamp (or per-session counter across files) — that IS a cached-key-shape change for a durable provider, so it requires a `CACHE_VERSION` bump (see the F4 key-prefix pin test and the carry-forward invariant comment in `parseProviderSources`). Add a two-journal fixture.

iamtoruk:
Closing as unreachable: the deep review of #1054 established that two journals for one session cannot both enter discovery (the discovery path dedupes the session before per-file occurrence counters ever see a second journal), so the leg-index collision this issue describes has no reachable trigger. The retargeted #1054 (merged, 67f085a3) adopted that analysis explicitly — it kept the `:<n>` occurrence keys and took only the lastEventTimestamp ordering for the call timestamp. If a future Codex/Copilot change makes multi-journal discovery reachable, reopen with a repro; the #946 branch's session-store keys make the whole question moot for store-era data anyway.

PR #1054, opened 2026-08-20
fix(copilot): prefer lastEventTimestamp for stampless shutdown calls

avs-io:
## Problem

A Copilot CLI journal with several `session.shutdown` legs that carry `sessionStartTime` but no per-leg timestamp billed every leg onto one date. Keying those rollups by timestamp (or timestamp + journal basename) collapsed the same journal onto one row.

Closes nothing. #1051's two-journal collision is unreachable through discovery (`discoverJsonlSessions` only yields `<sid>/events.jsonl`).

## Root cause

Shutdown timestamp fallback was `event.timestamp || sessionStartTime || lastEventTimestamp`. `sessionStartTime` is identical for every leg of a resumed session. Putting that stamp in the dedup key made three stampless legs share one key. A `:n` → journal-key migration then stripped cached occurrence keys and re-sent them under a new name (ledger double-count).

## Change

- Keep occurrence keys: `copilot:<sid>:shutdown:<model>:<n>`.
- Timestamp only: `event.timestamp || lastEventTimestamp || sessionStartTime`.
- Drop the journal-basename key, the `:n` strip, and the two-journal fixtures.

## User impact

Stampless multi-leg CLI journals keep every leg's input/cache tokens and land on the last stamped event's time, not session start. Already-synced `:n` keys are unchanged.

## Preservation

No LiteLLM rates. No store/OTel/JetBrains/transcript path changes. No `CACHE_VERSION` / fingerprint / ledger rewrite. Per-turn keys unchanged. #946 / #868 / #916 / #998 untouched.

## Testing

- 106 focused tests (`tests/providers/copilot.test.ts` + `tests/parser.test.ts`)
- New test: 3 stampless legs → keys `:1`/`:2`/`:3`, timestamps = lastEventTimestamp not `sessionStartTime`, token sum 3500
- Sabotage: sessionStartTime-first fails that timestamp assertion (`2026-07-15T07:54:00.274Z` vs last event)
- `tsc --noEmit` clean


iamtoruk:
Reviewed in depth against current `main` and against the #946 landing branch (`c41b14f1`), with fixtures. Requesting changes — the direction needs a rethink rather than patches.

**The collision being fixed is unreachable through discovery.** Only CLI jsonl sources emit `:shutdown:` keys (`isTranscript` sources `continue` at `copilot.ts:846`), `discoverJsonlSessions` hardcodes `<stateDir>/<sid>/events.jsonl`, and `sessionId = basename(dirname(path))` — so two shutdown-emitting sources can never share a session id. A fixture with `events-2.jsonl` alongside confirms discovery returns one source. The PR's two-journal tests hand-build a source shape the product cannot produce, and `journalId` is the constant `events.jsonl` in every real key.

**Meanwhile the new key regresses a reachable case.** `shutdownTimestamp` falls back to `sessionStartTime`, which is identical for every leg of a session — one journal with 3 stampless `session.shutdown` legs goes from 3 keys / 7,000 tokens on main to **1 key / 1,500 tokens** on this branch (`copilot.ts:862-864, 908`). That's #1051's failure mode moved from the unreachable case to a common one.

Two more blockers:
- **Migration destroys correct history:** `stripLegacyCopilotShutdownCalls` removes cached `:n` calls unconditionally before knowing the re-parse can reproduce them; a main-shaped cache with `:1`+`:2` for a stampless journal ends with one leg permanently gone (`parser.ts:2831-2839`).
- **Sync re-send:** the ledger is a raw key set with no retraction for usage spans; every already-synced `…:shutdown:<model>:<n>` returns under a new name and is pushed again — permanent receiver double-count. (#946's residual-key comment states this exact rule.)

And with #946 (either landing order): #946 reorders the stamp fallback to `event.timestamp || lastEventTimestamp || sessionStartTime`; under #1054 the stamp is *in the key*, so that reorder re-keys every stampless rollup, the #946 carry-forward union appends the re-parse, and the old key doesn't match `isLegacyCopilotShutdownKey` → silent duplicate.

What does hold up (fixture-verified): legacy keys with real per-leg stamps migrate cleanly and stably; orphaned legacy rollups survive untouched; `reconcileCopilotCalls` and the sync shape-freeze both treat the key tail as opaque and keep working.

**Recommendation:**
1. Land #946 first, unchanged.
2. Retarget this PR to the real defect it surfaced: the *timestamp fallback* collapses stampless legs — take the `lastEventTimestamp`-first ordering for the call's **timestamp only**, leaving the key as `:<n>`. ~1 line, no migration, no ledger churn.
3. Close #1051 as unreachable with the discovery analysis (happy to transfer the write-up there).
4. If the key shape ever genuinely must change: `CACHE_VERSION` bump + a ledger key-rewrite migration, not a targeted strip.

Also noted for whoever lands anything here: `verify:upgrade`'s corpus has no copilot jsonl sources, so it never exercises this class — worth adding one.

ozymandiashh (approved):
Reviewed the current head and effective merge. The fallback changes only the timestamp while preserving the :n identity keys, so cache and sync continuity remain intact. The realistic stampless JSONL repro retained all calls/tokens; 106/106 targeted tests, typecheck, and all CI checks pass. No blocking issue found.

PR #1055, opened 2026-08-20
fix(codex): read session_meta string fields at payload depth 1

### LINKED ISSUE #1045: codex: session_meta raw-field fast path is depth-agnostic for name/cwd/originator/session_id too
Follow-up from the #1040 review. #1040 fixes `model` by switching `session_meta` to the depth-1-only `getRawPayloadFieldWindow` scan (the depth-agnostic `getRawJsonStringField` could match `base_instructions.provenance.model`). The same depth-agnostic scan still applies to `session_meta`'s other raw fields — `cwd`, `name`, `originator`, `session_id` — so a nested field with the same key (e.g. `dynamic_tools[].name`) could shadow the top-level one. Harmless today (`payload.name` is only consumed for `function_call`/`custom_tool_call`), but it's the same latent class; the fast path for `session_meta` should read all its fields depth-1 for consistency. Add a fixture with a nested same-name key.

avs-io:
## Problem

#1040 made `session_meta.model` depth-1 on the compact Buffer path. The same scan still took the first `cwd`, `session_id`, `originator`, `name`, `forked_from_id`, and `model_provider` anywhere in the payload. A nested `dynamic_tools[].name` (or any same-named key under `base_instructions`) could steal the top-level field.

Closes #1045.

## Root cause

`parseCodexLine` uses a depth-agnostic regex over the first 64 KB of large lines. `session_meta` routinely embeds same-name keys inside nested objects. #1040 fixed only `model`.

## Change

- Every `session_meta` string field goes through the existing payload-depth-1 window (`getRawPayloadFieldWindow`). Other event types keep the cheap first-match scan, so `function_call.name` is untouched.
- Bump `PROVIDER_PARSE_VERSIONS.codex` (`session-meta-fields-v1`) and `CODEX_CACHE_VERSION` 9 → 10 so present sources re-parse. No daily-cache layout bump.

## User impact

Large Codex `session_meta` lines keep the real session id / cwd / originator instead of a nested shadow. Typical field order already put these keys first, so most corpora will re-parse with identical totals.

## Preservation

No LiteLLM rates. No models.ts. No store / OTel / JetBrains / transcript / Copilot path. No `CACHE_VERSION` layout change. #946 / #868 / #916 / #998 / #1051 untouched.

## Testing

- New fixture: 40 KB `session_meta` with nested same-name keys first; asserts `sessionId`, `workingDirectory`, `projectPath`, model, and `function_call` → Bash.
- Sabotage: test fails (`shadow-session`) without the depth-1 branch; passes with it.
- 42/42 `tests/providers/codex.test.ts`, plus cache-invalidation / throughput / rich-capture (55 focused).


iamtoruk:
Reviewed in depth — **approve**. Replayed the old vs new read over a real corpus (1,396 rollouts, 6,716 `session_meta` records, CLI 0.111.0 → 0.147.0-alpha.1.2, 24 distinct payload shapes): `cwd`/`model_provider`/`originator`/`session_id`/`forked_from_id`/`model` identical on all 6,716; the only change is `name`, which stops reading a nested `automation_update` — exactly the shadowing this PR kills, and `payload.name` is only consumed for function/custom tool calls, so it's inert. `models --provider codex -p lifetime` byte-identical on scratch caches. Both invalidation knobs bumped (CODEX_CACHE_VERSION 9→10 + parse-version suffix), and the version gate covers the stored resume state, so no mixed old/new reads inside a resumed parse. Gates green incl. the resume boundary differential.

Two nits, non-blocking: the >64 KB-head path works (synthetic 200–300 KB blobs verified) but ships untested — the corpus has no such line; and `getRawPayloadFieldWindow`'s 4 KB window returns `undefined` for a depth-1 string value longer than that (pre-existing property of #1040's helper, now on 7 fields; no plausible value approaches it).

**Coordination note:** this conflicts with #1056 in `codex-cache.ts`/`session-cache.ts` (both bump the same versions). Whichever lands second must resolve by **union** — version = max+1, *both* parse-version tokens appended. A "take ours" resolution silently drops one invalidation.

PR #1056, opened 2026-08-20
fix(models): price Codex activity ids via the official underlying model

### LINKED ISSUE #1047: models: 'Codex Auto Review' (659M tokens on one corpus) is unpriced / $0
Surfaced by the daily-cache re-derive during the #1040 review on a real corpus: the unpriced-model warning grew 9 → 12 entries, the largest being `Codex Auto Review` at 659M tokens, priced \$0. Pre-existing pricing-table gap (not caused by #1040 or #1033); worth deciding whether Codex's auto-review model id maps to a billable row (and which) or is genuinely included/flat. Check the exact model id string in the rollouts and LiteLLM's catalog.

avs-io:
## Problem

`codeburn models` / the unpriced warning reports `Codex Auto Review` at $0. On one real corpus that was 659M tokens (#1047). The id is a Codex *activity surface*, not a missing LiteLLM row and not a subscription SKU.

Closes #1047.

## Root cause

Rollouts store `codex-auto-review`. That string has a display name but no catalog row, so `calculateCost` returns 0. Official rate card (help.openai.com/articles/20001106): auto review uses the recommended review model and draws from the same Codex credit pool as other work. $0 is the wrong honesty.

Builtin aliases are not in `getModelAliasesConfigHash()`, so a warm Codex result cache exact hit and a finalized daily-cache day keep the pre-fix $0 forever.

## Change

Same class as `cursor-auto` / `copilot-openai-auto`:

- Alias only the observed id `codex-auto-review` to the existing `gpt-5.5` LiteLLM row
- Keep the activity label via `autoModelNames`
- Credits resolve the same alias
- Cache generations after the #1075 / #1085 / #1088 / #1090 / #1092 wave: `CODEX_CACHE_VERSION` 15 (main's v14 is the #1092 MCP/skills fix; v13 is #1079 tok/s; v12 is feat/core-extraction #1086), `PROVIDER_PARSE_VERSIONS.codex` keeps `codex-pricing-v1` + `codex-tps-v1` + `codex-mcp-skills-v1` and appends `activity-price-v1`, `DAILY_CACHE_VERSION`/`MIN_SUPPORTED_VERSION` 25 (v21 #946, v22 this PR's earlier claim, v23 #1075, v24 #1090; #1092 left daily alone)
- Do **not** invent `codex-code-review` (not in Codex source or real rollouts)

## User impact

Auto-review tokens price at GPT-5.5. The models table still says "Codex Auto Review". Existing caches re-derive; this is a lossless daily bump (v24 kept as baseline).

## Preservation

No new LiteLLM rates. No `codex-*` prefix collapse. No date-invented Terra mapping. No Keychain / Buzz / #868 / #916 / #998. No second Extra High.

**Note for #1050:** that PR lists `codex-auto-review` as a built-in flat-rate SKU. This PR says the opposite. When they meet, drop the id from the #1050 classifier — do not Extra High #1050 again.

## Testing

- 247 focused tests after merge-main (models, Codex provider, credits, suffix-cache, throughput, cache-invalidation)
- Object-identity: `getModelCosts('codex-auto-review') === getModelCosts('gpt-5.5')`
- Warm versioned $0 exact hit is discarded and repriced
- Unobserved siblings (`codex-code-review`, bare `code-review`) stay unpriced
- Extra High once: `/tmp/codeburn-xhigh-review/VERDICT-1047.md` → MERGE AFTER FIX. Named class closed. Do not Extra High again.


iamtoruk:
Reviewed in depth with real-data runs — this is the factually supported direction: OpenAI's own tracker says auto review **consumes normal model usage** (openai/codex#32224), so pricing it is right and #1050's `codex-auto-review` flat-rate entry should be dropped (its body already says so). On this corpus the row goes $0 → $390.12 over 659M tokens. Before merge:

1. **Blocker — daily-cache version collision:** main is at 20; this PR takes 21; **#946 (in landing) also takes 21** and edits the same `scripts/upgrade-path/run.mjs` filename constant. If both merge on green CI, the loser writes a v21 file with the winner's schema. #946 is further along — please take **22** here (and move `NEW_DAILY_CACHE` with it), or coordinate explicitly.
2. **High — the `gpt-5.4` pin is unverified and carries a 2× error bar.** The cited help-center page 403s; public evidence says `review_model` defaults to the **session model** with GPT-5.5 currently recommended, and this user's dominant Codex model is GPT-5.5 (credits 125 vs 62.5 per M) — the honest figure could be ~$780, not $390. Either pin to `gpt-5.5` with a comment that it tracks the recommended review model, or mark the cost estimated (`~`). Presenting it as exact is the thing this repo doesn't do.
3. **High — duplicate key after rebase:** main already has `'codex-auto-review'` in `autoModelNames`; the hunk re-adds it (post-merge it exists twice). Drop on rebase.
4. **Medium — priced in dollars, invisible in credits:** `codexCredits()` matches `'5.4'`/`'5.5'` substrings, so the biggest-token Codex surface contributes `credits: null` while the PR's own rationale is "draws from the same credit pool". Resolve the alias in `codex-credits.ts` too.

The three cache bumps (CODEX_CACHE_VERSION 9→10, parse `…-activity-price-v1`, daily) are correct and necessary — that part is exactly right.

ozymandiashh (dismissed):
This needs a rebase plus fresh cache generations; resolving the conflicts mechanically would create real stale-cache collisions. The branch uses Codex results v11 and the old unsuffixed file while current main already uses v11 for a different invalidation and now has version-suffixed cache files. It also uses daily cache v22 while main is v23, and drops `codex-pricing-v1` instead of composing it with `activity-price-v1` in the session parse token. Please rebase, allocate the next non-colliding versions (Codex v12 / daily v24 on the current main), preserve both invalidations, and rerun upgrade-path/cache migration coverage.

PR #1057, opened 2026-08-20
fix(cache): expose skipped-shard fingerprints on scoped loads

### LINKED ISSUE #1034: cache: scoped loads re-parse out-of-range files every run because their entries aren't visible to reconcileFile
Follow-up to #1032 (the republish half is fixed by the merge-shard no-op check).

Under #1007's month-scoped `loadCache`, files whose bucket is out of range have no entry in `section.files`, so `reconcileFile` sees them as uncached and the parser re-parses them on every ranged run (`status --format json` etc.), producing an identical entry. Observed on a real corpus: Claude files in an out-of-scope month, plus SQLite-backed sources (cursor `state.vscdb`, `warp.sqlite`) whose mtime moves constantly. It's now a wasted parse + one shard read, no longer a write.

Fix direction: have the scoped load expose the *fingerprints* (dev/ino/mtime/size/offset) of entries in skipped shards — cheap, e.g. a fingerprint index in the envelope or a per-shard sidecar — so `reconcileFile` can classify them `unchanged` without loading turn bodies. That's an envelope format change (lossless), so it belongs in its own PR with the usual migration/two-writer tests.

avs-io:
## Problem

Month-scoped `loadCache` (#1007) skips out-of-range shards, so those files have no entry in `section.files`. `reconcileFile` then treats them as uncached and the parser re-parses them on every ranged run.

Maintainer review on the first index: a 1.9 GB corpus `status --day` skipped **2 files / 0 MB**, wall time unchanged (~6.2 s), while the envelope grew **3 KB → 5.6 MB** because the index copied every in-memory file.

Closes #1034.

## Root cause

The envelope named skipped shards but did not expose their fingerprints. The first index then stored *every* file, including months already in memory.

## Change

- Additive v9 `providers[p].index`: path → `{ fingerprint, bucket }` for **skipped, still-named buckets only**.
- Full save **omits** the index.
- Index-only entries are never installed for attempted-but-unreadable shards, dropped buckets, or a full load (080bf62 / 3b52fc3).
- No `lastCompleteLineOffset` on new index rows.
- No `CACHE_VERSION` bump.

## User impact

Warm ranged runs skip unchanged out-of-range JSONL **when the skipped set is real**. A day query that barely skips anything no longer pays a multi-MB envelope. SQLite sources whose fingerprint moved still re-parse.

## Preservation

No shard layout change. Failed markers stay in `0000-00`. Durable providers still load in full. SQLite half of #1034 is still not claimed.

## Testing

- `npx vitest run tests/session-cache.test.ts tests/session-cache-shards.test.ts tests/scoped-load-index-only-parser.test.ts` — **130 passed**.
- Sabotage: full save has no `index`; scoped save keys are skipped buckets only.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -24,10 +24,12 @@ import {
   fingerprintFile,
   isCacheComplete,
   isCacheDirty,
+  isIndexOnly,
   loadCache,
+  lookupCachedFile,
   markCacheDirty,
   monthScopeForRange,
-  reconcileFile,
+  reconcileIndexedFile,
   saveCache,
 } from './session-cache.js'
 import { acquireCacheRefreshLock, type RefreshLockHandle } from './cache-refresh-lock.js'
@@ -1965,8 +1967,20 @@ async function scanProjectDirs(
       const fp = await fingerprintFile(filePath)
       if (!fp) continue
 
-      const cached = section.files[filePath]
-      const action = reconcileFile(fp, cached)
+      const cached = lookupCachedFile(diskCache, 'claude', filePath)
+      const indexOnly = isIndexOnly(diskCache, 'claude', filePath)
+      const action = reconcileIndexedFile(fp, cached, indexOnly)
+      if (indexOnly) {
+        // Out-of-range shard: fingerprint is enough to skip an unchanged file
+        // without loading turn bodies. Do not push empty-turn stubs into
+        // summaries or section.files (#1034).
+        if (readOnly || action.action === 'unchanged') {
+          if (readOnly && action.action !== 'unchanged') readOnlyServedStale = true
+          continue
+        }
+        if (!readOnly) changedFiles.push({ filePath, info: { dirName, fp, source } })
+        continue
+      }
       if (cached && (readOnly || action.action === 'unchanged')) {
         if (readOnly && action.action !== 'unchanged') readOnlyServedStale = true
         unchangedFiles.push({ filePath, dirName, source, cached: section.files[filePath]! })
@@ -3056,11 +3070,24 @@ async function parseProviderSources(
     const fp = await fingerprintFile(source.path)
     if (!fp) continue
 
-    const cached = section.files[source.path]
-    const action = reconcileFile(fp, cached)
+    const cached = lookupCachedFile(diskCache, providerName, source.path)
+    const indexOnly = isIndexOnly(diskCache, providerName, source.path)
+    const action = reconcileIndexedFile(fp, cached, indexOnly)
     // A cached parse failure at this same fingerprint stays skipped — don't
     // re-read a file that already threw and hasn't changed. It re-parses only
     // when the file changes (then `reconcileFile` reports non-'unchanged').
+    if (indexOnly) {
+      // Out-of-range shard: skip unchanged sources without loading turn bodies.
+      // Fingerprint growth cannot append (no body), so it falls through as a
+      // full re-parse (#1034).
+      if (readOnly || action.action === 'unchanged') {
+        if (readOnly && action.action !== 'unchanged') readOnlyServedStale = true
+        continue
+      }
+      if (!readOnly) changedSources.push({ source, fp })
+      else readOnlyServedStale = true
+      continue
+    }
     if (cached && (readOnly || (action.action === 'unchanged' && (cached.failed || !cachedFileNeedsProviderReparse(providerName, source.path, cached))))) {
       if (readOnly && action.action !== 'unchanged') readOnlyServedStale = true
       unchangedSources.push({ source, cached })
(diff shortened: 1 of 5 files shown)


iamtoruk:
Reviewed in depth on a rebased branch with real-data runs. The shape is right — an optional `index` on the existing envelope, add-only for v9, atomicity and two-writer safety inherited; verified with the concurrent-writer/stale-writer suites and the full upgrade harness (0.9.20 → this build migrates; shard-name stability from #1035 not regressed; March/July/day exports byte-identical to main). Two things gate merge:

1. **High (fix available):** an index entry the shards can't back silently drops the file — permanently in one path. (a) An in-scope shard that fails to load (corrupt, or unlinked by the concurrent republish+sweep race this code designs for) leaves its files index-only → classified `unchanged` → skipped from output *and* from the next save's shard; (b) an index entry whose bucket the envelope no longer names is index-only forever, invisible even on full loads. Pre-PR both fell through as uncached and re-parsed — self-healing. Both reproduced; the fix is to trust the index only for buckets this load *skipped* and the envelope still *names* (a full load ignores the index). I have a verified +11-line fix with two regression tests that fail on the unpatched tree — happy to push it to this branch if you flip maintainer-edits on, or paste the diff here.

2. **The perf case doesn't hold on a real 1.9 GB corpus.** After #1035, a ranged `status --day` on main re-parses **2 files (0 MB)** per run; this branch takes that to 0 with statistically identical timings (6.2 s ± noise both sides) — because `shardInScope` uses the shard's `until`, long-running sessions keep nearly every shard in scope, and only one month was ever skipped. Meanwhile the envelope grows **3.0 KB → 5.6 MB** (16k+ index entries), read on every `loadCache` (~12 ms vs 0.09 ms, × every serve request) and re-serialized on every save, plus a one-time full shard sweep on first save to backfill the index. On this corpus the PR costs more than it saves. Before merging I'd want a corpus/measurement where the win is real — or a narrower index (e.g., only providers/months where the skipped set is non-trivial, or fingerprints without offsets).

Smaller: the index-only branch drops the `cachedFileNeedsProviderReparse` condition (an out-of-range devin source stops re-enriching — undocumented behavior change, self-corrects when queried); index entries for deleted files in never-loaded buckets are carried forever (no growth bound); and the SQLite half of #1034 (cursor/warp mtime churn) is orthogonal — those shards are always in scope — worth stating on the issue so it doesn't read as half-fixed.

Cross-PR: merge-tree clean against the #946 landing branch; copilot is durable → never index-only, so the overlap is benign (re-run the copilot suite on whichever lands second).

ozymandiashh (changes requested):
The self-healing correctness fixes are present, but the stated performance gate is still unmet. The real 1.9 GB corpus measurement reduces only two reparses / 0 MB and leaves wall time effectively unchanged (~6.2 s), while the envelope grows from ~3 KB to ~5.6 MB and is read/serialized repeatedly. For a performance-only change this is currently a net cost without a demonstrated production win. Please narrow the index or provide a representative measurement where the saved work outweighs the persistent envelope overhead.

iamtoruk:
Rebased onto current main (clean merge, 134/134 tests pass) and re-ran the measurement your `8177456f`/`1a22f064` fixes were aimed at — this time work-based rather than wall-clock-only, on a different real corpus (~21k files across 9 providers).

Good news first: `8177456f` does what it says. The envelope no longer explodes — 3KB→220KB here (965 index entries) instead of the 3KB→5.6MB the original design produced. And the mechanism genuinely works: on a day-scoped `today`, main reparses the same 1,080 unchanged out-of-range files on *every single warm run, forever* — confirmed it never converges — while your branch cuts that to 115, with the other 965 served from the index. That's real, and it's exactly what #1034 describes.

But it still doesn't move wall-clock or CPU: 6.42s vs 6.44s wall, 5.57s vs 5.77s user CPU, steady state, 3-run averages. The 1,080 files the bug reparses are cheap — short, stale, abandoned sessions — and the dominant cost of a scoped run is the full-tree stat()/fingerprint discovery pass over all ~21k files, which is scope-independent and out of this PR's reach. Same shape as the original 1.9GB finding, at a different scale — which says it's structural, not a corpus artifact: this PR optimizes a step that was never the bottleneck.

Three data points (your corpus; mine by wall-clock; mine by CPU) land the same place. Leaning toward closing this and #1034 as "the fix is correct but the win isn't there" — your correctness follow-through was solid and the envelope fix answered the hold's regression completely. What would reopen it: a corpus/workload where reparse-avoidance is wall-clock-visible (far LARGER out-of-range files, not just more of them), or pairing it with work on the discovery sweep itself, which is where the ~6.4s actually lives.

iamtoruk:
Closing per maintainer decision, on the measurement two comments up: the mechanism is correct and the envelope regression is fully fixed (your follow-through here was exemplary — two same-day fix rounds, both verified), but the performance justification never materialized across three independent measurements on two real corpora. The 89% reparse cut is real but invisible behind the scope-independent discovery sweep that actually owns the runtime. Reopens with a workload where the win is wall-clock-visible, or as part of work on the discovery sweep itself.

PR #1058, opened 2026-08-20
Price and label OrcaRouter gateway model ids as the model they route to

XiaoHuo888-hue:
## Problem

OrcaRouter sessions show $0. The gateway writes route ids (`orcarouter/auto`, `orcarouter/fusion`, `orcarouter/fusion-flash`, `orcarouter/fusion-mini`) and nested upstream ids (`orcarouter/deepseek/deepseek-v4-pro`). Those were not routing wrappers, so they never hit a LiteLLM row.

## Root cause

`cmd/` and `antigravity/` already peel as routers. `orcarouter/` did not: it was neither in `ROUTER_PREFIXES` nor `EXTRA_NAMESPACES`, and the smart/fusion route ids had no alias or branded label.

## Change

Rebased the community mapping onto current `main` (was CONFLICTING). Files only: `src/models.ts`, `tests/models.test.ts`, `CHANGELOG.md`, `README.md`.

- Peel `orcarouter/` like the other routing wrappers.
- Alias `orcarouter/auto` → `claude-sonnet-4-5` (fast Sonnet estimate; the smart route rotates).
- Alias `orcarouter/fusion` / `fusion-flash` / `fusion-mini` → `openai/gpt-oss-120b` (verified live 2026-08).
- Nested upstream spellings price at the exact LiteLLM row.
- Branded labels: `OrcaRouter (auto)`, `OrcaRouter Fusion`, …
- Unknown vendor nested inside the route still fails closed.

## User impact

An OrcaRouter session reports the upstream model's spend instead of $0. Route ids keep a branded name rather than the current upstream's display name.

## Preservation

No parser, provider, doctor, cache, or Hermes changes. Existing wrappers (`cmd/`, `antigravity/`, `cp/`, …) are untouched. Unknown vendors stay unpriced.

## Testing

`HERMES_HOME=/tmp/cb-cleanhome npm test` — 3170 passed, 5 skipped.
`npx tsc --noEmit` — clean.

Unit tests cover pricing, fail-closed, and display labels for the route ids.

Thanks @XiaoHuo888-hue.


ozymandiashh (changes requested):
Blocking pricing issue: the patch maps `fusion`, `fusion-flash`, and `fusion-mini` to one `openai/gpt-oss-120b` completion. OrcaRouter documents Fusion as a panel plus judge and bills the sum of all completions; the variants also use different panels. Treating each route as one static model will materially understate cost, and `auto` is similarly not a fixed Sonnet route. Please model the billed multi-call composition (or leave these gateway IDs explicitly unresolved) and add fixtures for each route. References: https://www.orcarouter.ai/models/orcarouter/fusion and https://www.orcarouter.ai/models/orcarouter/fusion-flash.

iamtoruk:
Thanks for this, and thank you for disclosing your affiliation up front — that's exactly the right way to bring a vendor PR.

The **prefix-peel is the good part** and I'd like to land it. I confirmed OrcaRouter's zero-markup pass-through billing from your pricing page, which is what makes peeling `orcarouter/` to the upstream row correct — `orcarouter/deepseek/deepseek-v4-pro` prices at exactly the `deepseek/deepseek-v4-pro` row, and unknown nested vendors fail closed. That half is well built.

The four `BUILTIN_ALIASES` entries can't land as written, and I want to show you why with numbers.

**`orcarouter/auto`** is aliased to `claude-sonnet-4-5` = $3.00/M in, $15.00/M out. But your own PR body reports the route returning `MiniMax-M2.7` and `qwen3.6-flash`. Both are already in our catalog at $0.30/$1.20 and $0.19/$1.13. So the alias overstates input by 10-16x. The code comment says a Sonnet rate is "the honest estimate" for a flash model, and I don't think that holds — and since both real targets have rows, the $0 fallback you're avoiding wouldn't have triggered anyway. (Also: `/models/orcarouter/auto` 404s on your site — a canonical listing link would help.)

**The fusion routes** go the other way. Following @ozymandiashh's links: `fusion` runs Opus 4.8 + GPT-5.5 + Gemini 3.1 Pro with an Opus 4.8 judge, `fusion-flash` runs a *different* panel (Gemini 3.5 Flash + MiniMax M2.7 + GLM 5.1), and both bill as the sum of the completions that actually ran. A fanned-out 10k-in/2k-out request costs about **$0.384**; this patch prices it at **$0.0014** — 274x low. I suspect your live probe hit the cheap default that casual chat routes to, rather than the fan-out path.

A wrong-but-plausible number is worse for us than $0, because $0 trips the unpriced-model warning and tells the user to look, while a plausible number is silent.

Suggested path: **keep the peel, drop all four aliases.** Route ids stay unresolved and surface the warning; every `orcarouter/<vendor>/<model>` session gets correctly priced, which is the actual $0 bug. If you'd like to price the routes properly later, the panel-sum composition is a good follow-up PR on its own.

Three mechanical things: the branch needs a **rebase onto main** (it's ~10 cache-versions stale, and main has since moved the DeepSeek rows your tests compare against — `npm test` currently shows 3 failures on your base, though those are the stale base's, not yours). The new tests compare a value to itself, so they'd pass even if a rate changed underneath — the `zzz-namespace-probe` synthetic-override pattern just above in the file is the house style and stays hermetic. And could you move the orcarouter.ai link into a `docs/providers/` page like our other gateways, rather than the Pricing paragraph? If a dollar-moving mapping survives the rework, it'll also need a daily-cache ladder bump (finalized days persist cost) — flag it and we'll assign the number at merge time.

Happy to review a follow-up quickly. The peel is genuinely useful work.

iamtoruk:
Superseded by #1118 (merged), which implements the same peel with the review outcomes from both PRs applied: nested upstreams price at exact LiteLLM rows, fusion routes alias to their live-verified target with provenance in-code, and orcarouter/auto deliberately stays unpriced (the alias guesses here measured 16x-over/274x-under — fail-closed until a live probe pins the rotating target). Thanks for kicking this off — the peel approach itself is what shipped.

PR #1069, opened 2026-08-21
fix(web): lead session chart legends with title when unique

### LINKED ISSUE #997: Chart legend shows session UUIDs instead of session titles

avs-io:
## Problem

`codeburn web` Usage → Hourly buckets grouped by Sessions rendered each series as a truncated session UUID (plus a shared project prefix). In a monorepo every line looked the same. The Context tab and `codeburn sessions` already show titles.

Closes #997.

## Root cause

`buildGranularHistory` collected titles but always led the legend with `shortSessionId`. The chart clips labels at ~24 glyphs (`max-w-40`), so the human-readable title never made the visible prefix.

## Change

- Unique cleaned title in that visible window leads: `Refactor billing module (claude)`.
- Colliding titles or long shared prefixes stay short-id-first so truncated series stay distinguishable.
- Blank / missing titles keep the project-path fallback.
- Sanitise / 80-code-point cap unchanged.
- Visible uniqueness counts code points (`4065255`), matching the title cap — Extra High MERGE AFTER FIX on `088d264`.

## User impact

Session charts name the work, not the hex fragment, without collapsing two series into one truncated line.

## Preservation

- Untitled sessions unchanged.
- Model series unchanged.
- `sessionDisplayName` in sessions report / TUI already preferred title; not rewritten.
- Residual collision path still may include full path+id (pre-existing; not this class).

## Testing

`npx vitest run tests/granular-history.test.ts` — 17 passed.

### DIFF
--- a/src/granular-history.ts
+++ b/src/granular-history.ts
@@ -177,8 +177,13 @@ function preferredSessionTitle(titleCandidates: Map<string, SessionTitleCandidat
 
 // Chart legend is `max-w-40` at 10px ≈ 24–32 glyphs. A title that is unique in
 // that window can lead; otherwise the short id must stay in the prefix or
-// truncated series look identical (the #997 class).
+// truncated series look identical (the #997 class). Count code points, matching
+// the title cap — a UTF-16 slice can split an emoji and collide two titles.
 const VISIBLE_LEGEND_PREFIX = 24
+
+function visibleLegendPrefix(label: string): string {
+  return Array.from(label).slice(0, VISIBLE_LEGEND_PREFIX).join('')
+}
 
 function buildSessionLabels(inputs: Map<string, SessionLabelInfo>): Map<string, string> {
   // Stable raw-key order makes the residual used-label guard independent of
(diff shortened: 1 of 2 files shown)


ozymandiashh (approved):
Reviewed the current head and the effective merge with main. The unique-title legend behavior preserves collision fallbacks and sanitization; the targeted granular-history suite (17/17), typecheck, and all required CI checks pass. I found no correctness blocker.

PR #1071, opened 2026-08-21
fix(cache): accept completed-by-other as the stale-lock loser

### LINKED ISSUE #904: cache-refresh-lock-process: takeover-window race misclassifies completed-by-other even in a serial single-fork run

avs-io:
## Problem

CI quarantined `cache-refresh-lock-process` because the loser of a stale-lock contest sometimes reports `completed-by-other` instead of `timed-out` (#904). The suite then failed even though exactly one worker published.

Closes #904.

## Root cause

While the winner unlinks the stale primary and creates the successor, a poll can see missing lock + missing takeover guard. The lock classifies that as `completed-by-other`. That is honest. The test required `timed-out`.

## Change

Loser must not be `.parsed`. Allow `timed-out` | `completed-by-other` | `unavailable`. Do not change missing+missing to wait-out — that would delay a clean owner release.

## User impact

The process lock suite can gate again. Refresh still has one writer.

## Preservation

Lock implementation unchanged. Unit suite (`cache-refresh-lock.test.ts`) unchanged. #937 corrupt-lock recovery not touched.

## Testing

`npx vitest run tests/cache-refresh-lock-process.test.ts tests/cache-refresh-lock.test.ts` — 24 passed.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,14 +33,13 @@ jobs:
       # script since #948, so CI and a contributor's `npm test` can never drift.
       - name: Test suite (parallel)
         run: npm test
-      # Single forked worker, so lock contention comes only from the child processes the
-      # tests spawn deliberately. Quarantined (reports, never gates): the process
-      # suite still races its own takeover window even serially on slow runners -
-      # tracked in #904; drop continue-on-error once that race is settled.
-      # Step-level timeout so a stalled lock suite fails soft here instead of
-      # tripping the job's 15-minute budget, which kills the whole job as
-      # "cancelled" and hides the parallel suite's green result.
-      - name: Cache-lock suite (serial, quarantined)
-        continue-on-error: true
+      # Single forked worker, so lock contention comes only from the child
+      # processes the tests spawn deliberately. Serial because the suite is
+      # parallelism-sensitive (fails under full worker pressure, passes
+      # serially). After #904 the loser of a stale-lock contest may be
+      # timed-out or completed-by-other; this step gates CI.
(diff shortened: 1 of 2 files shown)


ozymandiashh (approved):
Reviewed the current head and its merge with main. The stale-lock loser outcomes are covered without changing runtime behavior; the lock suites (26/26), typecheck, and all available CI checks pass. No blocking issue found.

PR #1072, opened 2026-08-21
fix(report): require durable period; drop unreachable dailyMap

### LINKED ISSUE #1067: core-extraction: buildJsonReport's dailyMap fallback is unreachable dead code
Found during #929 review: `buildJsonReport` in `packages/cli/src/main.ts` takes `durable` from both call sites unconditionally (`buildDurablePeriod()` never returns null), so the `dailyMap` fallback branch (~30 lines) can never execute and cannot be covered by any test from any entry point. #929 originally patched this dead branch for the midnight-straddle rule; the hunk was dropped as unreachable. Follow-up: make `durable` a required parameter and delete the fallback branch (tsc + suite will prove nothing depends on it), or document why the fallback stays.

avs-io:
## Problem

`buildJsonReport` kept a live `dailyMap` fallback for callers that omit `durable`. Both JSON call sites always pass `durable` (`buildDurablePeriod` never returns null), so ~30 lines were unreachable and untestable. #929 had to drop a midnight-straddle patch to that branch for the same reason.

Closes #1067.

## Root cause

`durable` was typed optional after the durable cache became mandatory.

## Change

- `durable: DurablePeriod` is required.
- Headlines and `daily[]` come only from `durable.data` / `durable.days`.
- Delete the live `dailyMap` fallback.
- Proxied/net split remains live-session (not stored per day).

## User impact

JSON report shape unchanged. Dead code gone.

## Preservation

Call sites already passed durable. `tests/day-aggregator.test.ts` keeps its own per-call bucket mirror; not this function.

## Testing

`npx tsc --noEmit` (no `main.ts` errors). `npx vitest run tests/day-aggregator.test.ts` — 13 passed.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -472,7 +472,7 @@ program.hook('preAction', async (thisCommand) => {
   await loadCurrency()
 })
 
-function buildJsonReport(projects: ProjectSummary[], period: string, periodKey: string, durable?: DurablePeriod) {
+function buildJsonReport(projects: ProjectSummary[], period: string, periodKey: string, durable: DurablePeriod) {
   const sessions = projects.flatMap(p => p.sessions)
   const { code } = getCurrency()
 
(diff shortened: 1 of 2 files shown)


ozymandiashh (approved):
Reviewed the current head and effective merge. The report path now requires a durable period and removes the unreachable dailyMap branch without changing valid aggregation behavior; 22/22 targeted tests, typecheck, and the full CI matrix pass. No blocking issue found.

PR #1073, opened 2026-08-21
fix(sync): drop cleartext session id from attribution spans

### LINKED ISSUE #1070: sync attribution: drop cleartext ai.session_id from attribution spans (redundant with keyed traceId)
Found during the feat/core-extraction review of the attribution port (#932), but the behavior ships on main today: attribution session spans emit `ai.session_id` in cleartext (src/sync/otlp.ts), while the docs describe the join to usage spans as "shares the usage spans' traceId" — i.e. the HMAC-keyed traceId already provides the receiver-side join, so the raw field is redundant and could be dropped with no loss of function. Session ids are file/dir basenames or provider ids, not paths, so this is hygiene rather than a leak — but before attribution landed, the session id existed on the wire only as HMAC input. Fix on main first; the extraction branch will pick it up via sync, avoiding drift at the final merge. Related follow-up from the same review: two sessions sharing an id across projects clobber each other server-side because the wire keys sessions by ai.session_id alone (docs tell receivers to upsert by it).

avs-io:
## Problem

Attribution session and commit spans emitted `ai.session_id` in cleartext. Usage spans never did. The join was already the shared `traceId`. Docs told receivers to upsert session spans by `ai.session_id`.

Closes #1070.

## Root cause

`--attribution` copied the session id onto the wire as an attribute even though `deriveTraceId(sessionId)` already keys both usage and attribution spans. Before attribution, session id existed on the wire only as hash input.

## Change

- Drop `ai.session_id` from attribution span attributes (session and commit).
- Receivers upsert session spans by `traceId` (same id usage spans already carry).
- Pin usage spans still omit the field.
- Docs match the wire. Do not list session ids under What is NOT sent — this is hygiene, not a leak.

## User impact

Receivers that keyed session rows on `ai.session_id` must key on `traceId`. Join to usage is unchanged. Local sent-ledger keys still use session id.

## Preservation

Does not change `deriveTraceId`, ledger keys, commit upsert `(git.repo, git.sha)`, or the cross-project same-id clobber (that follow-up is not this slice).

## Testing

`npx vitest run tests/sync-attribution.test.ts tests/sync-ledger-otlp.test.ts` — 78 passed.


ozymandiashh (approved):
Reviewed the current head and its merge with main. The change removes cleartext session IDs from both attribution span types while preserving trace correlation; 78/78 targeted tests, typecheck, and the full CI matrix pass. No blocking issue found.

PR #1074, opened 2026-08-21
fix(core): bound the model field at the observation boundary, and make the privacy guards prove it

iamtoruk:
Lands #933 and #935 together, reconciled against the current `feat/core-extraction` and reworked per the adversarial review of each. Both PRs are by @ozymandiashh — this branch is their work; neither should be closed as superseded without their sign-off. They conflicted in both merge orders (`bridge.ts`, `session-cache.ts`, `content-smuggling.test.ts`) and were coupled: #935's copilot HMAC needs #933's privacy-key threading, and #933's copilot-adjacent change needs the parse-version discipline #935 carries.

## The design decision this branch makes

**Model normalization applies only at the observation boundary, never inside a dedup key.**

#933 normalized `model` in six decoders' key construction as well. That is dropped here. A dedup key is an *identity*, not an observation field:

- Collapsing two distinct raw models to `'unknown'` inside a key merges two genuinely different calls into one.
- It changes every cached key's value, so five more providers would need parse-version bumps and a daily-cache bump, for no privacy win.
- The win it claims is partial anyway: a dedup key is assembled from many raw components — paths, session ids, endpoints, ledger text — so cleaning the model component of a few providers proves nothing.

**Consequence, stated plainly:** the envelope's `dedupKey` still carries raw provider text for some providers (e.g. `vercel-gateway:<day>:<model>`). That surface is real and is deferred to the schema-wide dedup-key-hashing follow-up. The guardrail tests do not pretend otherwise: `serializeWithoutDedupKeys` excludes exactly that field, names why in its docstring, and is used only where a planted value rides a key component — so the exclusion shows up in the diff instead of hiding inside a passing assertion.

Because dedup keys are unchanged for those providers: no parity break, no cache churn, no daily-cache bump, and **no CLI number moves** (verified against a real corpus below).

## What landed

**From #933 — kept**
- `ModelIdentifier` bound (min 1, max 128, `^[A-Za-z0-9._:/@-]+$`) on `model` and `pricingModel`, in the zod schema and `schemas/observation-0.2.0.json`.
- `normalizeModelIdentifier` in all 34 providers' `toObservations` (the PR said 36; there are 34 `observations.ts` files, and a sweep confirms every `model:` assignment into a `CallObservation` is normalized).
- The smuggling-test flip: the inverted assertion that certified the hole is gone, a positive control proves a legitimate identifier still crosses, and every hostile case is guarded against vacuity.
- `sourceRefFingerprint` (new `source` HMAC domain) in the codebuff / zerostack / pi / omp / grok / lingtai-tui dedup keys, their six parse-version bumps, and `bridge.ts` threading `getHostPrivacyKey()`.

**From #933 — removed**
- Every `normalizeModelIdentifier` call inside key construction: `antigravity/decode.ts` (`usageSignature`), `copilot/decode.ts` (shutdown key), `forge/decode.ts` (`stableId` fallback), `quickdesk/decode.ts`, `vercel-gateway/decode.ts`, `lingtai-tui/decode.ts`. A grep of the merged tree confirms the normalizer now appears only in `schema.ts` and in `observations.ts` files.
- The `DAILY_CACHE_VERSION` 15→17 bump (base is at 25 and no rollup number changes).

**Review fix F6** — `pricingModel` got the hard bound in #933 but no normalizer, so the first producer filling it from a display name would hard-reject the whole envelope. It is documented as sharing the normalizer contract and pinned by a schema test. *No call site was added: no producer sets `pricingModel` today — it is CLI-side only and dropped at the boundary — so there is no boundary code to change without inventing one.*

**Review fix F4** — the six providers' keys now derive from the random per-install privacy key, which a parse version cannot see. A lost, rotated, or ephemeral (unwritable config dir) key would silently produce keys that never match the cached ones. `computeEnvFingerprint` now folds a digest of the key in for exactly the providers whose parse version carries the `source-ref-fingerprint-v1` token. Test proves both halves: same cache + different key ⇒ codebuff's fingerprint changes, claude's does not.

**Review fix F3** — the five bridge parity goldens computed their expected keys with the same function *and the same key* production uses, so they pinned nothing. They now assert key SHAPE (`/^codebuff:[0-9a-f]{16}:a1$/`), assert the raw fixture path appears nowhere, and re-derive the fingerprint **longhand** (`createHmac('sha256', pinnedKey).update('source:' + path)`) under a pinned test key — so the domain tag, key, and truncation are pinned rather than agreeing with themselves. Mutation-verified: restoring the raw path breaks 12 of them.

**From #935 — all of it**
- Diagnostics keyed-detail rework: required-key `sanitizeDetail`, `isolateRecords` strips caller-supplied details, `keyedDetail` omit-when-keyless helper, three decoders wired.
- Copilot JetBrains digest sha256 → HMAC, with its `cli-shutdown-cost-v1-skills-dedup-key-hmac-v1` parse version and the invalidation test.
- import-smoke strengthening and the architecture-gate schema walker + synthetic self-test.

**Review fix G1** — #935 recomputed the copilot JetBrains goldens under the bridge's *empty* key. With #933's real-key threading they become machine-dependent and permanently red on CI. They are recomputed under a **pinned** key: `tests/setup/fixed-privacy-key.ts` writes a fixed `privacy-key` file into the sandbox HOME that `tests/setup/env-isolation.ts` already mints, as a side effect of a first-position import. No production env-var override was added — accepting an externally supplied privacy key would be a far worse hole than the goldens are worth. Verified machine-independent: identical results with `HOME` pointed at an empty dir, at a dir holding a *different* real key, and at the developer's real home; and the goldens do move when the pinned key changes, so they are pinned to it and not to nothing.

**Review fix G4** — new core property test: the same JetBrains input under two different privacy keys yields different dedup keys, plus a direct anti-`createHash` pin. Mutation-verified.

## Correcting #935's import-smoke claim

#935's docstring read as though every parser, decoder, and detector body was exercised. It is **one representative each** (claude `parseJsonlLine`, codex `decodeCodex`, `junkReadsDetector`). Import purity is proven for every exports subpath; call-time purity is a spot check for those three. Both the test docstring and the harness comment now say so.

## Versions

`@codeburn/core` has a breaking diagnostics API change (version bump DEFERRED to the next release by maintainer decision — see below):  (`DIAGNOSTIC_DETAIL_MAX` deleted, `sanitizeDetail` signature, `DiagnosticDetail` retightened, `RecordOutcome.diagnostics` may no longer carry a `detail`) plus an in-place tightening of `observation-0.2.0.json`.

⚠️ **Maintainer decision (2026-08-21): versions stay at 0.9.20 on the branch.** `scripts/check-workspace-versions.mjs` requires root/cli/core to move together, so a core-only bump is not expressible; instead the bump owed for the core API break is recorded in the CHANGELOG heading and happens at the next actual publish (core before cli, per RELEASING.md).

Publish order at the next publish is **core before cli**, as always. Nothing is published by this PR.

## Goldens verified unchanged

- `packages/cli/tests/providers/antigravity-golden.test.ts` is **byte-identical to the branch base** (`git diff` against `feat/core-extraction` is empty) and all 15 tests pass, including G10 / G11 / G13 / G15. That is the direct consequence of dropping #933's `usageSignature` normalization: no antigravity key moved, so no parse-version or daily-cache bump was needed.
- The codebuff / grok / pi / zerostack / lingtai-tui bridge goldens intentionally changed (the raw path left the key) and are re-derived, not re-pasted.
- The copilot JetBrains goldens intentionally changed (sha256 → HMAC) and are recomputed under the pinned key.

## Verification

| Gate | Result |
|---|---|
| `@codeburn/core` suite | 37 files, **592 passed** |
| `codeburn` suite (locks excluded by script) | 215 files, **2588 passed**, 5 skipped |
| `npm run test:locks -w codeburn` | 2 files, **14 passed** |
| `tsc --noEmit` (core, cli) | clean |
| `npm run build -w @codeburn/core` | clean |
| `npm run check:workspace-versions` | agree at 0.9.20 (bump deferred to release) |

**Mutation checks**

| Mutation | Expected | Actual |
|---|---|---|
| `sourceRefFingerprint` returns the raw path | smuggling + bridge goldens fail | 9 core smuggling failures; 12 bridge-golden failures across all 5 files |
| copilot `createHmac` → `createHash` | property test fails | exactly the new property test fails |
| `isolateRecords` stops stripping caller details | diagnostics test fails | exactly the strip test fails |
| pinned test key changed | copilot literal goldens fail | 5 failures (the longhand re-derivations correctly track the key) |

**Real corpus, read-only** — `report --format json --from 2026-07-01 --to 2026-07-31`, cold `CODEBURN_CACHE_DIR` per run, base build (`86a6a54`) vs this build:

```
calls    43861  ==  43861
sessions 10013  ==  10013
cost     5965.901429949995  ==  5965.901429949995
estimatedCost 280.98596949999984 == 280.98596949999984
cacheHitPercent 99.3 == 99.3
models 16, projects 2142, daily 21, unpricedModels 1 — all equal
```

Byte-identical once the `generated` timestamp and the *current-month* budget fields are scrubbed. A base-vs-base control run at a third timestamp drifts in exactly those same fields (live sessions accrue between runs) and by a larger amount, confirming the drift is the clock and not the change. Every field in the July range is identical.

Closes nothing. #933 and #935 stay open for @ozymandiashh to review this reconciliation.


ozymandiashh:
Follow-up opened as draft: #1080.

While reviewing this reconciliation after merge, I reproduced a durable-cache regression that the green suite did not cover. On #1074's merged tree, a Copilot fingerprint change dropped the cached file entry whenever the JetBrains DB path still existed. That DB can retain the file while pruning individual conversations, so the cache may be the only remaining copy.

Exact pre-fix repro:

```text
same fingerprint:     2 calls / 15 output tokens / cached-only conversation present
changed fingerprint:  1 call  /  7 output tokens / cached-only conversation gone
persisted cache:       1 call
```

The #1074 invalidation test seeded only a DB-present turn, which explains why CI stayed green. #1080 adds a real Nitrite-like sequence with two calls (10 + 21 tokens): parse both, prune the 10-token row without deleting the DB path, rotate privacy key A → B, reload, and assert exactly 2 calls / 31 tokens throughout.

The draft fix:

- splits Copilot's stable local SHA identity from its public HMAC key, so DB-pruned rows can be re-keyed;
- adopts v5/v6/v7 durable history into cache v8, force-reparses present DBs, and preserves retry obligations on transient read failures;
- keeps former Copilot and source-ref keys as local-only aliases so already-sent calls are not uploaded again under new span IDs (`3` aliases survive full parse/cache reload; old ledger keys yield `0` unsent);
- restores the published `observation-0.2.0.json` byte-for-byte (`b616b250…`) and emits the tightened model contract as `observation-0.3.0` instead of mutating 0.2 in place.

Local validation on the PR head: Core 602/602; CLI 2623/2623 (5 infrastructure skips); lock suite 27/27; Core/CLI typechecks, workspace-version check, dist verification, and diff check all pass.

The PR body also documents the one unavoidable boundary: already-pruned rows written only by #1074's short-lived unreleased HMAC-v1 cache cannot be correlated cryptographically with released SHA rows because neither reply text nor the old identity was stored. Released caches and all caches written by #1080 migrate deterministically.

PR #1080, opened 2026-08-21
fix(core): preserve cache and schema compatibility after #1074

ozymandiashh:
## Summary

- Follow up #1074 with a durable Copilot cache migration that preserves DB-pruned conversations across parser-fingerprint changes and privacy-key rotation, retries transient reads without replacing good history, and reconciles current rows without double-counting.
- Preserve `sync push` idempotency when #1074 rotates public dedup keys: Copilot and the six source-ref providers retain their former public keys as host-only aliases. Raw legacy paths are persisted only in the local session cache, stay non-enumerable in report objects, and never enter observations or OTLP payloads.
- Restore the already-published `observation-0.2.0.json` contract byte-for-byte and move the new `model` / `pricingModel` bounds to an honest `observation-0.3.0` wire version.

## Reproduction and root cause

On the merged #1074 tree, a Copilot fingerprint change discarded the cached entry for every source path that still existed. JetBrains can prune individual rows while its Nitrite DB remains at that path, so the cache may be the only remaining copy of older conversations.

The pre-fix behavior repro was:

```text
same fingerprint:     2 calls, 15 output tokens, cached-only conversation present
changed fingerprint:  1 call,   7 output tokens, cached-only conversation gone
persisted cache:       1 call
```

Green CI did not catch it because #1074's invalidation test seeded only a DB-present row. The regression test here uses a real Nitrite-like JetBrains source with two distinguishable calls (10 + 21 output tokens), prunes the 10-token row while leaving the same DB path in place, rotates privacy key A to B, reloads from disk, and requires exactly two calls / 31 tokens at every phase.

The fix separates a stable local identity from the public keyed identity:

```text
released local identity: copilot:jb:<conversation>:<sha12(reply)>:<occurrence>
public v2 key:            copilot:jb:<conversation>:<hmac12(sha12)>:<occurrence>
```

That makes even a DB-pruned row deterministically re-keyable. Cache v8 adopts all valid durable Copilot files from v5/v6/v7, carries present paths, marks every carried file for a forced parse, and only clears that obligation after a successful reconciliation. An unreadable JSONL, chat session, or JetBrains DB throws into the durable retry path; QuickDesk's legitimate nullable durable arm keeps its existing behavior.

The source-ref migration had a second idempotency edge: old raw-path ledger keys and new HMAC keys produce different span IDs. A full parse/cache/ledger repro now reports:

```text
bridge aliases:                3
aliases after full parse:      3
unsent with old ledger keys:   0
```

The alias metadata is re-attached after pricing/project canonicalization, persisted locally, restored non-enumerably after a cache reload, and checked before a call is considered unsent. A follow-up parse-version suffix also invalidates caches already written by #1074, which had the new public keys but no ledger aliases.

## Compatibility boundary

The short-lived, unreleased #1074 HMAC-v1 cache keyed `HMAC(replyText)` directly and stored neither reply text nor the released SHA identity. Two already-pruned rows from the released and interim formats are therefore cryptographically uncorrelatable.

For the exact same-key interim fingerprint, v8 keeps an overlapping released-cache path authoritative (avoids deterministic duplicates), re-parses live DB rows, and preserves v7-only paths. This can omit an interim-only row that was already pruned on an overlapping path; a key-rotated interim cache is intentionally not guessed from an opaque fingerprint. Released v5/v6/v7 history and all caches written by this fix have the stable identity and are fully migrated.

## Schema compatibility

`@codeburn/core@0.9.20` already published `observation-0.2.0.json` with min-length-only model fields. Mutating that artifact in place made an archived 0.2 envelope fail under the same version. This restores the published artifact at SHA-256:

```text
b616b250f805ee2aa56d672dfc4dfd5c70a5a99911b3d82d77509f2f816211a7
```

The bounded identifier contract is emitted as `observation-0.3.0.json`, and drift/exports/architecture tests freeze both historical versions.

## Testing

- [x] `npm test` equivalent passes: Core 602/602; CLI 2623/2623 (5 infrastructure tests skipped by the suite)
- [x] Core build and Core/CLI `tsc --noEmit` pass
- [x] Lock suite: 27/27
- [x] `npm run check:workspace-versions`
- [x] `npm run verify-dist --workspace @codeburn/core`
- [x] `git diff --check`
- [ ] I have tested this locally against personal production data (the regression uses realistic committed/generated JetBrains cache fixtures)
- [ ] Full dashboard/app build was not required for this cache/schema-only change

### For new providers only

Not applicable.


iamtoruk:
Post-merge audit (deep, adversarial — the same pipeline every PR on this branch has gone through). First the good news, because there's a lot of it: **no privacy leak found** (a legacy path carrying a secret-like string never reaches observations, OTLP payloads, or exports; spread/structuredClone fail-open by dropping the alias), **#1074's invariants all hold** (rotation e2e passes and is stronger, no raw path in any observation dedup key, goldens still machine-independent), the alias reconstruction is exactly right for all six sourceRef key grammars, and your headline schema claim verified against npm itself — `npm pack @codeburn/core@0.9.20` yields `schemas/observation-0.2.0.json` byte-identical to your restored file (sha256 `b616b250…`), so #1074 really had mutated a published artifact and this was the correct fix. The migration tests also discriminate (four mutations, each caught).

That said, four things need a fix-forward, and two of them are final-merge blockers you couldn't have seen without cross-branch context we should have written down somewhere visible — that's on us, and here it is now:

**1. (blocker) `CACHE_VERSION = 8` collides with main's already-spent 8 — and main's 8 is a different shape.** `origin/main` commit `0bfdfa37` (2026-08-16) defined v8 as per-provider *shard directories*, and main now sits at `CACHE_VERSION = 9` (v9 = provider×month shards). The #946 landing branch (kelchm) is also at 9, with a fingerprint-sentinel carry-forward whose key formats `rekeyCopilotDurableCache` would silently no-op on. At the final branch→main merge, main's loader won't adopt this branch's v8 single file — silently dropping exactly the DB-pruned Copilot history this PR exists to preserve. Suggest: take **10** (clear of both lineages), and note the ladder in the comment the way daily-cache does (main 8/9 spent, #946 at 9).

**2. (blocker) `OBSERVATION_SCHEMA_VERSION = '0.3.0'` collides with `origin/harden/core-detectors`** (tip `1d452792`), which independently claims 0.3.0 **with an incompatible wire shape** (FingerprintHex 16→32, `dedupKey`→`callRef`, new envelope fingerprint-algorithm/host-key-id fields). Same filename `schemas/observation-0.3.0.json` on both sides; a merge silently picks one file's content instead of conflicting. Needs coordination with @laulpogan — either this branch takes 0.3.1/0.4.0, or the two 0.3.0 designs get reconciled into one before either lands further.

**3. (medium) Reproducible permanent double-count in one interim case.** A key-rotated #1074-interim v7 cache alongside a released v5: parse under the current key yields tokens `[10,10,21,21]` — both the DB-present and the DB-pruned row doubled, forever, in a store that never deletes. (Repro: seed `session-cache.v5.json` with legacy SHA keys + `session-cache.v7.json` at a fingerprint computed under a *different* key, parse under the current key.) The PR body predicts this case "can omit an interim-only row"; the actual behavior is 2× inflation. Only branch developers can trigger it today, but the wholesale carry-forward has no guard for unrecognized key shapes at still-present sources either. Suggest: correlate the interim cache or refuse to adopt it, rather than unioning it.

**4. (medium) Unbounded retry on permanently-unreadable durable sources.** `providers/copilot.ts:785/792/801` now throw where they returned null — but `readSessionFile` returns null for **oversize (>128MB)** and **empty** files too, so a 0-byte or oversize JetBrains DB sets `needsReparse` and bypasses the `failed` marker: re-read, re-thrown, re-warned on every run forever (verified across two consecutive runs). Transient and permanent need distinguishing.

Low-priority, take or leave: alias sets grow without bound across rotations (fine for the 7-day ledger, but persisted forever and O(rows×aliases) on every push — a cap or expiry would do); the non-enumerable descriptor at `parser.ts:1293` is `configurable:false` while `bridge.ts:137`/`parser.ts:1782` use `true` (a future re-definition throws in strict mode); and `schema-drift.test.ts` spot-checks the restored 0.2.0 file rather than pinning its content hash — one sha256 assertion closes that hole given we now know the npm artifact's exact bytes.

Housekeeping note: `feat/core-extraction` now has branch protection (merges restricted to the maintainer, the 9 CI checks required). Nothing personal — the same rule now applies to everyone including PRs we author ourselves, and today made a good case for making review mechanical rather than social in both directions: your #1086 merge picked up the pre-fix head and we had to chase it with #1087. Open the fix-forward PR whenever ready and we'll run it through the same review pipeline — happy to share the full audit transcript if useful.

PR #1085, opened 2026-08-21
fix(cache): version-suffix provider result caches

### LINKED ISSUE #1082: codex-results.json is not version-suffixed — mixed-version installs thrash a full re-parse every run
Found by upgrade-simulation review of #1078: daily-cache files are version-suffixed (daily-cache.v23.json) so binaries coexist, but codex-results.json is a single unsuffixed file with an internal version field. A v10 and a v11 binary (e.g. desktop app pinned to an older CLI alongside an upgraded brew CLI, or a rollback in flight) clobber each other's file and each re-parses all codex sessions on EVERY invocation — measured 9.2-9.4s vs 5.3s warm (+75%). Correctness is fine on both sides; this is pure cost. Fix: suffix the filename with its version like the daily cache.

avs-io:
## Problem

Daily and session caches already own a version-suffixed file so an old desktop binary and a newer CLI cannot clobber each other. The three per-provider result caches still used one unsuffixed filename with an internal version field. A v10 and a v11 binary rewrote the same `codex-results.json` (and the Cursor / Antigravity siblings) on every run and each re-parsed its whole corpus — measured 9.2–9.4s vs 5.3s warm on Codex.

Closes #1082.

## Root cause

`codex-results.json`, `cursor-results.json` and `antigravity-results.json` stored the schema version inside the file. Mixed-version installs shared one path. Daily-cache and session-cache already avoided this by minting `*.v<n>.json`.

## Change

- Write `*-results.v<n>.json` for Codex (v11), Cursor (v6) and Antigravity (v5).
- Never write or delete the unsuffixed file. Old binaries still own it.
- Adopt a matching-version unsuffixed file once when the versioned file is absent. A mismatched version is ignored.
- If a versioned file is present, it is the only source.

## User impact

Mixed-version installs (pinned desktop + upgraded brew CLI, or a rollback) keep their own warm cache. First run after upgrade adopts a same-version unsuffixed file so the corpus is not re-parsed just to rename it. Correctness is unchanged.

## Preservation

Does not bump any cache version, parse version, or pricing. Does not touch `models.ts`. Does not suffix `antigravity-statusline.jsonl` (event log, not a versioned result cache). Does not delete legacy files.

## Testing

`npx vitest run tests/result-cache-suffix.test.ts tests/cache-directory-switch.test.ts tests/codex-cache-invalidation.test.ts tests/providers/cursor.test.ts tests/codex-pricing-1075.test.ts tests/providers/codex-resume.test.ts tests/session-cache.test.ts tests/parse-workers.test.ts` — 140 passed.

`npx tsc --noEmit` — clean.


ozymandiashh (approved):
Reviewed the current head and its merge with current main. Version-suffixed provider caches isolate incompatible generations; the mixed v10/v11 repro preserved the v11 cache after a legacy rewrite. The targeted suite (141/141), typecheck, and all CI checks pass. No blocking issue found.

Issue #1083, opened 2026-08-21
core-extraction: port the #1078 codex pricing fixes (reasoning double-count + cache-write bucketing)


iamtoruk:
The feat/core-extraction branch needs #1078's fixes re-ported: pricing is centralized at packages/cli/src/pricing-pass.ts (~line 26: outputForCost = outputTokens + reasoningTokens, no provider carve-out) and packages/core's codex decode still computes totals with reasoning added. Port as a per-provider predicate (billableOutputTokens with the claude+codex inclusive set from main's src/models.ts) at the pricing-pass seam, plus the cache_write_input_tokens read + explicit-rate guard in the core codex decoder, plus the codex parse-version and cache bumps in the branch's own ladder (branch daily-cache is at 26). Without this the final branch→main merge would revert the fix.

avs-io:
Shipped on `feat/core-extraction` by #1089 (`0531cb6`). GitHub did not auto-close because the merge was not to the default branch.

Issue #1079, opened 2026-08-21
codex tok/s double-counts reasoning tokens in generated-token throughput


iamtoruk:
Split from #1075/#1078: the throughput path has the same reasoning-inclusion bug as the cost path fixed in #1078 — `activeGeneratedTokens`/`taskGeneratedTokens` in src/providers/codex.ts and `generatedTokens` in src/codex-throughput.ts:403 add reasoning tokens to output that already contains them, so displayed tok/s reads high (~+34% on a corpus where reasoning is ~1/3 of output). Display-only, no cost impact. Fix means rewriting ~7 existing throughput assertions; use `billableOutputTokens('codex', …)` from #1078 so the definition can't drift from the cost path.

PR #1086, opened 2026-08-21
feat(cli): rehome cline-cli and codex throughput (landing #940)

iamtoruk:
Rebuilds #940 by @ozymandiashh as a clean landing branch off the current `feat/core-extraction`. **#940 stays open** — this is its work, rebased onto a branch that has moved a lot since (#921-#937, #942, #1074), with the review's findings resolved and one design question settled by a maintainer decision.

Two commits, one per half.

## 1. cline-cli provider rehome

#940's cline-cli half was reviewed as excellent and is ported as-is: dedup keys identical to `main` character for character, all 34 upstream test cases carried, clean core/host split, and correctly kept separate from the shared vscode-cline tier.

Two changes on top:

- **Post-#1074 conventions.** The new core provider's `observations.ts` routes the model through `normalizeModelIdentifier` at the observation boundary like the other 34, respects the `ModelIdentifier` bound, and carries measured cost the way its siblings do instead of through a cast. Content-smuggling coverage matches the current suite.
- **Review fix: no un-asked-for billing change.** #940 routed the decoded `fetch_web_content` count into the pricing pass, which bills $0.01 each. `main` reports the count on every per-message call but passes a literal 0 as `calculateCost`'s sixth parameter (`src/providers/cline-cli.ts:299`; only the rollup path at `:349` emits 0, having no tool blocks to count). So the count always rides on the call (`packages/cli/src/providers/cline-cli.ts:76`) and the suppression sits where billing happens: `pricing-pass.ts:37` passes 0 web-search requests for cline-cli. `fetch_web_content` is a page fetch, not a billable provider-side search. Its test proves both halves — count emitted on the estimated and metered call, estimated cost equal to an identical session with no fetch.

Also restores the docs #940 dropped: `docs/providers/cline-cli.md`, the provider index row, the README provider table row, the `docs/providers/cline.md` cross-reference, and a CHANGELOG entry.

## 2. Codex throughput rehome — maintainer decision: main's resume design

Rehomed against `main` as it ships **today** (post #1078 and the codex polish PR), not against #940's snapshot.

**The resume question is settled in favour of main's design.** Timing state is snapshotted **only at `task_started` boundaries**, where every per-task accumulator is provably empty, and a task's calls are buffered until its window is known. **No recorded call is ever mutated after emission.** #940's alternative is dropped in full: the `lastCompleteLineOffset`-based timing resume, the task window threaded through serialized state, and `applyCodexTimingPatches`.

The branch's existing Phase-4 token-decode resume (`CodexDecodeState`, any-offset, round-trip proven) **stays untouched**. Marrying main's boundary snapshots to the split decode layout needed exactly one adaptation, because core decodes records and never sees bytes:

- the decoder reports its last `task_started` as a `checkpoint` (record index + call count + state);
- the CLI maps that index to a byte offset, stores it with a `callCount`, and on resume replays only the calls before the boundary — the still-open task re-derives from its own `task_started`;
- a pass that crosses no boundary keeps the previous one; a cold decode of a file with no `task_started` at all falls back to end-of-file with `taskOpen: false`, so a `task_complete` whose window this pass never saw attributes **nothing** rather than spreading a whole task's active time over part of its tokens.

**fad84662 review fixes restored** (all three were reverted in #940):

| fix | state |
|---|---|
| discovery fast path — `getCachedCodexProject` short-circuits before `isValidCodexSession` | already present on the branch (`packages/cli/src/providers/codex.ts:121`), verified unchanged |
| `duration_ms` precedence — payload-level wins over any nested one | `timingDuration ?? timingNumber('duration_ms')` per main (`packages/core/src/providers/codex/decode.ts`), + main's regression test |
| dashboard — `MIN_WIDE` stays **90** with a `showTps` gate on the Tok/s column, not 130 | `packages/cli/src/dashboard.tsx` |

Also ported: 7fe43209's fork-suppressed-`task_started` regression test, and the depth-1 `payloadString` helper (1d36f444/497f6556) the branch lacked, with both of main's nested-key tests.

**Dropped as superseded by #1074:** #940's carried #933 commit (`f5acac13`, model normalization + sourceRef fingerprinting) is not re-applied at all. #1074 landed the reconciled #933+#935, which also erases the old review's findings 1-2 (antigravity goldens, unbumped parse versions).

**Kept out of scope:** main's codex pricing work (`billableOutputTokens`, #1078) is tracked separately as #1083 and is deliberately absent — this PR contains no pricing changes. The tok/s reasoning question (#1079) is likewise untouched: this is a faithful port of main's current generated-tokens definition. Main's other unported parser changes (custom-tool transport, exact token counts and MCP names on oversized lines) are not dragged along either, which is why cost, calls and tokens are provably unchanged.

## Versions

| cache | value | why |
|---|---|---|
| `CODEX_CACHE_VERSION` | **12** | main's ladder reached 11 (#1078); 12 keeps the two payload shapes from ever being read as current by the other line. Codex parse version bumps in lockstep so `session-cache.json` cannot keep serving timing-less turns. |
| `DAILY_CACHE_VERSION` / `MIN_SUPPORTED` | **27** | cline-cli is a new provider whose historical sessions produce rollup data no older cache ever contained. |
| env | `CLINE_DIR`, `CLINE_DATA_DIR`, `CLINE_SESSION_DATA_DIR` added to the test env-isolation CLEARED list, `PROVIDER_ENV_VARS`, and the provider-env declaration guard. |

Both bumps carry revert-proof invalidation tests: revert `DAILY_CACHE_VERSION` to 26 and the daily test fails; drop the `-active-timing-v1` suffix and the codex test fails.

## Gates

- `packages/cli` suite: **225 files, 2678 passed**, 5 skipped, 0 failures
- `npm run test:locks`: **27 passed**
- `@codeburn/core` suite: **38 files, 616 passed**
- `tsc --noEmit`: clean in both packages

## Real-corpus A/B (read-only, scratch `CODEBURN_CACHE_DIR`)

Base branch vs this branch, `--from 2026-01-01 --to 2026-08-20` (today excluded so live usage cannot drift the comparison), 1397 codex rollouts on disk:

- **Full report JSON is identical** across every provider — every cost, call count and token bucket. The only fields that moved are `/plan` and `/plans/claude` month-to-date spend, and a base-vs-base rerun 36s apart drifts in exactly those same six fields, i.e. live usage, not code.
- **Codex model totals byte-identical**: 1326 sessions, per-model calls/cost/freshInput/cacheRead/cacheWrite diff clean.
- **Timing lands on real data**: 1302 of 1328 model slices now carry active timing — GPT-5.5 37.7 Tok/s over 490,337 active seconds, gpt-5.6-sol 43.2, gpt-5.6-luna 53.5, Codex Auto Review 23.3.
- Warm-run wall time within noise (base 10.2/11.4s, branch 11.5/11.5s over interleaved rounds).
- **cline-cli finds nothing here**: this machine has no `~/.cline`, so the new provider contributes zero rows — consistent with the identical totals above. It is exercised by its 35 ported tests instead.

## Adversarial-validation fixes (post-review, commits 3-4)

The checkpoint adaptation survived every attack (off-by-one, append splits, no-mutation, resume interaction), and three findings were fixed on top:

1. **cline-cli web-search faithfulness** (above) — the original instruction was itself corrected: main zeroes the pricing argument, not the emitted metric.
2. **`taskOpen` gate had no behavioral coverage** — deleting it left all 613 core tests green, because the existing cases pinned the flag's value, never its consequence. New case splits a task with two `token_count`s so the resumed pass holds a real call: without the gate its `task_complete` takes the whole 6s window for half the tokens. Verified failing with the gate removed.
3. **custom-tool transport** — the timing branches spoke only `function_call`/`function_call_output`; main and the ported `codex-throughput.ts` also count `custom_tool_call`/`custom_tool_call_output`. That is 11,390 tool pairs across 134 real rollouts whose wait was counted as active: GPT-5.5 506,373 → 490,337 active seconds, gpt-5.6-sol 66,194 → 54,120. Codex totals remain byte-identical to base.

The validator's append-split fixture landed as a permanent regression test (`packages/cli/tests/codex-checkpoint-resume.test.ts`): cold == warm at every line split point, a three-pass variant, and idempotence. It kills both resume-point mutations (untruncated `callCount`, naive EOF resume).

## Known gaps (deliberate, not regressions)

The branch's codex decoder still lacks main's oversized-line token-usage and MCP-name extraction, and does not yet name `custom_tool_call` tools (only their wait time counts). Two ported tests were adapted to the branch's record vocabulary rather than importing behaviour this PR is not porting.


ozymandiashh (approved):
Reviewed the exact head and its synthetic merge with feat/core-extraction. The Cline rehome preserves measured-versus-estimated billing semantics, and the Codex task-boundary resume path matches a cold decode without changing report totals. Full Core and CLI suites, lock tests, typechecks, targeted cache/resume tests, and a real-corpus A/B all pass. I found no blocking issue.

PR #1098, opened 2026-08-22
fix(desktop): say why the PR tab is empty

avs-io:
## Problem

Desktop Pull requests on Today/week/month looks broken: empty copy, no hint. On this Mac the extractor works — `status --format menubar-json --period all` returns 73 rows; today/month omit `pullRequests`. Default period is Today.

Closes nothing yet (no issue number). Class: empty-state / period honesty.

## Root cause

Empty copy did not name the selected period. Sessions only appear when a transcript contains a PR URL. Hermes/Grok work this week never writes those URLs.

## Change

- Period-aware empty note.
- If All has N rows, say so and point at the period control.
- No `buildPrAttribution` change. Default period unchanged.

## User impact

Today still empty if no linked sessions today. The tab explains why and how to see All.

## Preservation

No aggregator / gh-invented links.

## Testing

`npx vitest run renderer/sections/PullRequests.test.tsx` — 17 passed (including Today-empty + All-has-2).

### DIFF
--- a/app/renderer/App.test.tsx
+++ b/app/renderer/App.test.tsx
@@ -268,7 +268,7 @@ describe('App shortcuts', () => {
     expect(await screen.findByText('No sessions in this range yet.')).toBeInTheDocument()
 
     fireEvent.keyDown(document, { key: '3', ...chord })
-    expect(await screen.findByText(/PR links are captured as sessions are parsed/)).toBeInTheDocument()
+    expect(await screen.findByText(/No sessions in Last 30 days mentioned a pull request URL/)).toBeInTheDocument()
 
     fireEvent.keyDown(document, { key: '4', ...chord })
     expect(await screen.findByText('Cost flow · model → project')).toBeInTheDocument()
--- a/app/renderer/App.test.tsx
+++ b/app/renderer/App.test.tsx
@@ -485,6 +485,27 @@ describe('App shortcuts', () => {
     expect(screen.getByText('30D')).not.toHaveClass('on')
   })
 
+  it('names a selected custom range on the Pull requests empty note', async () => {
+    render(<App />)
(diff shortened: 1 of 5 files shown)


iamtoruk:
Reviewed in depth at `b095091a` (including your second commit — good self-catch on the `period` threading; my review found the same bug in the first commit independently, so that fix is confirmed necessary and correct). Verified: perf is clean — the lifetime probe rides the resident serve cache once per mount with the 5s coalescing layer, no spawn-per-poll; scope is exactly as claimed; full app suite 536/536 + tsc clean on your head. Two fixes before merge, one optional:

**1. Em dash in the rendered copy.** The `widerHint` string uses `—` ("…pull requests — switch the period control…") in the same file whose `spanLabel` comment states the rule ("never an en/em dash, per repo copy rules"). It's the only rendered em dash in `app/renderer` — split into two sentences or rephrase.

**2. `periodLabel` ignores a custom range.** `PrEmptyNote` computes `PERIOD_LABELS[period]` unconditionally, but when a custom date range is active, `period` is stale while `range` drives the query — so the empty note can claim "No sessions in Today…" for a custom-range view. That's the exact class of dishonesty this PR exists to fix, and `App.tsx:518` already has the pattern to copy: `customRange ? rangeLabel(customRange) : PERIOD_LABELS[period]` (`rangeLabel` is exported from `components/TopBar.tsx`). Your `canProbeWider` guard already handles `range` correctly — the label should follow the same guard. Please add a test passing `range` (none currently does).

**3 (optional):** one test for the `canProbeWider=true, widerCount=0` suppression branch (fresh install: today empty AND lifetime empty) — it's currently only covered implicitly via the lifetime-period path.

Everything else is exactly right — the empty state finally tells the truth, and "Life." matching the real tab label was verified before anyone could call it a typo.

PR #1099, opened 2026-08-22
fix(doctor): treat Buzz as a launcher, not a second Codex tree

avs-io:
## Problem
Buzz (and similar harnesses) sit on top of Codex. Adding a Buzz usage parser, or walking `~/.buzz/.codex` as a second Codex home, would count the same billed seat twice.

## Root cause
`codeburn doctor` only knows providers. A nest with no usage store still looks like “nothing found.” Codex discovery will happily parse any directory you hand it, including a launcher nest.

## Change
- Doctor lists Buzz (and Grok Bot when `~/.grok` is the real store) as **launchers** — no session count.
- Codex `discoverSessions` returns [] for a home nested under `~/.buzz` when a distinct primary Codex home exists.
- Sole Codex home under `.buzz` is still counted.

## User impact
`codeburn doctor` tells you Buzz is a launcher billed via Codex. Spend stays on the Codex row.

## Preservation
No Buzz parser. No rates. No ACP→Buzz. Hermes/Grok/OpenClaw providers unchanged.

## Testing
`npx vitest run tests/launcher-homes.test.ts tests/doctor.test.ts` — 29 passed.


iamtoruk:
Early feedback while this is still draft — the mechanics are clean (add-only doctor schema, env guard green, good three-state coverage, tsc/suite green), but two things must change before ready-for-review:

**1. The premise needs evidence.** Nothing in the repo, docs, or issue history establishes that Buzz actually sets `CODEX_HOME` to a nest under `~/.buzz` — the only prior Buzz reference is the Hermes ACP transport comment, which says nothing about Codex homes. And nothing shipped today walks `~/.buzz/.codex` as a second tree: `createCodexProvider()` has one production call site with no args, so the described double-count can only occur when a user's shell exports `CODEX_HOME` at the nest while `~/.codex` also exists — a narrower scenario than the PR body describes. A real report or a repro from an actual Buzz install would settle it.

**2. The suppression can silently drop real, unique sessions — and the test proves it.** `isNestedLauncherCodexHome` fires on a bare `existsSync(primaryDir)` with no content-overlap check, and `tests/launcher-homes.test.ts:64-89` writes two *different* rollouts (`rollout-primary.jsonl` vs `rollout-buzz.jsonl`) and asserts only the primary survives — i.e., the expected behavior is data loss. A user whose `CODEX_HOME` points at a Buzz nest and who also has a stub `~/.codex` (a `codex login` leftover is enough) would have all their real Buzz-driven usage silently hidden. Dropping real usage is strictly worse than the double-count this prevents. Fix: only suppress sessions whose ids/filenames actually overlap the primary tree — distinct histories must never be dropped.

Smaller notes: the `codexDir === undefined` proxy for "production singleton" is an unenforced contract (a future caller passing an explicit nest path gets silent `[]` — worth a guard or test); path containment uses `resolve()` not `realpath()` (untested symlink behavior in a feature that's entirely about same-vs-distinct homes). The Grok Bot doctor labeling half is fine as-is — no suppression there, no double-count vector exists.

iamtoruk:
Re-reviewed the rework (Qwen draft; both load-bearing findings re-verified against the diff by maintainer). Credit where due: the blanket-skip from the first draft is gone — unique nest rollouts are kept, which resolves the original data-loss objection *in the code that runs*. But:

1. **The overlap filter isn't wired into production.** Verified: `filterOverlap` requires `opts.primaryDir`, no production call site passes opts, and `defaultLauncherRoots()` is exported but unused in this diff. As shipped, default discovery behaves exactly as before this PR — the dedup only exists for a hypothetical second factory nobody constructs. Either wire the nest detection into real discovery (and show the test exercising the default path), or this PR is doctor-notes-only and the description should say so.

2. **Dedup is basename-only.** `listRolloutBasenames` compares filenames, not session identity: a nest-only session whose basename collides with an unrelated primary file gets dropped (dollar loss); a duplicated session that was renamed double-counts. Codex rollout names embed timestamps+uuid so collisions are unlikely — but 'unlikely' isn't the bar for a dedup that exists to protect dollars. Dedup on the session id inside the file (first line) or at minimum document why basename identity is sufficient for rollout files specifically.

3. Minor: the doctor 'LAUNCHER (billed via Codex)' note fires on `~/.buzz` existing, without checking a Codex store actually exists or that any nested tree was examined — fine as a heuristic, but the wording implies more than was verified.

Original objection #1 (unproven duplicate premise) is now moot in the current shape — precisely because nothing is skipped by default. Resolve finding 1 and the premise question returns; answer it then with session-identity evidence.

PR #1100, opened 2026-08-22
payload: add-only stale marker for read-only stale menubar serves

### LINKED ISSUE #771: menubar payload: data-freshness marker for stale readOnly serves

iamtoruk:
## Summary

Closes #771. The warm-refresh gate (#743) serves the prior complete snapshot when the refresh lock times out or is unavailable — correct and fail-closed, but the payload's only time field was the render timestamp, so a stale serve was indistinguishable from a fresh parse.

Since #771 was filed, `src/parser.ts` gained `isSessionHydrationComplete()` (PR #937) which already returns `false` exactly when the last completed parse in-process was a read-only serve that had to skip or stale real on-disk changes. This PR threads that signal through to the payload:

- `src/menubar-json.ts`: new optional top-level `stale?: boolean` on `MenubarPayload`; `buildMenubarPayload` gains a trailing optional `stale` param and sets `payload.stale = true` only when passed `true` — never emitted as `false`.
- `src/usage-aggregator.ts`: `buildMenubarPayloadForRange` passes `isSessionHydrationComplete() ? undefined : true`.
- `app/renderer/lib/types.ts`: mirrors the field (add-only, per this repo's CLI-app payload contract).
- `mac/Sources/CodeBurnMenubar/Data/MenubarPayload.swift`: mirrors the field in the Codable data model (`let stale: Bool?`, decoded via `decodeIfPresent` so older CLI payloads that never emit it decode fine as `nil`).

Absence of the field always means "assume fresh," including for payloads from a CLI version that predates it — the fail-safe direction for a marker like this.

## Follow-up (not in this PR)

The macOS menubar UI has no visible staleness indicator yet — this PR is data-model only for the Swift side. A subtle indicator (small dot/tooltip near the header, matching the pattern of `#1096`'s cold-state signaling) is a reasonable small follow-up once there's a view spot picked for it; scoped out here to keep this change minimal and reviewable.

## Test plan

- [x] `tests/menubar-json.test.ts`: new cases assert `stale: true` when passed, and `stale` is `undefined` (not `false`) on a normal build. Verified revert-proof — reverting the `src/menubar-json.ts` + `src/usage-aggregator.ts` changes fails the new `stale:true` test.
- [x] `tests/usage-aggregator.test.ts`: new assertion that a normal (env-isolated, always-complete) parse leaves `payload.stale` undefined.
- [x] `npm run test` (full suite, excluding the separately-gated lock tests): 225 files / 3141 tests passed, 2 files / 5 tests skipped (pre-existing skips, unrelated).
- [x] `tsc --noEmit` clean on both the CLI (`tsconfig.json`) and the desktop renderer (`app/tsconfig.json`).
- [x] `swift build` and `swift test --filter MenubarPayloadCombinedTests` clean in `mac/`.

### DIFF
--- a/app/renderer/lib/types.ts
+++ b/app/renderer/lib/types.ts
@@ -127,6 +127,9 @@ export type ClaudeConfigSelector = {
 
 export type MenubarPayload = {
   generated: string
+  // Optional: older CLIs omit it. Present and true only on a stale read-only
+  // serve; absent otherwise. Absence must always be read as "assume fresh."
+  stale?: boolean
   current: {
     label: string
     cost: number
--- a/mac/Sources/CodeBurnMenubar/Data/MenubarPayload.swift
+++ b/mac/Sources/CodeBurnMenubar/Data/MenubarPayload.swift
@@ -4,6 +4,11 @@ import Foundation
 /// `current` is scoped to the requested period; the whole payload reflects that slice.
 struct MenubarPayload: Codable, Sendable {
     let generated: String
+    /// Present and `true` only when this payload was assembled from a
+    /// read-only stale serve. Absent — never `false` — on a fresh payload;
+    /// absence must be read as "assume fresh," including for payloads from
+    /// a CLI version that predates this field.
+    let stale: Bool?
     let current: CurrentBlock
     let optimize: OptimizeBlock
     let history: HistoryBlock
@@ -15,8 +20,10 @@ struct MenubarPayload: Codable, Sendable {
          optimize: OptimizeBlock,
          history: HistoryBlock,
          combined: CombinedUsage?,
-         claudeConfigs: ClaudeConfigSelector? = nil) {
+         claudeConfigs: ClaudeConfigSelector? = nil,
+         stale: Bool? = nil) {
         self.generated = generated
+        self.stale = stale
         self.current = current
         self.optimize = optimize
         self.history = history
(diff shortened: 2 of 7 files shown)


iamtoruk:
Pushed a fix for the review's stale-global-read bug (commit e19c099b).

**Bug:** `isSessionHydrationComplete()` reads a module-level global in `parser.ts` that's only safe to read immediately after the parse it describes. The prior commit read it at the very end of `buildMenubarPayloadForRange`, after several `await`s — including, on the claude-config-scoped branch, a second `parseAllSessions` call (365-day history backfill) that runs *after* the parse producing the headline data. That let the history parse's hydration outcome silently overwrite the headline parse's.

**Fix:** capture `isSessionHydrationComplete()` into a local immediately after each branch's primary parse resolves (the scoped branch's Claude parse, or `buildDurablePeriod` on the non-scoped path), and thread that captured value to `buildMenubarPayload` instead of re-reading the global at the bottom. Also closes the cross-request race the review flagged (web-dashboard SWR / parallel MCP calls), since a synchronously-captured local can't be overwritten by another request's later parse.

**New test** (`tests/usage-aggregator-freshness.test.ts`): mocks `parseAllSessions`/`isSessionHydrationComplete` so the primary parse reports complete and a second (bystander) parse reports incomplete, then asserts the payload reflects the primary outcome. Verified it fails against the pre-fix code (`expected true to be undefined`, i.e. the bug reproduces) and passes against the fix.

Verified: full suite (226 files / 3142 tests passed, 2 files / 5 skipped — pre-existing), `tsc --noEmit` clean, `swift build` + `swift test --filter MenubarPayloadCombinedTests` clean.

PR #1103, opened 2026-08-22
fix: stamp a build/schema fence on session cache entries

### LINKED ISSUE #1095: session cache entries carry no build/schema fence — foreign-build caches mix silently
From vidoluco's #946 validation postmortem: their unpublished local build wrote synthetic calls into a shared cache dir; the published build read them back and served them (cost double-counted, 1.83x) because session-cache.v9 entries are fingerprinted on the SOURCE FILE only, with no record of the build/semantics that produced them. Two builds with different cost semantics sharing CODEBURN_CACHE_DIR mix silently and keep matching on fingerprint. Their words: 'a build or schema fence would have turned three days of hunting into a cache miss.' Fix idea: stamp entries (or the envelope) with a semantics token (e.g. the PROVIDER_PARSE_VERSIONS hash already exists — but their build shared it; consider a build-id or an explicit cache-semantics version distinct from parse versions) and treat mismatch as a cache miss for that provider.

iamtoruk:
## Summary

- Fixes #1095: two builds with different cost/accounting semantics sharing `CODEBURN_CACHE_DIR` could mix silently, because `session-cache.v9` entries were fenced only by source-file fingerprint + `PROVIDER_PARSE_VERSIONS` — and a fork that shared the same parse-version strings but different cost math (vidoluco's #946 postmortem) matched fingerprints and got served.
- Adds `CACHE_SEMANTICS_REVISION` (bumped when cost/accounting semantics change without a matching `PROVIDER_PARSE_VERSIONS` bump) and `computeCacheSemanticsToken()`, a hash of `PROVIDER_PARSE_VERSIONS ∪ DAILY_CACHE_VERSION ∪ CACHE_SEMANTICS_REVISION`, stamped on the v9 envelope on every save.
- Load-time behavior:
  - Envelope with **no** `semanticsToken` (pre-fix cache): loads unchanged, no forced re-parse. The next save stamps the token — adoption happens once, going forward, with zero special-casing.
  - Envelope with a **matching** token: no change.
  - Envelope with a **mismatched** token: every non-durable provider's cached files are dropped (`{}` — hard cache miss, the foreign numbers are never surfaced from `loadCache()`); durable providers (`copilot`, or any section with `durable: true`) keep entries whose source no longer exists (unrecoverable history) but force a full re-read on entries whose source still exists (mirrors the existing durable carry-forward already used for `envFingerprint` mismatches in `getOrCreateProviderSection`).
- Deliberately **not** a `CACHE_VERSION` or `DAILY_CACHE_VERSION` bump — this changes envelope *adoption policy*, not the shard layout or `daily-cache.v26` day-record shape, so there's no migration path to extend.

## Test plan

- [x] New `tests/session-cache-semantics-fence.test.ts`: same-token adoption, missing-token adoption + stamp, mismatched-token re-parse (durable carry-forward + non-durable drop, revert-proof), two-writer alternation (build A's token never lets build B's mismatched cost data leak into a caller-visible value, and republishing always re-stamps the local build's real token).
- [x] All four new tests verified to **fail** against pre-fix `src/session-cache.ts` (only the test file reverted-tested; confirmed each fails without the fence and passes with it).
- [x] Full suite: `npm test` — 226 files / 3143 tests passed, 5 skipped (pre-existing lock-suite/e2e skips, unrelated).
- [x] `npx tsc --noEmit` — clean.
- [x] `npm run verify:upgrade` — PASSED, parity `ok`. Confirms a cache written by the last published CLI (0.9.20, pre-fence, no `semanticsToken`) upgrades cleanly with no forced global re-parse and no payload drift beyond the pre-existing, already-documented codex/#1075 and grok/#1015 repricing.

🤖 Drafted with `cline-pass/qwen3.8-max` via local gateway; reviewed and verified by the operator.

### DIFF
--- a/src/session-cache.ts
+++ b/src/session-cache.ts
@@ -4,6 +4,7 @@ import { createHash, randomBytes } from 'crypto'
 import { join } from 'path'
 
 import { getCodeburnCacheDir } from './cache-dir.js'
+import { DAILY_CACHE_VERSION } from './daily-cache.js'
 import type { ToolCall } from './types.js'
 
 // ── Types ──────────────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


iamtoruk:
Closing without merge, maintainer decision: cache hygiene across unofficial builds is not CodeBurn's problem to solve — the product guarantees correctness for its own released builds, and anyone running a source-built experiment against the official cache dir owns that risk. The adversarial review also showed this draft couldn't have caught the motivating incident anyway (same-constants forks hash identically; pre-fence writers strip the token every write) while regressing per-provider cache invalidation into global wipes. The review findings are preserved on this PR for anyone revisiting the space. Standard practice for contributors: point experimental builds at their own CODEBURN_CACHE_DIR.

PR #1118, opened 2026-08-23
Price and label OrcaRouter gateway model ids as the model they route to

avs-io:
## Problem

OrcaRouter sessions show $0. The gateway writes route ids (`orcarouter/auto`, `orcarouter/fusion`, `orcarouter/fusion-flash`, `orcarouter/fusion-mini`) and nested upstream ids (`orcarouter/deepseek/deepseek-v4-pro`). Those were not routing wrappers, so they never hit a LiteLLM row.

## Change

Rebased the community mapping from #1058 onto current `main` (that PR was CONFLICTING). Files only: `src/models.ts`, `tests/models.test.ts`, `CHANGELOG.md`, `README.md`.

- Peel `orcarouter/` like the other routing wrappers.
- Alias `orcarouter/auto` → `claude-sonnet-4-5` (fast Sonnet estimate; the smart route rotates).
- Alias `orcarouter/fusion` / `fusion-flash` / `fusion-mini` → `openai/gpt-oss-120b` (verified live 2026-08).
- Nested upstream spellings price at the exact LiteLLM row.
- Unknown vendor nested inside the route still fails closed.

## User impact

An OrcaRouter session reports the upstream model's spend instead of $0. Route ids keep a branded name rather than the current upstream's display name.

## Preservation

No parser, provider, doctor, cache, or Hermes changes. Existing wrappers (`cmd/`, `antigravity/`, `cp/`, …) are untouched. Unknown vendors stay unpriced.

## Testing

`HERMES_HOME=/tmp/cb-cleanhome npm test` — 3170 passed, 5 skipped.
`npx tsc --noEmit` — clean.

Unit tests cover pricing, fail-closed, and display labels for the route ids.

Thanks @XiaoHuo888-hue.


iamtoruk:
Reviewed (MiniMax draft; findings verified against the diff by maintainer). The peel itself is sound — anchored, single-segment, nested upstreams hit exact LiteLLM rows, unknown vendors fail closed (all confirmed by your tests).

Three blockers before this can merge:

1. **Missing `DAILY_CACHE_VERSION` bump.** Verified: the diff touches no cache-version constant. This PR is dollar-moving (rows that priced $0 now price real money) — warm daily caches will keep serving $0 for OrcaRouter days until unrelated invalidation. Main is at 27; this takes 28.

2. **`orcarouter/auto` → `claude-sonnet-4-5` is contradicted by your own comment**, which says the route currently lands on a Qwen/Llama flash model. Sonnet rates would overprice that by roughly 3–30×. This repo has been burned by alias guesses before (16×-over and 274×-under, #1058). Either alias to the model the route actually serves (with the live probe evidence in a comment), or leave the route id unpriced/fail-closed until there's evidence.

3. **`orcarouter/fusion*` → `openai/gpt-oss-120b`: 'verified live 2026-08' appears only in the PR body.** Put the provenance where it survives — a code comment with the probe date/method next to the alias entries — so the next person can re-verify when OrcaRouter rotates targets.

Minor: the PR body's 'Branded labels: OrcaRouter (auto)…' line contradicts the code (labels resolve to the upstream model, which is also what the README text says). Fix the description either way.

PR #1121, opened 2026-08-23
fix(hermes): keep post-finalization session growth on the observation day

### LINKED ISSUE #916: Session-cumulative providers (hermes) lose post-finalization usage from the daily history

avs-io:
## Summary

Hermes `state.db` rows are lifetime totals stamped at `started_at`. After CodeBurn seals day D, later growth on the same session was still keyed to that sealed day and dropped. This adds a sidecar ledger (`hermes-session-ledger.v1.json` under `CODEBURN_CACHE_DIR`) that records last-seen lifetime totals per `(profile, sessionId)` and emits observation-time deltas.

- First observation: lifetime totals at `started_at`, behavioral (weight 1), key unchanged (`hermes:${profile}:${sessionId}`).
- Later growth: delta only, stamped now, `supplementaryAccounting: true` (weight 0), key `…:obs:n`.
- 150 → 0 is visible to the cursor (reconcile ledgered identities after the `> 0` discovery filter): no observation, last-seen 0. 0 → 40 is one `+40` weight-0 observation.
- Explicit `$0` is recorded. Ledger publish is atomic; failure is retryable and holds the daily watermark.
- No `CACHE_VERSION` bump, no Hermes parse-version bump, no sealed-day rewrite, no last-activity re-anchor.

Proof fixture: day D stays 100 input / $0.10 / 1 call / 1 session; the next observation records only +50 / +$0.05 with 0 calls and 0 sessions; lifetime is 150 / $0.15 while calls and turns stay 1.

Direction from ozymandiashh (2026-08-09 unpushed implementation notes).

## Test plan

- [x] P1 sealed-day split via `parseAllSessions` + `ensureCacheHydrated` + `aggregateProjectsIntoDays`
- [x] P2 same `sessionId` in `default` and `profiles/coder` — two cursors, no cross-profile delta
- [x] P3 150→0→40 through `discoverFromDb` → parser → ledger (not ledger-only)
- [x] P4 explicit `$0` actual is recorded
- [x] P5 seed from cached lifetime call does not dump lifetime at `now`
- [x] P6 unwritable ledger is retryable; `lastComputedDate` does not advance
- [x] P7 first key stays `hermes:default:<id>`
- [x] Existing `tests/providers/hermes.test.ts` + day-aggregator + daily-cache degraded completeness

Closes #916


iamtoruk:
Reviewed (DeepSeek draft; every finding re-verified against the diff by maintainer — two of the draft's findings were contradicted by the actual code and are NOT included here).

Verified positives first: cost-increase-with-flat-tokens **does** emit an observation (`costGrew`/`costDelta` computed before the no-op branch); ledger publish is atomic + retryable and holds the watermark; `CODEBURN_CACHE_DIR` respected; explicit $0 recorded; migration seeds cursors from the cached baseline so growth since the last warm parse is recovered.

Two items to resolve:

1. **DB-restore reset+regrowth overcounts (design question).** Restore `state.db` from a backup: row disappears (cursor reset to 0, no observation), reappears at 400 while the ledger already attributed 500 → +400 emitted, lifetime shows 900. The 'reset means new spend' semantic is documented for the 150→0→40 case, but a restore isn't new spend. Is a bounded guard worth it (e.g. skip the first post-reset regrowth up to the pre-reset high-water mark), or do you consider restores out of contract? Either answer is fine if it's explicit in the ledger doc comment.

2. **Recovery horizon without a cache bump (product decision, not a defect).** As shipped, recovery starts from the warm-cache baseline: growth dropped *before* the user's last parse stays dropped forever, because the seed uses cached totals and sealed days are never re-read. A one-time `DAILY_CACHE_VERSION` bump would re-parse lifetime rows and recover all historically-dropped dollars (first observation lands on the session's start day). Given this repo ships bumps for dollar-moving changes routinely, I'd take the bump — but flagging rather than demanding, since the cost is one cold re-parse for every user.

iamtoruk:
Resolving both open questions (owner ruling):

**1. Restores are out of contract.** If `state.db` is rolled back and the ledger is not, we overcount — accepted. A post-reset high-water-mark guard is rejected: it would eat genuine 150→0→40 resets, which are treated as new spend by design. The remedy for a restore is to wipe both stores together, or live with the overcount. One ask: the current diff doesn't mention restores anywhere — add a line to the ledger's doc comment stating exactly this contract ('state.db restored from backup without wiping the ledger ⇒ regrowth is re-credited; wipe both or accept it'), so the next person doesn't re-litigate.

**2. No cache bump on this PR — withdrawn, with the correct reasoning on record.** A bump would not recover historical growth onto observation days: those intermediate snapshots never existed. It would re-emit current lifetime at `started_at` and thereby rewrite sealed days — violating this PR's own sealed-days-stay-put invariant. The seed already captures growth since the last warm parse; stopping the leak is this PR's class, and any historical reattribution is a separate, explicit migration if ever wanted.

With the doc line from (1), this is merge-ready from the review side.

PR #1122, opened 2026-08-23
feat(doctor): probeRoots for five fixed-location providers (#899 Tier 2, batch 2)

avs-io:
Tier 2 of #899, batch 2: `probeRoots()` for `codebuff`, `devin`, `gemini`, `kiro`, and `mistral-vibe`.

Campaign plan and #938 conventions from therickfactr. Each provider shares one root-set resolver between `discoverSessions` and `probeRoots`, so doctor reports the exact paths discovery would read after the same resolution — not a lookalike parent.

## Per-provider

- **codebuff** — `getCodebuffBaseDir` is not the full root set. Default (and empty factory) reports all three CHANNELS before existence filtering: `~/.config/manicode`, `manicode-dev`, `manicode-staging`. Non-blank factory wins; otherwise non-blank `CODEBUFF_DATA_DIR`; empty factory is unset.
- **devin** — reports `join(cliDir, "transcripts")` and `join(cliDir, "sessions.db")`, not the parent. Empty factory is the default CLI dir. Same pair discovery reads.
- **gemini** — exported `getGeminiTmpDir()`; probe reports that parent only (`~/.gemini/tmp`). No `GEMINI_HOME`. Does not leak `tmp/<project>/chats`.
- **kiro** — shared pre-filter candidate list (`getKiroAgentDirCandidates`); probe does **not** call the Linux `existsSync`-filtering `getKiroAgentDir`. Empty agent/workspace fall back (same as unset). Empty CLI is skipped by both probe and discovery (not `''` or `.`). Empty v2 is not reported. Tests use overrides so they never stat the developer `~/Library/Application Support/Kiro`.
- **mistral-vibe** — same joined path as `getMistralVibeSessionsDir` (`VIBE_HOME` or `~/.vibe` + `logs/session`). Factory override goes through that helper.

`vercel-gateway` is unchanged (network, out of scope). #938 / batch 1 tests are untouched.

## Testing

- [x] `HERMES_HOME=/tmp/cb-cleanhome` vitest: tier-2 probe-roots + the five provider suites + kiro-projectpath + provider-turn-grouping — 161 passed
- [x] Assertions pin exact root sets, not substrings

Closes nothing on its own; this is the second Tier 2 batch of #899.


iamtoruk:
Reviewed (maintainer; the model draft for this one was unusable and discarded — its one 'critical' was a misquote of valid `??` code).

Verified the core contract directly: each of the five providers routes `probeRoots()` through the same resolver discovery uses (`getCodebuffRootSet` shared explicitly, `getDevinDiscoveryRoots`, kiro's shared pre-filter candidate list with the same empty-CLI skip, and the batch-2 mirror test suite asserts probe==discovery resolution per provider, including override and env-pin cases). Read-only, no discovery behavior change.

No findings. Merge-ready from the review side.

PR #1123, opened 2026-08-23
feat: Copilot AI Credit plan math from total_nano_aiu

### LINKED ISSUE #943: Request: Github Copilot AI Credit tracking
Currently the `codeburn plan` command supports subscription tracking for Claude and Cursor/Codex-style plans, but there's no preset for GitHub Copilot's AI Credits allotment (`codeburn plan set claude-max`, `codeburn plan set cursor-pro`, etc. — nothing for `copilot`).

Copilot moved to usage-based billing on June 1, 2026: every plan/seat gets a monthly allotment of AI Credits (1 credit = $0.01), and my org gives me a custom allotment of 20,000 credits/month rather than a default preset amount. Right now the only way to see "how many of my 20,000 credits have I used" is GitHub's own billing page — codeburn's Copilot numbers are a separate, locally-estimated dollar figure from session files, so they don't reconcile against the actual credit balance I'm being measured against.

GitHub added an `ai_credits_used` field to the Copilot usage metrics API on June 19, 2026, which reports real per-user credit consumption (not an estimate). It would be great if codeburn could either:
1. Add a `codeburn plan set copilot --credits 20000` (or similar) option so Copilot shows up in the plan overview alongside Claude/Cursor, using the same "credits used / credits remaining" framing GitHub itself uses, or
2. Pull `ai_credits_used` from the Copilot metrics API (where a token is available) so the Copilot numbers reflect the real credit ledger instead of a local token estimate.

This is basically the Github Copilot version of #408/#495 (https://github.com/getagentseal/codeburn/issues/408 and https://github.com/getagentseal/codeburn/issues/495).

avs-io:
Fixes #943

Plan math against real `total_nano_aiu`. Not #890. Not #1114 live quota. No `app/`.

## Behavior

- `copilot` is a plan provider. Presets use official allotments (fetched 2026-08-23), not sticker prices:
  - `copilot-pro` — 1500 credits / $15 equivalent
  - `copilot-pro-plus` — 7000 / $70
  - `copilot-max` — 20000 / $200
- `codeburn plan set custom --credits N --provider copilot` is the org path. Mixed units (`--monthly-usd` + copilot) are rejected.
- Copilot spend is `sum(finite nanoAiu) / 1e9`. Missing nanoAiu → `spentCredits = 0` + `creditsIncomplete`. Token-priced `costUSD` does not fill the bar.
- Claude / cursor / grok / custom-USD plans stay on `costUSD`.
- TUI headline is `N / B AI Credits`. Status still says “Not a live provider window.” No GitHub UTC-1st claim.

## Serve hole

`cachedCallToApiCall` now threads `nanoAiu` onto `ParsedApiCall`. Same field on the fresh-parse twin. `CACHE_VERSION` stays 9 — reconstruction from already-cached `CachedCall.nanoAiu` works.

## Proofs

1. `plan set copilot-pro` writes `{ id: copilot-pro, monthlyCredits: 1500, monthlyUsd: 15, provider: copilot, resetDay: 1 }`
2. `custom --credits 20000 --provider copilot` writes `monthlyCredits: 20000`, `monthlyUsd: 200`
3. `custom --monthly-usd 200 --provider copilot` exits non-zero
4. `nanoAiu = 1_500_000_000` → `spentCredits === 1.5`, `budgetCredits === 1500`, `percentUsed === 0.1`
5. Same tokens, no `nanoAiu` → `spentCredits === 0`, `creditsIncomplete === true`
6. Claude plan on the same fixture still sums `costUSD`. JSONL / rollup siblings without `nanoAiu` do not change the 1.5


iamtoruk:
Reviewed (MiniMax draft; findings verified against the diff by maintainer). Preset math, nanoAiu threading through the cache, no-app/no-quota scope claims: all **confirmed** (including `CACHE_VERSION` staying 9 with warm reconstruction asserted in tests).

Two changes requested:

1. **Add a `supplementaryAccounting` guard in `copilotCreditSpend`** (`src/plan-usage.ts`). Verified: the loop sums every copilot call's `nanoAiu` with no supplementary filter. Today's fixtures pass only because supplementary siblings happen to carry no `nanoAiu` — the moment a store attaches it to a paired rollup row, credits double-count. One line (`if (call.supplementaryAccounting) return`) closes the class.

2. **Retention predicate widened for every plan provider** (`src/plan-usage.ts`: `totalApiCalls > 0 || totalCostUSD > 0` → `sessions.length > 0`). Verified unconditional — Claude/cursor/grok plans now retain zero-spend sessions they previously dropped. Zero dollar impact, but it's an unannounced behavior change on providers this PR isn't about: gate it on `isCopilotCreditsPlan(plan)` or call it out in the description.

iamtoruk:
Merged. For the record on the retention-predicate change (`totalApiCalls > 0 || totalCostUSD > 0` → `sessions.length > 0` in plan views): this now retains zero-spend sessions for every plan provider, not only copilot. Zero dollar impact — arguably more correct, since behavioral-only sessions were silently dropped from plan views before. Documented here so the behavior change has a paper trail.

PR #1126, opened 2026-08-24
fix(sync): wire truth — no filesystem paths in ai.project, billable output tokens

iamtoruk:
Two sync wire-truth fixes, both diagnosed by a real-data pilot push against a local receiver (37.8k spans, $5.1k of real usage reconciled three-layer). Implementation by Kimi Code from written specs; reviewed and re-verified by maintainer.

## 1. ai.project must never carry a filesystem path
Every synced span carried `ai.project` as the slugified absolute path (`-Users-torukmakto-Projects-eywa`), contradicting docs/sync/README.md ('paths stay local') and COMPATIBILITY.md ('safe project basename'). Fixed with `wireProjectName()` (leaf directory name) at the single choke point after `parseAllSessions` — covers both wire paths; `codeburn yield` local output untouched. Never-lose: usage span ids + commit attribution keys carry no project (unchanged, no re-send); session attribution re-emits once per session and the receiver upserts by (org, trace) — replaced, not duplicated. Same-leaf directories collapse to one wire identity (the contract's definition); `git.repo` disambiguates.

## 2. ai.output_tokens must be billable output, not raw
Exclusive-reasoning providers were under-reported on the wire by exactly their reasoning volume (measured: opencode short 37,848 = summed reasoningTokens). Routed through `billableOutputTokens` — the same correction the display layer took in #1115. Inclusive providers (claude/codex/copilot) byte-identical. Cost unaffected (priced before export). Span ids carry no token counts → no re-send, no duplication.

## Tests
New wire-boundary tests: path-shaped project never reaches the wire (usage + attribution spans all equal the basename); exclusive provider output = output+reasoning, inclusive stays raw. Targeted sync suites 106/106; full suite 240 files / 3,253 green; tsc clean — all re-run by reviewer per commit.

Pilot note: the local pilot teams.db will be reset after merge (old rows carry path-shaped projects and raw outputs); no production receiver exists yet.

### DIFF
--- a/src/sync/cli.ts
+++ b/src/sync/cli.ts
@@ -23,7 +23,7 @@ import {
 import { createCredentialStore } from './credentials.js'
 import { readSyncConfig, writeSyncConfig, deleteSyncConfig, updateLastSync } from './config.js'
 import { collectUnsentCalls, collectUnsentAttribution, sendBatches, sendAttributionBatches, batchCalls, MAX_PER_PUSH, MAX_ATTRIBUTION_PER_PUSH, type PushResult } from './push.js'
-import { batchAttributionItems } from './otlp.js'
+import { batchAttributionItems, wireProjectName } from './otlp.js'
 
 export function registerSyncCommands(program: Command): void {
   const sync = program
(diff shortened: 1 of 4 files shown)


iamtoruk:
@Enclavet — you built this wire format, so flagging two value corrections here before they ship (no schema changes, span ids untouched): `ai.project` now sends the directory basename instead of the slugified absolute path (closing the gap with the README's 'paths stay local' promise), and `ai.output_tokens` routes through `billableOutputTokens` so exclusive-reasoning providers stop under-reporting by their reasoning volume (matches the #1115 display correction). Both were caught by a real-corpus push against a local receiver. If either interacts badly with how your side consumes the fields, would value your eyes before merge.

PR #1128, opened 2026-08-24
fix(sync): trust only proven outbound project metadata

avs-io:
## Summary

Follow-up to #1126. The billable-output correction in #1126 is sound, but its project basename mapper remained fail-open for Claude slugs, generated titles, home roots, relative paths, provider storage directories, credential-shaped basenames, and provider-owned container paths.

This PR moves the privacy decision to the OTLP boundary and requires explicit provenance:

- usage `ai.project` comes only from a trusted provider-recorded absolute cwd
- attribution resolves repositories only from that same trusted cwd and derives `ai.project` from normalized `git.repo`
- legacy cache cwd values without provenance fail closed
- relative cwd values cannot resolve against the directory running `sync`
- Hermes prompt-derived paths, including legacy Windows `Current working directory:` messages, remain local grouping labels but never become outbound provenance; this is an intentional privacy correction
- Goose propagates its dedicated `sessions.working_dir` field, while `/sessions/<container>` paths are rejected centrally at the final trust boundary
- provider, model, tool, and span-name identifiers are sanitized at serialization; real routed identifiers remain supported
- home, email, credential, encoded-path, slug/title, storage-path, and provider-container cases are covered

No Teams UI, desktop, auth, receiver, turn-metrics, or auto-sync scope is included.

## Relationship to #1126

#1126 is merged. This dependent hardening pass preserves its OTLP transport, deterministic trace/span IDs, and sent-ledger behavior while making the serializer the authoritative project-privacy boundary.

## Project identity precedence

`ai.project` remains optional. A usage span may carry the safe basename of a provider-recorded cwd; an attribution span may carry the basename of normalized `git.repo`. If both exist for one trace and differ (for example, a fork checkout named `my-fork` whose normalized repository is `widget`), the attribution repository basename is authoritative for project aggregation. Receivers must not count the provisional usage basename as a second project bucket.

## Review fixes at `cc721f00`

- reject Goose and other provider-owned `/sessions/<name>` container paths at the shared working-directory trust boundary
- document canonical repository precedence when usage and attribution labels differ
- accept legitimate `owner/repo` identities and reject credential-directory repo labels such as `.ssh`
- explicitly document removal of prompt-derived Hermes/Windows cwd as outbound provenance while preserving the local grouping label
- prove pre-fix cached canonical paths cannot bypass the serializer, and add positive coverage for real routed provider/model/tool identifiers

## Verification

- `npm test`: 240 files passed, 2 skipped; 3,258 tests passed, 5 skipped
- `npx tsc --noEmit`: passed
- `npm run build:cli`: passed
- `git diff --check`: passed

## Compatibility and cleanup

- `ai.project` is optional; receivers must accept an unattributed bucket
- legacy caches omit project labels until a trusted provider field is reparsed
- project labels already leaked and ledgered before this fix require receiver-side cleanup; the append-only local ledger cannot retract them by itself
- legitimate slash-separated model identifiers remain supported; path-, URL-, email-, credential-, and control-shaped identifiers fail closed


avs-io:
@Enclavet — following Resham's ping on #1126, this is the dependent hardening pass over the sync boundary you originally built in #660 and extended in #848.

#1128 preserves the OTLP transport, deterministic trace/span IDs, and sent-ledger behavior, but moves project privacy enforcement to the final serializer: `ai.project` is now omitted unless it comes from trusted provider-recorded provenance; attribution derives its project label from normalized `git.repo`; relative/legacy-cache paths fail closed; and provider/model/tool/span-name identifiers are sanitized before egress. There are no receiver or AWS changes in this PR.

Could you please sanity-check the receiver contract in particular:

1. the ingest/dashboard safely accepts spans with no `ai.project` and places them in an unattributed bucket;
2. no index or grouping logic assumes `ai.project` is always present or uses the old basename behavior;
3. identifier sanitization does not conflict with receiver validation or dimensions;
4. the existing upsert keys remain compatible, and whether already-ingested path-shaped project values need a receiver-side cleanup/reindex.

The PR is intentionally draft and dependent on #1126. Your eyes on those backend assumptions would be especially valuable before it is made ready.

iamtoruk:
Reviewed (two model drafts + maintainer verification; several draft findings were disproven and are omitted — notably a claimed compile error that is valid `??` code, tsc verified clean). The architecture is right and most claims verified: span-level never-lose holds (field omitted, span kept), legacy cache degrades to omission, hermes prompt paths stay local, goose working_dir propagates, relative-cwd resolution blocked. Verified findings to address before undraft:

1. **Goose container paths aren't filtered** (`src/providers/goose.ts` ~:222): `working_dir` flows into `workingDirectory` with only the home-root check — a container-style path (`/sessions/<name>`) is absolute, non-home, and reaches the wire as a basename. Claude's cowork filter has no goose equivalent. Add the same container guard on the goose path (or centralize it in `isTrustedAbsoluteWorkingDirectory`).
2. **Usage vs attribution project identity can diverge** (cwd basename vs repo basename) — e.g. a fork cloned as `my-fork` of repo `widget` shows two different `ai.project` values for the same session. The dashboard joins by trace so nothing breaks, but the projects breakdown will split one project in two. Either derive both from the same source when both exist, or document the precedence.
3. **`attributionProjectFromRepo` edges**: a repo whose last segment is credential-shaped-but-unmatched (e.g. `.ssh`) passes; and the `parts.length < 3` guard drops legitimate 2-part remotes. Tighten both.
4. **Windows prompt-derived cwd inference removed** (`tests/providers/hermes.test.ts` — the deleted 'infers projects from Windows cwd messages' behavior): if intentional, say so in the description; it's a silent feature regression otherwise.
5. Two smaller ones to double-check: cached `CanonicalCwd` surviving alongside a stripped `workingDirectory` (could bypass the boundary if a future path serializes it), and the identifier sanitizer's allow/deny overlap (`/`-containing ids that hit the system-dir prefix list get rewritten to `unknown` — spot-check real routed/model ids beyond `orcarouter/*`).

Plan on our side: 0.9.21 ships with #1126's baseline (already merged); this lands early in 0.9.22 once the above are addressed. The provenance model is the right end state — thank you for formalizing it.

avs-io:
Addressed all five review areas in `cc721f00` and pushed to this PR.

1. Goose container paths now fail closed at the shared `isTrustedAbsoluteWorkingDirectory` boundary. The provider regression carries `/sessions/synthetic-customer-secret` through the Goose parser and proves the final OTLP payload contains neither `ai.project` nor the container label.
2. Project precedence is now explicit in both sync contracts: for one trace, normalized `git.repo` attribution is authoritative and the usage cwd basename is provisional; receivers must not create two project buckets.
3. `attributionProjectFromRepo` now accepts legitimate `owner/repo` identities and rejects credential-directory labels such as `.ssh`, with wire tests for both.
4. The Hermes/Windows change is intentional and is now called out in the PR description and user/developer docs: prompt-derived cwd text remains useful only as a local grouping label and never becomes trusted outbound provenance.
5. The existing warm-cache provenance test confirms a cached canonical/project path cannot bypass the serializer without the provider-field marker. I also added a positive test for routed identifiers (`openrouter`, `anthropic/claude-sonnet-4.6`, `orcarouter/openai-compatible`) to prove the sanitizer does not erase real routing dimensions.

Verification on the pushed commit:

- full `npm test`: 240 passed, 2 skipped; 3,258 tests passed, 5 skipped
- `npx tsc --noEmit`: passed
- `npm run build:cli`: passed
- `git diff --check`: passed

@iamtoruk this is ready for your re-review. @Enclavet the receiver questions in the earlier note remain the backend confirmation gate, especially missing-project handling, attribution-over-usage project precedence, and cleanup of already-ingested path-shaped labels.

Issue #1143, opened 2026-08-25
TUI: q during the cold-start fill waits ~16.5s for the graceful index drain with no feedback


iamtoruk:
Follow-up from #1141/#1142. With the input-starvation fixed, q pressed during the post-paint background index is received instantly but the exit path drains the fill before quitting - a deterministic ~16.5s silent wait on a 21k-file corpus (measured at press times 1s/5s/10s after paint, all 16.5-16.8s). Ctrl+C exits in 0.1s via Ink's immediate path, and #1109 made abrupt exit kill-safe, so either: make q exit immediately too, or keep the drain but show a status line (finishing background index, press Ctrl+C to quit now). The drain does buy a completed index for the next launch, so the feedback option preserves that value.

ozymandiashh:
Shipped: `q` during the cold-start fill no longer waits silently for the index drain; the dashboard shows a status line and a second `q` (or Ctrl+C) exits immediately. See the comments in `src/dashboard.tsx` around the drain path. Closing.

Issue #1141, opened 2026-08-25
TUI: input deaf and unbuffered for ~15-20s after first paint on a cold cache - reads as a total freeze to new users


iamtoruk:
Reproduced from our side while investigating #1139, on 0.9.21, macOS, Node 26.7, pty-driven:

- Cold-cache interactive launch paints normally (fixture corpus: 0.7s via npx).
- For roughly 15-20s after paint, keyboard input is unresponsive: a single `q` or Ctrl+C pressed in that window is swallowed and never acted on (raw mode, not buffered). Pressed after the window settles, `q` exits in under a second. Output flow measurement shows no repaint loop (11KB then silence) - the event loop is starved by the post-paint background index (#1109 fill startup), not rendering.
- Ctrl+C DURING the pre-paint parse phase works; only the post-paint window is deaf.

Why it matters: a first-run user with a large history sees minutes of indexing, then a paint, presses Ctrl+C in the deaf window, concludes the app froze, and kills the terminal - so the first index never completes and every retry is cold again (#1139 is exactly this shape: npx and brew both 'freeze', menubar 'Couldn't load today' from its fetch timing out against the never-finished index). This was logged as the cold-quit-latency WARN in the 0.9.21 release sweep; it is a bug, not a latency footnote.

Fix directions to evaluate: yield to the event loop during fill startup (the fill itself is already async per #1109 - the starvation is in its synchronous setup/cache write), and/or buffer keystrokes arriving during the deaf window so a q/Ctrl+C is honored the moment the loop breathes. An immediate Ctrl+C should abort cleanly at any phase.

PR #1144, opened 2026-08-25
fix(tui): q during the cold-start fill shows a status line; second press quits immediately (#1143)

iamtoruk:
Completes the #1141 story for 0.9.22. First q during an active background fill arms a footer-styled confirmation ("Finishing background index so the next launch starts warm - press q or Ctrl+C again to quit now") and continues the graceful drain; a second q or Ctrl+C exits immediately through the #1109 kill-safe abrupt path. q with no fill active exits instantly with a pinned no-flicker test (unmount frame flushed before the assertion). Ctrl+C semantics unchanged. Status line renders only while quitArmed AND indexing, matching the existing StatusBar styling exactly. Suite 3712 green, tsc clean. Twin-built, verifier-compared (two verifier findings upheld and fixed, one refuted against the pre-existing StatusBar style).


iamtoruk:
nsis failure was the version-parity manifest guard (root 0.9.22 vs app 0.9.21), not this change - versions aligned on main. Re-running.

Issue #1148, opened 2026-08-26
macOS: menubar status item never appears - per-bundle-id state can be permanently poisoned (survives reinstall and reboot)


oscarbrey:
## Summary

On a Mac where `org.agentseal.codeburn-menubar` has accumulated bad status-item state, the CodeBurnMenubar icon never appears in the menu bar, on any launch, on any display. The app itself runs fine (process alive, CLI polling works); only the `NSStatusItem` placement fails. Re-signing the **identical binary** with a different `CFBundleIdentifier` fixes it instantly, which isolates the bundle id as the only variable.

## Environment

- macOS 26 (Tahoe), Apple Silicon
- Menu bar set to auto-hide (`_HIHideMenuBar = 1`), relevant, see below
- CodeBurnMenubar 0.9.19 (local build) and 0.9.22 (official release), both affected identically

## What I ruled out

All of the following were tried with the menu bar actively revealed during launch (auto-hidden bars make placement fail if the bar is hidden at launch time, but that alone isn't this bug):

1. Fresh reinstall via `codeburn menubar` (checksum-verified official 0.9.22): no icon
2. Deleting the app's preferences domain and all caches/state, then reinstalling: no icon
3. `killall ControlCenter` then relaunch: no icon
4. **Full reboot**, fresh launch with the bar revealed: no icon
5. Same binary, `CFBundleIdentifier` changed to a fresh id and re-signed: **icon appears on the first launch, every time** (reproduced with both 0.9.19 and 0.9.22)

## Diagnosis

macOS keeps per-bundle-id state for status items (Control Center's status-item scene service). That state can go permanently bad: once poisoned, `NSStatusBar.system.statusItem(...)` succeeds, `item.isVisible` reads true, the button has its image, but the item's window is parked at the screen's right corner behind the clock with legacy 22pt metrics, and AppKit never retries placement. The poison survives app reinstalls, preference deletion, and (new data point) reboots.

Likely contributor in this case: the app ran for weeks launching at login while the auto-hidden menu bar was hidden. Each such launch parks the item, and repeated parking appears to be what corrupts the id's state permanently. I've hit the same failure mode with three unrelated in-house menu-bar apps; the fresh-bundle-id fix worked first-try in every case.

## Suggested fixes

[...]

PR #1149, opened 2026-08-26
draft(menubar): repair poisoned installer identity

avs-io:
Fixes the installer/identity-repair half of #1148.

The self-contained native status-item recovery is now isolated in #1161. This draft intentionally contains only the larger installer identity-migration path so it cannot ride along with the safer Swift fix.

## What this draft contains

- Adds `codeburn menubar --repair-placement` and `--reset-placement`.
- Uses a persisted, validated recovery bundle ID and a transactional app/preferences swap; reset returns to the canonical identity.
- Preserves the installed bundle identity across `--force`, sanitizes only status-item placement defaults, and rolls back bundle/preferences/identity state on failure.
- Transfers Login Item state through the exact installed bundle with a capability-gated maintenance protocol.
- Treats `requiresApproval` conservatively as disabled because it can mean either pending approval or revoked consent; repair must not infer away a user's opt-out.
- Refuses local re-identification of Developer ID/notarized artifacts instead of invalidating the release identity.

## Verification on exact head `c112ad07ec47d472c525e5e9690924cd062ebd42`

- Installer tests: 65/65 passed.
- TypeScript type-check: passed.
- Focused Login Item policy tests: 2/2 passed.
- Full Menu Bar Swift suite: 211 tests in 32 suites passed.
- `git diff --check`: clean.

## Release gate: HOLD

Do not merge this installer subsystem on source or unit evidence alone. It needs an affected Tahoe machine to prove the complete identity migration and rollback path:

1. Refresh the installed app with `codeburn menubar --force`.
2. Run `codeburn menubar --repair-placement`.
3. Confirm the status item is visible and clickable.
4. Quit/relaunch and log out/in; confirm persistence and no duplicate app or Login Item.
5. Run `codeburn menubar --reset-placement`; confirm the canonical identity returns.
6. Exercise an interrupted/failed replacement and verify rollback restores the prior app and Login Item state.

The README documents that Developer-ID/notarized artifacts cannot be locally re-identified and must use reset/canonical identity instead.

Safe native counterpart: #1161.


iamtoruk:
The Swift placement fix here is clean and I'd happily take it as-is: `StatusItemPlacementPolicy.isParked` requires all three anomaly signals (legacy 22pt height + flush with the display's right edge + top band), so false positives are implausible, and recovery is a bounded ≤3 supported visibility pulses with no remove/recreate churn, cancelled on sleep/terminate. The unit coverage (placement policy, login-item policy, and the installer rollback paths — restore-on-commit-failure, keep-committed-on-launch-failure, state-restore-even-when-bundle-restore-fails) is genuinely good.

My one recommendation is to **split this PR**:

- Ship the Swift half now — `StatusItemPlacementPolicy.swift` + the `CodeBurnApp.swift` placement hunks (autosaveName, `isVisible` restore, bounded pulse recovery). It's self-contained and helps every affected Tahoe user immediately.
- Hold the +707-line installer identity-repair subsystem (`--repair-placement` / `--reset-placement`, bundle re-identification, Login-Item transfer, preference transactions) behind the on-device gate you called out yourself ("do not merge on source/unit evidence alone"). The unit tests are solid; what's unvalidated is whether macOS on a genuinely poisoned Tahoe machine accepts the fresh identity — which is exactly what needs a real-device pass before this half lands.

Two smaller notes for whenever the installer half proceeds:
- Once the project ever ships Developer-ID/notarized releases, a plain `codeburn menubar --force` will fail for any previously-repaired install (re-identify refuses non-ad-hoc and the persisted recovery id survives `--force`). It's fail-closed and the error names `--reset-placement`, but it's worth a README line.
- `--repair-placement` re-downloads the release even though the README's prescribed sequence just fetched it via `--force`.

avs-io:
Thanks for the careful split recommendation. Addressed.

- The self-contained Swift placement recovery is now isolated in #1161 at `d89ce218`; its focused and full Swift suites pass, and all GitHub checks are green.
- This PR is now installer/identity-repair only at `c112ad07`. The native placement files and hunks have been removed from its diff.
- The Developer-ID/notarization limitation is documented in `mac/README.md`.
- This installer half remains draft/HOLD exactly as suggested: no merge recommendation until the complete identity migration, restart persistence, reset, and rollback sequence is proven on an actually affected Tahoe Mac.

All checks on both PRs are green. I have not treated source/unit evidence as affected-device proof.

iamtoruk:
Closing this one. The status item fix from #1161 and the follow-up retry fix shipped, and #1148 is closed. The installer identity repair was always gated on a test on an affected Tahoe Mac, and we have no open report that needs it. If a poisoned identity shows up again, we can reopen and rebase. Thanks for the careful split.

PR #1150, opened 2026-08-26
fix(omp): correct OMP period scoping, cost attribution, and Grok 4.6 prompt tiers

sm0keyyy:
## Summary

This pull request corrects 3 defects in OMP-provider accounting and Grok 4.6 pricing.

### 1. OMP rows ignored the period filter

`codeburn models -p <period>` returned an identical OMP aggregate at every scope, while a comparison provider varied by period. Independent session output showed records across multiple dates. The row labelled `today` therefore reflected a cumulative total rather than a daily total.

Cause: `src/providers/pi.ts` stamped calls from `entry.timestamp`, but the session-header branch did not retain its timestamp. When `entry.timestamp` was absent, the range filter received an invalid date.

The shared range helpers remain unchanged: `callsInRange`, `classifiedTurnSlicedToRange`, and `filterProjectsByDateRange`.

### 2. OMP cost attribution stayed at zero

A valid alias plus price override still yielded a zero-cost OMP row, although the audit path recomputed a nonzero cost.

Cause: `src/providers/pi.ts` treated any finite `msg.usage.cost.total` as authoritative and skipped `calculateCost`. OMP writes `cost.total = 0` for `xai-oauth` IDs, so neither the alias table nor the price override applied. `audit` recomputes through `getModelCosts`, while `attributedCostUSD` sums the cached per-call value.

The alias loader remains unchanged. A zero reported cost now acts as absent. A nonzero reported cost keeps its existing behavior.

### 3. Grok 4.6 lacked the 200k prompt tier

xAI prices Grok 4.6 in 2 tiers. Once a request reaches 200k prompt tokens, xAI applies the higher rate to every token in that request.

| Tier | Input | Cached input | Output |
|---|---:|---:|---:|
| Under 200k prompt tokens | $2.00 | $0.50 | $6.00 |
| 200k or more prompt tokens | $4.00 | $1.00 | $12.00 |

Source: https://docs.x.ai/developers/models/grok-4.6 and https://docs.x.ai/developers/pricing, retrieved 2026-08-26.

The bundled rate card carried only the low tier, so large-context Grok traffic under-reported. Prompt tokens equal input plus cached input. The high tier triggers inclusively at exactly 200000 prompt tokens. An exact configured `priceOverrides` entry still wins over the built-in tier. `CACHE_SCHEMA_VERSION` moves to 3 so cached rendered costs recompute.

## Change evidence

| Change | Why | Receipt |
|---|---|---|
| Nested OMP agent discovery and payload metadata | Retain nested-agent identity, resolved model, and start time through the shared cache and menubar payload. | Ran `npx vitest run tests/providers/omp.test.ts`: 12 passed, 0 failed, duration 623 ms. Test title: "discovers nested per-agent sessions with their resolved model and start time". |
| Reported nonzero OMP cost retention | Preserve a provider-reported nonzero cost through parsing and caching. | Ran `npx vitest run tests/providers/omp.test.ts`: 12 passed, 0 failed, duration 623 ms. Test title: "preserves a provider-reported nonzero cost". |
| Removal of the unused Pi message field | Remove a message field that the parser does not consume. | The removed field was `message.provider` (`provider?: string`). A search of `src` for `msg.provider`, `msg?.provider`, `message.provider`, `message?.provider`, `entry.message.provider`, or `entry.message?.provider` returned 0 hits. No test applies to this type-only deletion. |
| Status snapshot render version | Invalidate a stale status snapshot after OMP payload semantics changed. | Ran `npx vitest run tests/cli-status-menubar.test.ts -t "invalidates the snapshot when a model is newly marked flat-rate"`: 1 passed, 0 failed, duration 1.49 s. Test title: "invalidates the snapshot when a model is newly marked flat-rate". It verifies invalidation behavior. No test asserts the literal `STATUS_SNAPSHOT_RENDER_VERSION = 2`. |
| Grok 4.6 prompt-tier pricing and pricing-cache version | Apply the vendor's high tier at and above the prompt threshold and invalidate cached pricing. | Ran `npx vitest run tests/models.test.ts`: 192 passed, 0 failed, duration 900 ms. Test titles: "uses the low tier below 200000 prompt tokens", "uses the high tier for every token at exactly 200000 prompt tokens", and "uses the high tier above 200000 prompt tokens". The cited tests do not assert `CACHE_SCHEMA_VERSION = 3` or rejection of a version-2 pricing cache. |
| OMP timestamp fallback and zero-cost repricing | Retain timestamps for date slicing and route a zero reported cost through aliases and overrides. | Ran `npx vitest run tests/providers/omp.test.ts`: 12 passed, 0 failed, duration 623 ms. Test titles: "prices a zero reported cost through its alias and price override" and "uses user attribution timestamps to date-slice records in one OMP session". |
| Aliased Grok price-override test | Confirm that an alias reaches an exact override instead of the built-in tier. | Ran `npx vitest run tests/models.test.ts`: 192 passed, 0 failed, duration 900 ms. Test title: "uses an aliased price override instead of the built-in prompt tier". |

The prior description recorded a passing model test command and a successful CLI build. This sanitation pass did not run either command. Those historical statements are not execution receipts for this metadata update.

## Notes

A retired catalog model intentionally retains zero rates.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -54,7 +54,7 @@ const { version } = require('../package.json')
 // Bump when the menubar payload's rendering semantics change without a package
 // release or daily-cache version change. The envelope version in session-cache
 // protects record shape; this protects the meaning of an otherwise valid one.
-const STATUS_SNAPSHOT_RENDER_VERSION = 1
+const STATUS_SNAPSHOT_RENDER_VERSION = 2
 const STATUS_SNAPSHOT_SEMANTIC_KEY = `${version}:render-${STATUS_SNAPSHOT_RENDER_VERSION}:daily-${DAILY_CACHE_VERSION}`
 import { loadCurrency, getCurrency, isValidCurrencyCode } from './currency.js'
 import { CodexThroughputReader, newestCodexSession, renderCodexThroughput } from './codex-throughput.js'
--- a/src/menubar-json.ts
+++ b/src/menubar-json.ts
@@ -327,7 +327,19 @@ export type MenubarPayload = {
     }
     tools: Array<{ name: string; calls: number }>
     skills: Array<{ name: string; turns: number; cost: number }>
-    subagents: Array<{ name: string; calls: number; cost: number }>
+    subagents: Array<{
+      name: string
+      calls: number
+      cost: number
+      agentName?: string
+      model?: string
+      startedAt?: string
+      inputTokens?: number
+      outputTokens?: number
+      cacheReadTokens?: number
+      cacheWriteTokens?: number
+      totalTokens?: number
+    }>
     mcpServers: Array<{ name: string; calls: number }>
     /// Every pull request with attributed spend, cost-descending, plus the
     /// multi-link-safe distinct total. Absent when no PR links were observed and
--- a/src/models.ts
+++ b/src/models.ts
@@ -71,7 +71,7 @@ const CACHE_TTL_MS = 24 * 60 * 60 * 1000
 // Also folded into getPricingGenerationKey() below: a resident/snapshot-caching
 // consumer needs the same "pricing behavior changed" signal this already gives
 // the on-disk LiteLLM cache, not just the on-disk cache itself.
-export const CACHE_SCHEMA_VERSION = 2
+export const CACHE_SCHEMA_VERSION = 3
 const WEB_SEARCH_COST = 0.01
 const ONE_HOUR_CACHE_WRITE_MULTIPLIER_FROM_FIVE_MINUTE_RATE = 1.6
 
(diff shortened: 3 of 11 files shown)


iamtoruk:
## Review Summary (verified locally + independent agent pass)

Checked out the branch, ran targeted suites (232/232 pass) and the full suite (3365 pass / 0 fail). A second independent agent re-ran everything and reached the same verdict: **Approve**.

### What looks solid
- Timestamp fallback chain (entry → user attribution → session header) fixes the period-scoping bug; date-slice test proves it end to end.
- Zero-reported-cost now falls through to `calculateCost`, so aliases and price overrides apply to OMP's xai-oauth rows. Nonzero reported costs still pass through untouched.
- Grok 4.6 tier boundary is correct per xAI's published pricing (≥200k prompt tokens, inclusive, applied to all buckets). Override-beats-tier is tested.
- All three cache versions bumped (`CACHE_SCHEMA_VERSION`, omp parse version, `STATUS_SNAPSHOT_RENDER_VERSION`) — stale pre-fix data can't be served.
- `validateCachedFile` covers the new cache fields.

### Non-blocking observations
1. **Grok tier counts `input + cacheRead`, not `cacheWrite`.** Checked xAI's live pricing page: they publish no separate cache-write rate for Grok, so cache-write tokens are presumably already inside `input`. Consistent with the vendor model — flagging in case you have docs saying otherwise.
2. **`if (!timestamp) continue` now silently drops fully-undated calls.** Intentional and reasonable, but no test documents the behavior change. A small test would pin it.
3. **`\u0000` literal in the ompSubagentMap key** works (verified it's a literal escape, not NUL), though `JSON.stringify([a,b,c])` would be more idiomatic.

### Follow-up commit
`18b2dcc0` applies three review patches: extracted `tieredCostsFor()` and `resolveMessageModel()` helpers for readability, and stopped formatting a display string into the OMP subagent row's `name` (the payload now carries `agentName` + structured fields; the menubar formats its own label). No behavior change; full suite still green.

iamtoruk (approved):
Verified locally: targeted suites 232/232, full suite 3365 pass / 0 fail. Independent agent review agrees. Non-blocking observations posted as a comment.

PR #1155, opened 2026-08-28
fix(desktop): harden packaged resident CLI routing

avs-io:
## Why

The packaged Desktop CLI can lose its resident stdio child or misroute ordinary CLI commands through the Electron launcher. That leaves Desktop refreshes brittle and can turn a closed pipe into an `EPIPE` failure.

## What changed

- recover safely when resident stdin closes: detach the dead child and fall back to one-shot execution
- preserve `serve --stdio` dispatch in the generated Electron-as-Node launcher without swallowing ordinary CLI arguments
- make the packaging gate prove both an ordinary CLI command and a held-open resident ready handshake
- run that packaged resident smoke under a disposable HOME/data/cache tree so packaging cannot arm watchers against the developer's real corpus
- document the packaged launcher behavior in the distribution notes

## Verification

- exact head: `20f3d29046ce7a8ebf82e970e430d0e504c63203`
- focused closed-pipe fallback regression: passed
- exact generated-launcher stage gate: passed; 160 production packages staged
- Desktop typecheck and Electron build: passed
- Desktop non-CLI test suite: 523 passed
- `git diff --check`: clean

## Review notes

This PR is intentionally limited to packaged CLI/resident lifecycle reliability. It does not change aggregation, pricing, sync, or product UI semantics.

The broader `electron/cli.test.ts` file still contains pre-existing graceful-kill timeout flakes under load; this PR does not claim to resolve those unrelated baselines.

---
Part of #1160.


iamtoruk:
LGTM — the argv carve-out (`argv[2] !== 'serve'`) matches main.js's pre-Commander serve dispatch exactly, and the stdin `EPIPE` guard correctly falls back to a one-shot spawn (the triggering request recovers, which is a bit stronger than the test asserts). Root cause, not symptom.

One minor, non-blocking nit: the new package-time smoke launches real resident `serve` inheriting the current env, so it arms provider watchers against the developer's real data dir during packaging. It's read-mostly and killed on the ready frame, but pointing it at an isolated `HOME`/data dir would keep packaging fully hermetic.

avs-io:
Thanks — the hermeticity nit was correct and is addressed at `20f3d290`.

The packaged resident smoke now creates a disposable root, supplies a minimal launch environment, points HOME/USERPROFILE/APPDATA/LOCALAPPDATA/XDG and CodeBurn cache/session paths at that root, and removes it in `finally`. It can no longer arm provider watchers against the packager's real corpus.

Focused fallback coverage, typecheck, the exact `stage-cli` packaging gate, and all GitHub checks pass. The PR description now records the updated head and verification.

PR #1157, opened 2026-08-28
perf(desktop): show exact totals while details refresh

avs-io:
## Why

On a large corpus, changing Today / 7D / 30D / Month / 6M / Life can make the Desktop app feel frozen while it rehydrates every detailed panel. It could also briefly display a total from the previous period, and the long activity timelines were difficult to read or opened on old dates.

## What changed

- paint the last exact cached headline immediately while the selected view refreshes in the background
- show honest progress copy and drill-down skeletons instead of a blank or misleading view
- pre-warm the standard time horizons without creating a provider-by-period parse storm
- key data, errors, and late responses to the selected period/provider so totals cannot leak across filters
- cap the session memo at 96 least-recently-used entries so date-boundary keys cannot grow forever
- invalidate in-memory and persisted progressive data safely after configuration changes
- reject partial resident hydration as an exact snapshot and keep boundary-sensitive Today/Month keys current
- keep weekday labels fixed, month labels aligned to the scrolling heatmap, and the newest dates visible first
- make long daily-spend labels readable

## Verification

- exact head: `fdae4b0f764d5067ee13bee076830e207705ab9a`
- focused renderer polling + app regressions: 64/64 passed
- Desktop tests excluding the known baseline CLI lifecycle file: 546 passed
- Desktop typecheck: passed
- exact arm64 package build: passed
- strict codesign verification: passed (ad-hoc local review signature)
- packaged Desktop click-through on this branch:
  - Today exact cached headline: `$75.40`, 765 calls
  - 30D: `$14,200.90`, 136,883 calls
  - 7D: `$2,906.60`, 28,392 calls
  - 30D and 7D remained distinct; the previous-period total did not leak
  - a settled 30D to 7D click plus accessibility capture completed in under one second
- arm64 DMG SHA-256: `8e64bced01db75a331295b35ef1571f8542446da98705f08a2e6b9c8fbf37f8f`
- arm64 ZIP SHA-256: `0adf43a0d56803abd40a3f5febe2a248e1cd376886cce6e2c1cea5e0d8d71357`
- `git diff --check`: clean

The capture round trip includes accessibility-tree and screenshot overhead; it is not presented as a pure renderer latency benchmark.

## Review notes

This PR changes only Desktop presentation, progressive caching, and request coordination. It does not change parser totals, price tables, sync semantics, or Teams behavior.

The locally built review app is intentionally not notarized, so Gatekeeper rejection of that development artifact is expected and is not a product regression.

---
Part of #1160.


iamtoruk:
Re-reviewed after the "preserve filter responsiveness" commit — the count-up replay and the provider-warm regression are both addressed now (`animate={!suppressHeroReplay}` scoped to the snapshot→live handoff via `pendingSnapshotHandoffRef`, and the other standard providers are warmed again). Nice.

One small item left, non-blocking:

**Unbounded module memo + stale comment.** `memoStore` in `app/renderer/hooks/usePolled.ts` is still "kept for the whole app session … never evicted," and the header comment says the keys are "bounded by (section × period × provider × range), so the map cannot grow without limit." That's no longer accurate: `overviewMemoKey` embeds the local date for `today` and the year-month for `month`, so a long-running session accrues one payload per standard key per day. It's slow and each entry is small, but it is unbounded, and the comment now reads the opposite of the behavior. Either add a light cap/TTL eviction (mirroring `overviewSnapshot`'s `MAX_ENTRIES`) or update the comment to describe the date-boundary growth honestly.

The leak-prevention mechanics all check out: synchronous key-mismatch masking, the epoch guard, per-key error masking, the Today/Month boundary baked into the key, config-change invalidation of both caches, and the `hydration.complete === false` refusal in `writeOverviewHeadline`.

avs-io:
Thanks — the unbounded memo finding was correct and is addressed at `fdae4b0f`.

`memoStore` is now a 96-entry LRU: reads and writes refresh recency, the oldest entry is evicted at the cap, and regression coverage proves both eviction and retention of recently used keys. The stale bounded-key comment was replaced with the actual policy.

The focused renderer polling/App suite passes 64/64, typecheck passes, and all GitHub checks are green. The PR description now records the updated head and verification.

PR #1159, opened 2026-08-28
perf(tui): show cached periods before refreshing sources

avs-io:
## Outcome

Makes the interactive TUI useful before a large local corpus has been fully re-indexed.

- paints an exact cached period first, clearly labelled as cached while source refresh continues
- makes Today, 7D, 30D, Month, 6M, and Lifetime immediately selectable from one normalized cached index
- stages cold-cache readiness in that order while parsing source contents once
- falls back from Today to 7D only when Today is exactly empty across calls, cost, savings, and tokens; unpriced/token-only activity remains on Today
- preserves provider and period truth across rapid navigation, custom ranges, day mode, and in-flight reloads
- keeps cached snapshots out of the parse memo, then atomically replaces the index after reconciliation
- preserves real q and Ctrl-C process termination

This is deliberately limited to the interactive TUI path. It does not change CLI JSON/CSV, Desktop, web, pricing, sync, or Teams semantics.

## Exact provenance

- base: `b79cf40169e4cde931aae10f50bf20cbe8aff49e`
- head: `c8c3f60af9607437fe885e13daa38d555819d282`

## Heavy-corpus evidence

Independent same-HOME Expect/PTTY first-visible comparison against exact built artifacts:

- base: 20,953.2 ms
- original progressive candidate: 1,891.5 ms

An alternating implementation check observed base runs of 27,260.4 / 25,010.3 ms and candidate runs of 2,283.3 / 3,893.7 ms (median 26.14 s -> 3.09 s, about 88% faster). These are acceptance observations on one Mac and corpus, not a general benchmark distribution.

## Verification

- exact-zero Today fallback and provider/day/custom-range regressions: 13/13 passed
- full suite: 253 files passed, 2 skipped; 3,381 tests passed, 5 skipped
- cache-lock family: 34/34
- TypeScript: clean
- production build: clean
- diff check and worktree: clean

## Review notes / follow-ups

These are intentionally left out of this reviewable slice:

- provider switching still rebuilds a provider-correct index; it needs its own real-corpus speed proof before claiming instant switching
- heavy detail tabs suppress automatic refresh and may age until navigation/manual refresh
- cold tiering repeats discovery/fingerprinting per readiness phase; preserve per-phase timing artifacts before optimizing this further
- the snapshot memo guard is defensive hardening; it is not presented as a reproduced user-facing stale-refresh fix

The larger provider-cache prototype is separately held: a 109.6 MiB shard made `JSON.parse` alone exceed the 250 ms interaction budget and independent review found filter-parity gaps. None of that unaccepted WIP is in this PR.

No issue or PR is auto-closed by this draft.

---
Part of #1160.

### DIFF
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -10,7 +10,8 @@ import { parseAllSessions, filterProjectsByDateRange, filterProjectsByName, setI
 import { isColdCacheOnDisk } from './session-cache.js'
 import { findUnpricedModels, isExpectedFreeModel, loadPricing } from './models.js'
 import { aggregateModelTotals } from './model-breakdown.js'
-import { buildDurablePeriod } from './usage-aggregator.js'
+import { buildDurableOverviewFromNormalizedIndex, buildDurablePeriod, hydrateDailyCacheFromNormalizedProjects } from './usage-aggregator.js'
+import { loadDailyCache, type DailyCache } from './daily-cache.js'
 import { getAllProviders } from './providers/index.js'
 import { classHeaderLine, classTotals, findingBasis, findingClass, scanAndDetect, type FindingClass, type WasteFinding, type WasteAction, type OptimizeResult } from './optimize.js'
 import { appliedFixGlyph, formatAppliedFix, type AppliedFix } from './act/types.js'
@@ -34,6 +35,7 @@ export type DailyActivityRow = {
 
 export const DAILY_ACTIVITY_PAGE_SIZE = 10
 export const INDEX_PROGRESS_TICK_MS = 1000
+export const LARGE_HISTORY_FILE_THRESHOLD = 1000
 export const INTERACTIVE_RENDER_OPTIONS = { alternateScreen: true } as const
 export const RESIZE_DEBOUNCE_MS = 150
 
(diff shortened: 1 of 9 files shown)


iamtoruk:
Reviewed the latest head. One blocking issue before this can go in, plus a merge-order note.

**Blocking — the `p` (provider-cycle) key wedges the dashboard in `--day` and custom-range (`--from/--to`) modes.**

In `src/dashboard.tsx` the `p` handler unconditionally does:

```
setProjects([]); setDurable(undefined); setLoading(true); setIndexing(true)
if (dayDate) void reloadData(period, next, dayDate)
```

It relies on the background-index effect to later clear `indexing`, but that effect early-returns when `!initialHistoryIndexing || initialHistoryIndex || isCustomRange || isDayMode`, and `initialHistoryIndexing` is `progressive`, which is `false` whenever `--day` or `--from/--to` is set (`progressive = isTTY && dayRange == null && customRange == null`). The only `setIndexing(false)` lives in that effect's `finally`, so it never runs in these modes.

Repro: `codeburn --from <a> --to <b>` interactive with ≥2 providers detected, press `p`:
- projects cleared, `loading` stays true forever, no reload is issued (`dayDate` is null),
- the "cached … source refresh pending" banner sticks,
- auto-refresh is disabled (gated on `indexing`),
- every period key then dead-ends on the `if (indexing) return` guard.

In `--day` mode the data does reload (dayDate is set) but `indexing` still never clears, so the banner sticks and all period switches stay dead.

Suggested fix: only enter the indexing path from `p` when the index effect can actually run — gate on `initialHistoryIndexing && !isCustomRange && !isDayMode`; otherwise keep the plain `reloadData(period, next, dayDate ?? null)` path so custom-range/day reloads normally without arming indexing.

**Non-blocking / merge-order:** this PR and #1156 both touch `src/dashboard.tsx` and `tests/dashboard.test.ts` (this one changes `INTERACTIVE_RENDER_OPTIONS`, which #1156's tests assert against). Whichever lands second will need a rebase + retest.

The rest looks good — the normalized-index gap-fill keeps the durable cache authoritative (no double-count), the cached headline is honestly labelled, and the q / Ctrl-C exit paths match the #1143/#1144 contract (and `hardQuitGuard` is correctly detached in the `finally`).

avs-io:
Thanks — the provider-cycle wedge was real and is fixed at `c8c3f60a`.

Day/custom-range provider switches now take the viable direct reload path instead of arming background indexing that cannot run. Regression coverage exercises both modes and verifies that indexing clears and subsequent period keys remain usable.

The same commit adds the requested empty-Today behavior: the TUI falls back to 7D only when Today is exactly zero across calls, cost, savings, and tokens. Unpriced/token-only activity correctly remains on Today.

The focused regressions pass 13/13, the full suite passes 3,381 tests with 5 intentional skips, and all GitHub checks are green. The PR description now records the corrected head and verification.

Issue #1160, opened 2026-08-28
Epic: make CodeBurn's local interactive surfaces fast, truthful, and resilient


avs-io:
## Outcome

Make CodeBurn's local interactive surfaces feel fast, truthful, and resilient on both normal and heavy corpora.

A user should see useful, correctly labelled data quickly; changing period or provider must never leak stale totals; terminal and Desktop surfaces must remain responsive; packaged resident failures must recover without breaking the app.

## Why these belong together

Each change remains an independently reviewable PR, but all slices came from the same 0.9.22 dogfood acceptance audit and share one user-facing quality bar. This issue is the acceptance ledger, not a request to merge them as one change.

CodeBurn Teams, plugin architecture, pricing semantics, sync policy, and feature-roadmap work are explicitly out of scope.

## Merged slices

- [x] #1155 — packaged Desktop resident CLI routing and fallback reliability
- [x] #1156 — terminal resize redraw / stale-panel cleanup
- [x] #1157 — Desktop exact totals first while detailed views refresh
- [x] #1159 — TUI cached-period first paint and progressive source reconciliation
- [x] #1161 — Menu Bar parked status-item recovery

## Review-ready draft slices

- [ ] #1163 — durable last-known-good Desktop destinations across restarts
- [ ] #1164 — repeatable 60-case release acceptance ledger and runner
- [ ] #1165 — TUI cache-lock cleanup on confirmed exit and interruption
- [ ] #1166 — Desktop resident/provider process-tree cleanup on real app quit

## Shared acceptance gates

- [ ] cached values are explicitly labelled until reconciled with source
- [ ] no stale totals cross period or provider boundaries
- [ ] already-indexed standard period navigation settles within 1 second on the frozen heavy corpus
- [ ] exact-SHA heavy-corpus first-useful timings are recorded for affected surfaces
- [ ] settled terminal resize leaves no ghost panels or redraw artifacts
- [ ] packaged resident failure falls back cleanly without EPIPE or a dead surface
- [ ] packaged app quit leaves no owned CLI/provider process or cache lock
- [ ] focused automated suites, full regression suite, and build/type checks pass
- [ ] installed-artifact manual visual, navigation, exit, and recovery gates are recorded before closing this epic

## Deliberately held, not hidden

[...]

PR #1163, opened 2026-08-28
perf(desktop): keep every destination useful across restarts

avs-io:
## Why

The merged progressive Overview work makes the headline useful quickly, but a long-open or restarted Desktop app can still greet Sessions, Spend, Optimize, Models, Compare, and Plans with a blank skeleton. Users should see the last exact result for the same period/provider/range immediately while CodeBurn revalidates it.

## What changed

- persist versioned, compressed report snapshots for every major Desktop destination;
- bound the in-memory LRU by both count (96) and serialized weight (12M characters), and bound durable storage to 72 snapshots;
- pre-cap restart persistence at 300k raw JSON characters so multi-megabyte reports are never synchronously compressed on the renderer thread;
- skip recompression and storage writes when a poll returns an unchanged report;
- reject stale and partially hydrated results from restart persistence while retaining them as last-good data in the current renderer;
- key Today and Month snapshots to local calendar boundaries so yesterday's answer cannot appear under today's label;
- paint an exact same-key snapshot immediately, label it as cached, and revalidate once after a renderer restart;
- retain Chromium's normal background throttling so Page Visibility and animation pausing work, while leaving data intervals registered and catching up on return;
- warm Today -> 7D -> 30D -> Month -> 6M -> Life serially at background priority, including each horizon's local first-click reports;
- preserve the reviewed current-period provider warm after the higher-value period/report queue;
- keep quota endpoint discovery user-driven (opening Plans), rather than triggering provider network calls from general launch warming;
- make the footer describe the selected destination and format old snapshots in hours/days;
- invalidate report generations after settings changes and reject old-config requests that resolve after invalidation.

This does **not** change parser totals, pricing, sync semantics, credentials, or Teams behavior. Renderer storage contains report output only. The separate universal provider-index prototype remains excluded because it failed responsiveness and exactness gates.

## Maintainer-review corrections

The review at the prior head correctly found that `backgroundThrottling: false` defeated the Page Visibility design and that unchanged large reports could be recompressed on every cadence. Head `c2350d6e` fixes those defects plus the partial/stale persistence and generation-race findings. The review's concern that the durable tests might not execute was independently falsified: they run under jsdom and are included in the passing focused suite. App tests are not yet enforced by the repository's root GitHub workflow; that is tracked as an acceptance-ledger follow-up rather than being hidden by green root checks.

## Verification

- base: `81f1eb3d6ae342baa6114ee5a725715af200fb24` (current `upstream/main` at validation)
- exact head: `c2350d6e01690de9b2ea9e03f40769bfbb082229`
- focused snapshot/visibility/App tests: **78/78 passed**
- complete Desktop suite excluding unchanged host-sensitive `electron/cli.test.ts`: **581/581 passed**
- complete Desktop suite including that file: **655 passed, 2 unchanged macOS SIGTERM-observation tests failed**; the same tests fail on the merged baseline on this Mac, and this PR does not modify `electron/cli.ts`
- Desktop typecheck: passed
- production renderer/electron build: passed
- packaged through the intended `package:arm64` entrypoint from the exact clean head: passed (arm64 and configured x64 artifacts emitted)
- strict deep codesign verification of the arm64 app: passed (ad-hoc local review signature; Gatekeeper rejection remains expected until Developer-ID signing/notarization)
- bundle: `org.agentseal.codeburn-desktop`, version `0.9.22`
- `git diff --check`: clean

Same-Mac synthetic compression calibration (observational, not a corpus claim): 250,011 raw characters took 97.0 ms; 500,011 took 258.9 ms. The 300k pre-cap is intentionally below the measured 250 ms boundary. The representative 272,891-character report fixture compressed in 34.1 ms.

Exact packaged artifact SHA-256:

- arm64 DMG: `8f340d84dc7282a8a4433cadac9cd14612b5de29cb9e3f9ce3a3806dc3ad4869`
- arm64 ZIP: `24321d9750a991bff65620918bab506205f65ac37a39f3dacf21d904b7173824`
- bundled CLI: `d13a755ef5314e9f54d4f879fae2bcd139004c5f5942a534af04cbe74f56f31f`
- `app.asar`: `bcf69181d5756722cdee1ce79cda0c340794058afe71d6695b79ee733456b664`

## Remaining installed-product gate

This remains a draft until the exact packaged app is exercised on the frozen heavy corpus for restart paint, cached/revalidating labels, minimized energy behavior, background CPU/memory, cancellation pressure during rapid switches, and click latency. Source tests prove cache custody and routing; they are not a substitute for that installed-surface gate.

Part of #1160.


iamtoruk:
Reviewed in depth (two independent passes). The key identity, durable revalidation (`!cached.durable`), generation invalidation, and the add-only `background?` IPC wiring (all `backgroundIndex` values verified against the preload signatures, `range` preserved) are sound. Two substantive items before this should merge, plus a few nits.

**1. `backgroundThrottling: false` silently disables the Page Visibility API — animations never pause when minimized.**
Electron's own doc for this flag: *"Whether to throttle animations and timers when the page becomes background. This also affects the Page Visibility API."* With it false, `document.visibilityState` stays `visible` when the window is minimized/occluded, so `pageVisibility.ts` never adds `page-hidden`, and the `html.page-hidden … { animation-play-state: paused }` rule in `plain.css` never fires. The result is the opposite of the comment "Animations remain separately visibility-gated in the renderer": a minimized app keeps running flame/shimmer animations (continuous compositor wakeups) **and** now polls the CLI every cadence tick. The `visibilitychange` catch-up in `usePolled` also becomes dead code. Suggest either keeping `backgroundThrottling: true` + the old hidden-tick guard, or gating animations on a signal that survives the flag and stating the battery cost explicitly. As written the PR misstates its own tradeoff.

**2. Persistence write path is main-thread jank on a perf PR.**
`memoSet` runs on every successful poll and (a) checks the 2.5M-char cap **after** `compressToUTF16`, so a heavy payload (e.g. Sessions/Lifetime) is fully compressed — measured ~2s on the renderer main thread — before it can be rejected; (b) has no dirty-check, so unchanged data is re-stringified + re-compressed every 30s cadence tick (and, per item 1, even while minimized); (c) `durableSnapshotCount`/`durableSnapshotRows` JSON-parse every stored envelope body on every write just to count/sort. Cheap fixes: a pre-compression length gate on `JSON.stringify(value)`, persist-only-on-change (or a per-key persist throttle), and reading `at` without parsing the body (the `at` field is written first, so an anchored `/^\{"at":(\d+)/` match suffices).

**Nit — partial/stale payloads are persisted durably by the poll path.** The warm loop filters `hydration.complete !== false`, but `usePolled`'s own success path calls `memoSet` unconditionally, so a mid-index partial overview (or a `stale: true` payload) is written to durable storage under the exact key and painted on next launch. It self-heals (the payload carries its own hydration/stale flags so the banner renders, and durable entries always revalidate), but it's inconsistent with #1157's "never persist a partial resident-hydration total as exact" principle. Consider applying the same completeness filter before durable persistence.

Minor: (a) a fetch in flight when `clearPolledMemo()` runs resolves afterward and stamps the old-config value with the new generation — bounded by the forced revalidate, worth a guard or a note; (b) the Today warm slice calls `getQuota` on every launch (provider-endpoint network contact even if Plans is never opened) — worth an explicit sign-off given the local-first positioning; (c) the footer passes `loading=false` always and `refreshedLabel` has no hours unit, so a restart-painted snapshot reads "refreshed 1440m ago"; (d) model-savings/flat-rate/proxy-path have no desktop mutation UI, so a config change there won't bump the snapshot generation — bounded by revalidate-on-view, worth a doc line; (e) please confirm CI actually executes the new `usePolled.test.ts` durable tests green — they use a bare `localStorage` that is undefined under the repo's jsdom setup locally, so the durable path may be silently untested.

avs-io:
Thanks — the two substantive findings were correct and are addressed at `c2350d6e`.

- Restored normal Chromium background throttling, so Electron now emits real Page Visibility changes and the existing `page-hidden` animation pause/catch-up path is live.
- Durable writes now have a 300k raw pre-cap, content fingerprinting (unchanged polls do not recompress/rewrite), metadata-only header reads, and a 12M-character/count-bounded memory LRU.
- Stale/partial reports remain current-renderer last-good data but cannot become restart snapshots.
- Old-config requests resolving after `clearPolledMemo()` are fenced out.
- Launch warming no longer calls quota endpoints; quota refresh stays tied to the Plans surface.
- Footer copy now uses hours/days.

The localStorage test-environment concern did not reproduce: the corrected focused suite is 78/78 and the complete Desktop suite excluding the two unchanged host-sensitive SIGTERM observations is 581/581. Typecheck, production build, exact-head packaging, and strict codesign verification pass. The PR body now carries the exact head, new artifact hashes, an observational compression calibration, and the remaining installed heavy-corpus gate. It stays draft until that gate is complete.

PR #1165, opened 2026-08-28
fix(tui): clean cache locks on confirmed exit

avs-io:
## What changed

- releases both cache-lock families before the TUI's direct process exit
- preserves the existing first-`q` confirmation and second-`q` fast exit while indexing
- uses conventional exit status 130 for Ctrl-C and 0 for `q`
- adds real child-process coverage for lock ownership and terminal restoration

## Why this is not a timing artifact

The initial small fixture sometimes completed hydration in the 50 ms between the two `q` keystrokes, so it could pass on unmodified main. The final test holds a real 60 MB / 2,000-file cold hydration open, waits until the first-`q` confirmation is visibly rendered, and verifies the owned lock is still live before the second key.

On the original exact base `81f1eb3d6ae342baa6114ee5a725715af200fb24`, the strengthened process test failed only because `hydrating.lock` remained after exit. This branch removes it while preserving terminal teardown.

## Current exact candidate

- rebased conflict-free on current main `9a929c87ab94b55079e53ff6cb29292c0317402c`, including merged #1163 and #1164
- exact head `38048078753376e41322d531e9a4fb099f39a310`

## Validation after rebase

- isolated direct-exit process boundary: 2/2
- serialized lock suite: 35/35
- full root suite, run without cross-suite contention: 258 files passed, 2 skipped; 3,482 passed, 5 skipped
- TypeScript: `tsc --noEmit`
- `git diff --check`
- process-level assertions: raw mode off, mouse tracking off, alternate screen off, cursor shown

An intentionally adversarial simultaneous full-suite + lock-suite run produced one transient leftover-lock assertion; the exact process test passed in isolation and the full root suite passed when rerun without the second lock suite competing. This is recorded as harness contention rather than hidden.

## Adversarial review

An independent Grok 4.6/high review was run against the original diff. It raised a possible hard-quit/Ink teardown ordering race, but repeatedly cited nonexistent file paths. The underlying claim was checked against the real TypeScript, vendored Ink lifecycle, and child-process teardown assertions. The hard-quit ordering predates this patch; this patch changes only the exit target from direct `process.exit` to ownership-checked cleanup plus the same exit. No new race reproduced.

Related to #1160.

### DIFF
--- a/src/cache-refresh-lock.ts
+++ b/src/cache-refresh-lock.ts
@@ -89,6 +89,13 @@ function removeOurLockSync(): void {
   }
 }
 
+/** Release every warm-refresh lock owned by this process before a caller uses
+ * `process.exit()`. Signal handlers already cover SIGINT/SIGTERM; the TUI's q
+ * path exits directly after restoring the terminal and needs the same cleanup. */
+export function releaseOwnedRefreshLocksForExit(): void {
+  removeOurLockSync()
+}
+
 // Arm once, only while we hold the lock: on a catchable termination (Ctrl-C, or
 // the menubar/desktop watchdog's SIGTERM) clean our lock before dying so a
 // killed refresh leaves no leftover. SIGKILL can't be caught, so that path still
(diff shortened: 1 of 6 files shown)


avs-io:
Updated after the latest upstream merges and revalidated on exact current ancestry.

- base now includes merged #1163 and #1164 at `9a929c87`
- exact head: `38048078753376e41322d531e9a4fb099f39a310`
- full root suite: 3,482 passed / 5 skipped
- lock suite: 35/35
- the new required Desktop CI job also passed, along with every other GitHub check

The PR body records the one deliberately induced cross-suite-contention falsifier and the clean isolated/full reruns, so the evidence is not being presented more strongly than it is.

PR #1201, opened 2026-08-31
fix(plan): estimate Copilot AI credits from tokens and say when the total is incomplete (#1199)

### LINKED ISSUE #1199: Codeburn undercounts GitHub Copilot Business AI Credits
## Summary

Codeburn is discovering my local Copilot usage and the correct Copilot credit plan, but the calculated AI-credit usage is much lower than what GitHub reports.

## Current hypothesis

The only raw events with `total_nano_aiu` are for the `claude-haiku-4.5` model. I suspect these may be coming from a personal Copilot Free account rather than my company-managed Copilot Business account.

Most of my other Copilot usage appears to come from a GitHub Enterprise Copilot account. Those sessions are discovered and included in the token and cost totals, but they do not appear to contain `nanoAiu` or `total_nano_aiu` in Codeburn's normalized export.

I am not sure whether the Enterprise Copilot service does not expose this field, or whether Codeburn is dropping it while parsing or normalizing sessions. Could the implementation or documentation clarify whether AI-credit metadata is expected for Enterprise Copilot sessions?

## Environment

- macOS `26.5.2`, Apple Silicon `arm64`
- VS Code `1.135.0`
- Built-in GitHub Copilot extension `0.63.0` (`github.copilot-chat`)
- GitHub Copilot CLI `1.0.82`
- Codeburn `0.9.23`
- GitHub Copilot Business seat managed by the company I work for, with a `20,000` AI-credit cycle
- Codeburn accounting profile: `copilot-max`

## Steps taken

1. Configured the closest available equivalent plan because it matches the same 20,000-credit cycle:

   ```sh
   codeburn plan set copilot-max
   ```

2. Refreshed the menubar:

   ```sh
   codeburn menubar --force
   ```

## What works

Codeburn discovers and parses local Copilot sessions:

```text
codeburn doctor --provider copilot
Copilot   Sessions 164   Parsed 8/8+   Cached 164   OK (164 sessions)
```

The configured Copilot accounting profile is visible in `codeburn plan`:

```text
copilot: Copilot Max (copilot-max)
  Budget: 20000 AI Credits
  Reset day: 1
```

## Discrepancy

GitHub currently reports roughly:

```text
19,326 / 20,000 AI credits used
```

Codeburn reports for the configured `copilot-max` plan:

```text
spentCredits: 3.925845
budgetCredits: 20000
creditsIncomplete: true
```

For August 2026, Codeburn sees:

- `54` sessions and `473` normalized Copilot request records
- `0/473` exported records with `nanoAiu` or `total_nano_aiu`
- only `4` raw usage events with `total_nano_aiu`
- those four events total `3.925845` AI credits
- estimated token-equivalent cost: `€88.06`

The raw events are in:

```text
~/.copilot/session-store.db
```

The `creditsIncomplete: true` flag means Codeburn found usage without the billing metadata needed to calculate AI credits. It looks like most of the locally discovered usage, including VS Code usage, cannot currently be converted into Copilot credits.

The plan was created at `2026-08-31T07:33:30.195Z`, but Codeburn filters by the plan reset period rather than the plan creation timestamp. So this does not look like it is just caused by counting only sessions created after `codeburn plan set copilot-max`.

## Detailed diagnostics

The current Codeburn plan result uses this period:

```text
periodStart: 2026-07-31T22:00:00.000Z
periodEnd:   2026-08-31T22:00:00.000Z
```

The complete Copilot plan result is:

```text
monthlyCredits: 20000
spentCredits: 3.925845
budgetCredits: 20000
creditsIncomplete: true
monthlyUsd: 200
spentApiEquivalentUsd: 0.03925845
```

The audit sees `473` Copilot calls across `13` models, with `46,642,530` input tokens and `370,880` output tokens. The API-equivalent cost is `€88.06` in the current export.

The raw billing table contains only `4` events across `2` sessions and `1` model:

```text
model             events  input_tokens  output_tokens  credits
claude-haiku-4.5  4       74452          406             3.925845
```

This shows that the token usage parser is seeing much more data than the AI-credit parser. The report should clarify whether `total_nano_aiu` is expected to be available for VS Code sessions, and if so, why those fields are not being captured or propagated into Codeburn's plan calculation.

## Local Copilot database inventory

I checked the local Copilot and VS Code storage locations for additional SQLite databases and session stores.

The relevant databases are:

```text
~/.copilot/session-store.db
~/.copilot/session-state/<session-id>/session.db
~/Library/Application Support/Code/User/globalStorage/github.copilot-chat/session-store.db
```

The main `~/.copilot/session-store.db` is the only database with an `assistant_usage_events` table. It contains `4` August 2026 usage events, all with `total_nano_aiu`.

The Copilot CLI session-state database contains only `todos`, `todo_deps`, and `inbox_entries`. It has no assistant usage events or billing fields.

The VS Code Copilot database contains `87` sessions, `767` turns, and `1,408` session-file references. It has sessions, turns, checkpoints, and file references, but no `assistant_usage_events` table and no `nanoAiu` or `total_nano_aiu` column.

I also found `158` VS Code `chatSessions/*.jsonl` files, of which `67` were modified in August 2026. None of the structured JSON objects in those files contained a key matching `nanoAiu`, `aiu`, or `credit`.

The workspace storage also contains `28` `state.vscdb` files. Their relevant entries are editor and extension state such as `GitHub.copilot-chat`, `github.copilot-github`, and `github.copilot-github-enterprise`. I did not find any additional Copilot billing database there.

Based on this inventory, I did not find another local GitHub Copilot database that contains the missing Enterprise usage credits. The VS Code sessions are stored separately, but their local records do not appear to include the raw AI-credit field that Codeburn's plan calculator currently requires.

## Expected behavior

- Copilot Business AI-credit usage from all supported local sources should be included when possible.
- Missing billing metadata should be reported clearly, rather than presenting a partial total as complete.
- Codeburn should document which sources and fields are included in Copilot credit accounting.
- Please clarify whether `total_nano_aiu` is expected for Enterprise Copilot sessions, and investigate why the sessions associated with my Enterprise account do not expose it if it should be available.

## Diagnostic commands

```sh
codeburn status --provider copilot --format json --period month --no-optimize --no-timeline | jq '.plans.copilot'
```

```sh
sqlite3 -header -column \
   "$HOME/.copilot/session-store.db" \
   "SELECT
       COUNT(*) AS events,
       SUM(CASE WHEN total_nano_aiu IS NOT NULL THEN 1 ELSE 0 END) AS with_nano_aiu,
       SUM(CASE WHEN total_nano_aiu IS NULL THEN 1 ELSE 0 END) AS without_nano_aiu,
       SUM(COALESCE(total_nano_aiu, 0)) / 1000000000.0 AS credits
    FROM assistant_usage_events
    WHERE created_at >= '2026-08-01'
       AND created_at < '2026-09-01';"
```

The database query returns:

```text
events  with_nano_aiu  without_nano_aiu  credits
4       4              0                 3.925845
```

Additional commands used to verify the coverage:

```sh
codeburn audit --provider copilot --period month --format json
```

```sh
codeburn export --provider copilot --format json --from 2026-08-01 --to 2026-08-31 -o /tmp/codeburn-copilot-august.json
```

The August export contains `473` Copilot records, but none contains `nanoAiu`, `total_nano_aiu`, or another AI-credit field.

iamtoruk:
Fixes #1199.

## Problem
Only the Copilot CLI's `~/.copilot/session-store.db` rows carry an exact credit value (`total_nano_aiu`). VS Code chat sessions, transcripts, the OTel database, JetBrains stores and the CLI session-state JSONL never do, so on a typical machine most Copilot requests have no exact figure. `copilotCreditSpend` summed only the exact rows and set `creditsIncomplete`, but that flag was a bare boolean in JSON: the `codeburn plan` headline showed a confident `3.9 / 20000 AI Credits` with no caveat while GitHub showed 19,326.

## Fix
- Requests without an exact figure are estimated from tokens at the model's listed API rate, 1 credit = $0.01. Verified against real session-store rows: that is exactly how GitHub bills credits (list rate with the cache-read/write split; `request_multiplier` is the legacy premium-request weight and never touches credits). On Copilot CLI events CodeBurn's pricing landed within about 15% of the exact bill; VS Code sessions record no cache split, so their estimate can land either side.
- New add-only plan fields: `estimatedCredits`, `creditRatedCalls`, `creditUnratedCalls`, `creditsNote` (`spentCredits` and `creditsIncomplete` unchanged in meaning).
- Headline says so whenever anything is estimated: `Copilot Max: ~9800 / 20000 AI Credits (estimated; 4 of 473 requests carry GitHub's exact figure)`; unchanged when every request is exact.
- The bar and `percentUsed` follow the estimate while any request is unrated (a bar at 0% beside a real 19k bill was the misleading picture), and the exact figure once every request carries one.
- A session that carries any exact figure is never estimated on top of, so paired store rows and JSONL twins cannot double count (tested).
- `docs/providers/copilot.md` documents which sources carry exact credits and how to read the note. The menubar payload carries no plan data, so it is untouched.

## Verification
- `npx tsc --noEmit` clean; `npm test`: 3905 passed, 5 skipped (9 new tests: estimate math, rated/unrated counts, pairing no-double-count, both headline states, JSON fields end to end).
- Dev run on a real machine: 11 Copilot CLI requests, `spentCredits` 9.334625 matches the session-store sum exactly, `creditsIncomplete` false, note reads complete.

Not in scope here: using the live GitHub quota figure as the authoritative number when a token is available (follow-up once the endpoint's AI-credit snapshot is confirmed).


iamtoruk:
Follow-up scope note: checked GitHub's quota endpoint on a live account after the June 2026 billing change. It reports only chat, completions and premium_interactions snapshots, no AI-credit figure, so a live authoritative credit number is not available from that endpoint today. The token-priced estimate in this PR stays the local best effort.

PR #1202, opened 2026-08-31
feat(mac): add Z.ai Coding Plan quota

MiguelMachado-dev:
## Summary

- add a native Z.ai Coding Plan quota adapter for the macOS menubar app
- show the 5-hour and weekly credit windows in Settings and Capacity Dock
- reuse an existing Pi Z.ai login read-only, with a provider-scoped Keychain override as fallback

## Testing

- [x] Tested locally against a real Z.ai Coding Plan account; the Pro tier and both quota windows loaded
- [x] `SDKROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX26.5.sdk mac/Scripts/package-app.sh 0.9.23.1`
- [x] `codesign --verify --deep --strict mac/.build/dist/CodeBurnMenubar.app`
- [x] focused Swift harnesses covered response decoding, live credential discovery, the real endpoint, and Capacity Dock dispatch
- [ ] Full `swift test --package-path mac` is blocked on this machine because its Command Line Tools install lacks `XCTest`, `TestingMacros`, and `SwiftUIMacros`; the universal app target builds successfully and macOS CI has the full test toolchain

### For new providers only

This is a live quota adapter, not a session-discovery provider. The session/cost parser checks in the template do not apply.

Sanitized live result:

```text
Z.ai service harness passed: Pro, windows=[5-hour, Weekly]
```


iamtoruk:
Reviewed and verified locally. Fetched the branch, ran the full gate on a machine with the complete toolchain: swift build clean, swift test 347 tests in 44 suites all passing (baseline on main is 341, your 6 new tests included), branch is based on current main. The endpoint shape matches what three independent community tools use for the GLM Coding Plan quota (opencode-glm-quota, GLM Usage Monitor, pi-glm-usage), and the unit=3 five-hour / unit=6 weekly mapping agrees with them.

Things I checked and liked: Deps-injected fetch with an ephemeral cookie-less session, the Pi login read is read-only through SafeFile with a synthetic-home test, terminal vs transient classification is wired into the dock's existing failure model, and the dispatch tests keep the provider-scoped key isolation honest.

Two small notes, neither blocking:
- The Authorization header sends the raw key without a Bearer prefix. Your live harness proves Z.ai accepts it, and the test pins it, so fine as is; if Z.ai ever tightens this, the fix is one line.
- The percentage fallback divides currentValue by the field named usage, which per your fixtures is the cap. The naming is Z.ai's, not yours, but a one-line comment saying usage means the window cap would save the next reader a double take.

Thanks for the sanitized live evidence and for the honest note about the blocked local suite; running it here was exactly what that note made easy.

iamtoruk:
One finding from a second live test before merge: Z.ai signals bad auth in the response body with HTTP 200, e.g. {"code":401,"msg":"token expired or incorrect","success":false}. The adapter only reads the HTTP status, so a body-level rejection surfaces as the generic "unrecognized quota response" instead of "Z.ai rejected this API key". Worth a fast follow-up: read code and success before parsing limits, and map body 401/403 to authenticationRejected. Merging as is since this only affects the error message, not working accounts.

PR #1203, opened 2026-08-31
fix(menubar): map Z.ai body-level auth rejections to real errors

iamtoruk:
Follow-up to #1202, as noted in review there.

Z.ai signals bad auth inside an HTTP 200 body:

```json
{"code": 401, "msg": "token expired or incorrect", "success": false}
```

The adapter only classified the HTTP status, so a body-level rejection fell through to "Z.ai returned an unrecognized quota response" (observed in a live test with a bad key). The decoder now reads the body envelope first: `success: false` with code 401/403 maps to authenticationRejected, 429 to rateLimited, anything else to providerUnavailable.

Verification: `swift build` clean, `swift test` 348 tests in 44 suites pass (4-case body-rejection test added).


iamtoruk:
Closing as duplicate: #1204 landed the same body-level auth mapping from the adapter's author. The extra 429/unavailable body mapping here can ride along in a future Z.ai touch if it ever shows up in practice.

PR #1204, opened 2026-08-31
fix(mac): map Z.ai body authentication errors

MiguelMachado-dev:
## Summary

- inspect Z.ai's body-level `code` and `success` fields before parsing quota limits
- map body codes 401 and 403 to `authenticationRejected`, even when HTTP status is 200
- keep other unsuccessful or malformed bodies classified as parse failures

Follow-up to #1202 and the maintainer's second live-test finding.

## Testing

- [x] red/green Swift harness reproduced HTTP 200 + body 401 returning `parseFailure`, then verified `authenticationRejected`
- [x] regression coverage for body codes 401 and 403
- [x] `SDKROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX26.5.sdk swift build --package-path mac --scratch-path /tmp/codeburn-zai-followup-build -c release --arch arm64`
- [x] `swiftc -parse mac/Sources/CodeBurnMenubar/Data/ZaiSubscriptionService.swift mac/Tests/CodeBurnMenubarTests/ZaiQuotaTests.swift`

The local Command Line Tools install still lacks the XCTest/Testing macro plugins required by the full Swift test target; macOS CI has that toolchain.

### DIFF
--- a/mac/Sources/CodeBurnMenubar/Data/ZaiSubscriptionService.swift
+++ b/mac/Sources/CodeBurnMenubar/Data/ZaiSubscriptionService.swift
@@ -110,6 +110,10 @@ enum ZaiSubscriptionService {
         guard let root = try? JSONSerialization.jsonObject(with: data) as? [String: Any] else {
             throw FetchError.parseFailure
         }
+        let bodyCode = jsonNumber(root["code"]).map(Int.init)
+        if bodyCode == 401 || bodyCode == 403 { throw FetchError.authenticationRejected }
+        if root["success"] as? Bool == false { throw FetchError.parseFailure }
+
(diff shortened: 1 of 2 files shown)


iamtoruk:
Verified on a machine with the full toolchain: swift build clean, swift test 348 tests in 44 suites all passing on your branch. The fix covers the reported case exactly (HTTP 200 body 401/403 now reads as a rejected key), and thanks for turning it around within minutes of the review note. Note #1203 fixes the same thing and will be closed as a duplicate if this one lands; the only delta there is mapping body 429 to rate-limited and other unsuccessful bodies to provider-unavailable instead of parse-failure, which can ride along in any future Z.ai touch if it ever matters in practice.

PR #1205, opened 2026-08-31
fix(mac): stop Capacity Dock hover drift

MiguelMachado-dev:
## Summary

- make AppKit the sole owner of Capacity Dock panel geometry
- disable `NSHostingView` intrinsic window resizing for the rail and detail panels
- prevent SwiftUI content-size updates from racing hover frame animations and accumulating position drift

## Testing

- [x] focused AppKit/SwiftUI harness: changing hosted content size leaves the controller-owned panel frame unchanged
- [x] `swiftc -frontend -parse mac/Sources/CodeBurnMenubar/CapacityDockController.swift`
- [ ] full Swift tests (local Command Line Tools installation lacks `SwiftUIMacros`; macOS CI has the complete toolchain)

### DIFF
--- a/mac/Sources/CodeBurnMenubar/CapacityDockController.swift
+++ b/mac/Sources/CodeBurnMenubar/CapacityDockController.swift
@@ -213,6 +213,8 @@ final class CapacityDockController {
         )
         let panel = CapacityDockPanel()
         let hosting = CapacityDockHostingView(rootView: view.environment(store))
+        // Panel geometry has one owner; intrinsic resizing races the frame animator.
+        hosting.sizingOptions = []
         hosting.autoresizingMask = [.width, .height]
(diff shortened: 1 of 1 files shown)


iamtoruk:
Verified on a machine with the full toolchain: swift build clean, swift test 348 tests in 44 suites all passing on your branch, which is based on current main. The reasoning holds up in the code: the controller owns panel geometry through ten setFrame call sites (the frame animator, drag, and detail-follow paths), while NSHostingView's default sizingOptions lets SwiftUI content-size changes resize the window underneath them, so two writers race exactly as described. Nothing relies on intrinsic sizing; both panels get their sizes from CapacityDockMetrics. Also running this build locally to confirm rendering is unaffected.

PR #1212, opened 2026-09-01
fix(mac): avoid provider presence deadlock

MiguelMachado-dev:
## Problem

Checking whether a provider's credentials are present could deadlock the menubar app — the presence check and the credential store's mutation path took the same lock in opposite order (reproducer: `presenceWritesDoNotDeadlockSwiftUICallbacks`).

## Fix

Reorder the lock acquisition in `CapacityDockProviderCredential` so presence reads never wait on the write path.

## Tests

- New `ProviderSettingsEditorStateTests` regression coverage (37 lines)
- Full menubar test suite passes locally (352 tests)

### DIFF
--- a/mac/Sources/CodeBurnMenubar/Data/CapacityDockProviderCredential.swift
+++ b/mac/Sources/CodeBurnMenubar/Data/CapacityDockProviderCredential.swift
@@ -193,30 +193,31 @@ enum CapacityDockProviderCredentialPresence {
         defaults: UserDefaults = .standard
     ) {
         guard CapacityDockProvider(rawValue: providerID) != nil else { return }
-        lock.lock()
-        var identifiers = providerIDsWithoutLock(defaults: defaults)
-        if present {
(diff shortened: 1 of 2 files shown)


MiguelMachado-dev:

iamtoruk:
Reviewed in depth and merging. Verified independently: your regression test fails on origin/main (both flags false, the writing thread is wedged) and passes on this branch, so the deadlock is real and the fix is at the root: the reentrant contains() from the synchronous UserDefaults notification now finds the lock free. Full suite 349 tests in 44 suites green. Two small things patched on top rather than sent back: a comment explaining that the main-queue hop is now what keeps the read-modify-write atomic (so nobody flips it to async later), and the test cleanup moved into defer with a 5 second bound so a loaded runner cannot fake a deadlock. One note on the description: the mechanism is single-thread reentrancy on a non-recursive NSLock, not two locks taken in opposite order.

PR #1213, opened 2026-09-01
fix(hermes): don't trust a $0 estimated cost — calculate from tokens

MiguelMachado-dev:
## Problem

Every Hermes session on a flat subscription (Codex plan, z.ai, Kimi) reports **$0.00** in CodeBurn despite real token usage.

Root cause: Hermes writes `estimated_cost_usd = 0.0` — not NULL — when its `cost_status` is `'unknown'` (no pricing data) or `'included'` (flat subscription, per Hermes's `usage_pricing.py`: "subscription-included; no provider invoice for usage"). `resolveHermesCost` treats any non-null estimate as a recorded amount, so the placeholder $0 wins over the token-based calculation forever.

I verified against a live `~/.hermes/state.db`: 17 sessions with full token counts, all `estimated_cost_usd = 0.0` with `cost_status` `'included'`/`'unknown'` and `cost_source = 'none'` — no measurement happened.

## Fix

Only trust **positive** recorded estimates; $0 falls through to `calculateCost`. Deliberately not keyed off `cost_status`: the `> 0` check also covers DBs from before the column existed, and matches the columns CodeBurn already reads.

Semantics preserved:
- `actual_cost_usd` non-null still wins — including explicit $0 (a real invoice amount is fact)
- genuinely free models still land on $0 — litellm prices them at 0, so the fallback calculation agrees

The ledger's cost-delta path makes existing sessions self-heal: on the next scan, calculated cost > lastSeen $0 emits a cost-only correction observation.

## Tests

- New `tests/hermes-cost-fallback.test.ts`: zero-estimate → calculated, positive estimate → trusted, zero actual → $0
- Existing hermes suites pass (12 tests total)

### DIFF
--- a/src/providers/hermes.ts
+++ b/src/providers/hermes.ts
@@ -463,11 +463,16 @@ function resolveHermesCost(
     tokens.cacheReadTokens,
     0,
   )
-  // This slice treats explicit $0 as recorded. null still falls through.
+  // actual is a recorded invoice amount; explicit $0 still counts as recorded.
   if (row && row.actual_cost_usd != null) {
     return { costUSD: row.actual_cost_usd, costIsEstimated: false, costBasis: 'actual' }
   }
-  if (row && row.estimated_cost_usd != null) {
+  // estimated_cost_usd = 0.0 is not a measurement: Hermes writes it when cost_status
+  // is 'unknown' (no pricing data) or 'included' (flat subscription), so trusting it
+  // would zero out every subscription session. Only positive estimates are trusted;
+  // anything else falls through to the token-based calculation. A genuinely free
+  // model still lands on $0 either way, since litellm prices it at 0.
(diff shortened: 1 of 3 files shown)


MiguelMachado-dev:

iamtoruk:
Reviewed in depth and merging. The fix is right, and the objection that a literal zero might be a genuine free-model zero does not hold in practice: free, local and flat-rate models calculate to zero anyway, API-equivalent pricing for subscription turns is the same convention every other provider uses, and docs/providers/hermes.md has said since the provider landed that only a positive recorded cost is trusted. Your new fixture test is genuine red/green (fails on main with expected 0 to be greater than 0). One thing patched on top: PROVIDER_PARSE_VERSIONS.hermes in src/session-cache.ts was not bumped, so sessions whose state.db fingerprint has not changed would have kept serving the cached zero; bumped it so they re-cost. tsc clean, hermes and session-cache suites green.

PR #1214, opened 2026-09-01
feat(menubar): keep token-only and flat-rate providers visible in agent tabs

MiguelMachado-dev:
## Problem

Agent tabs filter strictly on `cost > 0`, so a provider whose sessions carry tokens but no recorded dollar cost vanishes from the strip entirely — its usage exists but is invisible. This hits Hermes on a flat subscription, free-tier models, and any unpriced model.

## Fix

The menubar payload now carries an **activity signal** per provider (real usage, not just cost), and the tab strip keys visibility off it:

- providers with actual usage stay listed even at $0 recorded cost
- providers merely *discovered on disk* stay hidden
- older payloads without the signal keep the previous detected-provider behavior (backward compatible)

Applied consistently across the three frontends: mac menubar (`AgentTabStrip`/`MenubarPayload`), Electron renderer types, Windows payload types.

## Tests

- New `AgentTabProviderVisibilityTests.swift`: zero-cost providers with usage remain active, idle/discovered-only providers stay hidden, legacy payloads fail open
- Extended `menubar-json`, `usage-aggregator`, `cli-status-menubar` suites — 52 tests pass
- Full menubar Swift suite: 352 tests pass

### DIFF
--- a/app/renderer/lib/types.ts
+++ b/app/renderer/lib/types.ts
@@ -174,8 +174,9 @@ export type MenubarPayload = {
     localModelSavings: LocalModelSavings
     providers: Record<string, number>
     // Optional: older CLIs omit it. `id` is the internal provider name (round-trips
-    // as --provider), `label` the display name. Fall back to `providers` when absent.
-    providerDetails?: Array<{ id: string; label: string; cost: number }>
+    // as --provider), `label` the display name. `hasUsage` distinguishes active $0
+    // providers from detected-but-idle providers when present.
+    providerDetails?: Array<{ id: string; label: string; cost: number; calls?: number; hasUsage?: boolean }>
     topProjects: Array<{
       name: string
       cost: number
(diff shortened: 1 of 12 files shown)


MiguelMachado-dev:

iamtoruk:
Reviewed in depth and merging. The mac change is correct: hasUsage is a strict superset of the old cost filter, activeKeys falls back to the providers map for older CLIs, and the payload change is add-only with no cache bump needed (Swift 351 tests green). One gap patched on top: app/renderer/sections/Settings.tsx still filtered the Providers pane on cost, the same bug this PR exists to close, one line away from the field it adds; it now reads hasUsage with a cost fallback (typecheck clean, sections suite 150 green). Two notes for a follow-up: the Windows payload type declares hasUsage but nothing reads it there, and the middle branch in the Swift ProviderDetail decoder (calls present, hasUsage absent) cannot occur since both fields ship together.

PR #1216, opened 2026-09-01
test(budget): seed current-month spend in the past, not at noon UTC

MiguelMachado-dev:
## Problem

`tests/cli-budget.test.ts` fails on every branch when CI runs between 00:00 and 12:00 UTC (observed on four unrelated PRs opened at 01:46 UTC on Sept 1, all failing identically on `keeps overview budget lines only on unfiltered overviews` with `No usage found for September 2026`).

`currentMonthTimestamp` seeds at today 12:00 UTC — a deliberate month-safety choice, but before noon UTC that timestamp is **in the future**, and future-dated rows never surface in the overview. The seeded spend vanishes, the budget line never renders, the assertion fails.

Reproduced locally at 02:30 UTC Sept 1; passes again after 12:00 UTC, which is why it doesn't fail every day.

## Fix

Seed five minutes in the past, pinned to month start on the 1st so the row can't slip into the previous month:

```ts
const monthStart = new Date(now.getFullYear(), now.getMonth(), 1)
const fiveMinutesAgo = new Date(now.getTime() - 5 * 60_000)
const base = fiveMinutesAgo < monthStart ? monthStart : fiveMinutesAgo
```

The remaining race sliver is ~2 minutes at month-start midnight UTC (the assistant line seeds at +1min), down from 12 hours daily.

## Tests

All 7 `cli-budget` tests pass at 02:40 UTC (inside the previously broken window), including the previously failing one.

### DIFF
--- a/tests/cli-budget.test.ts
+++ b/tests/cli-budget.test.ts
@@ -41,10 +41,12 @@ function currentMonthTimestamp(offsetMinutes: number): string {
   const now = new Date()
   // Noon-today-UTC was month-safe but lands in the future for the 12 hours
   // before it, and rows timestamped in the future don't surface in the
-  // overview — every CI run before 12:00 UTC went red. Seed five minutes
-  // back instead, pinned to month start on the 1st so the row can't slip
-  // into the previous month.
-  const monthStart = new Date(now.getFullYear(), now.getMonth(), 1)
+  // overview on the 1st, so every CI run on the 1st before 12:00 UTC went
+  // red. Seed five minutes back instead, pinned to month start on the 1st so
+  // the row can't slip into the previous month. The child runs with TZ=UTC,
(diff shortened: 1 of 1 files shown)


iamtoruk:
Reviewed in depth and merging. The flake is real: the touched test was red on main this morning (the 1st, before 12:00 UTC) and green on this branch in the same minute, stable over five runs. One thing patched on top: the month start was computed in local time while the CLI child is pinned to TZ=UTC, so the clamp only worked at offset zero and a few-minute window survived at the start of each month in every other zone; it now uses Date.UTC, verified under Pacific/Kiritimati and Pacific/Pago_Pago. Also tightened the comment: it was every run on the 1st before noon UTC that failed, not every run. Worth a follow-up: the weekly test at the bottom of the same file seeds from getDateRange in the parent zone and hands it to the UTC child, the same class of issue with a larger window.

PR #1218, opened 2026-09-01
fix(daily-cache): stop a frozen under-read from outranking the sources on disk

### LINKED ISSUE #1217: bug: an under-read day inside the settle window is frozen forever — Overview headline stays below the sum of its own Daily Activity rows
**Summary:** On 0.9.23 the dashboard's Overview headline for `7 Days` sat ~27% below the sum of the Daily Activity rows printed directly beneath it (**27,346 calls** in the headline vs **32,608** summed from its own visible day rows). The two panels read different sources by design — Overview takes every historical day from the durable daily cache, Daily Activity / By Model parse the live session logs — and five cached days had been frozen at a third to two-thirds of what their session logs actually contain.

The logs were never missing. Re-parsing those dates today returns the full numbers, and the previous cache generation still on disk (`daily-cache.v17.json`) holds them correctly. Once a day is written into the durable cache and the watermark passes it, nothing ever looks at it again — so a single bad derivation is permanent.

**Environment**

- codeburn 0.9.23 (Homebrew CLI), macOS 26 / darwin 25.6.0
- providers: claude, codex, grok, pi, cline-cli, warp
- `~/.claude/projects` 1.3G, `~/.codex/sessions` 2.5G, daily cache v29 with 222 days

**What the cache held vs. what the logs contain**

Calls only (the cost columns diverge in the same proportion):

| day | cached (Overview) | live logs (Daily Activity, and a re-parse today) |
|---|---|---|
| 2026-08-24 | 149 | 1,123 |
| 2026-08-25 | 257 | 1,650 |
| 2026-08-26 | 892 | 2,429 |
| 2026-08-27 | 2,174 | 4,458 |
| 2026-08-28 | 1,893 | 3,191 |

`daily-cache.v17.json`, written before the upgrade, has the correct figures for all five (e.g. 2026-08-26 claude 2,363 calls, against 826 in v29). Single-day and 6-month parses today both return the live column, so this is not a parser gap — the durable cache is simply holding a stale under-read.

**Root cause: two individually reasonable rules that combine badly**

1. `mergeDayEntries` lets a *smaller* fresh slice replace a larger baseline slice whenever the day is inside the settle window — `isPartialSurvival()` (`src/daily-cache.ts:1033`) only rescues the baseline for `date < today - SETTLE_DAYS` (7). This is deliberate and covered: *"recent days stay authoritative: a shrink inside the settle window is honored"*, `tests/daily-cache-carry-forward.test.ts:852`.
2. `ensureCacheHydrated` never revisits a day at or before `lastComputedDate` — the gap parse starts at the watermark + 1 (`src/daily-cache.ts:1393`), so a day is derived once and then final.

The window in which a fresh derivation is trusted unconditionally is exactly the window that is never re-derived. If any one complete-marked derivation under-reads a day while that day is still inside the settle window, the undercount is written, the watermark moves past it a day or two later, and no subsequent run ever re-examines it.

The upgrade path is what exposed it here. With no `daily-cache.v29.json` present, `adoptOlderDailyCaches()` unions the older files (v17 among them), stamps `complete: false`, and the same run does a full re-derive. That re-derive under-read Claude on the five days above — all inside the settle window at the time — so the truncated slices won and the correct v17 slices were discarded. The other providers on those days produced no fresh slice at all, so they were carried, which is why the days ended up flagged `carried: true`.

I want to be straight about the limit of the diagnosis: **I cannot reproduce the under-read itself.** A parse of those dates today, narrow or wide, returns the right numbers. Whatever made that one derivation partial (it ran during an upgrade, on a large corpus, and `sessionComplete()` evidently returned true) was transient. This report is about the second half — that a transient under-read becomes permanent state with no path back.

**Second-order symptom: the footnote asserts the opposite of the truth**

Because those days took carried slices, the Overview footnote read `includes $… preserved from expired session logs`, and the amount was *exactly* the sum of the five frozen days. So the UI advertised the corrupted days as preserved history whose sources were gone, while every one of their transcripts was intact on disk. After the repair below, that footnote drops to $0.00.

**Deterministic repro (no real corpus needed)**

Against a cache marked complete through yesterday, with a day 3 days old holding a full slice:

1. Run `ensureCacheHydrated` with a complete parse that returns a truncated slice for that day → the day is overwritten with the truncated numbers. *This step is intended behavior (rule 1).*
2. Run `ensureCacheHydrated` again with a parse that returns the correct, full slice → **the day stays truncated**, because the watermark is current and no parse covers it. Every later run does the same.

Step 2 is the bug.

**Workaround for anyone hitting this**

Set `"complete": false` in `~/.cache/codeburn/daily-cache.v29.json` and run codeburn once. That takes the re-derive path, where today's correct parse wins per (date, provider) and days whose sources really are gone stay carried. Verified here: all five days healed to match a fresh parse exactly, and the carried footnote went to $0.00. **Do not delete the file instead** — `MIN_SUPPORTED_VERSION = 28` means an older generation cannot be loaded back, so a delete throws away every carried day whose sources have since expired.

**Proposed fix** (flagging before writing much, per CONTRIBUTING)

Re-derive the unsettled tail rather than freezing it: start the gap parse at `min(lastComputedDate + 1, today - SETTLE_DAYS)`. That leaves every merge rule untouched — a fresh derivation still wins inside the settle window, days past it stay frozen and still get `isPartialSurvival` protection, and a partial parse still cannot shrink anything — but it makes the window self-healing instead of write-once, so the same transient failure costs one launch instead of forever. The cache write needs a no-op guard alongside it so the extra pass doesn't republish the file on every run (cf. #1032).

Happy to send that as a PR if the approach looks right, or to take it another way if you'd rather the settle window not be re-parsed on every launch — the alternative I considered is recording per-day whether the derivation that sealed it was source-complete, and re-deriving only suspect days, which is a schema bump.

therickfactr:
## Summary

- A day is derived into the durable cache once and then frozen behind `lastComputedDate`, while `isPartialSurvival` deliberately lets a fresh derivation **shrink** a day inside the settle window. Together, one parse that missed sources becomes permanent — my `7 Days` Overview headline read 27% below the sum of the Daily Activity rows printed directly under it (27,346 calls vs 32,608), on five days whose transcripts were intact the whole time.
- The read path already parses the dates it reports on, so it now reconciles the two derivations per `(date, provider)` and keeps whichever explains **more calls**. Carry-forward is untouched — nothing live can outbid a day with no surviving source — and ties go to the cache, so a re-pricing still lands only through the write path.
- The write path is unchanged: what to freeze stays its decision, a still-settling day is still defined by its fresh parse, and nothing re-parses.

Fixes #1217.

## Testing

- [x] `npm test` passes — 1 failed | 3,500 passed, and that one failure is `cli-budget > keeps overview budget lines only on unfiltered overviews`, which fails identically on `main` (1 failed | 3,494 passed). It is the pre-noon-UTC seeding flake #1216 fixes; CI here ran at 08:05 UTC and hit the same one. Zero new failures.

Also run: `npx tsc --noEmit` clean, and `npm run test:locks` matching `main` (1 failed | 35 passed, the known `cache-refresh-lock` heartbeat flake, same test both sides).

Against real data, on the machine that hit this: the headline now matches a fresh parse day for day, and the `includes $… preserved from expired session logs` footnote — which had been naming exactly the five corrupted days — drops to $0.00.

```
before   $6,415.71   27,346 calls    includes $4,108.78 preserved from expired session logs
after    $8,880.38   34,527 calls    (no carried footnote)
```

## Detail

**Evidence it is the cache, not the parser.** `daily-cache.v17.json`, still on disk from before the upgrade, holds the correct figures (2026-08-26: 2,429 calls; the v29 row says 892). A single-day parse and a 6-month parse of those dates both return the full numbers today. The under-read itself was transient and I could not reproduce it — that limit is written up in #1217. This PR is about the other half: that a transient under-read becomes permanent state with no path back.

**Why not fix it on the write side.** I first tried re-deriving the unsettled tail on each run (`gapStart = min(watermark + 1, today - SETTLE_DAYS)`). That breaks six tests which deliberately guarantee the opposite — `does not recompute yesterday after it has already been cached`, `trusts a stamped watermark over an idle tail — no re-derive treadmill`, and four more. Those tests are right, so I left them alone; I flagged the change of approach on the issue before rewriting.

**Cost.** No new parse and no new file read. `buildDurablePeriod` aggregates a parse it already performed, `buildDurableOverviewFromNormalizedIndex` reuses `normalizedDays` it already computes, and when cache and parse agree the merge finds nothing to swap. A partial index can never lower a total under this rule, since a thinner slice never wins — which matters for the progressive-startup path, where the index is deliberately incomplete early.

**Implementation.** The reconcile is `mergeDayEntries`'s existing partial-survival machinery under a new `'prefer-richer'` mode, so slice/day totals stay reconciled by the code that already does that.

`tests/durable-underread-reconcile.test.ts` fails 4/6 against `main` and passes 6/6 here. It builds a real session fixture on a past date plus an under-reading cache row for it, then asserts the durable headline matches the live parse — along with the three cases that must **not** change: a genuinely expired day still carries, a richer cached slice is not dragged down to a partial live one, and an equal-call re-pricing leaves the durable value alone.

**Note for anyone already affected:** this stops the symptom wherever a live parse happens, but a cache that already froze an under-read stays wrong on disk. Setting `"complete": false` in `~/.cache/codeburn/daily-cache.v<N>.json` and running once repairs it (verified: all five days healed exactly). Deleting the file instead is not equivalent — `MIN_SUPPORTED_VERSION` makes older generations unadoptable, so a delete drops every carried day whose sources have since expired.

### DIFF
--- a/src/daily-cache.ts
+++ b/src/daily-cache.ts
@@ -1034,6 +1034,15 @@ function isPartialSurvival(date: string, baseline: ProviderDaySlice, fresh: Prov
   return date < settleCutoff && fresh.calls < baseline.calls
 }
 
+/// The read path's rule: keep whichever derivation explains MORE calls, on
+/// every date, with ties going to the baseline. Unlike the write path this is
+/// not deciding what to freeze, so it has no reason to prefer a thinner slice
+/// on a recent day — but it is also not the place a re-pricing lands, so an
+/// equal-call slice leaves the durable value alone.
+function baselineExplainsMore(baseline: ProviderDaySlice, fresh: ProviderDaySlice): boolean {
+  return fresh.calls <= baseline.calls
+}
+
 /// Index `freshUnderOldTz` (the same parse re-aggregated under the cache's OLD
 /// tzKey) by date then provider, so the merge can subtract exactly what the
 /// fresh parse still explains under the old bucketing.
@@ -1088,7 +1097,14 @@ export function mergeDayEntries(
   /// baseline wins on settled days (`isPartialSurvival`). The adoption union
   /// leaves it off - both sides are cache generations there and the newer
   /// schema deliberately wins per (date, provider).
-  guardPartialSurvival = false,
+  ///
+  /// `'prefer-richer'` is the read-path union's rule (`unionDaysForPeriod`):
+  /// on EVERY date, keep the slice explaining more calls, ties to the baseline.
+  /// That path is choosing what to REPORT, not what to freeze, so a thinner
+  /// derivation is never the better answer there — while the write path still
+  /// lets a recent day shrink, on the grounds that a still-settling day is
+  /// defined by its fresh parse.
+  guardPartialSurvival: boolean | 'prefer-richer' = false,
   /// Providers whose baseline slices were recorded under an accounting where a
   /// call meant something else, so a shrink is not evidence of source loss for
   /// this one re-derivation (see PENDING_REDERIVE_PROVIDERS). Only consulted
@@ -1141,7 +1157,10 @@ export function mergeDayEntries(
         }
       }
       if (existingSlice && hasSliceData(existingSlice) && !residual) {
-        if (!guardPartialSurvival || pendingRederive?.has(provider) || !isPartialSurvival(day.date, slice, existingSlice, settleCutoff)) continue
+        const keepBaseline = guardPartialSurvival === 'prefer-richer'
+          ? baselineExplainsMore(slice, existingSlice)
+          : isPartialSurvival(day.date, slice, existingSlice, settleCutoff)
+        if (!guardPartialSurvival || pendingRederive?.has(provider) || !keepBaseline) continue
         // The baseline holds more evidence than the sources can still produce:
         // swap the fresh slice back out for it (inverse of addSliceIntoDay, so
         // the day's totals and nested maps stay reconciled with its slices).
--- a/src/usage-aggregator.ts
+++ b/src/usage-aggregator.ts
@@ -15,7 +15,7 @@ import { scanUserCorrections, medianTimeToFirstEditMs, aggregateFileChurn, compu
 import { buildPrAttribution, aggregateByBranch } from './sessions-report.js'
 import { scanAndDetect } from './optimize.js'
 import { callBillableOutputTokens, sessionBillableOutputTokens } from './session-output.js'
-import { getDaysInRange, ensureCacheHydrated, emptyCache, BACKFILL_DAYS, toDateString, type DailyCache, type DailyEntry, type ProjectDayStats, type ProviderDaySlice } from './daily-cache.js'
+import { getDaysInRange, ensureCacheHydrated, emptyCache, mergeDayEntries, BACKFILL_DAYS, toDateString, type DailyCache, type DailyEntry, type ProjectDayStats, type ProviderDaySlice } from './daily-cache.js'
 import { buildGranularHistory } from './granular-history.js'
 
 // Row caps for the by-PR / by-branch payload aggregations, ranked by cost.
@@ -393,6 +393,10 @@ function unionDaysForPeriod(
   periodInfo: PeriodInfo,
   daysSelection: Set<string> | null,
   sliceHistorical?: (day: DailyEntry) => DailyEntry,
+  /// Historical days from the parse this period already ran. They are evidence
+  /// about the same dates the cache is answering for, so where one explains
+  /// more of a day than the other, that one is used (#1217).
+  liveHistoricalDays: DailyEntry[] = [],
 ): DailyEntry[] {
   const now = new Date()
   const yesterdayStr = toDateString(new Date(now.getFullYear(), now.getMonth(), now.getDate() - 1))
@@ -406,8 +410,24 @@ function unionDaysForPeriod(
   // never reaches the slicer (which tallies what it could not attribute).
   const selectedCacheDays = daysSelection ? cacheDays.filter(d => daysSelection.has(d.date)) : cacheDays
   const historicalDays = sliceHistorical ? selectedCacheDays.map(d => sliceHistorical(d)) : selectedCacheDays
+  // A cached day is derived once and then frozen behind the watermark, so a
+  // derivation that missed sources stays the answer forever — the Overview
+  // headline reading below the live panels beneath it (#1217). This run already
+  // parsed these dates. Reconcile the two per (date, provider), keeping
+  // whichever explains MORE calls: a cached day whose transcripts have expired
+  // still wins (nothing live can outbid it), and an under-read cached row stops
+  // suppressing evidence that is sitting on disk. Only dates the cache already
+  // holds are reconciled — filling absent dates is a separate decision each
+  // caller makes for itself.
+  const cachedDates = new Set(historicalDays.map(d => d.date))
+  const liveForCachedDates = liveHistoricalDays.filter(d =>
+    cachedDates.has(d.date) && (!daysSelection || daysSelection.has(d.date)),
+  )
+  const reconciledDays = liveForCachedDates.length > 0
+    ? mergeDayEntries(liveForCachedDates, historicalDays, true, undefined, 'prefer-richer')
+    : historicalDays
   const todayInRange = todayAllDays.filter(d => d.date >= rangeStartStr && d.date <= rangeEndStr)
-  const unfiltered = [...historicalDays, ...todayInRange].sort((a, b) => a.date.localeCompare(b.date))
+  const unfiltered = [...reconciledDays, ...todayInRange].sort((a, b) => a.date.localeCompare(b.date))
   return daysSelection ? unfiltered.filter(d => daysSelection.has(d.date)) : unfiltered
 }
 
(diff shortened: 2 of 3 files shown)


therickfactr:
Both `test` legs are red on `tests/cli-budget.test.ts > keeps overview budget lines only on unfiltered overviews` — `expected 'CodeBurn September 2026\n\nNo usage …' to contain 'Monthly budget:'`. That is the pre-noon-UTC seeding flake #1216 describes, not this change: CI ran at 08:05 UTC, inside the window, and it is the same single failure I get on `main` locally. Everything else in both legs passes (258/259 files), the new suite included.

[test (22.13.0)](https://github.com/getagentseal/codeburn/actions/runs/33485267119/job/99783745966) · [test (22)](https://github.com/getagentseal/codeburn/actions/runs/33485267119/job/99783745618)

Local gate for the record — `npm test` on `main`: 1 failed | 3,494 passed; on this branch: 1 failed | 3,500 passed, same test. Happy to rebase once #1216 lands so this goes fully green.

therickfactr:
Retitled to the `fix(daily-cache):` convention and rewrote the body against `.github/PULL_REQUEST_TEMPLATE.md`. The force-push carried **no code change** — `efa06184` and `f2e82139` have the same tree, `b3715ed389a99657f222c696c1ac07fb57a4e5a9`, so only the commit subject moved.

That makes the two CI runs a clean A/B on identical content, and they failed differently each time:

| run | `cli-budget` | `Cache-lock suite (serial)` |
|---|---|---|
| [08:05 UTC](https://github.com/getagentseal/codeburn/actions/runs/33485267119) | fail (pre-noon-UTC seeding, #1216) | pass |
| [15:19 UTC](https://github.com/getagentseal/codeburn/actions/runs/33525072915) | pass | timed out at 5m |

Same tree, disjoint failures — both are flakes, neither is this change. In the second run the parallel suite itself passed whole (261 files, 2 skipped, my new suite included) and the job died in the serial lock step: `The action 'Cache-lock suite (serial)' has timed out after 5 minutes`. That is `the fence never loses to its own heartbeat (in-process serialization)` burning its 10 retries — it takes ~330s on my machine against `main` too, which is already over the step's 5-minute budget. Looks like the #904 / #1196 neighbourhood.

This PR touches `src/daily-cache.ts`, `src/usage-aggregator.ts` and one new test; the lock path is `src/parser.ts` / `src/session-cache.ts`, untouched. I don't have re-run rights here — a maintainer re-running `test (22)` should settle it.

ozymandiashh (changes requested):
The bug is real and the layer is right: a day frozen behind `lastComputedDate` while `isPartialSurvival` lets a fresh parse shrink it inside the settle window, so one transient under-read becomes permanent. Reconciling per `(date, provider)` against the parse the read path already ran is the correct shape, and carry-forward is untouched. I resolved the `src/usage-aggregator.ts` conflict locally and the merged tree is green (`tsc` clean, 129/129 across the 11 related suites).

One blocker before merge:

**`markSecondaryCarried: true` marks healthy days as carried.** In `unionDaysForPeriod` (`src/usage-aggregator.ts:453`) the call is `mergeDayEntries(liveForCachedDates, historicalDays, true, undefined, 'prefer-richer')`. With `'prefer-richer'`, `baselineExplainsMore` returns true on `fresh.calls <= baseline.calls`, so the tie case, which is every normal day inside retention, goes through the swap branch and sets `existing.carried = true`. `carriedCostUSD` then sweeps up the whole period and Overview / dashboard / menubar print `includes $X preserved from expired session logs` for money whose transcripts are on disk. Reproduced with a cache row carrying the same 4 calls the live parse finds: merge-base `carriedCostUSD 0`, this branch `3.16` of a `3.16` day. Same root cause hits the provider-filtered path through the `!existing` branch.

Fix: pass `false`, matching what `overlayProviderDaySlices` does on main for the same read-path purpose (`3e48f8d0`). Please add a case to `tests/durable-underread-reconcile.test.ts` where cache calls == live calls and `carriedCostUSD === 0`; the current suite only pins the strict-live-wins case, which is why 3,500 green tests miss this.

Rebase notes: the import line takes main's verbatim (main already imports `mergeDayEntries`); in `buildDurablePeriod` your `liveHistoricalDays` line and the sixth argument to `unionDaysForPeriod` need to be combined with main's `overlayProviderDaySlices(allDays, freshDaysInSelection, pf)` routing, not chosen over it.

Non-blocking, for the PR description or a follow-up: this is read-path only, so the corrupt row stays on disk and the 365-day history / heatmap in the menubar payload still come from raw cache days while the headline in the same payload is reconciled. A write-back of the richer day (data already in hand, no extra parse) would close that. Also worth noting the extra `aggregateProjectsIntoDays` + `structuredClone`-per-day merge runs on every `buildDurablePeriod`, twice per `status`.

iamtoruk:
Landed as #1304 (cfa0f9fa) with your commit and authorship intact, rebased onto current main, plus two follow-ups found during real-data verification: the reconcile no longer rewrites a day's carried mark (the original passed markSecondaryCarried=true, which flagged every agreeing day as preserved from expired logs), and days where no live slice can win are skipped rather than cloned. Verified on a real 137-day corpus: a frozen under-read day restores to the live figure, healthy days and the carried footnote are byte-identical to before. We could not push to your fork, hence the superseding PR. Thank you for the diagnosis and the fix.

PR #1222, opened 2026-09-01
feat(menubar): add native Grok Build quota provider

avs-io:
## Summary

- add a native CodeBurn Grok Build quota adapter
- provide the same single-click connection flow as the other local providers
- surface plan, quota usage, reset timing, and actionable connection errors

## Privacy and security

- reads the Grok CLI-owned auth file read-only from `$GROK_HOME/auth.json` (or the default Grok home)
- keeps the bearer token in memory only for the request
- uses an ephemeral, cookie-free URL session
- does not copy credentials into CodeBurn storage or depend on another application/library at runtime

## Verification

- 361 tests across 46 suites passed
- optimized macOS build passed
- live smoke tested against an existing local Grok Build login
- branch contains only the Grok provider changes; PR #1220 performance work is intentionally excluded

Base: `7d5bcd4d9a4c103af8331af01dd2abeaf3304222`


iamtoruk:
Reviewed and verified locally: based on current main, merges clean, swift build clean, swift test 361 tests in 46 suites passing (12 net new), no attribution trailers, no em-dashes. Structure matches the ClinePass and Z.ai adapters: injected fetch, ephemeral cookie-free session, terminal vs transient classification wired into the dock failure model, and the auth.json read is read-only with the documented snake_case keys (auth_mode, expires_at) and OIDC scope precedence, with an expired login surfaced as a re-login state rather than a retry loop.

Two things to settle before merge:
1. Same bar as #1202: please paste a sanitized live harness line (plan label, windows, reset) from your smoke test. The cli-chat-proxy.grok.com/v1/billing?format=credits shape (config.creditUsagePercent, currentPeriod.end) is not documented anywhere public that I could find; the public write-ups describe the gRPC-web GetGrokCreditsConfig path, so the live evidence is what backs the schema.
2. The wire-level analysis of the Grok CLI shows /v1/settings being called with POST; this adapter uses GET for both requests. If the plan label comes back empty in your live run, that is likely why. It degrades gracefully (label optional), so not blocking, but worth a check.

One note: the catalog change also drops .web and .cookieOrWebSession from the Grok entry. Fine if intentional (the adapter only supports the CLI login), just calling it out since it narrows what Settings offers.

PR #1225, opened 2026-09-01
fix(menubar): launch the Windows tray exe by its real binary name

iamtoruk:
First run of `codeburn menubar` on a real Windows 11 install: the MSI installs fine, but the launch step fails with spawn ENOENT.

Cause: `parseInstalledWindowsMenubar` builds the exe path as `InstallLocation + 'CodeBurn Menubar.exe'` (the product name), while Tauri names the binary after the crate: `codeburn-menubar.exe` (windows/src-tauri/Cargo.toml). Verified in the guest: `C:\Program Files\CodeBurn Menubar\codeburn-menubar.exe` is the only exe the MSI ships, and launching it directly works.

Fix: join with a `WINDOWS_EXE_NAME` constant instead. The DisplayIcon fallback path is unchanged. Test fixtures updated to the real name.

### DIFF
--- a/src/menubar-installer.ts
+++ b/src/menubar-installer.ts
@@ -27,6 +27,8 @@ const SUPPORTED_OS = 'darwin'
 /// rewrites the spaces in the bundle name to dots when it stores the asset, so both the asset
 /// name and its download URL carry `CodeBurn.Menubar_...`.
 const WINDOWS_PRODUCT_NAME = 'CodeBurn Menubar'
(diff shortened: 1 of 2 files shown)


iamtoruk:
Superseded by #1228, which landed the same fix. The tray binary is now named by its own constant rather than derived from the product name (`WINDOWS_BINARY_NAME` in `src/menubar-installer.ts`), so `codeburn menubar` finds `codeburn-menubar.exe` on Windows. Closing as already fixed on main.

PR #1229, opened 2026-09-01
fix(parser): avoid marking partially scanned caches complete (#912)

earbona23:
A provider-scoped parse could mark the whole session cache complete while another provider still had discoverable sessions that the run had not scanned. This change withholds the completeness stamp in that case, while preserving it on single-provider installations so warm-refresh snapshot fallback keeps working.

The additional discovery runs only for writable, previously incomplete scoped scans that did not defer files for first paint. It discovers sources across providers without parsing their full sessions; discovery may still read metadata or query databases. The demonstrated impact of the original bug is transient: a subsequent full refresh rediscovers the omitted provider and repairs the cache.

Adds ordinary regression tests for the omitted-provider case and recovery on a full refresh, with isolated provider roots. Addresses the provider-scoped completeness case in #912.

Validation at `0e8a379e5c5aacc2e6c3ac3fb8b0c6d6e240f9e0`:

- 34 targeted tests and TypeScript pass locally, both on the PR head and after integrating current `main` (`eaed4240`).
- All 15 applicable GitHub checks pass, including both Node 22 test jobs, the serial cache-lock suites, desktop, upgrade paths, and package builds.
- The release-only asset check is skipped as expected.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -5491,7 +5491,24 @@ async function runParseInner(
   // it keeps the normal stamp.
   const deferredForFirstPaint = firstPaintDeferredThisRun > 0
   const wasComplete = isCacheComplete(diskCache)
-  if (!readOnly && !wasComplete && !deferredForFirstPaint) diskCache.complete = true
+  // A provider-scoped run walks only its own provider's sessions, so it never
+  // saw whatever the providers it skipped hold on disk. Stamping the WHOLE
+  // cache complete off that partial view writes a wrong "done" to disk (#912):
+  // per the marker's own contract above, a complete cache stops being re-read
+  // as cold, so the unscanned providers are not revisited and the gap stops
+  // looking like a gap. The guard is conditioned on real on-disk data, not on
+  // scoping alone: when every provider the run skipped has NO discoverable
+  // sessions, the scoped run really did see the whole corpus (a claude-only
+  // machine, and exactly what the warm-refresh snapshot tests rely on), so the
+  // stamp is correct and stands. Discovery here is a bounded directory walk,
+  // not a parse — the scoping win (skipping the other providers' PARSE) holds.
+  const scopedRun = !!providerFilter && providerFilter !== 'all'
+  let skippedProviderHasSessions = false
+  if (scopedRun && !readOnly && !wasComplete && !deferredForFirstPaint) {
+    const corpusSources = await discoverAllSessions()
+    skippedProviderHasSessions = corpusSources.some(s => s.provider !== providerFilter)
+  }
+  if (!readOnly && !wasComplete && !deferredForFirstPaint && !skippedProviderHasSessions) diskCache.complete = true
   if (!readOnly && (isCacheDirty(diskCache) || (!wasComplete && !deferredForFirstPaint))) {
     try {
       const published = await saveCache(diskCache, refreshLock?.verifyStillOwner)
--- a/tests/scoped-run-completeness.test.ts
+++ b/tests/scoped-run-completeness.test.ts
@@ -46,10 +46,10 @@ async function writeClaudeSession(): Promise<void> {
 }
 
 describe('provider-scoped run and the whole-cache completeness marker', () => {
-  // `it.fails` pins current (wrong) behaviour without turning the suite red while the
-  // defect is open. When the stamp learns about scope this starts passing, vitest reports
-  // "Expect test to fail", and that is the signal to drop `.fails` and land the guard.
-  it.fails('does not stamp the cache complete when providers were left out of scope', async () => {
+  // Regression for #912: the stamp now learns about scope. A run scoped to one provider
+  // must not mark the whole cache complete while a provider it skipped still has sessions
+  // on disk it never scanned.
+  it('does not stamp the cache complete when providers were left out of scope', async () => {
     await writeClaudeSession()
 
     // Scoped to 'codex': discovery is filtered and the cached-provider loop skips every
@@ -62,8 +62,8 @@ describe('provider-scoped run and the whole-cache completeness marker', () => {
     // anyway and the rest proves nothing.
     expect(Object.keys(raw?.providers?.claude?.files ?? {}).length).toBe(0)
 
-    // The finding: the stamp at the end of runParseInner is guarded on
-    // readOnly / wasComplete / deferredForFirstPaint — not on whether the run was scoped.
+    // The guard: the stamp at the end of runParseInner refuses to mark the whole cache
+    // complete when a skipped provider (claude here) still has discoverable sessions.
     expect(raw?.complete ?? false).toBe(false)
   })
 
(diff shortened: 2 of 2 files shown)


ozymandiashh:
Thanks — the first invariant is reproduced, and I independently got the same flip at `ca541de`: the normal assertion receives `true`, while guarding the global stamp makes it pass.

Before turning this into a fix, please tighten two things:

1. Make the fixture hermetic. The test calls `parseAllSessions(undefined, 'codex')`, but redirects only `CLAUDE_CONFIG_DIR`, `CODEBURN_CACHE_DIR`, and `CODEBURN_DESKTOP_SESSIONS_DIR`. Codex discovery still defaults to `CODEX_HOME` / `~/.codex`, so on a developer machine this test reads the real local corpus. Point Codex discovery at an empty temp root and restore the environment in teardown.

2. Pin the actual consequence and narrow the claim. I simulated a later process (`clearSessionCache()`, then `parseAllSessions(undefined, 'all')`); Claude was discovered and written to cache. An ordinary all-provider refresh therefore rehydrates the omitted provider, so the current "nothing re-derives them" / persistent-zero claim is not established by this test.

Please add either an assertion documenting that the next full run repairs the cache plus a real first-paint/snapshot repro demonstrating the transient mislabel, or another real caller-path repro that proves persistence. Once the scope is accurate, the conservative guard can come into this PR and `it.fails` can become a normal regression test.

earbona23:
Both addressed, pushed.

**1. Hermetic.** Good catch — the scoped run is `codex`, and Codex discovery falls back to `CODEX_HOME` / `~/.codex`, so the test was reading the real local corpus on a dev machine. It now sets `CODEX_HOME` to an empty temp root and restores the prior value (including the unset case) in teardown.

**2. Claim narrowed — you are right, and I have dropped the overstatement.** Your `clearSessionCache()` → `parseAllSessions(undefined, 'all')` check is correct: an ordinary full refresh rediscovers the omitted provider, so "nothing re-derives them / persistent zero" was not established, and it is gone. There is now a test asserting exactly that bound — a plain `parseAllSessions()` after the scoped run repairs the cache and claude reappears. The mislabel is **transient**, not persistent.

On the snapshot angle you offered as the alternative: I traced it and it is real — `canServeCompleteSnapshot()` (parser.ts L5190) gates only on `isCacheComplete()`, so a first-paint `all` render after a scoped run can serve a snapshot that structurally omits the unscanned provider: one wrong paint, corrected by the background full refresh. I did **not** ship it as a test, because pinning it needs a non-empty Codex section in the right on-disk format, and I would rather not contort a fixture to force it. Flagging it as reasoned-from-source, not proven — if you think it is worth pinning, point me at the Codex session fixture helper and I will add it properly.

The `it.fails` invariant is unchanged and still mutation-sensitive. Whenever you want the conservative guard (`&& !scopedRun` on the stamp) in this PR, say so and I will add it with `.fails` removed — I have it ready and verified in both directions.

ozymandiashh:
Go ahead and bring the conservative scoped-run guard into this PR, remove it.fails, and keep the tests as normal regressions. The narrowed transient-impact claim is sufficient; no additional first-paint fixture is needed. Thanks.

earbona23:
I implemented the guard, ran the **full** suite before pushing it, and it regresses four legitimate tests — so I have held it back rather than land a green-looking fix that breaks your snapshot path. Reporting precisely, because the blunt version is wrong and the reason matters.

## What breaks

Guarding the stamp with `&& !scopedRun` fails `tests/parser-cache-refresh-timeout.test.ts` (4 tests), e.g.:

- `serves the prior complete snapshot and leaves the holder cache untouched`
- `does not report a complete hydration when the served snapshot is stale`
- `does not report a complete hydration when a session file is missing from the snapshot`

They all drive `parseAllSessions(undefined, 'claude')` — a **scoped** run — and depend on it stamping the cache complete so the warm-refresh timeout can serve a prior complete snapshot. On clean `main` they pass 4/4; with `!scopedRun` they fail. My repro's `scopedRun` guard is too broad: it forbids a case your design relies on.

## Why the simple guard is wrong

The two situations are genuinely different, and `providerFilter != null` does not separate them:

| | scoped provider | other providers with sessions on disk | stamp complete? |
|---|---|---|---|
| your tests | `claude` | none | **yes — correct** |
| issue #912 | `codex` | `claude` has sessions, unscanned | no — the bug |

The bug is not "a scoped run stamped complete." It is "a scoped run stamped complete **while another provider had on-disk sessions it never scanned**." When the scoped provider is the only one with data, stamping complete is right — which is exactly what your tests pin.

The hard part: knowing an *unscanned* provider has sessions is precisely the discovery a scoped run skips for performance, so the cheap `!scopedRun` check cannot see it. A correct guard needs one of:

1. a bounded check at stamp time for whether any non-scoped, known provider has discoverable sessions (costs part of the scoping win), or
2. moving completeness to **per-provider** rather than one whole-cache flag, so a scoped run marks only its own provider complete and `canServeCompleteSnapshot` composes across providers, or
3. accepting the transient mislabel as documented and keeping this PR test-only.

That is an architecture call on your cache, and you know it far better than a first patch should assume. **Which direction do you want?** If (2), I am happy to sketch the per-provider flag and carry the snapshot logic with it. If (3), the branch already stands green as the reduced-claim regression tests you approved. I would rather ask than push a guard that quietly narrows what "complete" means for every other caller.

_(Full suite run locally, `--exclude tests/cache-refresh-lock*`, before and after the change; the 4 regressions are deterministic and reproduce on toggling the one line.)_

earbona23:
Pushed the conservative guard (`0e8a379e`).

Instead of gating on scoped-vs-`all`, the whole-cache `complete` stamp is now suppressed only when the scoped run left **another provider's sessions on disk unscanned**. So `#912` (scope `codex`, `claude` sessions present) stops stamping, while the `parser-cache-refresh-timeout` cases (scope `claude`, no other provider present) still stamp and stay green — those four failed under the naive `!scopedRun` version precisely because their second run relies on the complete flag to serve the prior snapshot. The probe is a directory walk, not a parse, so the scoping speedup is preserved.

`it.fails` is gone — both are normal regressions now. The Codex fixture is hermetic (discovery points at an empty temp root, restored in teardown) and the claim is narrowed to the transient mislabel that the next full run repairs, per your note. `tsc --noEmit` clean; `scoped-run-completeness` + `parser-cache-refresh-timeout` green (6/6), and the scoped non-claude suite (22/22).

earbona23:
Friendly bump — this has been ready since the conservative guard landed in `0e8a379e`, per your last note.

Recap of where it stands: the whole-cache `complete` stamp is suppressed only when a scoped run left another provider's sessions on disk unscanned, so #912 (scope `codex` with `claude` sessions present) stops stamping, while the `parser-cache-refresh-timeout` cases (scope `claude`, no other provider present) still stamp and stay green. The fixture is hermetic — `CODEX_HOME` is redirected to an empty temp root and restored in teardown, including the unset case. The tests are plain regressions and the transient-impact claim is the narrowed one.

Full suite is green. Nothing pending on my side.

ozymandiashh (approved):

ozymandiashh:
Merged — thanks for the fix and for catching the snapshot regression in the initial guard. The focused tests and full CI are green. Appreciate the careful follow-through and your patience!

PR #1232, opened 2026-09-02
fix(copilot): include copilot legacy chat sessions

JaylyDev:
## Summary

Adding support for VS Code's GitHub Copilot chat sessions prior to February 2026 to copilot's parser.

How it works:

The parser scan for `chatSessions/*.json` and parses the legacy chat sessions if there is a `requests` array in root JSON object. The schema of those JSON files and its definitions like `IToolCallRound` and `ThinkingData` can be found in https://github.com/microsoft/vscode-copilot-chat.

This old format reads all of input, output, reasoning messages and tool calling outputs to determine number of tokens used by a model, which is also stored in the request object. It reads exact token counts (`promptTokens`, `completionTokens`, `usage`) when available, otherwise it fallbacks to character-based estimation.

This parser also maps the model used in auto model selection to the actual model used.

## Testing

- [x] `npm test` passes
- [x] `npm run dev -- models --provider copilot` shows correct model names and pricing

**Terminal output of running `npm run dev -- models --provider copilot -p lifetime`**

Before:

image

After:

image


ozymandiashh (changes requested):
Thanks for taking on the legacy format. I can't merge this as it stands: the tests pass, but the defects are all inside untested branches, and several change numbers for users on the current format too. Everything below was measured on the branch merged with current `main` (clean merge, full suite green apart from the known #1216 flake).

1. **Reasoning tokens are billed as web-search requests.** `src/providers/copilot.ts:702-709` passes `reasoningTokens` as the sixth argument to `calculateCost`, which is `webSearchRequests`. A session with an 8,000-char thinking block on claude-sonnet-4-6 comes out at $20.04 where the true cost is about $0.03, with `costIsEstimated: false`. Every other call site in the file passes `0` there.

2. **The rewrite regresses the current `.jsonl` path.** The old parser read `metadata.promptTokens` and `metadata.outputTokens || completionTokens` because the two fields do not live in the same object (see `chatSessionSampleRequest` in the tests). The new code requires both in one object, so the repo's own real-data shape (`metadata.promptTokens: 32543` + root `completionTokens: 490`) falls through to char estimation: 32,543 input tokens become 2. Rows with no token fields, previously skipped, are now emitted as estimates. That's not a behavior-preserving refactor.

3. **Model ids get mangled instead of resolved.** `normaliseLegacyModelId` turns every dotted version into dashes and the repo has no dash-to-dot normalisation, so anything not in the aliases you added loses or changes pricing: `gemini-2.0-flash-001` $0.15/Mtok to $0, `gpt-4.5-preview` $75 to $30 (matches a different row), `gpt-5.3-codex-spark` folded into the non-Spark SKU. Resolve through the alias mechanism; don't rewrite ids at the provider.

4. **Unresolvable model now falls back to `claude-sonnet-4-5`.** Main returned `unknown` and $0 plus the unknown-model warning. The branch invents premium spend ($3.00 on a 1M-token request with no model info) and suppresses the diagnostic.

5. **Double-count with transcripts.** `hasChatSessionFiles()` (line 3358) still matches `.jsonl` only, so a workspace with only legacy `.json` sessions no longer suppresses `GitHub.copilot-chat/transcripts/`. Both sources are discovered, dedup keys are in different namespaces, and `docs/providers/copilot.md:366` documents this exact invariant. Extend the gate to `.json`, and key legacy sessions on `session.sessionId || basename` so the two formats of one session dedup against each other.

6. **Scope.** `models.ts` display-name changes for `copilot-*-auto` (affects dash, menubar, JSON), `deriveGeminiShortName` now overriding the short-name table, eight JetBrains model tokens, the `inferJetBrainsProject` regex rewrite (which drops `C:/repo` on Windows because the downstream filter still requires a leading slash), and a `win32` early-return added to an unrelated existing test. Please move all of that out; this PR should be the legacy chat sessions and nothing else.

There is no test coverage for the new cost path (`grep -c 'renderedUserMessage|toolCallRounds|CacheBreakpoint' tests/` is 0), which is why CI is green. Fixtures exercising 1-5 are required for a re-review.

iamtoruk (changes requested):
Still the same six. No commits since 2026-09-04, so nothing here is addressed.

I re-measured on main plus this branch:

- reasoning passed as `webSearchRequests`: $20.00 where the real cost is $0.0045, and the same tokens are also added into `outputTokens`
- the existing `.jsonl` chatSessions path now runs through `parseLegacyChatSession`, so a request with `metadata.promptTokens: 32543` and root `completionTokens: 490` exports as 2 in / 4 out, $0.10 to $0.00
- `gemini-2.0-flash-001` $0.15 to $0.00, `gpt-4.5-preview` $75 to $30, `gpt-5.3-codex-spark` $1.75 to $1.25
- unknown model $0.00 to $3.00 per 1M with the warning gone
- `hasChatSessionFiles` is still `.jsonl` only while discovery takes `.json`, which breaks the invariant in docs/providers/copilot.md:366

One more: the `else if (resultObj.usage)` branch sets `foundExact = true` for a `usage` object with no token fields, so char estimates ship as `costIsEstimated: false`.

On my machine this PR changes nothing (no legacy sessions), but every one of those defects is live for anyone who has them. The full suite passes because no test asserts a cost.

Scope is also unchanged: 833 lines in copilot.ts and no fixtures. The `copilot-*-auto` display-name rename, `deriveGeminiShortName`, the JetBrains tokens, the `inferJetBrainsProject` regex and the `win32` early return in an unrelated test all need to come out. models.ts should be pricing aliases for legacy ids and nothing else.

Fix the six, add fixtures that fail without the fix, drop the rest, then I will look again.

iamtoruk (changes requested):
Re-measured all six on d92c065c. All six are fixed, and the fixtures do fail without the fix (13 of the new legacy tests, plus the display-name test, fail when run against main's copilot.ts):

1. reasoning: the 6th arg to calculateCost is a literal 0 at all 6 call sites; an 8000-char thinking block on claude-sonnet-4-6 is $0.030033 with webSearchRequests 0.
2. .jsonl split tokens: metadata.promptTokens 32543 + root completionTokens 490 exports 32543 in / 490 out, $0.104979, same as main. Rows with no token fields are skipped again.
3. pricing per 1M input: gemini-2.0-flash-001 $0.150000, gpt-4.5-preview $75.000000, gpt-5.3-codex-spark $1.750000. All match main, ids kept verbatim.
4. unresolvable model: "unknown", $0.00, warning back.
5. hasChatSessionFiles now takes .json as well as .jsonl.
6. a usage object with no token fields now ships costIsEstimated true.

Real Copilot data is unchanged: report --format json --provider copilot --period all is byte-identical between main and the branch apart from the generated timestamp. tsc clean, full suite green, locks 36.

Still blocking. All of it lands on the current .jsonl format and none of it was asked for:

- tools. extractChatSessionTools was deleted and .jsonl now runs through the legacy extractor, which adds toolCallRounds[].summary, .phase and the raw toolCalls[].id as tool names. A round with summary "Reading the file", phase "execute" and tool call id call_9f3a2b7c1d goes from ["Read"] to ["Reading the file", "execute", "Read", "call_9f3a2b7c1d"].
- .jsonl cost. Thinking text is now char-estimated into reasoningTokens and billed at output price on .jsonl too. A request with exact 32543/60 tokens plus an 8000-char thinking block goes $0.098529 to $0.128544, and costIsEstimated stays false. Main ignores thinking text there.
- .jsonl userMessage was '' and is now the first 500 chars of the message, which changes the JSON export.
- display names. modelDisplayName matching changed from model.includes(key) to exact-or-prefix, the fallback changed from the raw id to getShortModelName, 30 entries were added and 'gpt-4o' was dropped. Measured: gemini-2.5-pro "gemini-2.5-pro" to "Gemini 2.5 Pro", claude-opus-4-6 raw to "Opus 4.6", claude-3-5-sonnet raw to "Sonnet 3.5", claude-sonnet-4.5 "Sonnet 4" to "Sonnet 4.5", gpt-5.6-luna "GPT-5" to "GPT-5.6 Luna", gpt-5.3-codex-spark "GPT-5" to "GPT-5.3 Codex" (still the wrong SKU). This is the display-name scope from the last two commits, which is the thing I asked to cut.
- dead code: the llm_request | llm.request CopilotEvent variant is declared, never read, and carries data?: any; LegacyResultMetadata is declared and never used; four whitespace-only edits in loadSpanAttributesFromTable, setReplayValue, inferTranscriptModel and createJsonlParser.

copilot.ts goes 3042 to 3799 lines. The legacy parser and the .json discovery are the part I want. Keep the .jsonl path behaviour identical to main except for the split-token fix, and take the display-name work to its own PR. If dotted legacy ids need pricing and display, do it as aliases in models.ts rather than by changing matching for every Copilot model.

JaylyDev:
> dead code: the llm_request | llm.request CopilotEvent variant is declared, never read, and carries data?: any; LegacyResultMetadata is declared and never used; four whitespace-only edits in loadSpanAttributesFromTable, setReplayValue, inferTranscriptModel and createJsonlParser.

That part was largely due to me copying the code from the VSCode copilot repo, I'll remove the dead code later.

I'll also remove the model naming procedure from the PR later, but I want to note that it's odd to have 2 processes on getting display names on models.ts and some providers. I would like to have more details on that choice

PR #1233, opened 2026-09-02
feat(cursor-agent): discover and reconstruct sessions from store.db (fixes #986)

### LINKED ISSUE #986: Cursor Agent provider misses CLI sessions stored in `~/.cursor/chats/**/store.db`
## Summary

CodeBurn misses recent Cursor Agent CLI sessions because the `cursor-agent` provider discovers only transcript exports under `~/.cursor/projects/**/agent-transcripts/`. Recent sessions are stored locally in per-session SQLite databases at:

```text
~/.cursor/chats/<hash>/<session-id>/store.db
```

The separate `cursor` provider does not cover these databases either; it reads Cursor IDE state from the platform-specific `globalStorage/state.vscdb`.

The `agent-tools/*.txt` files observed alongside recent sessions are terminal-output captures, not conversation transcripts or usage records.

## Current CodeBurn behavior

[`src/providers/cursor-agent.ts`](https://github.com/getagentseal/codeburn/blob/main/src/providers/cursor-agent.ts) probes only:

```text
~/.cursor/projects/
~/.cursor/ai-tracking/ai-code-tracking.db
```

Its discovery logic enters only directories named `agent-transcripts` and parses JSONL or legacy TXT transcripts.

[`src/providers/cursor.ts`](https://github.com/getagentseal/codeburn/blob/main/src/providers/cursor.ts) reads only Cursor IDE `state.vscdb`.

Neither provider discovers `~/.cursor/chats/**/store.db`, so store-only sessions—and sessions whose transcript export is absent, delayed, or incomplete—are silently omitted.

## Observed `store.db` format

Each session store contains:

| Table | Columns | Purpose |
| --- | --- | --- |
| `meta` | `key TEXT PRIMARY KEY`, `value TEXT` | Root metadata under key `0`; the value is hex-encoded UTF-8 JSON. |
| `blobs` | `id TEXT PRIMARY KEY`, `data BLOB` | Content-addressed conversation-state graph; IDs are 64-character hexadecimal strings. |

Decoded root metadata includes:

| Field | Meaning |
| --- | --- |
| `agentId` | Session ID; matches the containing directory. |
| `latestRootBlobId` | Current conversation-state root. |
| `name` | Session title. |
| `mode`, `isRunEverything` | Execution mode. |
| `createdAt` | Unix milliseconds. |
| `lastUsedModel` | Last selected model. |
| `blobEncryptionKey` | Local decryption material. |

Resolving the graph from `latestRootBlobId` yields structured user, assistant, and tool records, including conversation text, reasoning blocks, tool calls, tool results, request/model fields, and internal timestamps. Some observed stores also contain token- or context-related fields.

`blobEncryptionKey` is sensitive local material and must never be included in logs, diagnostics, errors, cache keys, exports, snapshots, or committed fixtures.

## The attribution DB is not a fallback

`~/.cursor/ai-tracking/ai-code-tracking.db` tracks AI-authored code attribution and related metadata. It may help enrich a session with model or project information, but it does not contain the complete current conversation stream or reliable request-level usage records.

It cannot substitute for reading `store.db`.

## Expected behavior

CodeBurn should discover and reconstruct Cursor Agent sessions from `~/.cursor/chats/**/store.db`, including sessions that have no matching transcript export.

For the same session UUID, source precedence should be:

1. `store.db`
2. JSONL transcript
3. Legacy TXT transcript

A malformed or unsupported store may fall back to a matching transcript, but a successfully decoded store and its transcript must not both be counted.

## Suggested implementation

- Add `~/.cursor/chats` to `cursor-agent` discovery and `probeRoots()`.
- Discover `chats/*/*/store.db` without following arbitrary symlink trees.
- Open each SQLite database read-only and validate the expected `meta` and `blobs` schema.
- Hex-decode `meta['0']`, validate `agentId`, and begin reconstruction at `latestRootBlobId`.
- Reconstruct the reachable blob graph while tolerating unknown protobuf fields and message block types.
- Use validated internal request/message timestamps when present.
- Use session metadata only as an explicitly session-level timestamp fallback; do not report database file mtime as an exact request timestamp.
- Prefer explicit request-level input, output, and cache counters only after validating their semantics.
- Do not treat a running context-window gauge as exact billed token usage.
- Estimate only missing token components and preserve exact/estimated provenance in the emitted calls.
- Deduplicate matching stores and transcripts using the session UUID plus stable request ID or turn index.
- Include each database’s WAL state in cache invalidation so active sessions refresh correctly.
- Retain the existing JSONL/TXT parser as a compatibility fallback.

## Acceptance criteria

- A store-only Cursor Agent session appears under `--provider cursor-agent`.
- User and assistant messages are reconstructed in order.
- Structured tool calls and tool results are retained.
- Valid internal timestamps determine date attribution.
- Explicit, semantically validated token counts take precedence over estimation.
- Estimated values remain visibly identified as estimated.
- A store and matching transcript produce one set of calls.
- An unsupported or corrupt store can fall back to its transcript without double-counting.
- `blobEncryptionKey` cannot reach logs, errors, caches, exports, snapshots, or fixtures.
- Active WAL changes invalidate cached results.

Tests should cover:

- An empty chats directory.
- A valid minimal store.
- A multi-turn blob graph.
- Malformed hex metadata.
- Missing or invalid root/blob references.
- Unknown protobuf fields and message variants.
- Seconds versus milliseconds timestamp validation.
- Exact, partial, gauge-only, and absent token data.
- Store/transcript deduplication.
- Transcript fallback after store decoding fails.
- WAL-aware cache invalidation.
- Secret-redaction behavior.

maheshsingh20:
## Summary

Fixes #986.

The `cursor-agent` provider previously missed CLI sessions stored in `~/.cursor/chats/**/store.db`. This PR adds full discovery and reconstruction for those sessions, with proper source precedence, deduplication, WAL-aware caching, and `blobEncryptionKey` redaction.

---

### `src/providers/cursor-agent.ts`

**Discovery**
- `discoverSessions()` now scans `~/.cursor/chats/<hash>/<uuid>/store.db` before the existing transcript walk, without following symlinks beyond two directory levels.
- `probeRoots()` gains a `chats` entry so `codeburn doctor` reports it.

**Schema validation**
- Checks for required `meta` + `blobs` tables before any data is read. A missing or incompatible schema emits a warning and falls through to the transcript fallback.

**Metadata decoding**
- `meta['0']` is hex-decoded from UTF-8 JSON to extract `agentId`, `latestRootBlobId`, `name`, `createdAt`, and `lastUsedModel`.
- `blobEncryptionKey` is **explicitly excluded** from the decoded object and never appears in logs, caches, exports, snapshots, emitted calls, or stderr output.

**Blob-graph reconstruction**
- BFS walk from `latestRootBlobId`, following `nextBlobId` chains and `childBlobIds` arrays.
- Classifies blobs by `role` (user / assistant / model / request).
- Handles both plain `text` strings and content-block arrays (`[{type:'text', text:'...'}]`).
- Handles both `toolCalls: [{name}]` and content-block `{type:'tool_use', name}` formats.
- Unknown fields are silently ignored for forward compatibility.

**Token provenance**
- Explicit per-request `inputTokens` / `outputTokens` used when present and plausible (<=2,000,000). Values above that threshold are treated as context-window gauges and replaced by char-based estimates.
- `cacheCreationInputTokens` and `cacheReadInputTokens` recorded when present.
- `costIsEstimated: true` set on any call where at least one count was estimated.

**Timestamps**
- Internal blob/request timestamps preferred; session `createdAt` used as session-level fallback. File mtime is **never** reported as an exact request timestamp.
- Both epoch-ms (> 1e12) and epoch-s (< 1e12) values are normalised correctly.

**WAL-aware in-memory cache**
- Per-session cache keyed on a fingerprint combining store.db mtime and the `-wal` sidecar size/mtime. Active sessions always trigger a re-parse when the WAL changes.

**Source precedence and deduplication**
- store.db sources are emitted first in `discoverSessions()`.
- A successfully decoded store records a `cursor-agent-store-decoded:<uuid>` sentinel in `seenKeys`; transcript parsers for the same session UUID check it and skip themselves.
- A store that fails schema/metadata validation does **not** set the sentinel, so the JSONL/TXT transcript fallback runs normally without double-counting.

---

### `tests/providers/cursor-agent-store.test.ts` (new, 44 tests)

| Category | Tests |
|---|---|
| Discovery | Empty/missing chats dir, valid store found, multiple sessions, non-UUID dirs skipped, no store.db skipped, probeRoots |
| Valid minimal store | Single-turn parse, model resolution, costUSD > 0 |
| Multi-turn blob graph | 3-turn nextBlobId chain, childBlobIds branching, stable dedup keys, seenKeys respected |
| Malformed metadata | Bad hex, invalid JSON, missing agentId, short blobId, agentId mismatch, missing meta table, missing blobs table |
| Missing/invalid blobs | Absent root blob, broken nextBlobId (partial graph still yields), corrupt blob JSON |
| Unknown fields | Ignored without crash, content-block array format, toolCalls array format |
| Timestamp normalization | Epoch-ms, epoch-s promoted to ms, session-level fallback |
| Token provenance | Exact both, input estimated, output estimated, gauge rejected, cache tokens, all absent |
| Store + transcript dedup | One set of calls per session UUID, transcript fallback on corrupt store, store-only session appears |
| WAL cache | Re-parse on file change, cache hit on unchanged fingerprint |
| blobEncryptionKey redaction | Absent from emitted calls, absent from stderr, absent from source paths |

---

## Test results

```
Tests  44 passed  <- new (cursor-agent-store.test.ts)
Tests  17 passed  <- existing cursor-agent.test.ts, no regressions
```

### DIFF
--- a/src/providers/cursor-agent.ts
+++ b/src/providers/cursor-agent.ts
@@ -1,11 +1,17 @@
 import { createHash } from 'crypto'
-import { existsSync } from 'fs'
+import { existsSync, statSync } from 'fs'
 import { readdir, readFile, stat } from 'fs/promises'
 import { join, basename } from 'path'
 import { homedir } from 'os'
 
 import { calculateCost, getShortModelName } from '../models.js'
-import { openDatabase, type SqliteDatabase } from '../sqlite.js'
+import {
+  openDatabase,
+  blobToText,
+  isSqliteAvailable,
+  isSqliteBusyError,
+  type SqliteDatabase,
+} from '../sqlite.js'
 import { normalizeContentBlocks } from '../content-utils.js'
 import { estimateTokensFromChars } from '../token-estimate.js'
 import type {
@@ -16,28 +22,15 @@ import type {
   ProbeRoot,
 } from './types.js'
 
-type ConversationSummary = {
-  conversationId: string
-  model: string | null
-  title: string | null
-  updatedAt: string | null
-}
-
-type AssistantTurn = {
-  body: string
-  reasoning: string
-  tools: string[]
-}
-
-type ParsedTurn = {
-  userMessage: string
-  assistant: AssistantTurn
-}
+// ---------------------------------------------------------------------------
+// Shared constants
+// ---------------------------------------------------------------------------
 
 const CURSOR_AGENT_COST_MODEL = 'claude-sonnet-4-5'
 const MAX_USER_TEXT_LENGTH = 500
 const DIGITS_ONLY = /^\d+$/
 const UUID_LIKE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i
+const HEX64 = /^[0-9a-f]{64}$/i
 const USER_MARKER = /^\s*user:\s*/i
 const ASSISTANT_MARKER = /^\s*A:\s*/
 const THINKING_MARKER = /^\s*\[Thinking\]\s*/
@@ -45,7 +38,19 @@ const TOOL_CALL_MARKER = /^\s*\[Tool call\]\s*(.+?)\s*$/i
 const TOOL_RESULT_MARKER = /^\s*\[Tool result\]\b/i
 const USER_QUERY_OPEN = '<user_query>'
 const USER_QUERY_CLOSE = '</user_query>'
+
+// Marker on SessionSource.path so the parser knows this is a store.db source
+// rather than a transcript. The marker is a private implementation detail;
+// it never appears in output.
+const STORE_SOURCE_PREFIX = 'cursor-agent-store:'
+
+// Sentinel written into seenKeys to record that a store.db was decoded
+// successfully for a given session UUID. Transcript parsers check for this
+// and skip themselves to prevent double-counting.
+const STORE_DECODED_KEY_PREFIX = 'cursor-agent-store-decoded:'
+
 const warnedUnrecognizedTranscripts = new Set<string>()
+
 const CONVERSATION_SUMMARY_QUERY = `
   SELECT conversationId, model, title, updatedAt
   FROM conversation_summaries
@@ -67,6 +72,10 @@ const modelDisplayNames: Record<string, string> = {
   default: 'Auto (Sonnet est.)',
 }
 
+// ---------------------------------------------------------------------------
+// Path helpers
+// ---------------------------------------------------------------------------
+
 function getCursorAgentBaseDir(baseDirOverride?: string): string {
   if (baseDirOverride) return baseDirOverride
   // Windows paths unverified; tracked as Open Question 3 in issue #55.
@@ -77,9 +86,72 @@ function getProjectsDir(baseDir: string): string {
   return join(baseDir, 'projects')
 }
 
+function getChatsDir(baseDir: string): string {
+  return join(baseDir, 'chats')
+}
+
 function getAttributionDbPath(baseDir: string): string {
   return join(baseDir, 'ai-tracking', 'ai-code-tracking.db')
 }
+
+// ---------------------------------------------------------------------------
+// Type definitions
+// ---------------------------------------------------------------------------
+
+type ConversationSummary = {
+  conversationId: string
+  model: string | null
+  title: string | null
+  updatedAt: string | null
+}
+
+type AssistantTurn = {
+  body: string
+  reasoning: string
+  tools: string[]
+}
+
+type ParsedTurn = {
+  userMessage: string
+  assistant: AssistantTurn
+}
+
+/** Decoded root metadata from meta table key '0'. */
+type StoreRootMeta = {
+  agentId: string
+  latestRootBlobId: string
+  name: string | null
+  mode: string | null
+  createdAt: number | null
+  lastUsedModel: string | null
+  // blobEncryptionKey is intentionally excluded — it must never be stored,
+  // logged, cached, exported, or appear in any emitted value.
+}
+
+/** Provenance tag for token counts. */
+type TokenProvenance = 'exact' | 'estimated'
+
+/** One reconstructed request/turn from the blob graph. */
+type BlobTurn = {
+  userMessage: string
+  outputText: string
+  reasoningText: string
+  tools: string[]
+  inputTokens: number
+  outputTokens: number
+  reasoningTokens: number
+  cacheCreationTokens: number
+  cacheReadTokens: number
+  inputProvenance: TokenProvenance
+  outputProvenance: TokenProvenance
+  timestamp: string | null
+  requestId: string | null
+  model: string | null
+}
+
+// ---------------------------------------------------------------------------
+// Utility functions
+// ---------------------------------------------------------------------------
 
 function estimateTokens(charCount: number): number {
   if (charCount <= 0) return 0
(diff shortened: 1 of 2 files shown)


ozymandiashh (changes requested):
Thanks for the work on #986. Requesting changes; as it stands the feature does not run in the CLI, and CI would fail at the first step. Details, all verified on the branch and on a merge with current `main` (which is clean):

1. **Store sources are discovered but never parsed.** `appendStoreSources` emits `path: "cursor-agent-store:<dbPath>:<uuid>"`. `src/parser.ts:3333` fingerprints every source first, and `fingerprintFile` (`src/session-cache.ts:1456`) only strips virtual suffixes, so it stats the literal `cursor-agent-store:/...` string, gets `null`, and `parser.ts:3348` skips the source silently. `createSessionParser` is never reached for any `store.db`. The 44 new tests pass only because they call `createSessionParser().parse()` directly. Fix is the existing convention: real path first, synthetic data as a suffix (`<dbPath>#cursor-store=<uuid>`, like `cursor.ts` does with `#cursor-ws=`). That also gets `-wal` folding for free and makes the hand-rolled `fingerprintStore` / `storeCache` unnecessary.

2. **`npx tsc --noEmit` has 6 new errors** in `src/providers/cursor-agent.ts:835-863`: `let hashDirs: Awaited<ReturnType<typeof readdir>>` resolves the Buffer overload. Annotate `Dirent[]` or let inference from the `withFileTypes: true` call site work. `tests.yml` runs tsc before vitest.

3. **`tests/provider-probe-roots.test.ts` fails** on the branch (passes on main) because the new `chats` probe root was not added to the existing assertion.

4. **Precedence only holds within one pass.** The `cursor-agent-store-decoded:<uuid>` sentinel is set when the store parser runs, but `parser.ts` only parses changed sources. In the #986 scenario (transcript export lands after the store), the store is unchanged, not re-parsed, no sentinel, and the transcript's calls are added on top of the cached store turns. Key namespaces are disjoint so cache dedup cannot catch it. Precedence has to be decided at discovery time.

5. **Signs this was not run against a real store:** `decodeStoreMeta` decodes `lastUsedModel`, `name`, `mode` and none is used; the model fallback is `source.project`, i.e. the session UUID, which then flows into `calculateCost`; the timestamp fallback is `new Date()`, which CONTRIBUTING forbids for parsers; `fetchBlob` assumes JSON where the issue describes protobuf; token fields are tried across four aliases each. The PR template's real-data section (`npm run dev -- today`, `models --provider cursor-agent`, terminal output against a real `~/.cursor/chats/**/store.db`) is missing entirely. CONTRIBUTING is explicit that guessing storage schemas is not acceptable; please install Cursor, generate a session, and paste what the CLI shows.

Credit where due: reusing `src/sqlite.ts` means read-only open, `busy_timeout` and `SQLITE_BUSY` handling are correct, the `blobEncryptionKey` redaction is real, and there is no path traversal or injection. The skeleton is fine; it needs to be wired in and proven on real data.

iamtoruk:
Closing this. I ran it against my own 22 store.db files.

The blob graph is protobuf, not JSON. All 22 root blobs are binary. `nextBlobId`, `childBlobIds`, `inputTokens`, `outputTokens` and `toolCalls` appear zero times across 5,426 real blobs. Pointing the parser straight at real stores yields 0 calls. Only the meta hex-JSON decode matches reality.

Before and after are identical here: 20 sessions, 25 calls, $0.0145. The sources never reach a parser, as flagged three weeks ago.

All five earlier points are still open. tsc still fails with 6 errors, probe-roots still fails, so CI stops at step one.

Two things worth knowing. Every store session on my machine already has a transcript, so there are no store-only sessions to recover. What I actually lose is 4 transcripts skipped as unrecognized format. That is the cheaper fix and probably the real issue.

Also, reasoning tokens are always estimated but not counted in provenance, so estimated tokens get priced as exact.

1,984 lines, roughly 140 usable. Please reopen #986 against a real store dump and land discovery plus the format work separately.

iamtoruk:
To be clear about the line count: the problem is not duplication. You reuse the token, cost and model helpers correctly, and the SQLite handling through src/sqlite.ts is right. The blob-graph code has to go because it parses a format that is not there, not because it repeats anything.

Also please pull the four modelDisplayNames additions into a separate PR.

PR #1242, opened 2026-09-02
fix(mac): treat omitted Grok credit percent as 0% in an active window

mh7whmyz6z-alt:
## Summary
- SuperGrok Heavy unified billing (`GET /v1/billing?format=credits`) drops `creditUsagePercent` after a weekly reset. The Capacity Dock treated that as an unrecognized payload and showed no usage, even though grok.com still shows **0%**.
- During an active `currentPeriod` (`start <= now < end`), an omitted percent is now **0% used**, matching Grok's web client / proto3 default. A published percent or a usable used/cap ratio still wins. An expired window without a percent is not invented as 0%.
- Live check on SuperGrok Heavy: ring showed 0% right after reset, then 1% after more usage. No 0% screenshot (state already moved on); the 1% state matched grok.com.

## Testing
- [x] `swift test --filter GrokBuildSubscriptionServiceTests` passes (8 tests)
- [ ] `npm test` — not run; this change is mac menubar only
- [ ] `npm run build` — not run; no CLI/JS change

### DIFF
--- a/mac/Sources/CodeBurnMenubar/Data/GrokBuildSubscriptionService.swift
+++ b/mac/Sources/CodeBurnMenubar/Data/GrokBuildSubscriptionService.swift
@@ -202,26 +202,28 @@ enum GrokBuildSubscriptionService {
             credential: credential,
             deps: deps
         )
-        let billing = try decodeBilling(billingData)
+        let billing = try decodeBilling(billingData, now: now)
         let settingsPlan = try await fetchPlan(credential: credential, deps: deps)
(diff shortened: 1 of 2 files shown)


mh7whmyz6z-alt:
Before this fix, when (GET /v1/billing?format=credits) drops creditUsagePercent the error caused CodeBurn to disconnect grok from the list of Providers, and attempting to reconnect would fail until usage was greater than 0%.  

This fix is to prevent the disconnection issue.

PR #1265, opened 2026-09-07
Show per-model cached tokens alongside cost in desktop and menubar

ozymandiashh:
## Summary

- Show per-model input, output and cache-read tokens alongside cost in the desktop Overview and macOS menubar Models rows. Keep observed token counts visible for unpriced models in the desktop Models tables.
- Carry counts through fresh-session and durable-day aggregation, using the same model-name grouping as cost and the existing billable-output normalization.
- Invalidate pre-change status snapshots once so an unchanged session corpus also receives the new fields.

Previously, the menubar payload discarded per-model token counts, Overview had no cache-read column, and the desktop Models table hid usage when a model had no attributed cost. A menubar row now includes a secondary line such as `12.3K in · 4.5K out · 67.8K cache read`, with its cost on the primary line.

Cache-read means reused input. **Cache pricing was already included in attributed cost**; this change exposes usage without changing pricing. Cache-write remains a separate payload field and is included in the macOS accessibility description. Reasoning is added to output only where the provider reports it separately.

The new fields are optional. Older payloads still decode, missing counts remain unknown, and known zero stays zero. Alias merges do not present partial counts when a contributor lacks data. Overview uses the period/provider-scoped model rows when available, with its previous daily-history source retained for older payloads. The existing 20-model cap remains. Daily/session cache formats and stored history are unchanged.

## Testing

- [x] Tested locally against real data
- [x] `npm test` passes in the completed branch validation run

Validation against the unchanged `main` base (`4a9d8852`):

- Root: main **3,738 passed**, branch **3,748 passed**, five skipped on each. An independent full rerun at 00:01 UTC hit three existing time-sensitive assertions in `cli-status-menubar.test.ts`: their fixture offsets placed calls in the future. The entire file passed **23/23** on rerun after that window; the other 3,725 tests passed in the full run.
- Desktop: main **827 passed / 27 failed / 4 skipped**; branch **829 passed / 27 failed / 4 skipped**. The same 21 `usePolled` and six `overviewSnapshot` failures occur under local Node 26 with unavailable jsdom localStorage; no new failures.
- macOS: **433 tests in 55 Swift Testing suites passed**. This is a partial run: 12 XCTest-importing suites cannot compile with the local Command Line Tools and were excluded, then restored. Full Swift-suite validation requires Xcode.
- Root/app typechecks and builds pass; `git diff --check` is clean.

The new CLI regression plants a render-v5 snapshot without token fields, verifies that the new binary recomputes the counts, then requires the next identical call to leave the snapshot byte-identical. Pipeline fixtures also cover fresh versus durable history, provider/range selection, alias merging, cache-only usage and unpriced models.

Real Hermes data returns `input=139, output=393, cacheRead=30,592, cacheWrite=0` with the existing attributed cost unchanged. A 15-model real corpus also reconciled model counts with the period totals.

Layout was checked using the actual branch components with fixture data: native SwiftUI views at the 360pt popover width, and desktop Overview/Models at 1200px and 900px. An additional 900px check reserved the 186px sidebar and applied the real content padding. Cache values, costs and zero-cost usage remain visible. These are isolated component/native-view renders, not validation of an installed release; no installed applications were replaced.


ozymandiashh:
Rebased on top of #1267, which should land first; against that branch this PR's own contribution is 18 files.

Render revision is now **8**, taken by bumping the shared `STATUS_SNAPSHOT_RENDER_VERSION` in `src/status-snapshot-semantic.ts` rather than a third inline edit, so `src/main.ts` drops out of this diff entirely. This PR's own test planted `:render-5:`, which main alone already satisfied, so it had stopped proving anything; it now plants the derived previous tag and asserts the current one.

Also fixed from the earlier review: the double per-session call walk is folded into one `sessionBillableOutput()` returning `{total, byModel}`, and Swift `exact()` no longer renders a negative as `--1,234,567`.

Two findings disclosed rather than changed, both documented at the call sites: the `mergeCount` unknown-vs-zero guard is genuinely unreachable on the durable path, and fixing it means widening `ModelDayStats` plus four daily-cache arithmetic sites, which is a durable-cache contract change too broad for a stacked PR; and the all-provider Overview table did silently gain the 20-row cap, so I corrected the PR body rather than the code.

tsc clean, swift build complete, full suite 3996 passed / 0 failed.

iamtoruk:
Landed as #1315 (ef97600d) with your two commits and authorship intact, rebased onto #1314 (the snapshot bump that lived in your merge resolution was restored into the first commit), plus two assertions pinning the per-model reasoning normalization. Verified on frozen dates: 16 added token fields, zero changed values, every cost, call and session identical to the digit; rows sum exactly to period totals and match models --format json. The Overview all-provider table now capping at 20 rows is tracked separately. Thank you.

PR #1267, opened 2026-09-07
Show cache-read usage and quota pace in Capacity Dock

ozymandiashh:
## Summary

- Show provider-scoped cache-read tokens in the Capacity Dock’s Today section, alongside input, output, calls and the existing cost. Known zero displays as `0`; missing or incomplete historical accounting stays unknown. Cache reads are already included in cost, so this adds visibility without changing pricing.
- Add per-window quota pace and estimated exhaustion using the existing whole-window average calculation. Captions require valid durations and fresh data; short windows retain deficit/reserve guidance without an exhaustion ETA. Four quota windows use a two-column grid to keep scope labels, reset times and estimates readable.

The JSON field is optional for compatibility with older clients and histories. Status snapshot revision 7 invalidates older cached payloads without purging daily history. No additional polling is introduced.

## Testing

- [x] Tested against a copied real Codex session in an isolated environment: CLI Today and the direct parser both reported **11,148,288 cache-read tokens**, 84 calls and one session.
- [x] `npm test`: **3,748 passed, 5 skipped**, versus **3,738 passed, 5 skipped** on the base revision; zero failures. Covers durable/fresh history, provider/date/config filters, known zero, incomplete legacy data and snapshot invalidation.
- [x] `npm run build` and `tsc --noEmit` passed.
- [x] Native `CapacityDockDetailView` render harness: **62 checks passed** across 0.9×, 1× and 1.25× scales and all four dock edges; rendered layouts inspected visually.
- [x] Swift Testing subset: **457 tests in 55 suites passed**, versus **424 tests in 53 suites** on the base. Includes fresh/stale quota samples, invalid windows, layout sizing, legacy payload decoding and overlapping/cancelled refreshes. `swift build --jobs 2` passed.

Local Swift validation used an isolated copy of the Swift Testing suites because this machine has Command Line Tools without XCTest. [macOS CI](https://github.com/getagentseal/codeburn/actions/runs/34076709968) subsequently passed the full suite: **129 XCTest tests and 457 Swift Testing tests**, plus release packaging.

All 18 CI checks passed on `98ad43fe`, including Windows tests and packaging (the release-only asset check is skipped).


ozymandiashh:
Rebased onto current main (after #1304, #1305, #1308, #1309) and absorbed the worthwhile parts of #1287, which I am closing in favour of this.

Ported: the agent-tab hover card's verdict placement and the plainer wording (`Lasts until reset` / `Runs out in 1d 20h` / `Limit reached`), both expressed through this PR's `QuotaPacePresentation` and its real `windowSeconds`. Deliberately NOT ported: `inferredWindowSeconds(label:)`, which is the Grok Build bug you found. New regression test `grokBuildMonthlyCycleLabeledWeekly` pins it: a monthly cycle at 72% with six days to reset, labelled `Weekly`, now reads `Lasts until reset` against its real 30-day window, where the label-inferred 7-day window would have printed `Runs out in 9h 20m`. A nil `windowSeconds` with the same label stays silent, which is what catches a re-introduced fallback.

Merge notes: render revision 7, and main's `rejects a pre-count-basis render-5 snapshot` test asserted `:render-6:`, which auto-merged clean and was wrong once this moved to 7. The suite now derives the tags from `STATUS_SNAPSHOT_RENDER_VERSION` so this cannot recur; the deliberately historical plants stay literal. `addProviderSlice` keeps your `Math.max` on sessions and sums `cacheReadTokens` beside it, with a comment on why one is additive and the other is not.

tsc clean, full suite 3986 passed / 0 failed, swift build complete. The gates are untouched: connection, ten-minute freshness, validated duration, percent-range and clock-skew rejection, 3% elapsed silence.

PR #1270, opened 2026-09-07
test: stabilize process smoke and cache-lock regressions

avs-io:
Three process-test fixtures can fail for reasons unrelated to their assertions: a chatty Node child gets only 600ms to boot, a hostile-model output check inherits color decoration, and a degraded-cache regression guesses when another process has reached the refresh lock.

The watchdog fixture writes immediately after boot and then regularly beyond a two-second silence budget; boot time still counts. The hostile-model child disables color while retaining every source-safety assertion. The degraded-cache case moves into the existing serial lock suite and waits for an observed failed exclusive lock acquisition before releasing the holder. Its child has bounded termination and is stopped before fixture cleanup.

The degraded response must still report stale, retain the older call count and write no snapshot; the next clean response must include the new call and persist its snapshot. Reintroducing the bad save gate makes the same test fail at the no-snapshot assertion. The watchdog fixed-cap negative control also fails as intended.

Validation on current main 0b1cf56f plus this branch: 3,745 CLI tests and 37 serial lock tests passed, with 5 existing CLI skips. The unchanged watchdog/hostile-model changes retain their earlier desktop and negative-control evidence. Independent review and root reproduction covered the lock handshake, assertion preservation and child cleanup. No production source differs from current main.


iamtoruk:
Landed via #1271 (159198be), which carried these commits. main already matches this branch on all five files.

Issue #1260, opened 2026-09-04
Same working directory splits into multiple projects across different providers (not covered by #217)


systemize-ai:
## Summary

A single physical working directory shows up as multiple separate "projects" in CodeBurn's `sessions`/`models --by-task` output when different providers/tools are used against it, because each provider's own path-sanitization convention produces a different project key for the same directory. This splits cost, call count, and session count across what should be one bucket in `By Project` views.

This is a different root cause than #217 (which was a single-provider, Windows-only case+slash variant issue and is closed). Here the paths are already consistent per-provider — the split happens *across* providers on the same Linux host, because each provider's own discovery code sanitizes `cwd` differently.

## Environment

- Platform: Ubuntu 24.04, Node 24.20.0
- CodeBurn: 0.9.24
- Providers involved: `claude`, `codex`, `hermes`, `kimicode`, `pi` — all pointed at the same physical directory `/root/vault`

## Repro

Run multiple different AI coding tools (Claude Code, Codex CLI, Hermes agent, Kimi Code CLI, Pi coding agent) against the same working directory, then:

```bash
codeburn sessions -p 30days --format json
```

## Actual output

The single directory `/root/vault` surfaces under three distinct project keys, one per sanitization convention:

```
-root-vault                 -> provider: claude   (Claude Code's own ~/.claude/projects/-root-vault/ naming, leading dash)
root-vault                  -> providers: codex, hermes   (no leading dash)
vault                       -> providers: kimicode, pi    (basename only)
```

Each of these accumulates its own separate cost/call/session totals in `By Project` views (`codeburn` TUI, `codeburn sessions`, `codeburn models --by-task`, etc.), even though they all refer to the exact same repo/directory. A user comparing "which project costs the most" gets a misleading answer — the true per-project total is silently split 3 ways.

This isn't specific to `/root/vault` — the same pattern reproduces for every directory used across more than one of these providers, since it's driven by each provider's own path-to-project-key logic, not by anything project-specific.

## Expected

Project keys derived from the same underlying absolute directory path should normalize to one canonical project across providers — e.g. resolve each provider's sanitized key back to a real filesystem path where possible (most providers already store or can derive the real cwd), then group by that resolved path rather than by the provider-specific sanitized string.

## Suggested fix

- Where a provider's session data includes (or can derive) the real absolute `cwd`, use that as the grouping key instead of (or in addition to) the provider's own sanitized folder-name convention.
[...]

PR #1271, opened 2026-09-07
fix(usage): preserve project identity and qualify session counts

avs-io:
Project summaries can split the same working directory across providers or attach a legacy pathless summary to the wrong project. Session totals can also look exact when the available history cannot establish unique session identities.

This preserves working-directory identity through discovery, parsing, caching and aggregation. It coalesces the same directory across providers, keeps distinct POSIX paths separate, preserves Windows path normalization, leaves ambiguous legacy summaries unallocated, and sums project savings correctly. Project details and charts use the same identities, including collision-safe rows and accessible full-path labels.

Counts now say **“At least N”** when history only supports a lower bound; unsupported per-session averages are omitted. Combined-device views show **“Session count unavailable”** when identities cannot be reconciled. The native menu uses the same count qualification and exposes project expansion as an accessible button.

**Review order:** this branch includes the two test-only prerequisite commits from #1270 because the existing process watchdog smoke test reproduced its startup race during verification. Please review/land #1270 first; I will remove those duplicate prerequisite commits when rebasing. The project/counting change is the separate `f5fd47e8` commit.

Validation on current main (`0b1cf56f`) plus this change:

- CLI build and full suite: 3,856 passed, 5 skipped. Also rerun from an unbuilt source export after correcting a regression test that originally assumed `dist/cli.js` existed: 3,856 passed, 5 skipped. The test now spawns the source CLI with an isolated temporary fixture and preserves the same assertions.
- Desktop full suite with the #1270 prerequisite: 864 passed, 4 skipped; typecheck and renderer build passed.
- Native focused count/payload tests: 8 passed across two suites.
- Independent adversarial source review, including comparison with the previously reviewed patch series.
- Personally exercised the freshly built renderer using actual CLI fixture outputs: ambiguous/case-distinct projects, independent expansion, keyboard collapse, provider filtering, and a legacy-to-current refresh without leftover rows. This uses an isolated read-only bridge, not the installed app's transport.

Related to #1260. No dependency or workflow changes.

Upgrade handling: invalidate render-5 status snapshots written before the count metadata existed, including same-version rebuilds. Daily/session accounting caches are retained. Independently reproduced the old-snapshot failure, verified the correction with the unchanged reproduction, and passed 50 focused/adjacent CLI tests covering empty zero, nonempty totals, carried historical accounting and subsequent warm snapshot reuse.


iamtoruk:
Pushed df38e638 on top of the branch:
- spendProjectIdentity now folds Windows case like normalizeAbsProjectPathKey; cache C:\Work\Vault + live c:/work/vault was producing two payload rows
- overview projectAggKey uses the same helper, so POSIX case stays identity in the CLI table
- --json rows carry sessionCountBasis and drop avgCostPerSession when the basis is not exact; overview block gets the basis too
- deleted the five root-*.verify.test.ts scratch files (console.log, tautological asserts); ported the two uncovered assertions into existing suites
- renamed the codex-review-* tests, scrubbed REVISE markers, removed unused label helpers, Set in addSessionOnce, compact session label on the Windows and Mac project rows
Root 3853 pass, tsc clean, app typecheck clean, swift build + SessionCountLabel tests pass, CI green.

iamtoruk (approved):
Identity, qualification, cache invalidation and privacy paths verified; patch df38e638 closes the Windows/overview identity gaps and the --json qualification. Merging.

PR #1272, opened 2026-09-07
fix(doctor): report filesystem roots for seven providers

avs-io:
`doctor` cannot explain an empty or missing installation when a filesystem provider does not expose its search roots. Seven providers currently have this gap.

Add `probeRoots` for Forge, Mux, Open Design, OpenClaw, ZCode, Zed and Zerostack, using the same paths as their discovery code. OpenClaw reports the agents directory, allowing an existing empty root to be distinguished from a missing one. Vercel Gateway remains a network-only exception.

Validated against current main (`0b1cf56f`): 53 focused and adjacent provider/doctor tests passed, and TypeScript checking passed. The added census covers all non-network providers and explicitly checks the Vercel exception. Parsing and accounting behavior is unchanged.

Addresses the filesystem-root coverage portion of #899.

### DIFF
--- a/src/providers/forge.ts
+++ b/src/providers/forge.ts
@@ -5,7 +5,7 @@ import { join } from 'path'
 import { extractBashCommands } from '../bash-utils.js'
 import { calculateCost } from '../models.js'
 import { getSqliteLoadError, isSqliteAvailable, openDatabase, type SqliteDatabase } from '../sqlite.js'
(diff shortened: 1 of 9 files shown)


iamtoruk (approved):
Verified: paths match discovery for all seven, census + openclaw + doctor tests pass, tsc clean. Pushed b2f6df19 to trim the review-note comment and pin the OpenClaw empty-root probe.

PR #1274, opened 2026-09-09
fix(ci): prepare star-history updates on a review branch

avs-io:
Star History currently fails when its generator pushes directly to protected main (run 34311396184, GH013). Generate charts without the action's commit/push step, then update a dedicated review branch and put a comparison link in the run summary. A maintainer opens or reviews the chart PR. Runs on other branches are skipped.

This respects the repository's existing prohibition on bot-created PRs and retains the existing contents-write permission. The generator stays pinned to its current immutable revision.

Validation: checked the pinned action contract and YAML/shell syntax; executed the publishing script against an isolated bare Git repository with checkout-style shallow fetches. First update, repeated update and unchanged-tree no-op passed; main remained unchanged, no PR API was called, and both updates produced the expected comparison URL. Remote execution remains to be verified after merge.

### DIFF
--- a/.github/workflows/star-history.yml
+++ b/.github/workflows/star-history.yml
@@ -7,7 +7,6 @@ on:
 
 permissions:
   contents: write
-  pull-requests: write
 
 concurrency:
   group: star-history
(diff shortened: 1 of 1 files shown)


iamtoruk (approved):
Checked the pinned action supports commit:false and the changed output; the job has failed daily on GH013. Merging.

Issue #1273, opened 2026-09-09
no matter how many times i disable copilot in the menubar, it re-appears


AbdelrhmanHamouda:
After the recent huge update, copilot auto appeared on my menubar and its always red because "i'm out of quota" butthe thing is, i don't havwe copilot active and i have disabled it a million times 

<img width="343" height="43" alt="Image" src="https://github.com/user-attachments/assets/92d4abb1-4dab-4fd0-ad5b-88134491a68a" />

<img width="861" height="489" alt="Image" src="https://github.com/user-attachments/assets/2f00a427-2b6b-4985-ada2-5c949c814622" />

avs-io:
Thanks for reporting this. Fixed in #1278, which is now merged: disconnecting Copilot stays in effect across refreshes and app restarts until you explicitly connect again.

Closing as fixed on `main`. The latest release predates this change, so the fix will be available in a subsequent release.

AbdelrhmanHamouda:
thank you and i realise that my bug report can read a bit rough. i was having a bad day and i shouldn't have let that seep into the bug report. 
my apologies and thank you for such a great tool and for your kindness + timely response

PR #1278, opened 2026-09-09
fix(mac): preserve explicit Copilot disconnect

avs-io:
Refs #1273.

After Disconnect, discoverable Copilot credentials could reactivate quota tracking on the next refresh or relaunch. Persist the explicit opt-out, preserve it during automatic refresh, and clear it only when the user connects again. Credentials and historical usage are preserved. Plan, Settings, and Capacity Dock distinguish an explicit disconnect from missing credentials.

Validation: the initial fix passed 46 focused/adjacent standalone tests and 47 integrated tests, including in-flight cancellation and same-defaults relaunch. The presentation follow-up passed 16 focused tests. A universal native QA build passed compilation and strict ad-hoc signature verification. Actual UI checks exercised Disconnect, Refresh, normal Quit, same-profile relaunch remaining disconnected, and explicit Connect resuming discovery. Capacity Dock's incorrect OAuth guidance was reproduced and the corrected message and Connect action verified in the app. Tests used a separate bundle identity, a fictional token, and blocked network/Keychain access; no live-account or release-signing acceptance is claimed.

### DIFF
--- a/mac/Sources/CodeBurnMenubar/AppStore.swift
+++ b/mac/Sources/CodeBurnMenubar/AppStore.swift
@@ -166,8 +166,9 @@ final class AppStore {
     var copilotError: String?
     // Same file-based activation as Kimi/Gemini — reading
     // Copilot discovery never raises a keychain prompt, so we start dormant and
-    // auto-activate on the first refresh tick.
-    var copilotLoadState: SubscriptionLoadState = CopilotSubscriptionService.hasCredential ? .dormant : .notBootstrapped
+    // auto-activate on the first refresh tick unless the user explicitly
+    // disconnected. That opt-out is persisted and applied in `init`.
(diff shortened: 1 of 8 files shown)


iamtoruk:
Pushed 1ffa3ec6: removed the six post-await isSet guards. disconnectCopilot bumps copilotRefreshGen in the same MainActor block it sets the flag, so the existing generation guard on the line above already returns; all 8 focused tests pass without them. Full swift test 442/442. Note the Dock's "Complete Copilot OAuth, then click Retry" copy still shows for the no-credentials and terminal-failure cases; Kimi, Gemini and Antigravity have the same resurrect-on-relaunch bug. Merging.

iamtoruk (approved):
Verified: bug reproduced on main, fix single-writer and provider-namespaced, 442/442 swift tests.

PR #1282, opened 2026-09-10
fix(kiro): emit workingDirectory so sync attribution can resolve the repo

Enclavet:
## Summary

- Every kiro session was invisible to `codeburn sync push --attribution` — it reported `Attribution: 0 facts total` at every window, so no session→commit correlation ever reached the endpoint.
- Cause: `project-path-v1` wired the session's directory to `projectPath`, but sync attribution reads `session.workingDirectory`. `computeAttributionRecords` calls `buildRepoGroups` in `"trusted-session-cwd"` mode, resolves the repo from that field, then filters to sessions whose own directory resolved (`group.ownIdentity`). With it unset, `isTrustedAbsoluteWorkingDirectory(undefined)` returns false on its first clause, every session is dropped, and `attributeCommits` joins the repo's commits against an empty session list. The path was being carried the whole time, just on the field nothing consumed.
- Fix: all three kiro parsers now emit `workingDirectory` from the same provider-recorded value they already use for `projectPath`, plus a `PROVIDER_PARSE_VERSIONS.kiro` bump so warm caches re-parse.

Downstream, three things change for kiro sessions, each the same treatment every other provider with a working directory already gets: plain `sync push` usage spans gain `ai.project` (the cwd basename, scrubbed by `safeProjectAttribute`) and `ai.subscription_covered`; and sessions become eligible for working-directory PR correlation, so a kiro session sharing a cwd with a PR-anchored session within 6h picks up its `prLinks`. The absolute path itself never leaves the machine: attribution records carry the normalized repo remote, shas and timestamps, with `project` derived from the remote's last segment.

`parser.ts` already stamps `workingDirectoryProvenance: 'provider-field'` on any provider-set value that clears the trust check, and that marker is what the consumer requires — an unmarked value is treated as synthesized from `projectPath` and fails closed.

The three sources, unchanged in meaning from what each parser already read:

| Parser | Field |
|---|---|
| CLI (`~/.kiro/sessions/cli/<id>.jsonl`) | companion `.json` → `cwd` |
| v2 IDE (`~/.kiro/sessions/<hash>/sess_<id>/`) | `session.json` → `workspacePaths[0]` |
| workspace sessions | record → `workspaceDirectory` |

The cache bump matters as much as the parser change: entries cached under `project-path-v1` hold `projectPath` but no `workingDirectory`, so without it the fix would miss every user who already has a warm cache — that is, everyone upgrading.

## Testing

- [x] `npm test` passes

Measured against a real corpus — 2457 CLI sessions and 8 v2 IDE sessions — same command, same window, patched build vs installed:

```
before:  [dry-run] Attribution: 0 facts total, would push 0 (0 sessions, 0 commits)
after:   [dry-run] Attribution: 190 facts total, would push 190 (21 sessions, 169 commits)
```

Then a real push, and what the receiving endpoint recorded:

```
Synced 11 calls ($4.85) to https://<endpoint>
  Attribution: 190 facts synced

receiver: attribution: received=190 written=190 | rejected=0
stored:   OTEL_ATTR_SESSION 21 | OTEL_ATTR_COMMIT 169
          ai_tool=kiro (169)  ai_origin=ai-generated (169)
```

Commits previously recorded as `human` by the CI census were corrected to `kiro` / `ai-generated` on the same key, which is the interaction this field was always meant to enable.

On `npm test`: 3860 pass, and `tests/perf-harness-phase0.test.ts` fails **identically on clean `main`** in this environment — the perf harness refuses to write fixtures under a real home directory (`assertIsolatedHome`). Verified by stashing the change and re-running. Unrelated to this diff.

New tests cover `workingDirectory` for all three session shapes, re-parse of a `project-path-v1` cache, and the home-root rejection — the last is not hypothetical: 5 of the 8 real IDE sessions carried `workspacePaths: ['/home/<user>']`, which `isUserHomeRoot` rejects by design, so the field must be absent rather than present-and-untrusted.

Provenance is asserted implicitly rather than directly: `parser.ts` only promotes a call's `workingDirectory` onto the session when the marker is present, so a populated `session.workingDirectory` is itself proof the marker was stamped. The marker lives on cached calls, which `parseAllSessions` does not surface.

## Scope

This recovers sessions whose recorded `cwd` sits inside a git work tree — 70 of the 2457 in the corpus above. Sessions run from directories outside any repo are unaffected, and no parser change can reach them; that is a separate question about what kiro records, not about this field.

### DIFF
--- a/src/providers/kiro.ts
+++ b/src/providers/kiro.ts
@@ -510,7 +510,12 @@ function parseCliSession(meta: KiroCliSessionMeta, entries: KiroCliEntry[], seen
       userMessage: pendingUserMessage,
       sessionId,
       project,
-      ...(meta.cwd ? { projectPath: meta.cwd } : {}),
+      // Both fields, from the same provider-recorded value: projectPath drives
+      // display/grouping, while workingDirectory is what sync attribution reads
+      // (buildRepoGroups' "trusted-session-cwd" mode). Setting only projectPath
+      // leaves every session attribution-blind — see the note on
+      // PROVIDER_PARSE_VERSIONS.kiro.
+      ...(meta.cwd ? { projectPath: meta.cwd, workingDirectory: meta.cwd } : {}),
     })
     turnIndex++
   }
@@ -668,7 +673,10 @@ async function parseWorkspaceSession(record: Record<string, unknown>, source: Se
     userMessage: pendingUserMessage,
     sessionId,
     ...(typeof record['workspaceDirectory'] === 'string' && record['workspaceDirectory']
-      ? { projectPath: record['workspaceDirectory'] as string }
+      ? {
+          projectPath: record['workspaceDirectory'] as string,
+          workingDirectory: record['workspaceDirectory'] as string,
+        }
       : {}),
   })
 
@@ -782,7 +790,9 @@ async function parseV2Session(source: SessionSource, seenKeys: Set<string>): Pro
           userMessage: turnUserMessage,
           sessionId,
           project: source.project,
-          ...(meta.workspacePaths?.[0] ? { projectPath: meta.workspacePaths[0] } : {}),
+          ...(meta.workspacePaths?.[0]
+            ? { projectPath: meta.workspacePaths[0], workingDirectory: meta.workspacePaths[0] }
+            : {}),
         })
       }
     }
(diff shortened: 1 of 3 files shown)


iamtoruk (approved):
Traced the attribution chain end to end, confirmed the trust check and cache bump, full suite 3861/0, new tests fail against main's kiro.ts. Pushed 7d6388a0 to shorten the inline note and amended the description with the three downstream span changes. Merging.

Issue #1075, opened 2026-08-21
Codex: price cache_write_input_tokens, avoid reasoning-output double count, and apply long-context tiers


chr-evensen:
- Codeburn: `0.9.20`
- Provider: Codex Desktop and CLI
- Model: `gpt-5.6-terra`

**Observed behavior**
For Codex `token_count` events, the session telemetry includes:
- `input_tokens`
- `cached_input_tokens`
- `cache_write_input_tokens`
- `output_tokens`
- `reasoning_output_tokens`

The parser currently:
1. derives uncached input as input_tokens - cached_input_tokens;
2. ignores cache_write_input_tokens and passes zero cache-creation tokens to pricing;
3. prices output_tokens + reasoning_output_tokens;
4. uses one static model-rate record, without selecting OpenAI short- versus long-context prices per call.

**Why this is inaccurate**
- Cache-write tokens are a subset of uncached input, not an additional fifth token category. They must be split out of uncached input and charged at OpenAI’s cache-write rate—without double-counting their volume.
- In Codex telemetry, reasoning output appears to be included in `output_tokens`: e.g. `23,318 input + 251 output = 23,569 total`, while reasoning output is `53`. Adding reasoning to output therefore double-bills those tokens.
- OpenAI’s model pricing has different short- and long-context matrices. The parser already has each call’s input volume, but Codeburn applies a single LiteLLM rate. Long-context calls can consequently be materially underpriced.

**Concrete test session**
- 59 calls, all `gpt-5.6-terra`
- 7,748,962 input tokens total
- 7,568,256 cached input tokens
- 180,706 uncached input tokens
- 180,430 cache-write input tokens
- 27,027 output tokens
- 10,098 reasoning-output tokens
- 44 calls had input totals of at least 128k tokens; maximum was 180,433

**Suggested correction**
- Read cache_write_input_tokens.
- Clamp it to uncached input, charge that portion at the cache-write rate, and charge only the remainder at the standard input rate.
- Treat reasoning_output_tokens as an output breakdown for Codex unless the source explicitly establishes it is exclusive.
- Add model-specific short/long-context threshold handling rather than a single static price record.

iamtoruk:
Thanks for the detailed report and the concrete session numbers — they made this verifiable. I ran all three claims against a large real corpus (1,396 rollout files, 136,094 `token_count` events, 45,638 priced calls) plus OpenAI's published docs. Verdicts:

[...]

PR #1284, opened 2026-09-11
Fix DSH v0–v3 session discovery and usage accounting

### LINKED ISSUE #1281: DSH v3 sessions are not detected
## Problem

CodeBurn does not detect DSH v3 session logs, leaving their usage out of reports. The files are named `session.v3.jsonl.zstd`, which the DSH provider never scans. When legacy logs are also present, `doctor` reports the provider as OK without warning about the omitted v3 files.

## Reproduction

Environment: CodeBurn 0.9.24 (Homebrew), macOS, Node v26.8.1, DSH `0.1.5-rc.1` source checkout writing Session format v3.

I reproduced this with existing local sessions using:

```sh
codeburn doctor --provider dsh --json
codeburn sessions --provider dsh --period all --format json
```

The directory layout can include both generations of a session:

```text
~/.dsh/sessions/<project>/<session-id>/
  session.jsonl.zstd
  session.v3.jsonl.zstd
```

## Suggested fix

The [current reader](https://github.com/getagentseal/codeburn/blob/0fea397d1446d7ec4db7a3215dfd4e233a69a5e5/src/providers/dsh.ts) only looks for `session.jsonl.zstd` / `session.jsonl` and accepts format version 0. Versioned files never reach its unsupported-version check.

DSH's [format policy](https://github.com/deepseek-ai/deepseek-harness/blob/master/docs/cookbook/adding-a-session-format-version.md) freezes released formats and adds an adjacent migration for future structural changes. Versioned logs are part of that storage design. Supporting the current formats requires decoding each version and selecting one generation per session; simply scanning every file would double-count migrated logs. Unknown future versions should produce a warning.

Would you be open to a follow-up to #1001 adding v0–v3 support? I can validate it against DSH's own token totals on existing sessions and contribute sanitized fixtures.

xiaozhenliu:
## Summary

Fixes #1281. Implements the v0–v3 follow-up approved in [the issue discussion](https://github.com/getagentseal/codeburn/issues/1281#issuecomment-5624243938).

- Discover canonical `session.jsonl` and `session.v<N>.jsonl` files, including `.zstd`, and select the numerically highest generation **within each session directory**. An unreadable or unsupported highest generation never falls back to an older snapshot.
- Parse v0/v1 usage events and v2/v3 embedded assistant streams, preserve retry usage, exclude inherited fork usage, and preserve raw output tokens and use CodeBurn's shared inclusive-reasoning rule.
- Report unsupported-generation skips in doctor's JSON and text output, warn once per unknown version, and invalidate stale session and finalized daily caches.

**Separate legacy and newer-format sessions are all counted.** A newer-format session never supersedes a different legacy session. Only multiple generations inside the same session directory are alternatives: `session-A/session.jsonl` and `session-B/session.v3.jsonl` both contribute; `session-C/session.jsonl` alongside `session-C/session.v3.jsonl` contributes C's v3 usage once.

## Testing

- [x] `npm test` passes

Full-suite comparison on Node v26.8.1:

| Revision | Passed | Skipped | Failed |
| --- | ---: | ---: | ---: |
| Upstream base `6a0f830` | 3,863 | 5 | 0 |
| This PR branch | 3,913 | 5 | 0 |

`npm run test:locks`: 36 passed. `npx tsc --noEmit` and the complete CLI/dashboard build also pass.

The four sanitized fixtures (`tests/fixtures/dsh/v0.jsonl` through `v3.jsonl`) pass DSH's official strict format restoration and `tokenUsage` reducer at DSH commit `c291e7961a515f6d7af9304e7fd1d257929aef26`. All four produce uncached input 100, full output 20, cache read 30, and cache write 5; CodeBurn tests assert those values, including raw output and informational reasoning.

The mixed-generation regression creates six log files across five session directories and verifies that exactly five sessions and their token totals survive cold, warm, and cache-reloaded reads. Further tests cover non-adjacent settlements across all four versions, retries that reset only the matching step, non-fork legacy seed headers, fork boundaries, missing usage, multi-frame zstd and torn tails, unknown/unreadable successors, model attribution, date filtering, parser-cache upgrades, and finalized daily re-derivation.

Local real-data validation compared matching session/log watermarks against DSH's persisted `tokenUsage` projection, excluding stale snapshots. No private logs, session identifiers, or usage history are included.

## Review notes

- The optional discovery callback reports each unsupported session to doctor independently of warning deduplication, without retaining shared diagnostic state between scans.
- `PROVIDER_PARSE_VERSIONS.dsh` forces warm session caches to reparse. Daily cache v32 uses the existing provider-scoped repair mechanism so finalized DSH days can adopt corrected totals even when selecting the authoritative generation reduces the call count.
- This supports JSONL persistence. Missing provider usage and reasoning detail absent from pi-ai logs are not inferred; pricing continues to use CodeBurn's existing tables.

The additional `v3-retry.jsonl` fixture includes failed and successful attempts separated by a scheduled retry, with exact `totalTokens`. DSH's official strict restore and reducer produce 110 input, 24 output, 33 cache read, and 7 cache write tokens (174 total). See the [public TokenUsage contract](https://github.com/deepseek-ai/deepseek-harness/blob/c291e7961a515f6d7af9304e7fd1d257929aef26/docs/subsystems/llm-streaming.md#tokenusage) and [native adapter](https://github.com/deepseek-ai/deepseek-harness/blob/c291e7961a515f6d7af9304e7fd1d257929aef26/packages/llm/llm-deepseek/src/translate.ts#L47-L72) for inclusive reasoning semantics. The parser fingerprint is bumped again to reparse caches containing the previous exclusive-output representation.


ozymandiashh (changes requested):
Discovery, doctor and the cache bump correctly address #1281, and the touched suites, `tsc` and the full run are green. Three things block the merge, two of them silent accounting regressions I reproduced against `main`.

1. **v0 steps double-count when settlements are non-adjacent.** The per-bucket "final report wins" rule became one global `projectionState.last` slot. Any usage event for a different `(turn, step)` between a step's `assistant/chunk` sample and its `assistant/message` settlement closes the slot, so the settlement is pushed as an extra `:attempt:2` observation instead of replacing the sample. Plain v0 log, no retry events, order `chunk 1:1, chunk 1:2, message 1:1, message 1:2`: main 2 calls / 300 input, this branch 4 calls / 600 input. Same for v2/v3 with `assistant/attempt` + `assistant/message` interleaved. With `DAILY_CACHE_VERSION` 32 re-deriving finalized days, that would be written into sealed history. Fix: key the slot per `${turn}:${step}` (a `Map`, cleared by `llm/retry-started` for that key) and add a test with non-adjacent settlements.

2. **The v0/v1 seed cut lost the `parentSession` guard.** Main only discards `seq <= seedLength - 1` when `parentSession` is present; the branch does it whenever `header.seedLength` is a number. A non-forked v0 header with `seedLength: 4`: main 400 input, this branch 300, first call gone. Not mentioned in the body, not covered by a test (the only fork test supplies both fields). Restore the guard or justify in a code comment plus test why `seedLength` alone implies a replayed prefix.

3. **The reasoning-tokens flip needs a source.** Main treats DSH reasoning as additive at the output rate; the branch treats it as a subset of `outputTokens` and pre-subtracts inside the provider. That lowers every user's historical DSH cost retroactively, and nothing in this repo can verify the premise (no fixture carries a total). Point me at the DSH `TokenUsage` definition that says reasoning is included in output. If it is, implement it the way #1075 established: add `dsh` to `REASONING_INCLUDED_IN_OUTPUT` in `src/models.ts` and keep the provider's raw `outputTokens`. The numbers are identical, and `dsh.outputTokens` stops meaning something different from every other provider.

Minor: `v1.jsonl` is byte-identical to `v0.jsonl` except the version field and `v3` differs from `v2` only by a missing `request/header`, so "one fixture per version validated against DSH totals" holds nominally. A retry / `assistant/attempt` fixture would make the new paths actually load-bearing.

xiaozhenliu:
Thanks for the concrete repros. Addressed all three blockers in cd2e0dd:

1. Replaced the global settlement slot with a Map keyed by `(turn, step)`. Only `llm/retry-started` for that key resets its slot. Added v0–v3 interleaved-sample/final regressions, both with and without a retry on one of the steps. The no-retry case now yields exactly two calls / 300 input. I ran the regressions before the fix and reproduced the double count.
2. Restored the non-empty `parentSession` guard for legacy v0/v1 seed exclusion. Added tests showing that `seedLength` without a parent retains all four calls / 400 input; existing fork tests still pass. These also failed before the fix.
3. Added `dsh` to `REASONING_INCLUDED_IN_OUTPUT`, retained raw provider `outputTokens`, and routed fresh pricing through `billableOutputTokens`. Cache and daily/session aggregation tests cover inclusive output, and the parser version is now `session-formats-v0-v3-attempts-v5` to reparse the prior representation.

The public DSH [TokenUsage contract](https://github.com/deepseek-ai/deepseek-harness/blob/c291e7961a515f6d7af9304e7fd1d257929aef26/docs/subsystems/llm-streaming.md#tokenusage) explicitly defines reasoning as detail included in output. The [native adapter](https://github.com/deepseek-ai/deepseek-harness/blob/c291e7961a515f6d7af9304e7fd1d257929aef26/packages/llm/llm-deepseek/src/translate.ts#L47-L72) preserves `completion_tokens` as output, derives total from prompt plus completion, and separately preserves reasoning detail. The earlier SHA in the fixture notes was a local checkout commit unavailable upstream; I replaced it with this public revision after checking the relevant codec/reducer sources match.

Also added `v3-retry.jsonl` with an `assistant/attempt`, retry scheduling/start, and successful settlement. It includes exact totals and passes DSH's official strict restore and tokenUsage reducer: input 110 / output 24 / cache read 33 / cache write 7, total 174. The fixture regression checks two separately counted attempts and shared billing totals.

Validation: full suite 3,913 passed / 5 skipped / 0 failures; unchanged upstream base 6a0f830 has 3,863 passed / 5 skipped / 0 failures. Lock tests 36/36, type check, and full build pass. The PR description has been updated to match the revised implementation.

ozymandiashh (approved):
Re-reviewed cd2e0dd4. All three blockers are fixed the right way: the settlement slot is a `Map` keyed by `(turn, step)` and my interleaved v0/v2/v3 probes now match main (2 calls / 300 input, retry on one step still additive); the non-empty `parentSession` guard is back for v0/v1 (400 input on a non-fork header, forks still cut their prefix); `dsh` sits in `REASONING_INCLUDED_IN_OUTPUT` with raw `outputTokens` preserved, and the cited `llm-streaming.md` at c291e796 does define reasoning as detail included in output. The `v3-retry.jsonl` totals corroborate it in-repo.

Mutation check with `src/` reverted to a2f1fe20: 22 of the new tests fail, so the coverage is load-bearing. Full suite on the head: 3913 passed / 0 failed, `test:locks` 36/36, `tsc` clean. Approving; merging once CI is green.

Issue #1276, opened 2026-09-09
Desktop app cannot be scoped to a subset of projects


mathieustan:
### What I am trying to do

I want to demo the desktop app to other people without showing every project on my machine. The sidebar and every section list all of them, client work included, and nothing in the UI narrows that down.

### What I found looking for a workaround

The README says:

> The `--project` and `--exclude` flags work on all commands and can be combined with `--provider`.

That is not the case. Only `report`, `overview`, `web`, `status`, `today`, `month` and `export` declare them. Seven others do not: `sessions`, `models`, `compare`, `audit`, `optimize`, `yield` and `spend`.

```
$ codeburn sessions --project my-company
error: unknown option '--project'
```

So a filter passed down from the Electron layer only reaches the one `status` call. The Sessions list keeps showing everything, which is what sent me looking at the CLI in the first place.

The second thing I hit is subtler. `filterProjectsByName` treats every pattern as a raw substring, so excluding a project by its path also takes out the neighbors that share the prefix:

```
$ codeburn sessions --exclude=/Users/me/work/my-company
# my-company-kit disappears too
```

There is no pattern I can type that means this project and not its siblings. `isProxiedPath` already handles the same problem for `proxyPaths` by checking a path-segment boundary.

### What I would like to propose

The flag gap looks cheap to close. All seven commands already funnel through `parseAllSessions`, so the existing `filterProjectsByName` can be applied at that one point. `yield` and `spend` parse inside `computeYield` and `computeSpendFlow`, which would take the patterns as optional arguments.

For the substring problem, an absolute path (POSIX or Windows drive) would match the way `isProxiedPath` matches: equal to the project path, or a prefix of it on a path-segment boundary. Anything that is not a path stays the loose substring people actually type, so `--project my-company` still covers `my-company-kit` and worktrees. I read this as a bug fix rather than a behavior change, since anyone passing a full path today already gets neighbors they did not ask for.

Then a Projects pane in Settings, one switch per detected project. I put the filter in `~/.config/codeburn/app-filter.json` and read it in the main process, since that is where the CLI argv gets assembled and renderer storage never reaches it. No file, or empty lists, means no filter, so nothing moves for anyone who never opens the pane.

[...]

PR #1285, opened 2026-09-11
fix(cli): honor --project/--exclude on every reporting command, and anchor absolute-path patterns

mathieustan:
The CLI half of #1276, split out as you asked. The desktop Projects pane follows as a second PR on top of this branch.

## What is in here

**1. The seven commands that dropped the flags now honor them.** `sessions`, `models`, `compare`, `audit`, `optimize`, `yield` and `spend` declare `--project`/`--exclude`, and the filter is applied at the one point they all funnel through (`parseAllSessions`). `yield` and `spend` parse inside `computeYield`/`computeSpendFlow`, so those take the patterns as optional arguments.

**2. An absolute-path pattern is anchored to a path-segment boundary.** Built on `normalizeAbsProjectPathKey` / `foldIdentifiedWindowsPath` from #1271, as you asked, rather than a second copy of the `isProxiedPath` logic. Both the pattern and the project path go through the same helper, so one identity rule holds for both:

- identified Windows drive and UNC paths casefold, POSIX case stays identity (`/a/Vault` and `/a/vault` are two projects)
- a Codex-style stripped absolute path (`root/vault`) keys the same as the `/root/vault` a user types
- rootedness alone picks the branch, so `/` names no project instead of falling through to a substring that matches every path

A pattern that is not rooted stays the loose substring people actually type: `--project my-company` still covers `my-company-kit` and worktrees, and a fragment like `alice/projects` still matches inside a longer path.

The same rule reaches the carry-forward day cache: `dayProjectMatches` in `src/usage-aggregator.ts` used to carry its own lowercasing substring copy, so a filtered headline anchored today's live days and substring-matched every cached day inside one total. It now calls `matchesProjectPattern`, and the callers stop pre-lowercasing.

**Disclosure:** this makes rooted POSIX patterns case-sensitive. `--exclude /users/me/vault` no longer matches `/Users/me/vault`, because pattern and project path now go through the identity rule from #1271, where POSIX case is identity. A Windows drive or UNC pattern still folds case. The README says so now.

**3b. A rooted pattern that names no project warns on stderr.** A rooted pattern selects one project or nothing, so a typo does not fail — it reports a total over a set the caller did not mean, which looks like a correct answer:

```
codeburn: no project matches /Users/gone/apps — an absolute path has to be a project's path, or a prefix of it on a segment boundary
```

Each pattern is judged on its own against the unfiltered project list, so an earlier `--project` cannot mask a later one that selects the same project, and an `--exclude` is not tested against what `--project` already removed. The message is period-scoped because a range-scoped parse is what was checked, and the durable path passes the carried days' project identities alongside the live parse, so a path naming only an expired-source project stays silent. Nothing is reported under `serve`, where stderr is routed into the progress-frame emitter. A loose word never warns. Exit status is unchanged.

## One behavior change, resolved in review

A rooted pattern that is not a real path from the root used to substring-match and now matches nothing: `--exclude /Web/clients` selects no project, where before it excluded every project whose path contained that fragment. Anchoring cannot keep both meanings. Per review, the strict rule stays and the silence is what got fixed — see 3b above.

## Tests

`tests/parser-filter.test.ts` covers the sibling case in both directions, Windows casefolding against POSIX identity, the stripped-absolute path, trailing slashes and backslashes, and the slash-only pattern.

`tests/project-filter-durable-totals.test.ts` adds the two carried-day shapes where a second copy of the rule diverges from the anchored one inside a single headline.

`tests/cli-project-filter-commands.test.ts` spawns the real CLI over two path siblings: `sessions` for the commands that filter at the call site, `spend` and `yield` for the two that take the patterns as arguments, plus the warning and the cases that must stay silent. Dropping the argument from any of those three fails the suite.

Full suite green: 4425 passing, `tsc` clean.

## What the second PR needs from you

`status --scope combined` rejects `--project`/`--exclude`, since paired devices report their usage unfiltered. In the pane I made the filter win: Combined is unavailable while a project filter is set, because dropping the filter instead would put the hidden projects back inside the combined total, which is the one thing the pane exists to prevent. If you would rather the combined view stayed available and simply ignored the filter, that is a one-line change before I open it.


ozymandiashh (changes requested):
The plumbing half is right: all seven commands declare both flags, filtered at the one point each funnels through `parseAllSessions`, and the set matches the body. Suites, `tsc` and `npm run build` are green on the branch and on a merge with current `main`. Not mergeable yet, for one blocking reason and two disclosure gaps.

**Blocking: the anchored rule was applied to `filterProjectsByName` only.** `src/usage-aggregator.ts:379-386` carries `dayProjectMatches`, documented as "Mirrors parser.filterProjectsByName exactly ... so a filter selects the same projects whether it is resolved against a fresh parse or against the day cache". It still lowercases and substring-matches. `buildDurableOverviewFromNormalizedIndex` (lines 535-546) and `buildDurablePeriod` (645, 707-725) call both matchers on the same `opts.project`, so today's live days anchor while every cached day substring-matches, inside one headline total. Reproduced on the `tests/project-filter-durable-totals.test.ts` harness with a carried day split across `/Users/gone/app` ($30) and `/Users/gone/app-kit` ($70):

- `--exclude /Users/gone/app` on the carried day: expected 70, got 0 (sibling swallowed).
- `--exclude /users/gone/app`: expected 100, got 0 (case variant still excludes).

That regresses `report`, `overview`, `status`, `today`, `month`, `export`, dash and menubar, and breaks the headline-vs-By-Project reconciliation that suite exists to pin. Fix: route the cached path through `matchesProjectPattern` (it already takes `{ project, projectPath? }`), drop the pre-lowercasing at 535-536 and 707-708, update the comment at 379-381, and add the sibling + case-variant cases to that suite.

**Disclose:** rooted POSIX patterns are now case-sensitive (`/users/me/vault` no longer matches `/Users/me/vault`; your own test pins it). That may be the right rule given #1260, but it's a second behavior change and neither the body nor the README mentions it.

**Decision on the deferred question:** keep the strict rule (a rooted pattern that is not a real prefix matches nothing), but a rooted pattern that matches zero projects should warn on stderr instead of failing silently. Wrong totals that look right are the worst outcome here.

Also: the new tests are unit tests of `filterProjectsByName` only, and the yield suites stub the filter to identity, so dropping the argument on any command would still pass. One integration case per command group please.

mathieustan:
Thanks, the cached-day copy was a real miss. All four points are addressed, and the branch is three commits now, one idea each.

**Blocking, fixed.** The rule is one closure (`makeProjectFilter`) applied to the live parse and to the carry-forward cache alike, so `dayProjectMatches` and its pre-lowercasing are gone rather than re-implemented. Your two shapes are in `tests/project-filter-durable-totals.test.ts`, on a carried day split across `/Users/gone/app` ($30) and `/Users/gone/app-kit` ($70):

- `--exclude /Users/gone/app` → 70, the sibling survives
- `--exclude /users/gone/app` → 100, the wrong-case path selects nothing

Both fail against the old rule and pass against this one.

**Disclosure, added.** The README now states that an absolute POSIX path is case-sensitive, that a Windows drive or UNC path folds case, and that a leading `~` is the shell's to expand. Same note in the PR body.

**Your decision, implemented.** The strict rule stays, and a rooted pattern that matches nothing warns on stderr, exit status untouched:

```
codeburn: no project in this period matches /Users/gone/apps (an absolute path has to be a project's path, or a prefix of it on a segment boundary)
```

Each pattern is judged on its own against the unfiltered list, so an earlier `--project` cannot mask a later one that selects the same project, and an `--exclude` is not tested against what `--project` already removed. The durable path passes the carried days' project identities alongside the live parse, so a path naming only an expired-source project stays silent. Nothing is reported under `serve`, where `runCaptured` routes `process.stderr.write` into the progress-frame emitter and a warning would arrive as scan progress.

**Integration coverage, added.** `tests/cli-project-filter-commands.test.ts` spawns the real CLI over two path siblings: `sessions` for the commands that filter at the call site, `spend` and `yield` for the two that take the patterns as arguments, so dropping the argument anywhere fails the suite. It also pins the shapes that must stay quiet: a pattern that selects something, a loose word, a second `--project` overlapping the first, and an `--exclude` naming a project `--project` had already removed.

Full suite green: 4426 passing, `tsc` and `npm run build` clean, and `git merge-tree` against current `main` is conflict-free. The branch was rewritten while addressing this, so the commits you reviewed have new hashes.

mathieustan:
Reviewed my own warning properly before asking you to look again, and it was wrong in four ways, each of them warning about a path that is in fact correct:

- `include.some()` short-circuits, so a second `--project` selecting a project an earlier pattern already claimed never recorded a hit.
- The exclude pass ran against the include-filtered list, so an `--exclude` whose project `--project` had already removed was never tested.
- A range-scoped parse judged a valid path against that one window, with no way to say so.
- The hit sets were process-global and never cleared, so a `codeburn serve` child carried one request's patterns into the next, and the exit hook was not installed on that branch at all. Had it been, `runCaptured` would have delivered the warning to the desktop app as a scan-progress line, since it routes `process.stderr.write` into the progress-frame emitter.

It is a pure function now. `unmatchedRootedPatterns()` tries each pattern on its own against the unfiltered list, and the report runs once per command over the projects that command actually parsed. The durable path passes the live parse plus the carried days' identities, so a pattern naming only an expired-source project stays quiet. The message reads "no project in this period matches", which is what was checked, and nothing is reported under `serve`.

Two more from the same pass: the rule is compiled once per filter instead of being re-normalized inside the per-day, per-provider loop it now runs in, and the include-then-exclude composition lives in one `projectPassesFilters` that both `filterProjectsByName` and the day-cache slice call.

The new tests cover the shapes that must stay silent as well as the ones that must warn: a pattern that selects something, a loose word, a second `--project` overlapping the first, and an `--exclude` naming a project `--project` had already removed.

I squashed the broken-then-fixed pair so the branch reads as one change per idea, so the commits you reviewed have new hashes. Nothing else moved: the tree is identical to the pre-squash one, 4426 tests pass, `tsc` and `npm run build` are green, and `git merge-tree` against current `main` is conflict-free.

ozymandiashh (changes requested):
Re-reviewed 648fa40e merged onto current `main`: `tsc` and `npm run build` clean, full suite 3878 passed / 0 failed, `test:locks` 36/36. All four points are delivered and I mutation-checked each: reintroducing the substring rule inside `sumMatchingProjects` fails exactly the two new durable tests; dropping the patterns on `sessions` or the arguments on `computeYield` fails the new CLI suite. Compiling the filter once in `makeProjectFilter` is a better shape than what I asked for, given how often the day cache runs it.

One new blocker, small and localized, in the warning:

- **It fires into the Ink dashboard.** `reportUnmatchedProjectPatterns` only guards `process.argv[2] === 'serve'`, but `src/dashboard.tsx:390` calls `buildDurablePeriod` with the filter, so `codeburn report --project /typo` (and bare `codeburn`, since `report` is the default command) writes a raw line over the rendered TUI. The repo already has the mechanism for this: `setInteractiveScanUI`, which `parser.ts:3229` uses for the same reason. Route the warning through it.
- **The `serve` guard is fragile.** `main.ts` declares program-level `--verbose` and `--timezone`, so `codeburn --verbose serve` shifts `argv[2]` and the warning becomes a progress frame in `runCaptured`, and `reported` then outlives the request. Nothing passes `--project` to `serve` today, but the Projects pane follow-up will. Key it off the actual command, not the argv position.

Non-blocking: indentation at `src/main.ts:2241-2242`, a type declared above its import in the new module, and the expired-source-only warning path has no test. Fix the two above and this is good to go.

mathieustan:
Both blockers are fixed in 740ec551 and d8dfd728, pushed on top so the delta since your review is readable.

**The Ink screen.** Routed through the mechanism you pointed at: the warning now honours the same `setInteractiveScanUI` latch `parser.ts:3229` checks before printing scan progress. The TUI arms it before `render()` (`dashboard.tsx:2421`, `compare.tsx:516`), so `codeburn report --project /typo` writes nothing over the dashboard, while `report --format json` still warns.

**The serve guard.** It no longer reads argv: `runStdioServe` calls `suppressProjectFilterWarnings()` at startup, so it holds for the life of the process whatever the command line looks like. One correction to your repro, since it matters for where the guard belongs: `codeburn --verbose serve --stdio` does not serve at all. The serve entry is dispatched before Commander parses, and the registered `serve` command is a no-op action, so that invocation exits without a request loop and `runCaptured` never runs. Your point stood anyway, an argv position is not a key.

The three non-blocking items are done: the indentation at the `compare` call site, the type moved below its import, and the expired-source path now has a test. Writing that test turned up a fourth defect, which is the other half of d8dfd728: `report`, `overview` and `status` judged a pattern against the live parse plus the day cache, while the seven commands that filter a bounded parse judged it against the live parse alone. So on a project whose transcripts have expired, `overview --project <path>` billed it while `sessions --project <path>` called the path wrong. The check now falls back to the day cache, and only for the patterns the live parse missed, so the normal path reads no cache.

740ec551 is separate: a quoted `'~/work/app'` reaches the CLI unexpanded, and a pattern typed into a field has no shell behind it, so it matched nothing and reported nothing. The README was steering people into exactly that. A leading `~` is now expanded and anchored like any absolute path. Same commit stops the compiled filter from computing the project path key when no pattern is rooted, which the day cache runs per project, per day, per provider slice.

`SERVE_OPTIONS` also gains `--project`/`--exclude` for the seven commands that just declared them, as the comment above the table requires. Without it a filtered request falls back to a cold spawn instead of the resident, which the Projects pane follow-up would have hit on every fetch.

One hole I left alone, tell me if you want it here or as a follow-up: `compare --format json --project` filters its model rows, but `scanSelfCorrections` still walks every provider's session directories, so the self-corrections count aggregates the whole machine. The hole predates this PR for `--provider`; `--project` inherits it. Scoping it means mapping discovered directories back to filtered projects across every provider, which I did not want to land unverified in this PR.

Full suite 4431 passing, `tsc` and `npm run build` clean.

ozymandiashh (approved):
Re-reviewed d8dfd728 merged onto current `main`: `tsc` and `npm run build` clean, full suite 3883 passed / 0 failed, `test:locks` 36/36. Both blockers are fixed, and the structural fix is better than what I asked for: `buildDurablePeriod` no longer warns at all and the command layer resolves patterns once against `knownProjects`, so the Ink leak is gone by construction. Ran the real TUI with a typo'd pattern: stderr empty, dashboard renders, `--format json` still warns with valid JSON.

You are right about `codeburn --verbose serve --stdio`; the serve dispatch happens before Commander parses, so my repro could not have leaked. The argv key was still the wrong mechanism and `suppressProjectFilterWarnings()` is the right one.

Good catch on the fourth defect (`overview` vs `sessions` disagreeing on an expired-source path). Mutation-checked the fallback both ways: disabling it fails one test, making the cache read throw fails only the four unmatched-pattern tests and nothing on the normal path. Tilde expansion in `compile()` rather than at the CLI boundary is correct for the same live-vs-cache reason.

Two non-blocking notes for later: `unmatchedRootedPatterns` recompiles per entry (typo path only), and `compare --format json` counting self-corrections machine-wide is pre-existing for `--provider` too; worth a tracking issue. Approving; CI is now approved to run on this branch.

Issue #726, opened 2026-07-17
menubar: quota windows show no pace — add run-out projection (deficit, lasts-to-reset / runs-out-in)


avs-io:
## Problem

The Plan tab shows each quota window's used% and reset time, but not the one thing the user actually wants to decide: **am I going to make it to the reset?** "Weekly window 68% · resets in 5 days" forces the user to do the pace math themselves, every time they look.

## Proposal — phase 1, linear pace (no new data required)

Everything needed is already in each window: `usedPercent`, `resetsAt`, `limitWindowSeconds`.

- `elapsed = duration − (resetsAt − now)`; `expected% = elapsed/duration × 100`
- **Deficit/reserve**: `Δ = used% − expected%`, rendered as "N% in deficit" / "N% in reserve" / "on pace" (staged at |Δ| ≤ 2 / ≤ 6 / ≤ 12 so the wording is legible, not twitchy).
- **Run-out ETA**: whole-window linear rate `used%/elapsed`; `eta = (100 − used%)/rate`. If `eta ≥ time-to-reset` → "lasts until reset"; else → "runs out in ~X".
- **Noise guards** (these matter more than the math): show nothing until `expected% ≥ 3` (kills early-window nonsense projections), skip when `elapsed ≤ 0` (reset-time skew) or the window is already exhausted, clamp everything to sane ranges.
- **Rendering**: the plumbing already exists — `UtilizationRow` accepts a `projection` and `ProjectionCaption` renders it; the Codex Plan rows currently pass `projection: nil`. Phase 1 is largely "compute a `WindowProjection` for Codex windows and stop passing nil", keeping the caption explicitly estimate-flavored.

## Phase 2 — historical forecast (separate follow-up, only if wanted)

The menubar already persists per-window usage snapshots (`SubscriptionSnapshotStore`). With ≥3 complete past windows, a nonparametric forecast becomes possible: reconstruct each completed window as a monotone cumulative curve, shift each curve vertically through the current (elapsed, used) point, and read off (a) the recency-weighted median crossing of 100% as the ETA and (b) the smoothed weighted fraction of curves that end ≥ 100% as an explicit run-out **risk percentage**. With thin history it blends toward the phase-1 linear baseline — and is capped so expected usage never exceeds the linear baseline, meaning a habitual heavy user is never flattered with "in reserve". Filing the design detail here so the scope is visible, but phase 1 stands alone and ships first.

## Tests

[...]

PR #1287, opened 2026-09-11
feat(menubar): say whether a quota window lasts until its reset

### LINKED ISSUE #1215: menubar: bring the lasts-until-reset verdict to the Capacity Dock and quota hover card
## Proposal

#726 shipped deficit/reserve + lasts-to-reset/run-out captions on the Plan tab's utilization rows (#728). The two other quota surfaces — the **Capacity Dock rows** and the **agent-tab hover card** (`QuotaDetailRow`) — still show only "Resets in Xd Yh", so the am-I-going-to-make-it answer is missing exactly where most users now look.

Add the same verdict there, one line under the bar:

- `Lasts until reset` — linear projection ≤ 100% at reset
- `Runs out in 2d 3h` — overflow with the linear ETA (weekly/monthly windows)
- `Won't last until reset` — overflow on ≤6h windows, keeping the existing ETA-suppression guard so short windows don't cry wolf
- silent when early in the window, reset unknown, or already exhausted

## Implementation sketch

Working proof exists on my fork — happy to PR it if the direction is right.

- `QuotaPace.inferredWindowSeconds(label:resetsAt:)`: window length from the normalized provider labels ("Weekly", "5-hour", "3-day", "Monthly usage limit"). The adapters derive labels from exact API durations, so the round trip is lossless except calendar months, which compute back from the reset date in UTC (mirroring `monthlyWindowSeconds`).
- `QuotaSummary.Window.paceVerdict`: `(text, willOverflow)` on top of the existing `QuotaPace.evaluate` — no new math, no new data.
- Rendered in `CapacityDockQuotaRow` and `QuotaDetailRow`, orange when overflowing.

Open questions for maintainers:

1. Label inference vs threading exact `windowSeconds` through `QuotaSummary.Window` (~15 call sites). I picked inference — zero call-site churn, exact for every label the adapters emit today — but threading is more explicit if you'd rather.
2. Wording on the ≤6h overflow ("Won't last until reset") vs deficit/reserve-only there, per the #726 review note.
3. Whether the dock's compact rows should show it for every window or just the weekly headline.

Tests: 9 cases in `QuotaPaceVerdictTests.swift` (verdict text, overflow flag, every silence guard, label inference including "Weekly · Opus" and "Monthly usage limit").

ozymandiashh:
## Summary

- The Capacity Dock's window columns and the agent-tab quota hover card now answer "am I going to make it?", not just "when does this reset?": one line under the bar reading `Lasts until reset`, `Runs out in 2d 8h`, or `Won't last until reset` on windows of 6 hours or less, where a linear run-out ETA is not defensible.
- It is the same projection the Plan tab's deficit/reserve caption has used since #726/#728 — `QuotaPace.evaluate` stays the only pace math in the app, and this is a `(text, willOverflow)` reduction of its result — so the three surfaces cannot disagree. All of #726's guards come along: silent in the first 3% of a window, on an exhausted window, without a reset time, and on clock skew, with the <= 6h ETA suppression kept rather than crying wolf over one burst.
- `QuotaSummary.Window` carries no window length, so rather than thread one through ~15 adapter call sites, `QuotaPace.inferredWindowSeconds(label:resetsAt:)` reads it from the normalized label the adapters already derive from the exact API duration (`Weekly`, `5-hour`, `Weekly · Opus`, `Monthly usage limit`). Calendar months are measured back from the reset date in UTC, mirroring how the Codex spend window is derived. A label that implies no length (Copilot's `Premium requests`, Cursor's `Team pool`) keeps the row silent instead of guessing.
- The dock reserves the line's height whether or not a column has a verdict, because the glance panel's frame is computed rather than fitted — the same reason the staleness notice has its own reserved height — so a projection crossing in or out of silence cannot resize the bubble under the pointer. The hover card is fitted, so there the line collapses.

Closes #1215

## Testing

- [x] `npm test` passes — not applicable, no TypeScript touched; `cd mac && swift build` succeeds (Build complete, 53.76s)
- [x] `npm run build` succeeds — not applicable; macOS target only

`mac/Tests/CodeBurnMenubarTests/QuotaPaceVerdictTests.swift` adds 12 XCTest cases over the pure
`(utilization, elapsed fraction, window length) -> String?` mapping: all four outcomes
(lasts / runs out with an ETA / won't last on a 5h and on an exactly-6h window / silent), the
early-window silence at 2% and 1% elapsed with 5% speaking, exhausted, missing reset and clock
skew, the countdown wording, the `percent` 0...1 conversion, and label inference including
`Weekly · Opus`, `GPT-5.3-Codex-Spark · 5-hour`, `Gemini Models · Five-hour`,
`Monthly usage limit` against a February-in-UTC reset, and the bucket labels that must stay
silent. `CapacityDockGlanceTests` is updated for the three hard-coded panel heights the
reserved line moves.

Note: my machine has a CLT-only Swift toolchain with no swift-testing module, so the test
target does not compile locally (`error: no such module 'Testing'` from an existing test file)
and I could not run `swift test`. To avoid shipping unverified assertions I executed every
expectation in the new file against the real `QuotaPace.swift` through a standalone `swiftc`
harness — 43/43 pass — but CI's `swift test` is the real check here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Added (macOS)
+- **The Capacity Dock and the agent-tab quota card now say whether a window will last to its reset, not just when it resets.** Each quota window gets one line under the bar: `Lasts until reset` while the linear projection lands at or under 100% at reset, `Runs out in 2d 8h` when it overflows on a window long enough for a whole-window rate to be defensible, and `Won't last until reset` when it overflows on a window of 6 hours or less, which keeps the run-out ETA suppressed there rather than crying wolf over one burst. The verdict is the same projection the Plan tab's deficit/reserve caption uses, so the two surfaces cannot disagree, and it stays silent in the first 3% of a window, on an exhausted window, with no reset time, or when the window's own label does not imply a length (a Copilot premium-requests bucket, say). The dock reserves the line's height whether or not a column has a verdict, so a projection crossing in or out of silence cannot resize the bubble under the pointer. (#1215)
+
 ## 0.9.24 - 2026-09-04
 
 ### Added
--- a/docs/design/capacity-dock.md
+++ b/docs/design/capacity-dock.md
@@ -101,6 +101,13 @@ V1 does not include:
   countdown. The most constrained available window supplies the ring value;
   this matches the reference's glance-first use and avoids understating a
   provider whose secondary window is closer to exhaustion.
+- Under that, one line says whether the window lasts: `Lasts until reset`,
+  `Runs out in 2d 8h`, or `Won't last until reset` on windows of 6 hours or
+  less, where a linear run-out ETA is not defensible. It is the same projection
+  the Plan tab's pace caption uses, and it is silent early in a window, on an
+  exhausted window, without a reset time, or when the window's label implies no
+  length. Its height is reserved whether or not a column has a verdict, because
+  the panel's frame is computed rather than fitted.
 - Stale or retrying data remains visible and is labeled/dimmed. A terminal
   authentication/configuration failure provides a Connect/Reconnect action in
   the bubble itself. Network, rate-limit, parse, and provider outages remain
--- a/mac/Sources/CodeBurnMenubar/Data/QuotaPace.swift
+++ b/mac/Sources/CodeBurnMenubar/Data/QuotaPace.swift
@@ -61,3 +61,116 @@ enum QuotaPace {
         )
     }
 }
+
+extension QuotaPace {
+    /// The one-line "am I going to make it?" answer for surfaces with no room
+    /// for the Plan tab's deficit/reserve caption: the Capacity Dock's window
+    /// columns and the agent-tab hover card (#1215). Same projection as #726 —
+    /// `evaluate` stays the only math here — reduced to a verdict.
+    struct Verdict: Equatable {
+        let text: String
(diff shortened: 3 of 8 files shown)


iamtoruk:
Leaving this open rather than merging, and it is not because the code is wrong: the math checks out on every real window on this machine and the tests are mutation-proven. Two reasons. First, #1267 writes a pace caption into the same slot under the same window columns, carries a real windowSeconds from the adapter instead of reverse-engineering the length from the display label, has a freshness and connection gate, and treats short windows as on pace / deficit / reserve the way #726 intended. The two cannot both ship, and on those axes #1267 is the stronger base. Second, the label round trip breaks on Grok Build: its label is chosen from the distance to reset, so a monthly cycle passing through the 4 to 12 day band is labeled Weekly and the pace is computed against a 7 day window. The agent-tab hover-card placement and the plainer wording here are worth porting onto #1267's QuotaPacePresentation. Happy to take a rework in that direction.

ozymandiashh:
Closing in favour of #1267, exactly as you suggested. The hover-card placement and the plainer wording are ported there on top of `QuotaPacePresentation`, using the adapter's real `windowSeconds`.

You were right about Grok Build, and it is worse than a wording issue: deriving the window from the display label means a monthly cycle sitting in the 4-to-12-day band is measured against 7 days, so the caption is confidently wrong on a real provider. #1267 carries a regression test for that exact shape, and a nil `windowSeconds` stays silent so the fallback cannot come back unnoticed. Thanks for catching it before it shipped.

PR #1288, opened 2026-09-11
feat(menubar): switch the Capacity Dock gauge between usage windows

### LINKED ISSUE #1243: Feature Request: Allow switching Claude usage periods in the collapsed Capacity Dock
Description

It would be great if the collapsed state of Capacity Dock could provide a way to switch between Claude's 5-hour and weekly usage limits.

Currently, when Capacity Dock is collapsed, it would be useful to see the usage information at a glance, but being able to switch between the 5-hour and weekly limits without expanding the dock would make it much more convenient.

Proposed behavior

When the Capacity Dock is collapsed:
Allow clicking/tapping the Claude usage indicator to switch between 5h and weekly usage.

Persist the selected period so the preference is retained.
Keep the interaction minimal so the collapsed dock remains compact.

This would make the collapsed Capacity Dock much more useful as an always-visible usage monitor without requiring the full dock to be opened.

ozymandiashh:
## Summary

- The resting Capacity Dock gauge always reported the weekly (else monthly)
  billing window. Clicking the provider that already rests in the rail now
  switches that gauge to the provider's short rolling window — Claude's 5-hour
  limit, Codex's 5-hour or daily window, any `Hourly` / `Daily` / session row an
  adapter reports — and clicking again switches back. Closes #1243.
- The choice is stored per provider under `CodeBurnCapacityDockGlanceWindows`
  alongside the other dock preferences, so Claude can sit on its 5-hour window
  while Codex stays weekly, and it survives relaunch. It is deliberately not a
  Claude-only switch: the dock's collapsed row is provider-agnostic.
- A new pure `CapacityDockGlanceWindow` policy decides which window is drawn.
  `billing` reuses `QuotaSummary.headlineWindow` so the app keeps one definition
  of the weekly glance; `burst` matches short-horizon labels only after
  excluding weekly/monthly ones, so a per-model row like `Weekly · Opus` is
  never mistaken for a 5-hour window. A provider reporting one window keeps the
  existing click-to-pin behaviour, and a stored horizon the provider stops
  reporting falls back to the window it does report instead of blanking the
  gauge to `--`.
- The rail's geometry is untouched, the click can only pin and never collapses
  (so hover expansion, the drag suppression window and Escape / outside-click
  dismissal all behave as before), and keyboard and VoiceOver users get the same
  switch as a named action on the provider cell, whose accessibility value now
  names the window feeding the ring.

## Testing

- [ ] `npm test` passes

`mac/` change only; no TypeScript touched.

- `cd mac && swift build` — passes.
- `cd mac && swift test --filter CapacityDock` — could not run on this machine:
  every test file, including ones this branch does not touch, fails with
  `error: no such module 'Testing'` on a CLT-only toolchain. Relying on CI for
  the suite.
- New coverage in `mac/Tests/CodeBurnMenubarTests/CapacityDockGlanceWindowTests.swift`:
  the per-provider preference round-trip (including defaults and rejection of
  foreign stored values) and the toggle policy — which window is shown for a
  given stored preference and set of published windows, that `Weekly · Opus` is
  not a short window, and the fallbacks when only one window exists or the
  stored horizon disappears.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Added (macOS)
+- **The Capacity Dock gauge can report the short usage window instead of the weekly one, without expanding the dock.** The resting rail shows one number per provider, and that number was always the weekly (else monthly) billing window. Clicking the provider already resting in the rail now switches its gauge to the provider's short rolling window — Claude's 5-hour limit, Codex's 5-hour or daily window, any `Hourly`, `Daily` or session row an adapter reports — and clicking again switches back. The choice is stored per provider under `CodeBurnCapacityDockGlanceWindows`, so Claude can sit on its 5-hour window while Codex stays weekly, and it survives relaunch. A per-model row such as `Weekly · Opus` is never read as a short window, a provider that reports only one window keeps the plain click-to-pin behaviour, and a stored horizon the provider stops reporting falls back to the window it does report rather than blanking the gauge to `--`. The rail's geometry is untouched, VoiceOver and keyboard users get the switch as a named action on the provider cell with the window named in the cell's value, and Escape or a click outside still unpins the dock. (#1243)
+
 ## 0.9.24 - 2026-09-04
 
 ### Added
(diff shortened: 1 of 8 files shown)


iamtoruk:
Landed as #1309 (b0e53cfc) with your commit and authorship intact, rebased onto current main after #1308; changelog conflict only. Verified before merging: single call-site blast radius on the resting gauge, every adapter's real label shape checked, click writes a preference only with no fetch, persistence normalized, mutation-verified tests, 473/473. Thank you.

PR #1289, opened 2026-09-11
feat(menubar): i18n support + Simplified Chinese localization for the macOS menubar

### LINKED ISSUE #1219: [Feature Request] Add i18n support / Chinese localization
### Is your feature request related to a problem? Please describe.

CodeBurn is a great local tool for tracking AI coding token usage, but the UI (CLI output and the local dashboard) is hardcoded in English only. It is quite popular among Chinese developers, and a localized version would make it much more accessible to them.

### Describe the solution you would like

Add internationalization support:

- Extract user-facing strings from `src/` and `dash/src/` into resource files
- Support a `--lang` / config option (e.g. `LANG` env or `codeburn config set lang zh-CN`)
- Ship at least `en` and `zh-CN` (Simplified Chinese) locales
- Apply to both the CLI/Ink UI and the web dashboard

### Describe alternatives you have considered

- A third-party translation patch fork (fragile, hard to maintain)
- Keeping it English-only (fine, but limits adoption)

### Additional context

- The codebase currently has no i18n library (no i18next / react-i18next in dependencies)
- Happy to help with implementation / PR review if the maintainers are open to it

I am happy to submit a PR if the direction is agreed. Thanks for the great tool!

ozymandiashh:
## Summary

Adds internationalization to the macOS menubar app and ships a complete Simplified Chinese
translation. This is the menubar half of #1219, the split @TheCrazyAnt proposed in the issue
thread; the CLI/Ink output and the web dashboard are a different stack and stay out of scope.

No third-party library. One `Localizable.strings` table per locale, shipped as SwiftPM target
resources for `en` and `zh-Hans`, with the language chosen by AppKit from the user's system
preferences — there is no picker inside CodeBurn, because macOS already owns that choice.
`CFBundleLocalizations` in both packaging scripts lists the app under System Settings >
General > Language & Region, so the language can be overridden for CodeBurn alone.

533 keys cover the popover, the Capacity Dock, the status-item menu, the update alerts and
all of Settings. **The key is the English copy**, so English stays the development language:
an untranslated string renders as correct English rather than a dotted identifier, and
`en.lproj` is an identity table kept so the bundle advertises `en` and so the two tables can
be diffed.

One design point worth a close look: every lookup names `Bundle.module` explicitly via
`L(_:)` rather than relying on SwiftUI's implicit `LocalizedStringKey`. SwiftPM emits target
resources into a sibling bundle that the packaging scripts copy into `Contents/Resources`, so
`Bundle.main` has no `.lproj` at all and `Text("literal")` would always miss. Naming the
module bundle is the one form that resolves identically under `swift run`, under `swift test`
and in the packaged `.app`.

Enum raw values that double as persistence or cache keys (`Period`, `MenubarScope`,
`InsightMode`, `AccentPreset`, `ProviderFilter`) keep their raw value and gained a separate
`displayLabel`, so nothing a user has saved changes meaning. Three display-only date
formatters that were pinned to `en_US_POSIX` with fixed patterns now follow the locale, and
the calendar's weekday row comes from the locale's own symbols, so a Chinese UI reads
`2026年9月` and `一 二 三` instead of `September 2026` and `Mo Tu We`. Numbers and currency
are untouched — they already went through the existing formatters, and `L(_:_:)` only
substitutes values those produced.

Left verbatim in every locale: provider, model and plan names; units; currency codes; shell
commands and paths; anything the `codeburn` CLI produces; and the `QuotaSummary.Window`
labels, because `headlineWindow` selects the Capacity Dock's glance window by
case-insensitive English substring match on `week`/`month` and translating them would
silently break that selection.

`build-local.sh` also now copies the SwiftPM resource bundle into the app, using the
named-bundle + `[[ -d ]]` pattern `package-app.sh` already has. The feature needs it to find
its strings at runtime, and that script's output already trapped at launch without it — the
resource-bundle half of #1262.

Three commits: mechanical extraction, then the translation, then tests and docs.

Closes #1219

## Testing

- `cd mac && swift build` — clean build from an empty `.build`, exit 0, no new warnings (the
  only warnings are pre-existing `SecKeychainGetStatus` deprecations in a file this PR does
  not touch).
- New `LocalizationCatalogTests` (10 tests): key sets equal in both directions, no blank
  values, `en` is an identity table, argument specifiers identical in count and order, `%%`
  counts equal, no key is only specifiers, shipped localizations agree with
  `L10n.supportedLocalizations`, plus four cases resolving representative presentation
  strings in both locales and checking the English one against what the presentation struct
  returns.
- Verified runtime resolution from the real resource bundle in both locales (`3 sessions` /
  `3 个会话`, `Reconnect Claude` / `重新连接 Claude`), and from a hand-assembled `.app`
  laid out the way `package-app.sh` produces it (main bundle reports `["en", "zh-Hans"]`,
  both tables reachable through `Contents/Resources`).
- Checked format-string arity at all 711 `L(...)` call sites — `L(_:_:)` is variadic
  `CVarArg`, so the compiler cannot catch a mismatch; 0 found.

`swift test` could not run on my machine: it is Command Line Tools only with no Xcode, so
swift-testing is unavailable — pristine `origin/main` fails there identically with 124 ×
`no such module 'Testing'`. Same reason `package-app.sh` could not run locally (its two-arch
build needs xcbuild). I ran the new tests' assertions through a `Testing`-free harness
derived from the test file instead (1621 assertions, 0 failures), but the macro expansions
themselves are only exercised by CI. No visual QA in a Chinese UI; text fit in the
fixed-width Capacity Dock and period-strip controls is unverified, though Chinese is
typically shorter than English.


ozymandiashh:
Rebased onto current main and extended, because #1308 and #1309 landed new user-facing strings that bypassed the catalog entirely (zero `L(` calls in either diff, e.g. `Toggle("Second row")`). Those are now routed and translated: **18 new keys, 533 to 551**, covering the second-row Settings section and picker, the dock window-switch accessibility action and value, and their help text.

Two things worth a reviewer's attention. The merge had a trap: main moved the status-button tooltip out of `composeStatusTitle` into `refreshStatusButton`, and taking main's side of that conflict compiles and passes every existing test while silently dropping the `L(...)` routing. Re-applied at the new location, and I verified mechanically that no other lookup was lost and that outside `mac/` the tree is byte-identical to main.

Second, the parity test only proved the two catalogs agreed with each other, which passes even when a literal never reaches the catalog at all, which is exactly how this drift happened. There is now a guard with three passes (call sites, display-label properties, catalog round-trip) that fails CI when a user-facing literal is not routed. It immediately found pre-existing drift of its own in `CodexUsage`'s credit-limit copy, which I localized rather than denylisted.

One gap left for a follow-up: `LocalizedError.errorDescription` is unlocalized across all ten `*SubscriptionService.swift` files. Main's two new Copilot error strings follow that same existing rule, so doing Copilot's alone would be worse drift.

swift build complete; catalog harness 53/53, en 551 / zh-Hans 551.

iamtoruk:
Landed via #1330 (merge 5498818) with your five commits intact and credited. What changed on top of your branch: rebased onto main and added the 14 keys new strings needed, a Language picker in Settings (System / English / 中文) so a Chinese system can still run the app in English, weekday row now uses two-character short symbols so Chinese weekdays fit, the second menubar row budgets width per glyph instead of per character, and QuotaPacePresentation, EarlyQuotaReset and CodexBankedResets are localized with whole-sentence keys. English output is byte-identical to main on the 70 rendered strings I diffed, except four date formats that now come from the locale. Argument-order drift in the all-%@ keys is tracked in #1331. Thanks for the groundwork, the catalog and coverage tests carried the whole review.

Issue #1252, opened 2026-09-04
Feature request: configurable multi-row macOS menu bar layout


ILOT-code:
## Motivation

I would like to see subscription quota and local usage together in the macOS menu bar without opening the popover. For example, weekly quota remaining and time until reset on the first row, with today's all-agent cost and total tokens on the second.

## Proposal

Add an optional native layout editor with:

- Multiple items per row, optional aligned columns, a live preview, and a few presets.
- Configurable content: icons, cost, token totals/breakdowns, calls, sessions, cache hit, and supported subscription quota/reset metrics.
- Independent Provider, Period, and Local/Combined scope for usage items, reusing the existing data interfaces and restrictions.
- Optional labels, decimal precision for cost/token items, row/item reordering, text size, and column spacing.

The local prototype supports up to three rows and six items per row. Two rows are the practical default because of the menu bar's height. Subscription choices use already-connected providers with the requested quota window; Combined usage retains the existing All-providers restriction.

## Contribution

I have a working local Swift/SwiftUI prototype and would be happy to prepare a focused PR if this direction is welcome. I would keep the existing display available and leave the popover unchanged. Is there a smaller initial scope or an alternative settings approach you would prefer?

The prototype has passed debug/release builds and 88 targeted tests. I have not yet completed the required full branch-versus-main test comparison; I will complete that before opening a PR. This issue is to discuss the approach first, per the contributing guide.

## Screenshots

The first image shows the layout editor; the second shows the resulting two-row menu bar item.

<img width="1548" height="1114" alt="Image" src="https://github.com/user-attachments/assets/6808ca7c-3a19-40dd-b3c7-d44c4204ec79" />
<img width="204" height="60" alt="Image" src="https://github.com/user-attachments/assets/96dabe38-70d0-4c92-8b0c-d66e14c66fac" />

PR #1290, opened 2026-09-11
feat(menubar): optional second row in the macOS menu bar item

ozymandiashh:
## Summary

- Adds an optional second line to the macOS menu bar status item, off by default, configured from
  the existing Settings → General → Display section: a "Second row" toggle plus a picker for what it
  shows — quota remaining with its reset countdown for whichever connected provider is nearest its
  limit, today's all-provider cost, today's total tokens, or running sessions. Every metric is one
  the app already computes; nothing here adds a fetch or a refresh cadence.
- The two lines render as one attributed title at 9pt with both line heights clamped to 10pt and the
  inline flame reduced to an 8pt symbol, which measures 20pt inside the standard 22pt menu bar. With
  the setting off, or whenever the chosen metric has no data yet, the title falls back to the
  existing single-row composition unchanged. Settings persist as `CodeBurnMenubarSecondRowEnabled`
  and `CodeBurnMenubarSecondRowMetric` alongside the other menubar defaults keys.
- This is a deliberately reduced slice of #1252. Not included: the layout editor, presets, live
  preview, three rows, multiple items or aligned columns per row, per-item provider/period/scope,
  optional labels, decimal precision, reordering and text size. The row formatting is a pure type
  (`MenubarRowFormatter` over `MenubarRowSettings` + `MenubarRowSnapshot`) so that surface can be
  layered on later without touching the rendering path.

## Testing

- [ ] `npm test` passes

`mac/` change, so no TypeScript was touched and the CLI suites are not affected.

- `cd mac && swift build` → Build complete.
- New suite `mac/Tests/CodeBurnMenubarTests/MenubarSecondRowTests.swift`: 19 tests covering each
  metric, the off state, the no-data fallback to one line, the reset-countdown shape against an
  injected clock, the nearest-limit provider selection and its tie-break, UserDefaults round-trip
  including an unknown stored metric, and the two-row geometry — including one test that composes
  the real attributed title (flame attachment plus the paragraph clamp) for 16 row combinations and
  asserts each measures 20pt, inside the 22pt menu bar.
- `swift test` could not run on the machine this was written on: it has the Command-Line-Tools
  toolchain only, which ships no swift-testing module, so every existing test file in the target
  fails with `no such module 'Testing'`. To avoid shipping unexecuted assertions, all of the new
  suite's expectations were re-run as a standalone `swiftc` binary (52/52 pass), and the height
  assertions were re-run the same way against AppKit (all 16 combinations measure 20.0pt). CI runs
  the real suite.
- Not verified: the rendered look on a real menu bar (no GUI session available), menu bars whose
  thickness is not 22pt, and behaviour when macOS runs out of menu bar space.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Added (macOS)
+- **The macOS menubar item can show a second line.** Settings → General → Display gains a "Second row" switch, off by default, and a picker for what that line shows: quota remaining with its reset countdown for whichever connected provider is nearest its limit, today's all-provider cost, today's total tokens, or the number of running sessions. Both lines render as one attributed title at 9pt with their line height clamped to 10pt, so the pair fits the standard 22pt menu bar, and the second line hides itself whenever its metric has no data yet, leaving the existing single-row figure exactly as it was. This is a deliberately small first slice of the multi-row layout request: no layout editor, no presets, no live preview, no per-item provider or period scoping. The setting persists as `CodeBurnMenubarSecondRowEnabled` and `CodeBurnMenubarSecondRowMetric` in the app's own defaults domain alongside the existing menubar period, scope and metric keys. (#1252)
+
 ## 0.9.24 - 2026-09-04
 
 ### Added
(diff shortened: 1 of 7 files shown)


iamtoruk:
Landed as #1308 (aad1ca3b) with both your commits and authorship intact, rebased onto current main; the only conflict was the changelog heading. Verified before merging: default-off rendering byte-identical to main, two unclipped rows inside the real 22pt bar at 2x, no new fetch or timer, placement recovery untouched, disconnected providers excluded, 478/478 tests. #1310 tracks the headline-window versus worst-window wording and the width cap. Thank you.

PR #1291, opened 2026-09-11
fix(menubar): follow the credential's GitHub host for Copilot quota

### LINKED ISSUE #1286: [Feature Request] Support Copilot Enterprise on custom `*.ghe.com` hosts
## Motivation

CodeBurn currently assumes Copilot quota requests use `github.com` / `api.github.com`.

For GitHub Enterprise Cloud with data residency, Copilot can instead be tied to a dedicated hostname such as:

```text
{hostname}.ghe.com
```

GitHub uses a tenant-specific API host for these environments:

```text
https://api.{hostname}.ghe.com
```

In my case, CodeBurn detects local Copilot usage correctly, but live quota lookup fails with:

```text
available: false
error: Temporarily unavailable.
```

The current implementation hardcodes:

```text
https://api.github.com/copilot_internal/user
```

and does not retain the hostname associated with credentials read from `hosts.json`.

## Proposal

Make Copilot quota lookup hostname-aware:

* retain the hostname together with the discovered credential
* keep the existing `api.github.com` behavior for `github.com`
* for `{hostname}.ghe.com`, use the corresponding tenant API host:
  `https://api.{hostname}.ghe.com/copilot_internal/user`
* optionally allow an explicit Copilot hostname override when automatic discovery is ambiguous
* never send a GHE.com credential to `api.github.com`

## Data residency pricing

GitHub also documents a **10% increase in AI Credit consumption** when the enterprise enables the policy restricting Copilot to data-residency compliant models.

It would be useful to support a configurable option such as:

```text
copilot credit multiplier = 1.10
```

This multiplier should apply **only to locally estimated credits**.

If GitHub provides an exact value such as `total_nano_aiu`, that value should remain authoritative and must not receive the multiplier again.

## Current example

```text
Copilot Enterprise
Host: {hostname}.ghe.com
Budget: 37,600 AI Credits
Reset day: 1

CodeBurn:
estimatedCredits: ~13,037
percentUsed: 34.7%

GitHub:
~96% used
```

The large discrepancy is not necessarily caused only by the 10% multiplier, but CodeBurn currently cannot retrieve the authoritative live quota from the GHE.com account.

## Expected behavior

For Copilot Enterprise on `{hostname}.ghe.com`:

1. Detect or configure the GitHub Enterprise hostname.
2. Associate credentials with their hostname.
3. Query the tenant-specific Copilot quota endpoint.
4. Display the live quota when available.
5. Optionally apply a configurable 1.10 data-residency multiplier to estimated credits only.
6. Preserve existing `github.com` behavior.

Related to #1199, but this request is specifically about **GHE.com hostname support and data-residency pricing**.

ozymandiashh:
## Summary

- Copilot live quota now follows the GitHub host its credential came from: `api.github.com` for `github.com`
  (and for any rung that carries no host — an app-name `apps.json` key, `COPILOT_GITHUB_TOKEN` / `GH_TOKEN` /
  `GITHUB_TOKEN`, `gh auth token`, a pasted token), and `https://api.<tenant>.ghe.com/copilot_internal/user`
  for a GitHub Enterprise Cloud data-residency host. Both readers hardcoded `api.github.com` and dropped the
  host, so a `*.ghe.com` enterprise could only ever report "Temporarily unavailable". Fixes #1286.
- The token and the host always come from the same credential entry (`hosts.json` is keyed by host, newer
  `apps.json` files by `<host>:<app id>`). With several hosts signed in `github.com` wins, otherwise the
  first `.ghe.com` tenant in sorted order; with exactly one, that host is used. A host neither rule can
  address — a self-hosted GitHub Enterprise Server install — fails with a message naming that host rather
  than sending the credential to dotcom, and unreachable-host and HTTP failures name the host that was tried
  (the CLI prints it instead of the bare "Temporarily unavailable").
- macOS Settings says which host answered in the Copilot connection row; `docs/providers/copilot.md` gains a
  GHE-hosts section. Tokens are still never logged and `sanitizeForUI` is unchanged. The issue's second ask,
  a configurable 1.10 data-residency credit multiplier for estimated credits, is not part of this PR.

## Testing

- [x] I have tested this locally against real data (not just unit tests) — dotcom path only; no `*.ghe.com`
      enterprise account was available, so the tenant path is covered by unit tests and by compiling the new
      host-policy type standalone against the expected values.
- [x] `npm test` passes
- `cd mac && swift build` succeeds. `swift test` cannot run on this machine (Command Line Tools only, no
  swift-testing module — the test target fails to compile in files this PR does not touch), so the new
  XCTest cases rely on CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+- **Copilot live quota works for GitHub Enterprise Cloud enterprises on a `*.ghe.com` host.** Both readers hardcoded `https://api.github.com/copilot_internal/user` and threw away the host their credential came from, so a data-residency enterprise signed in on `<tenant>.ghe.com` could only ever report `available: false` with "Temporarily unavailable". A discovered credential now carries its host — `hosts.json` is keyed by host and newer `apps.json` files key by `<host>:<app id>` — and the request follows it: `api.github.com` for `github.com` and for any rung that carries no host of its own (an app-name `apps.json` key, `COPILOT_GITHUB_TOKEN` / `GH_TOKEN` / `GITHUB_TOKEN`, `gh auth token`, a pasted token), and `https://api.<tenant>.ghe.com/copilot_internal/user` for an enterprise host. The token and the host always come from the same entry, with `github.com` preferred when several hosts are signed in and otherwise the first `.ghe.com` tenant in sorted order; a host neither rule can address, such as a self-hosted GitHub Enterprise Server install, fails with a message naming that host instead of sending the credential to dotcom, and unreachable-host and HTTP failures name the host that was tried. The macOS Settings connection row now says which host answered. (#1286)
+
 ## 0.9.24 - 2026-09-04
 
 ### Added
--- a/docs/providers/copilot.md
+++ b/docs/providers/copilot.md
@@ -315,8 +315,9 @@ surfaces, add a reader with a captured fixture.)
 ## Live quota (Copilot subscription)
 
 Separate from the log parser above: the macOS menubar reads live quota from
-`GET https://api.github.com/copilot_internal/user` with a GitHub token. Usage
-tracking never needs this token; only the quota bars do.
+`GET https://api.github.com/copilot_internal/user` with a GitHub token — or from
+the tenant's own host on GitHub Enterprise Cloud, see below. Usage tracking
+never needs this token; only the quota bars do.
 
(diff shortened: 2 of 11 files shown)


iamtoruk:
Landed as #1305 (ef863486) with your commits and authorship intact, rebased onto current main, plus one hardening commit: the normalized host is character-validated before URL construction, since a hosts.json key such as evil.com?.ghe.com passes the suffix check but builds a URL whose host is api.evil.com. Verified on this machine for github.com (same URL as main, HTTP 200, expected keys); the ghe.com path is covered by tests only, no enterprise account here. #1306 tracks the rungs that still carry no host. Thank you.

PR #1292, opened 2026-09-12
Add cohort model comparison (Compare → Cohorts) over an explicit observation population

ozymandiashh:
Add Compare → Cohorts alongside the classic comparison. Users choose two models, a period, a project and an activity category, then inspect the observations behind cost percentiles, retry rates and token-volume statistics. A volume band recomputes the statistics over the selected observations.

An observation is an edit turn owned by exactly one behavioral model. Mixed-model turns are excluded and disclosed. The owning model retains all its recorded cost and tokens, including supplementary accounting, without counting supplementary rows as requests or giving them behavioral ownership. Unknown prices stay distinct from free models.

Project facets expose canonical ids. The desktop passes these as repeatable `--project-id=<id>` arguments to `compare --format cohort-json`, preventing prefix-similar projects from entering the selection. Existing loose `--project` matching remains available. A selected project with no activity for one of the chosen models produces an empty cohort for that model.

Validation: full root suite compared with upstream/main `a1805393` (3,962 passed; the same four Node 26 DEP0205 stderr failures as main); full app suite (902 passed, four skipped); root/app typechecks and CLI/Electron/renderer builds. All 33 core/CLI cohort tests passed again after the final empty-cohort adjustment. Bridge and component tests verify exact project ids, including leading-hyphen fallback ids. Concrete fixtures retain $1 + $9 supplementary spend as $10 and select alpha without alpha-backend; the real CLI also returns a zero-observation second cohort for a one-model project. The main project-filter suite passes all nine tests on Node 24.

Sample rows expose their full data here; wiring their navigation adapter to the separate drill-through feature remains part of cross-feature integration. Details: `docs/compare-cohorts.md`.


iamtoruk:
Landed as #1324 (e0247585) with your commits and authorship intact, rebased onto current main, plus fixes found on real data: the stub sampleNavigation is deleted and sample rows open the real session through the drill-through navigation; observations now carry provider and the session's own project so the row key actually resolves; both new fetches carry the Projects-pane filter; serve --stdio allowlists the new compare options (they were refused before); invalid dates get the friendly error; an unrelated CODEBURN_USER_DATA_DIR hunk was dropped. Verified on frozen data: 172 and 301 observations for the two models in use here, histograms sum exactly, retry rate is the existing formula, percentiles pinned, band recomputation exact, compare --format json and the TUI byte-identical across every period. One divergence disclosed in the docs: a turn touched by two models is excluded here and given to the first model by the classic view, one turn in 526 on this corpus. Thank you.

PR #1294, opened 2026-09-12
feat(app): add a Projects pane that scopes the desktop app to chosen projects

mathieustan:
The desktop half of #1276, on top of the CLI half that merged as #1285.

## What is in here

Settings gets a Projects pane: one switch per detected project, plus a
field for hiding a pattern that matches nothing detected yet. The filter
lives in `~/.config/codeburn/app-filter.json` rather than renderer
storage, because the main process is where the CLI argv is assembled. No
file, or empty lists, means no filter, so nothing changes for anyone who
never opens the pane. `CODEBURN_APP_FILTER` overrides the path and an
empty value disables it outright, which is what the app test suite sets
so a real filter file cannot reach an assertion.

Two rules the pane is built around:

- **Its own list is fetched WITHOUT the filter**, and over the whole
  history rather than the period on screen. Scoping it either way would
  hide the projects the checklist exists to switch back on.
- **Patterns go out as `--project=<value>`**, never as two argv entries.
  A Claude project directory is encoded as `-Users-me-Web-thing`, and a
  leading dash in its own argv slot parses as a flag.

A row is one project as the parser sees it, which is one session's
starting `cwd` resolved by `resolveCanonicalProjectPath`. A linked
worktree folds back into its main repo; a subdirectory of a repo does
not, by that function's own rule, so a monorepo can show `packages/api`
as its own row beside the repo root. Hiding the repo path hides both:
a rooted pattern is anchored on a segment boundary, so it covers the
project and everything inside it.

## Combined scope, as flagged in #1285

`status --scope combined` rejects `--project`/`--exclude`, because paired
devices report their usage unfiltered. The filter wins here: Combined is
unavailable while a project filter is set, and says why. Dropping the
filter instead would put the hidden projects back inside the combined
total, which is the one thing the pane exists to prevent. The offer from
#1285 still stands: if you would rather Combined stayed available and
simply ignored the filter, that is a one-line change.

`buildOverviewArgs` keeps the same rule as a guard, so a stale renderer
cannot produce the rejected argv either.

## Five holes a local adversarial pass found, each closed with its test

The pane commit is the one I wrote; these are what a review pass against
it turned up before this PR opened. Each is its own commit.

1. **A failed filter read unhid every project.** `appFilterCache` is
   module level and null on the first read of a launch, and `statSync`
   rejects `EACCES` the way it rejects `ENOENT`. An unreadable
   app-filter.json read back as an empty filter, so every fetch ran
   unfiltered and the renderer re-offered Combined, with no error
   anywhere. Only a missing file means "no filter" now; everything else
   fails closed into the panel's error state.

2. **The filter write skipped the atomic rename** that `saveConfig` uses
   in the same directory, and for the same reason. Staged and renamed
   now, with the staged file cleaned up if the rename cannot land.

3. **The pane's list was bounded to the period on screen.** A filter has
   no period, so an exclude for a project with no activity this period
   read as `matches nothing detected` next to a Remove button, under a
   footer saying `Nothing is hidden.` Lifetime now, and the period
   argument is gone from the channel.

4. **An all-hidden export reported success.** `export` prints prose and
   exits 0 when every period is empty; the toast named a folder the CLI
   never created. Success reads the saved-path line instead.

5. **The list bypassed the resident serve child.** `report` was in
   neither routing table, so the heaviest read in the app was also the
   only panel query that always paid a cold start. Routed, in its JSON
   form only.

Finding 5 is why this PR touches `src/serve.ts` and `tests/` rather than
only `app/`: the routing table lives on both sides of the wire.

## Tests

15 added. I mutation-tested every one of them: each of the nine defects
above was put back one at a time, and the test that covers it fails
without its fix. Two of my first attempts at the atomic-write test passed
against the unfixed code (the old path never created a temp file, so
"no temp file is left behind" was true by accident); they are replaced by
one that asserts the live path is never a write target.

- `app/electron/main.test.ts`: the two read-failure shapes, the fetch and
  the overview failing closed rather than spawning unfiltered, the
  channel reporting the failure instead of an empty filter, the staged
  write and its cleanup, both export outcomes, and the lifetime horizon.
- `app/renderer/sections/Settings.test.tsx`: the list is asked for once,
  with no period, and survives a period change without refetching.
- `app/electron/cli.test.ts` and `tests/serve-stdio.test.ts`: report
  rides the resident child, and its TUI form and `--refresh` are refused.
- `tests/cli-export-project-filter.test.ts`: the real CLI, exercised in
  both directions, pinning the stdout contract the app now depends on.

CLI suite 3918 passing plus the serial lock suite, desktop suite 919
passing, `tsc` clean on both projects. Every intermediate commit is green
on its own, so the stack stays bisectable.


iamtoruk:
Landed as #1307 (ce59a3af) with all your commits and authorship intact, rebased onto current main, plus four follow-ups found while driving the built app against a real corpus: the 365-day history and month-to-date are now sliced by the same project filter as the headline (they were unfiltered under a filtered headline), the renderer re-reads the filter when it changes outside the window so the Combined label cannot sit over a local filtered query, the pane gets a search box and cost ordering (2,759 rows here), and the instant-switch memo is dropped when the filter changes externally. Every argv shape, the fail-closed read, and the local-only scope decision verified live. We could not push to your fork, hence the superseding PR. Thank you, this was careful work.

PR #1295, opened 2026-09-12
Add drill-through: from aggregates to the sessions that compose them

ozymandiashh:
Connect desktop aggregates to the sessions behind them: day, model, category, project and PR selections open a filtered session list with a detail drawer and navigation history. The view separates each session's contribution to the selection from its full cost.

`sessions --format json --contributions` supplies canonical project ids and call-day segments with model costs and actual per-model call/token counts. Provider identity participates in row matching; ambiguous identities remain explicitly unattributable. Each call follows its own timestamp across midnight. Model usage is independent of model price, including free models and supplementary accounting. PR selection retains the existing even-share attribution rules. Older cached payloads without model usage are excluded with a coverage disclosure when a model filter needs that data, and the report cache namespace is refreshed.

Concrete regression fixtures now return:

- Same session id in Claude and Codex: costs $2 and $9 remain attached to their own rows.
- One turn crossing midnight: $3 belongs to the first day and $5 to the next.
- A $9 model with one call/100 tokens versus a $1 model with one call/900 tokens: selecting the first returns exactly one call and 100 tokens.

Validation: full root suite compared with upstream/main `a1805393` (3,948 passed; the same four Node 26 DEP0205 stderr failures as main). The full app run had 905 passes and three timing failures in unchanged `electron/cli.test.ts`; rerunning that entire file alone passed all 81 tests (four skipped). Root/app typechecks and CLI/Electron/renderer builds passed. The production contribution engine and renderer selection functions were replayed over the concrete fixtures above; all corrected values reconcile. The main project-filter suite passes all nine tests on Node 24.

Branch, period-comparison and work-unit integrations across the other feature PRs remain separate work. Details: `docs/drill-through.md`.


iamtoruk:
Landed as #1316 (2d18da09) with your commits and authorship intact, rebased onto current main, plus three fixes found on real data: the new sessions handler now carries the Projects-pane filter from #1307 (it landed beside it with no textual conflict and silently bypassed it), rawCategory is now actually produced so category drill-through works (it was declared and forwarded but never written), and the model drill keys off the display name so it matches the segments (0 of 19 matched before). Contribution partition verified exact over 1929 sessions. One gap left open and tracked in #1317: the pull-request row is not wired. Thank you.

PR #1296, opened 2026-09-12
Add period-vs-period comparison (Compare periods) with contribution drill-down

ozymandiashh:
Add Compare periods with explicit A/B ranges, project and model contributions, and session detail behind each contribution. Differences use B minus A; cost per day and cost per 100 calls use each side's own denominator. Unknown pricing, unavailable denominators and aggregate-only retained history are disclosed separately.

Project contributions and session drill-down use canonical project identities, so two projects with the same display name retain both costs. Daily normalization recomputes the percentage and direction: $100 over one day versus $160 over ten days displays $100/day versus $16/day, a decrease of 84%. Session-detail cache keys include both ranges, so changing B fetches fresh details after remount.

Validation: full root suite compared with upstream/main `a1805393` (3,956 passed; the same four Node 26 DEP0205 stderr failures as main, zero additional failures); full Electron/renderer suite (893 passed, four skipped); root/app typechecks and CLI/Electron/renderer builds. Concrete fixtures exercise duplicate project labels, unequal durations, and changing B from $160 to $250 with two detail fetches. The main project-filter suite passes all nine tests on Node 24.

The Sessions navigation adapter passes the selected date range. Cross-feature integration with the separate drill-through PR remains a separate step.


iamtoruk:
Landed as #1323 (2373f553) with your commits and authorship intact, rebased onto current main, plus fixes found on real data: both new fetches now carry the Projects-pane filter, compare-periods --format sessions honours --project/--exclude, the stub inspectRange (which no longer compiled after the drill-through landed) is replaced by the real navigation with project and model filters, invalid dates get the app's friendly error instead of a stack trace, and the Totals card discloses carried history inline. Verified: B minus A equals the diff for every component on frozen ranges, lenses reconcile to 1e-12, totals match models --format json to 2e-12 and the history figure matches report to the digit, drill-down rows reconcile exactly, and every existing command is byte-identical. Thank you.

PR #1297, opened 2026-09-12
Add Spend by branch & worktree lens (project x branch attribution)

ozymandiashh:
Add Spend → By branch: select a project and inspect its AI spend per branch, then expand recorded worktrees and session contributions. Branch identity combines the canonical project id and branch name, so equal branch names in separate projects remain distinct. Worktree evidence comes from recorded historical working directories.

The CLI exposes `branches --format json` through the typed Electron bridge. Attribution carries the last observed branch across turns and preserves unknown metadata explicitly. The existing menubar branch contract remains intact. Desktop development uses the standard port 5173 in both Vite and the Electron launch command.

Validation: full root suite compared with upstream/main `a1805393` (3,938 passed; the same four Node 26 DEP0205 stderr failures as main); full app suite (888 passed, four skipped); root/app typechecks and CLI/Electron/renderer builds. Branch fixtures cover equal branch names across projects, branch switches, pre-range carry, worktree evidence, missing metadata and token components. The main project-filter suite passes all nine tests on Node 24.

Branch metadata coverage follows what providers record. The feature does not infer branch value, merge state or branch-to-PR links. Details: `docs/by-branch.md`.

### DIFF
--- a/app/electron/main.test.ts
+++ b/app/electron/main.test.ts
@@ -45,6 +45,7 @@ const CHANNELS = [
   'codeburn:getCompare',
   'codeburn:getYield',
   'codeburn:getSpendFlow',
+  'codeburn:getBranchSpend',
   'codeburn:getOptimizeReport',
   'codeburn:getDevices',
   'codeburn:getDevicesScan',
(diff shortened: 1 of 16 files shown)


iamtoruk:
Landed as #1319 (003854c7) with your commits and authorship intact, rebased onto current main, plus one fix: getBranchSpend was the only main-process fetch without the Projects-pane filter, so a hidden project still showed its branches; now carries projectArgs() like every sibling, with the scoping test extended. Verified on a frozen range: 1,927 projects and 1,746 branch rows sum exactly to report totals, and every (project, branch) key matches the drill-through contributions to the cent. Branch-switching sessions are attributed per turn. flow-json byte-identical. One pre-existing leak it surfaced is tracked in #1321. Thank you.

PR #1311, opened 2026-09-12
fix(menubar): pick the second row's provider by its worst quota window

### LINKED ISSUE #1310: menubar second row: 'nearest its limit' uses the headline window, not the worst window
Follow-up to #1290. The second-row selector picks the provider via QuotaSummary.headlineWindow, which prefers weekly then monthly and only falls back to the max, so Cursor's API window at 100% and Claude's 5-hour window at 18% are discarded for Monthly 9.5% and Weekly 5% on a real machine. Meanwhile the flame tints on the worst window across the six legacy providers, so the comment at MenubarSecondRow.swift:92 (same provider the flame tints for) is wrong on both counts. Either switch the candidates to the worst window or reword README, CHANGELOG and the Settings help to say headline window. Smaller items from the same review: transientFailure providers drop out of the candidates while the dock still shows their dimmed data, so the row can vanish and the item resizes during a backoff; no width cap (GitHub Copilot 12% left · 6d 3h makes the item 2.3x wider); no accessibility label for the two-line title, which now carries a raw newline; menubarRowSnapshot is built on every refresh even when the feature is off.

ozymandiashh:
## Summary

- The second row's "quota remaining" metric picked each provider's `headlineWindow` — a billing horizon (weekly, else monthly, else busiest) that is right for the Capacity Dock rail and wrong for a row meant to name the provider nearest its limit. On a real machine it threw away Cursor's API window at 100% and Claude's 5-hour window at 18% for Monthly 9.5% and Weekly 5%. Each provider now contributes its **worst** window, per-model rows included, which is exactly what `aggregateQuotaStatus` (the menu-bar flame tint) already uses — so the row and the flame agree, and README / CHANGELOG / the Settings help are true as written. The comment claiming the row followed the flame was wrong on both counts and is gone. Fixes #1310.
- A provider in `transientFailure` keeps its last-known window instead of dropping out of the candidates, matching the Capacity Dock, which shows the same data dimmed during a backoff. The row no longer vanishes and resizes the status item for the length of a retry.
- Three smaller fixes from the same review: the row is capped at 24 characters, shortening a long provider name rather than letting "GitHub Copilot 12% left · 6d 3h" more than double the item's width; the two-line title gets a VoiceOver label that reads as one phrase instead of a string with a raw newline in it; and the snapshot feeding the row is no longer built on refreshes where the feature is off. With the setting off the title is byte-for-byte what it has always been.

## Testing

- [ ] `npm test` passes

`mac/Tests/CodeBurnMenubarTests/MenubarSecondRowTests.swift` gains cases for the worst-window selection (including the exact real-machine shape from the issue: Cursor API 100%, Claude 5h 18%, Monthly 9.5%, Weekly 5%), per-model windows, the reset instant following the worst window, tie-breaking, all six connection states, the width cap on every metric, and the accessibility label.

`cd mac && swift build` passes. `swift test` cannot run on this machine — `error: no such module 'Testing'` on a CLT-only toolchain, which is equally true on `main` — so the pure logic was additionally verified with a standalone `swiftc` harness over the real source, and each new behaviour was mutation-checked by reverting it and confirming the harness fails (headline-window selection: 6 failures; dropping `transientFailure`: 2; removing the cap: 3; a passthrough accessibility label: 3). The test file was type-checked against the built module with the `Testing` macros shimmed out.


iamtoruk (approved):
Verified on this machine: the selector now names Cursor at 0% left (API window exhausted) where main named Codex 60% left; off path byte-identical; 506/506, mutation-proven. One prose note: the flame's aggregate covers only the six legacy providers, so Cursor's row and the flame can still disagree. Merging.

PR #1312, opened 2026-09-12
fix(copilot): follow the credential's GitHub host on every quota rung

### LINKED ISSUE #1306: Copilot quota: gh auth token, env var and pasted-token rungs always target api.github.com
Follow-up to #1291. Rungs that carry no host (COPILOT_GITHUB_TOKEN, GH_TOKEN, GITHUB_TOKEN, gh auth token, and a pasted token) are always sent to api.github.com. A user whose only Copilot credential is a GitHub Enterprise Cloud login via gh auth login --hostname <tenant>.ghe.com, or GH_TOKEN with GH_HOST, or a pasted enterprise PAT, gets a 401 and a terminal failure. Options: parse gh auth status / hosts.yml to resolve the host for the gh rung, honour GH_HOST for the env rung, and add a host field next to the pasted-token input. Also update the hardcoded api.github.com string in the dormant Settings detail.

ozymandiashh:
## Summary

- Every Copilot credential rung now carries its own GitHub host. #1286 fixed
  `hosts.json` and `apps.json`; the rungs that carried none —
  `COPILOT_GITHUB_TOKEN` / `GH_TOKEN` / `GITHUB_TOKEN`, `gh auth token`, and a
  pasted token — were still sent to `api.github.com`, so a user whose only
  Copilot credential is a GitHub Enterprise Cloud login got a 401 and a
  terminal failure. The environment rung reads `GH_HOST`, the `gh` rung reads
  the host out of gh's own `hosts.yml` (`GH_CONFIG_DIR`, then
  `XDG_CONFIG_HOME/gh`, then `~/.config/gh`, picked the way gh picks the host
  for `gh auth token`, with no second subprocess), and macOS Settings gains a
  GitHub host field beside the pasted token, saved in the same Keychain record
  so the token and the host can never drift apart.
- Hardening carried forward from #1305: `GH_HOST`, gh's `hosts.yml` keys and
  the pasted host are three new untrusted inputs, and each reaches the existing
  character validation untouched, so a crafted value such as
  `evil.com?.ghe.com` fails closed with no request rather than pointing the
  Authorization header at `api.evil.com`. Settings refuses an unaddressable
  host at the field instead of storing it, and the dormant connection row no
  longer promises `api.github.com` before anything has been fetched.
- `codeburn quota` mirrors the environment and `gh` rungs so the CLI and the
  Windows Capacity Dock report what the menubar reports; its `gh` rung reads
  the token and the host out of one `hosts.yml` entry instead of spawning gh.
  The pasted rung stays macOS-only — it lives in that app's Keychain item.
  Behaviour change worth noting: a shell exporting `GITHUB_TOKEN`, or a gh
  login without Copilot, now reports a Copilot fetch failure where it used to
  report "disconnected", which is what the menubar has always reported for
  those credentials.
- Adding a field to the shared provider credential record required a
  hand-written `init(from:)` using `decodeIfPresent`: Swift's synthesized
  decoder ignores property defaults and throws on a missing key, so every
  credential saved before the field — for every provider — would otherwise stop
  decoding. A test pins that.

## Testing

- [ ] I have tested this locally against real data (not just unit tests) — see
  the note below: this machine has a `github.com` Copilot credential only.
- [x] `npm test` passes

`cd mac && swift build` passes. `swift test` cannot run on this machine (CLT
only, no `Testing` module — also true on pristine `main`), so the new pure
logic and all five rung resolutions were verified with a standalone `swiftc`
harness compiling the real sources and asserting exactly what the XCTest files
assert; each of seven mutations (dropping the host from any rung, dropping the
character guard, reading indented YAML keys as hosts, making `host` a required
decode key) was caught. `npx tsc --noEmit` is clean and `npm test` reports 3987
passing. Verified on this machine against a real `github.com` credential only —
there is no GitHub Enterprise Cloud account or enterprise `gh` login here, so
the `*.ghe.com` paths are unit-tested, as in #1305.

Closes #1306


iamtoruk:
Landed as #1313 (4a316dac) with your three commits and authorship intact, rebased onto current main, plus one commit dropping the CLI rung additions: #1306 was about the menubar's hostless rungs, and adding env and gh rungs to codeburn quota would turn an exported GITHUB_TOKEN or a Copilot-less gh login into a fetch failure where it said disconnected before. The macOS half is unchanged. Verified on this machine: winning rung is gh auth token, hosts.yml lists github.com only, resolved URL equals main, HTTP 200 with the expected keys; GH_HOST=acme.ghe.com resolves to api.acme.ghe.com, github.acme.net and evil.com?.ghe.com are refused with no request. 506 + 159 tests, mutation-proven. Thank you.

Issue #725, opened 2026-07-17
Epic: quota intelligence (live remaining, burn-rate forecast, pre-flight awareness, invoice reconciliation)


iamtoruk:
## Motivation

The single loudest unserved ask across every voice-of-customer source is a live "how close am I to the wall" view with a forecast: how much of my plan is left, and when will I hit the limit at this pace. Users describe it viscerally ("my lizard brain wants to hoard it"; weekly limits "gone in one or two days"), and the structural reason no one has built it is that **no vendor exposes an official quota API** for consumer plans. A local tool that already reads the session logs and the live rate-limit surfaces is the only thing that can physically see the signal, and see it **across vendors** in one place.

CodeBurn is already most of the way there without having named the target:

- `src/plan-usage.ts` models plan periods, per-provider utilization, and a month-end projection (`projectMonthEnd`, `PlanStatus = under | near | over`).
- `src/budget.ts` has run-rate projection (`projected = spent * totalDays / elapsedDays`) and the guard hooks enforce caps in-session.
- The menubar already fetches live quota windows: `ClaudeSubscriptionService` (`api/oauth/usage`) and `CodexSubscriptionService` (`wham/usage`), with #702 handling 429s and #724 surfacing Codex limit-reset credits.

What's missing is the synthesis: a single quota model per plan, a burn-rate forecast against the *plan window* (not just the calendar month), pre-flight cost awareness, and one clear surface that answers "am I about to run out." This epic scopes that, decoupled so pieces can land independently.

## Non-goals (this epic)

- No new network polling loops. Everything rides the existing refresh lifecycle (single-flight, backoff, Retry-After), the rule #702/#724 established.
- No request-path proxy or model routing (that is the separate deferral/routing track, #614 part 3).
- No scraping of undocumented endpoints beyond the ones the official clients already call with the same auth we already hold.
- No hard enforcement of vendor quotas (we cannot block a vendor request); this is visibility and pre-flight nudging. Guard already owns local-cost enforcement.

### Part 1 — Quota model (`src/quota.ts`, foundation, no UI)
[...]

PR #1320, opened 2026-09-12
menubar: detect early quota resets and say so

ozymandiashh:
## Summary

Vendors sometimes reset a usage window ahead of schedule as a goodwill gesture. CodeBurn showed the new
percentage but never said it had happened, so the window of free capacity went unused.

Two signals now catch it on the refresh lifecycle that already runs, with no new polling (#725's hard
non-goal): a reset time that jumps to a genuinely new cycle while the stored one still had time to run, and
usage emptying — a fall of at least 40 points landing at or under 10% — while the advertised reset time
stands still. Both carry the same identity for the cycle they cut short, so the second cannot re-announce
what the first already said.

Three surfaces: a system notification through the existing notifier naming the provider, the window and the
lead ("Claude's weekly limit reset 18h early. You're back to 100%."); a Capacity Dock band saying the same
for twelve hours, drawn and reserved exactly like the staleness notice; and, in the quota hover card, this
Mac's own pattern derived from the 30 days of snapshots already on disk ("Last 3 weekly resets came ~18h
early") — from the stored reset times alone, no network and no external feed. A new `Quota Alerts` switch in
Settings turns the notification off, default on; the dock band still appears with it off.

It stays silent on a normal scheduled reset, a plan change, clock or timestamp skew, a window appearing or
disappearing between fetches, a window's first observation, a reconnect after a terminal failure or a fresh
bootstrap, and a window whose duration the adapter does not validate. A missing or malformed stored reading
is no opinion rather than an event.

One design note for review: the baseline is the previous fetch's readings, kept alongside the announcement
record, not `SubscriptionSnapshotStore`. The store keeps each cycle's *highest* reading, which is right for
"what did the cycle end at" and wrong for "what did the last fetch say" — used as a baseline it would
re-announce a usage-drop reset on every later fetch of that cycle, and make the fetch after a reconnect look
like a fresh drop. The store keeps its stated job and now also backs the history summary; it is otherwise
untouched.

Claude is the provider wired up today, because it is the one whose quota readings are persisted. The CLI has
no equivalent persisted quota store — `src/quota/*` fetches live and persists nothing between invocations —
so it is untouched rather than given an invented one.

`Refs #725` rather than `Closes`: that issue is a five-part epic and this is one slice of its surfaces.

## Testing

- `cd mac && swift build` — Build complete, exit 0 (pre-existing Keychain deprecation warnings only).
- 30 new tests in `mac/Tests/CodeBurnMenubarTests/EarlyQuotaResetTests.swift` (swift-testing, matching the
  neighbouring notification suite), covering both signals, every listed false-positive case as its own test,
  notification coalescing and no-repeat-after-relaunch, the preference, the bounded dock band, the history
  summariser, and the reserved panel height.
- `swift test` cannot run on this machine — `error: no such module 'Testing'`, which reproduces on files this
  branch does not touch and on pristine main. The suite was syntax-checked with `swiftc -parse`, and the
  logic was executed instead through a standalone `swiftc` harness built from the real source files, which
  mirrors every assertion in the suite (including the monitor's, over an isolated `UserDefaults` suite with a
  recording notifier): all pass.
- Mutation-checked: 20 mutations of the new logic — every false-positive guard, both thresholds, the history
  rules, the notice expiry, the preference, the dedupe and the coalescing — each rebuilt and re-run against
  the harness. All 20 are caught. The first pass left three survivors, which produced two additional tests
  and corrected one wrong expectation; it also surfaced a real bug, an exact-timestamp dedupe key that would
  have re-announced the same reset when a vendor replica's timestamp differed by minutes, fixed before commit.
- Not verified: no notification was actually delivered and the dock band and hover caption were not seen in a
  running app; the band's reserved height is asserted arithmetically, as the neighbouring dock tests do.


iamtoruk:
Landed as #1329 (5110eb83) with your four commits and authorship intact, rebased onto current main, plus follow-ups from a real-data review: the usage-drop signal has its own wording (the fetch that fires it shows the reset time did not move, so 'reset early' misstated it; it now says the usage was cleared before its reset), the lead rounds to the unit it prints instead of truncating, two constants the tests read back are now pinned with literals so their mutations fail, the three notification sections in Settings are one Notifications section, and a test asserts the alert identifier never carries the update-notice prefix (tapping it must not install an update). Replay of this Mac's 2,351 snapshots over 30 days: one notification, at the one real event on Sep 1, nothing else. Full Swift suite the author could not run: 606 pass. Thank you, the detector is careful work.

PR #1322, opened 2026-09-12
feat(codex): notify when OpenAI banks a limit reset, and show it everywhere quota is shown

ozymandiashh:
## Summary

- **CodeBurn now tells you when OpenAI banks a limit reset on your Codex account.** These grants are sometimes announced and sometimes not, and they are the one capacity signal you can act on before it is spent — but until now the only way to notice one was to go looking. The reset-credit inventory CodeBurn already reads on every Codex quota refresh is compared against the previous reading; a credit seen for the first time posts one notification naming what was granted, when it landed, and how many you have available to use. Settings → Notifications has a switch for it, on by default.
- **The awkward cases are the feature.** The first reading after connecting is a baseline, not news. A credit that disappears because you spent it says nothing. A failed fetch or a malformed payload is treated as *no opinion* rather than as an empty account, so reconnecting cannot re-announce what you were already told, and a credit that flickers out of one response and back into the next cannot announce itself twice. The seen set persists to `codex-banked-resets.json` in the CodeBurn cache directory — written through `SafeFile` behind an actor, exactly the way `subscription-snapshots.json` is, not a second persistence mechanism — so a relaunch does not repeat a notice. Detection is a pure function over an observation and the state it is judged against; all of the above is unit-tested and mutation-checked.
- **The count and the most recent grant now show up wherever Codex quota shows up**: the Plan tab row #724 added, the agent-tab quota hover card, and a `Limit resets · …` line in `codeburn quota` (text and `--format json`). One formatter produces the sentence, mirrored line-for-line in Swift and TypeScript down to the compact-age rules, so the CLI and the menubar print the same characters. `codeburn quota` gains a `notes` channel for provider facts that are not a window; only Codex sets it, so every other provider's output is unchanged.
- **This adds no network polling loop and no new data source**, per #725's non-goals. There is no new endpoint, no new timer and no new credential: the fields read here come off a response CodeBurn already fetches on the existing refresh cadence — the rule #702 and #724 established. The macOS side reads the companion response already fetched inside the usage 200 branch; the CLI reads the `rate_limit_reset_credits` block of the usage response it already fetches and never calls the companion endpoint.
- **It is a notice, not an action.** Nothing here ever spends, redeems, or requests a credit, and nothing logs a token or a payload.
- One thing worth recording for whoever picks up the rest of #725: **nothing in the payload distinguishes a granted-but-not-yet-usable credit from a usable one.** There is no `available_at`, no pending status, and `granted_at` has only ever been observed in the past. The earliest warning this source can give is "it just landed", not "it lands at 5pm". A `granted_at` ahead of our clock is worded as "lands in 2h" defensively, but no observed payload produces it. Refs #725; no issue is closed by this PR.

## Testing

- [x] I have tested this locally against real data (not just unit tests) — not honestly checkable: there is no test-account plumbing for chatgpt.com in the repo, so the payloads exercised are the committed live-capture fixtures and #723's documented shape, not a live account.
- [x] `npm test` passes — the relevant suites were run rather than the full sweep: `npx vitest run tests/quota-codex-banked-resets.test.ts tests/quota-providers.test.ts tests/quota.test.ts tests/quota-codex-refresh.test.ts app/electron/quota/codex.test.ts` → **94 passed**. `npx tsc --noEmit` is clean.
- [x] `npm run build` succeeds — `cd mac && swift build` is clean on a local swift.org toolchain.

Detail:

- **New Swift tests** (`CodexBankedResetTests.swift`, swift-testing, matching `UpdateNotificationTests`) cover new-credit detection, first-observation baseline, consumed-credit silence, the flicker case, reconnect-no-refire, relaunch-no-repeat over a shared store, malformed/missing payload as no-opinion at both cold and warm state, one post per credit when two land together, toggle-off silence with no backlog on re-enable, denied authorization, retention pruning, and that the credit id never reaches the notification text. Seven more cases extend `CodexResetCreditsTests.swift` (XCTest, matching the file) over the new parser fields.
- **New TypeScript tests** (`tests/quota-codex-banked-resets.test.ts`, 14) cover the wording parity with the Swift expectations, the usable-now clause, soonest-expiry, the available-status filter, the identity fallback, six malformed shapes leaving the quota read connected, and the `codeburn quota` rendering path.
- **Mutation-checked.** Eight Swift mutations and seven TypeScript mutations were each applied to the real source and each produced a failing test: baseline guard removed, consumed id pruned immediately, `nil` treated as an empty inventory, usable-now guard removed, zero no longer hiding the row, unidentifiable credit given a constant id, redeemed credits accepted as grants, applicable-count merge removed, `min`→`max` on expiries, `floor`→`ceil` on ages, future-grant wording removed, `notes` dropped from the command provider. Two of these initially survived and exposed weak tests (a single-credit expiry case where min and max agreed; exact-multiple ages where floor and ceil agreed); both tests were strengthened.
- **`swift test` cannot run on this machine** (CLT-only, no `Testing`/`XCTest` module) — pre-existing and true on pristine main, affecting the whole `mac/Tests` suite, the same limitation #724 reported. The pure logic was instead verified by a standalone `swiftc` harness that compiles the real `CodexBankedResets.swift`, `CodexUsage.swift` and `CodexSubscriptionService.swift` against small stubs for the app-only dependencies and re-runs every assertion from both test files: **59/59 pass**. CI is the gate for the test-runner itself.

Not covered: a live end-to-end fetch against chatgpt.com (no test-account plumbing in the repo). The payload shape used is the one #723 documented and the committed live-capture fixtures show; new test payloads are inline JSON with fabricated credit ids and no account identifiers.


iamtoruk:
Landed as #1328 (66ea03e4) with your three commits and authorship intact, rebased onto current main, plus one fix in the shared notification delegate: since #1283, tapping any notification ran a full app update, so a tap on the new Codex alert would have installed one; the delegate now acts only on the update notice's identifier prefix. Verified: no new endpoint or timer (the companion fetch already existed on main), Swift and TypeScript formatters byte-identical on 34 well-formed cases, quota --format json unchanged for every provider, 569 Swift and 4090 root tests pass, six mutations killed. This Free account carries no reset credit, so the non-zero path rests on fixtures. Two things worth a follow-up: the typed Swift parser now hides the row on a wrong-typed field where main rendered it, and retention keys off first-seen rather than last-seen. Thank you.

PR #1327, opened 2026-09-12
feat(codex): forecast the chance of a usage-limit reset, with its range

ozymandiashh:
## Summary

- **`codeburn quota` now estimates the chance of an OpenAI Codex usage-limit reset landing soon, and never prints that number without its range.** A Codex section reads `Reset forecast: 24% chance in the next 24h (10 to 51%), 4% in 6h (1 to 14%). 12h since the last global reset; typical wait 2.2d. Working hours in SF: yes.`, with the numbers also carried structurally under `providers[].resetForecast` in `--format json`. The same two lines appear in the macOS Codex quota hover card and in the Plan tab, beside the pace captions that answer the same question one step nearer.
- **The model is a pure module, `src/reset-forecast.ts`.** Empirical conditional survival over the inter-reset waits — Kaplan-Meier reduced to its uncensored form — weighting the last 120 days three times as heavily because the cadence sped up (mean wait 201h across the whole record, 116h across the last 120 days), shrunk toward a memoryless rate where the at-risk set thins to one or two waits, ranged by a Wilson interval on the *unweighted* counts, and tilted by the observed hour-of-day density in San Francisco. In the record shipped here nothing lands between 02:00 and 07:00 Pacific and 95% lands between 07:00 and 23:00, any day; the weekday spread is flat enough that treating weekends differently would describe a pattern the record does not show.
- **The honesty rules are the feature.** The prior can tilt a probability but never takes it to exactly zero; no estimate is allowed to reach certainty, because multiplying a hazard by up to 24 otherwise manufactures a 100% that 44 events cannot support; the 24-hour figure is never reported below the six-hour one inside it; a missing or unusable record produces a stated reason rather than a zero; a record older than 14 days is called stale on the line itself; and the wording is "chance", never "expected" and never "will".
- **`confidence` is earned by a walk-forward backtest, and on this record it comes out `low`.** The backtest rebuilds the model from each prefix, asks it at fixed offsets for the chance the next reset lands within 24 hours, and scores it against what happened and against the same prefix's base rate. Over 220 probes the model scores Brier **0.1324 against the base rate's 0.1288** — conditioning on elapsed time adds nothing across these 44 events. So the shipped label is `low`, a test asserts exactly that, and a second test proves the label can reach `moderate` on a record with a real cadence, so the mechanism is live rather than hard-wired. If a future refresh of the record flips the result, that test fails and somebody looks at the claim instead of quietly upgrading it.
- **The dataset is published from this repo.** `src/data/codex-reset-history.json` carries 44 resets and 9 banked-credit grants from codex-reset.com's public timeline — a community tracker, **not operated by or endorsed by OpenAI** — as ids, instants, a type and a reset kind, with every text field dropped before anything is written. `.github/workflows/refresh-codex-reset-history.yml` pushes it to a `data/codex-reset-history` branch every 30 minutes and opens a weekly pull request to keep the copy bundled in releases recent. It never pushes to `main`, and has no `push` trigger, so pushing the data branch cannot retrigger it.
- **Two inputs, two freshness stories, and the second one is the one that mattered.** The *distribution of waits* can be days old without harm: 44 waits do not change shape in a week. The *last-reset clock* cannot, and it was frozen at the last release. It now comes from this machine's own signals — the early-reset detector (#1329) and the banked-credit watcher (#1328) — so a reset that lands here moves the clock within one quota refresh cycle. With neither having fired, it falls back to the dataset's newest reset. On top of that, the dataset itself is refreshed at runtime:

  > #725 exception granted by the maintainer on 2026-09-13: a client-side, first-party, conditional fetch of the reset-history dataset from this repository on GitHub, at most hourly, carrying no user data; no request is ever made to codex-reset.com or any third party from the client.

  One host — `api.github.com`, which the app already contacts for update checks; `raw.githubusercontent.com` would have been a new one, so the contents API is used instead. A conditional GET with the stored ETag, so the usual answer is a 304 with no body and one request against the unauthenticated 60-per-hour limit. At most once an hour, on the refresh that already runs, detached so nothing in the UI waits on it, and only while Codex is connected. No credential, no cookie, no account id, no plan, no usage. A fetched record is adopted only if it passes the same validation the workflow applies — on the raw bytes, because `Decodable` would silently drop a `text` field — **and** is strictly newer than what is held. Offline, rate-limited, malformed or switched off, the record compiled into the build is used exactly as before. **The set of hosts this client contacts is unchanged: `api.github.com`, and nothing else.**
- **A macOS notification, opt-in and off by default** — the only quota notice here that is, because the others report something that already happened and this one reports a probability. You pick a threshold (default a 50% chance within six hours). It fires at most once per crossing and re-arms only once the estimate falls back under the threshold; the fired state persists the way the subscription snapshots do, so relaunching cannot repeat it; a moved threshold or a new reset starts a new crossing; a failed refresh is no opinion rather than a re-arm; a stale record or a wait already past the end of the record never fires at all. The text says what it is before it says a number — "A statistical estimate from public reset history, not an announcement from OpenAI" — and it carries no button, no link and no side effect. Nothing in this feature spends, redeems, requests or refreshes anything.
- **The menubar mirrors the model in Swift because it cannot read this off the CLI payload.** `codeburn status --format menubar-json` carries cost, sessions, providers and pricing but no quota block at all; the menubar's quota comes from its own native adapters. So unlike pricing, the forecast has to be computed there, over a byte-identical copy of the same committed record (SwiftPM resources must live inside the target directory, so the file exists twice and a test pins them together). The two implementations are mirrored constant for constant, and three tests pin the Swift sentences to the *literal* output of the TypeScript module on the same fixture.
- **Local resets are a seam, not a dependency.** When this machine has seen a reset itself — the early-reset detector (#1320) or a banked-credit grant (#1322) — and it is more recent than the public record, "since last reset" counts from yours and the sentence says so. It is an input, so neither PR has to merge first; with no local events the forecast conditions on the global record exactly as it does today. Worth flagging for whoever merges: this adds a Settings section at the same two anchors in `GeneralSettingsTab` that #1320 and #1322 both use, so expect a textual conflict there if more than one lands — the resolution is to keep all the sections.
- `docs/codex-reset-forecast.md` gives the arithmetic, the guards, the provenance, the backtest numbers and a section on what the forecast is not. `Refs #725` rather than `Closes`: that issue is a five-part epic and this is one slice of its surfaces.

## Testing

- [x] `npx tsc --noEmit` is clean.
- [x] `npx vitest run tests/reset-forecast.test.ts tests/reset-forecast-dataset.test.ts tests/quota-codex-reset-forecast.test.ts tests/quota.test.ts tests/quota-providers.test.ts tests/quota-codex-refresh.test.ts` → **130 passed**. 78 of those are new: 47 on the model, 24 on the dataset and the refresh transform, 7 on the command wiring.
- [x] `cd mac && swift build` → **Build complete!**, exit 0, no warnings from the new files. The first run failed — two static `ISO8601DateFormatter`s tripped StrictConcurrency's non-Sendable global check — and was fixed by building parsers where they are used, the pattern every other subscription service in the target already follows.
- [ ] I have tested this locally against real data — not honestly checkable. The dataset is the provided seed capture; `codex-reset.com/api/timeline` has never been called from this branch, so **the exact live response shape is unverified**. The normalizer is tolerant about where the upstream puts its fields and strict about what comes out, and the workflow fails loudly rather than committing a bad file.

Edge cases covered as their own tests, on both sides: empty history, a single reset, two resets, a stale record, an undated record, clock skew in both directions (`now` before the last reset, and a record dated in this machine's future), a wait longer than any in the record, the empty 01:00-to-08:00 Pacific window, the local-event fallbacks (absent, older, future-dated, unparseable), range containment in both directions, the probability ceiling, the recency weighting, the 24h-contains-6h rule, and every threshold-hysteresis case: first crossing, silence above, re-arm below, exactly at the threshold, a new reset, a moved threshold, a failed refresh, a stale record, and a relaunch over a round-tripped state.

**`swift test` cannot run on this machine** — `error: no such module 'Testing'`, which reproduces on files this branch does not touch and on pristine `main`, the same limitation #1320 and #1322 reported. `CodexResetForecastTests.swift` (52 tests, swift-testing, matching the neighbouring pace-presentation suite) was syntax-checked with `swiftc -parse`, and the logic was executed instead through a standalone `swiftc` harness built from the real source files against small stubs for the app-only dependencies, mirroring every assertion in the suite: **163/163 pass**, including the character-for-character parity with the TypeScript output.

**Mutation-checked: 57 mutations, all 57 caught.** The first pass left six survivors, each of which was a real hole and produced an additional test — a clock-skew guard masked by the guard that runs before it, a confidence floor no fixture was small enough to exercise, recency weighting that every flat-cadence fixture was blind to, a probability ceiling pinned on only one of the two sides, staleness tests that used the threshold constant symbolically and so moved with the mutation, and an upward range-widening branch that needed a burst-then-outage record to reach at all. The list covers both range widenings, the Wilson interval, the hour-prior floor, shrinkage, every clock and local-event guard, all five crossing-detector rules, the notice's disclaimer and its range, the render block and its attribution, and eight rules in the refresh script including the one that refuses a response holding fewer resets than the committed file.

Not verified: no notification was delivered and no surface was seen in a running app; the workflow has never executed, and `gh pr create` may be barred org-wide, in which case it falls back to a pushed branch plus a comparison URL in the step summary, the way `star-history.yml` already does.
```

---

# Follow-ups (2026-09-13)

Four rounds on top of the original five commits.

## `4de2a8e8` — the two CI failures

**macOS `swift test`.** Five pre-existing `CodexQuotaSummaryTests` asserted Codex's
`footerLines` exactly and got three lines. Fixed as a design problem, not a test
problem: `QuotaSummary` gains `forecastLines`, defaulted empty, so every other
adapter's construction site and every other provider's summary are byte-identical
to main. The hover card draws it under its own divider. The five tests are
**unchanged** and now structurally cannot be affected — the Codex `footerLines`
array is built from the credit rows alone. Three new tests pin that.

**Windows `test-platforms`.** The dataset-parity test compared `datasetPaths()`
against paths written with `/`. Fixed by building the expected side with `join`,
pinning the invariant as *tail segments* rather than a spelling, and adding a case
that runs the splitter over a Windows path. The byte-identical check also
normalizes line endings and compares parsed documents, since a Windows checkout
can hand back CRLF for a file the script wrote with `\n`.

## `6b53b99d` — the last-reset clock

The maintainer's objection was correct. `codexResetForecast(localEvents:)` defaulted
to `[]` and nothing passed anything, so "since last reset" came only from the
bundled dataset — days stale by construction.

`CodexResetForecastLocalEvents` now reads both sibling stores. Three things worth
recording, all different from what the brief assumed:

- **#1320 persists to `UserDefaults`**, one JSON record per provider under
  `codeburn.quota.earlyReset.state.<providerID>`, seconds-since-1970 — not a file
  under `CodeBurnCacheDirectory.resolve()`. Only `latestEvent` is retained.
- **`detectedAt`, not `scheduledResetAt`.** The latter is when the cut-short cycle
  *would* have reset: in the future, and not when anything happened.
- **#1322's store keeps `firstSeenAt`, not `grantedAt`.** `grantedAt` lives only on
  that branch's transient event and never reaches disk. So the timestamp errs late
  by up to one refresh cycle, never early.

**Also worth flagging: #1320 only records Claude.** `detectEarlyResets` hard-codes
`CapacityDockProvider.claude`. Until it also records Codex, the early-reset half of
this is correct and dormant; the banked half works today. The Codex filter is
enforced twice anyway — by storage key and by the id inside the record — because an
Anthropic early reset moving the Codex clock would wreck the forecast silently.

Wording now names the clock: `12h since the reset observed on this machine at 14:30`
rather than `since the last global reset`, mirrored in TypeScript.

**The CLI is untouched.** `src/quota/*` persists nothing between invocations, so
there is no local store to read; `codeburn quota` stays conditioned on the dataset.

## `d9c898c0` — the merge with `origin/main` (5110eb83)

Five conflicts, all resolved as unions, nothing from either side dropped:

| File | What collided |
| --- | --- |
| `CHANGELOG.md` | both sides' entries, kept in their sections |
| `SettingsView.swift` | main folded the notification switches into one `Section("Notifications")` (`138ef342`); the forecast's notice, its threshold picker and the refresh toggle moved **into** it and the standalone section is gone |
| `AppStore.swift` | five hunks: both announcers, both post-fetch observe calls, both disconnect cleanups, and the Codex summary carrying main's banked footer line alongside `forecastLines` |
| `HeatmapSection.swift` | main replaced `resetCreditsLabel` with `CodexBankedResetPresentation`; that deletion kept, forecast rows kept |
| `src/quota/index.ts` | main's `notes` and this branch's `resetForecast`, both kept — deliberately separate, which is what `4de2a8e8` was about |

**On the mirror types.** Switched to main's real `Codable` types
(`EarlyQuotaResetMonitor.ProviderState`, `EarlyQuotaResetEvent`,
`CodexBankedResetState`), so drift is now a compile error. **Kept the by-path read**,
because neither store exposes a usable one: `CodexBankedResetStore.load()` is
`async` and hard-wired to the real cache path with nothing injectable, while the
forecast is evaluated synchronously from view bodies; and
`EarlyQuotaResetMonitor.visibleEvent` applies the twelve-hour *dock-visibility*
window, which is the wrong filter — a reset twenty hours ago is no longer worth a
band in the dock and is still exactly what "since last reset" counts from.

The `UserDefaults` key prefix stays a literal because `EarlyQuotaResetMonitor` is
`@MainActor` and its static cannot initialise a nonisolated one; a `@MainActor` test
asserts the two strings are equal.

The trade: a record that does not satisfy their shape now reads as nothing at all
rather than being partially decoded. That is the safe direction, and it is tested.

## `c51eef32` — the runtime fetch

Exception statement, verbatim, in the workflow header, the docs page and the PR body:

> #725 exception granted by the maintainer on 2026-09-13: a client-side,
> first-party, conditional fetch of the reset-history dataset from this repository
> on GitHub, at most hourly, carrying no user data; no request is ever made to
> codex-reset.com or any third party from the client.

**Workflow.** Two crons: `*/30 * * * *` pushes to `data/codex-reset-history`
(`contents: write`, `--force-with-lease`, never `main`); `23 6 * * 1` additionally
opens or updates the pull request to `main`. No `push` trigger, so it cannot
retrigger itself. Validated: YAML parses, 7 steps, all four shell blocks pass
`bash -n`.

**Client.** `api.github.com` contents API with `Accept: application/vnd.github.raw+json`
and `If-None-Match`; hourly gate on disk; single-flight; `Retry-After` pushes the
next attempt out rather than sleeping; cookies refused; detached from the UI; only
while Codex is connected. Validation runs on the **raw bytes**, because `Decodable`
would silently drop a `text` field and "nothing beyond the four fields" is the rule
that matters most. A record is adopted only if it validates **and** is strictly
newer. The bundled copy is the floor.

**CLI.** Same fetch, same hourly disk cache and ETag, gated by
`CODEBURN_PRICING_SNAPSHOT_ONLY` — the existing knob that pins pricing to its
bundled snapshot, rather than a second one to discover. `codeburn quota` prints
`Record: <generated_at>, bundled with this build|refreshed from this repository on GitHub.`

**Hosts contacted by the client: `api.github.com`, and nothing else.** Unchanged
from before this feature. codex-reset.com is contacted only by the Action.

Settings placement was corrected in `4d6b2d59` — see below.

## `4d6b2d59` — Settings placement

The refresh toggle is a data switch, not a notification, so it moves out of
`Section("Notifications")` into its own **"Reset Forecast Data"** section, together
with the fact a user actually wants next to that switch: which copy of the record is
in use and when it was built. `CodexResetForecastPresentation.datasetLine` renders it
in the same words and the same raw ISO instant `codeburn quota` prints, so the two
surfaces cannot disagree about which record feeds the number; an undated record still
names the copy rather than showing an empty date. The opt-in notice toggle and its
threshold picker stay in Notifications, where they do post notifications. **The
preference key is unchanged**, so nothing migrates.

Docs page and CHANGELOG updated to name the new section. Verified: `swift build`
clean, harness **245/245**, three new mutations on the dataset line all caught,
`swiftc -parse` clean, `tsc --noEmit` clean.

## Verification on `c51eef32`

- `npx tsc --noEmit` — clean.
- `npx vitest run` over 8 suites (the 4 new plus `quota`, `quota-providers`,
  `quota-codex-banked-resets`, `quota-codex-refresh`) — **164 passed**.
- `cd mac && swift build` — **Build complete!**, exit 0, no warnings from the new files.
- Standalone `swiftc` harness, now built against **main's real**
  `EarlyQuotaReset`/`CodexBankedResetState` declarations rather than stubs —
  **241/241 checks pass**, including the three character-for-character
  TypeScript-parity sentences.
- `swiftc -parse` clean on both touched Swift test files (95 `@Test` functions
  across them).
- **Mutation checks, all caught:** 13 loader/wording + 17 fetch-policy on the Swift
  side, 20 on the TypeScript side, on top of the original 57. Six earlier survivors
  each produced a test; the two that went `SKIP` after the switch to real types were
  re-expressed against the new code and now fail.

## Not verified

- `swift test` still cannot run here (no `Testing` module; pre-existing on pristine
  main). **The five `CodexQuotaSummaryTests` expectations could not be executed** —
  they need `AppStore`, which the harness cannot compile. What can be said is that
  they are textually unchanged and that Codex's `footerLines` is now assembled
  without the forecast. CI is the gate.
- No running app: no notification delivered, no Settings section or hover card seen.
- The workflow has never run; `codex-reset.com/api/timeline` has never been called
  from this branch, so **the live response shape is still unverified**. The data
  branch does not exist yet, so the client's first fetch will 404 until the workflow
  runs once — which is silence, and the bundled record.
- `npm test` in full was not run; the eight relevant suites were.

## Process note

`git stash` is **not safe in this worktree**. A `stash push -u` / `pop` pair
returned another worktree's stash (muse-code files, unrelated to this branch), which
had to be unpicked by hand; the stash list is shared across worktrees of this repo
and another session is using it. The runtime-fetch work was parked on a temporary
branch and cherry-picked instead. No foreign files reached any commit here; the
strays were preserved under `scratchpad/stray-muse-code/` rather than deleted.


ozymandiashh:
Heads-up for review order: the live "since the last reset" clock in this PR reads Codex early-reset events from the #1329 detector, which until now only ever watched Claude. #1339 extends the detector to every provider with quota windows (Claude and Codex announce today; the others are wired but stay silent until their adapters report a window duration). Without #1339 the forecast still works on the global record and on banked-credit events, but its local early-reset input stays dormant for Codex. Merge order that makes the clock fully live: #1339, then this.

iamtoruk:
Closing. 6,200 lines and two parallel implementations for a forecast whose own backtest (Brier 0.1324 against a 0.1288 base rate) does not beat guessing. The always-show-a-range framing is the right instinct. If the model earns real skill on more data, a CLI-only line at a tenth of this size is the shape I would review.

PR #1338, opened 2026-09-12
fix(app): open the sessions behind a pull request row

### LINKED ISSUE #1317: drill-through: the pull-request row does not open the session list
Follow-up to #1316 (drill-through). Day, model, category and project selections open the filtered session list. The pull-request row does not: prFilters is imported and onInvestigate is threaded through four component levels in the Pull requests section, but the row's only handler is expand/collapse, so neither is ever invoked. investigation.test.ts covers prFilters in isolation, which is why it looked wired. Needs a click target on a row that already toggles (a dedicated affordance rather than the row itself). Two minor items from the same review: Sessions.tsx pagination reset is dead under the controlled prop, and FilterChips.tsx keys chips by a truncated label that can collide.

ozymandiashh:
Fixes #1317.

You were right on all three counts, @iamtoruk.

**The PR row.** The plumbing from #1316 was there — `prFilters` imported, `onInvestigate` threaded through `PullRequestsContent` → `PullRequestsPage` → `PrTable` → `PrRowView` — and nothing ever called it. The row's only handler was the expand/collapse toggle, so the section that names actual shipped work was the one aggregate with no way into the sessions behind it.

I gave it a control of its own rather than making the row the trigger, since the row is already a toggle. An expanded PR now shows `View sessions for this pull request →` above the work breakdown, which selects the PR by its URL (the aggregation key the by-PR report is built on) and lands in Sessions with the chip applied and each session's contribution reconciled. Clicking it or pressing Enter on it leaves the expansion exactly as it was — it lives in the detail cell, not in the trigger, so nothing bubbles into the toggle. Rows with no per-turn detail get it too.

On placement: I put it in the expanded detail rather than in the collapsed row so it matches the project entry point in Spend one for one, and because `plain.css` already reserved `.pr-drill` for it next to `.spend-proj-drill` — #1316 shipped the style and not the button. That also kept the `.pr-card-trigger` grid untouched. If you would rather have it visible before the row is expanded, say so and I will move it; it needs a grid column and a `stopPropagation`, nothing more.

**Pagination reset.** `Sessions` reset its visible count to the first page on a selection/sort/search/data change, but it wrote the internal fallback state while the app passes `visibleCount` as a controlled prop (pagination depth lives in the nav history so Back/Forward can restore it). So after `Show more`, re-sorting or removing a chip left you 240 rows deep in a list that had just been reordered. The reset now happens where the count is owned, in the same `commitNav` as the change that caused it — one history entry, and nothing fires during a Back/Forward restore to discard the depth that was just restored. The effect stays for the uncontrolled wiring and now says so.

**Chip keys.** Chips were keyed by the label they display, which truncates a session id to 12 characters and a project path to its last segment. Two sessions of one provider sharing an id prefix, or `/work/app` and `/personal/app` both on `main`, collapsed into one React key — a duplicate-key warning and a real risk of removing the wrong chip. New `filterChipKey` returns the composite identity the selection is already normalized and removed by (provider + session id, project + branch); the visible label is unchanged.

Tests: a render test clicks the new affordance, asserts `onInvestigate` is called with `prFilters(url)` and that the row is still expanded afterwards; a keyboard variant covers a legacy row with no per-turn detail; one asserts no affordance appears when the section has no destination wired. `FilterChips` gets its own file — two chips whose labels collide must render as two chips with no duplicate-key warning, and removing the second must remove exactly it — plus a unit test that `filterChipKey` separates what `filterChipLabel` collapses. An App-level test reveals the extra rows, changes the sort and asserts the list is back on the first page. All four failed before the change; each fix reverted individually puts its test back to red.

`app` suite and `tsc --noEmit` are clean per file; the pre-existing full-suite timeouts on my machine are unrelated (they reproduce on `main`).


iamtoruk:
Landed via #1350 (merge 68ce480) with your four commits intact and credited; only the changelog needed a conflict resolution after the rebase. Verified by driving the branch build over CDP on this machine: the button on the #1341 row lands in Sessions with the PR chip and one session, $10.57 of $16.57 attributed, row still expanded; lifetime view with 9,005 sessions goes 120, 240 after Show more, back to 120 after a re-sort. Placement in the expanded detail is right, leave it there.

PR #1339, opened 2026-09-13
menubar: early-reset notification for Codex

ozymandiashh:
## Summary

- Codex gets the same early quota-reset notification Claude gets. After each successful Codex fetch or bootstrap, its windows go to the existing `EarlyQuotaResetMonitor` under provider id `codex`. A bootstrap only seeds a baseline, and disconnecting Codex clears its record. No dock band, no caption, no new polling.
- Codex windows have no keys of their own, so they are keyed by a pre-localization `storageLabel` when the row has one (the monthly credit row). Otherwise they use the display label, which is a period or model name and does not translate. That way a language switch or the `· limit reached` suffix can't drop the baseline at the moment the reset happens.
- When a provider reports absolute usage, signal 2 requires the absolute figure to fall too. A spend-cap increase drops the ratio but not the usage, so it stays silent. Claude is percent-only and keeps the ratio test unchanged.
- The detector's type doc now spells out that fixed windows are enforced by the `windowSeconds` contract, not by the anchor inequality.

## Testing

- [x] `cd mac && swift build` clean.
- [x] Early-reset test file typechecked under Swift 6 strict concurrency and run locally against the built module with a stub `Testing` module (the machine has no swift-testing): 42/42. Three new tests drive `refreshCodexReportingSuccess()` end to end. Mutations checked: removing the Codex refresh call or dropping `storageLabel` fails them.
- [ ] CI `swift test`.


iamtoruk (changes requested):
Replayed all 3,274 Claude snapshots on my machine through both branches. Identical: 2 detections, both the Sep 1 incident, one notification. Claude is safe.

The rest I cannot check, because nothing but Claude keeps quota-window history on disk here. Before this merges I need your own replay numbers per provider, and the false-positive check on scheduled resets.

Three things first.

Only Codex can actually fire. Kimi, Gemini, Copilot, Antigravity and Cursor all pass `windowSeconds: nil`. That is a lot of code for one provider. Convince me it should not be Codex-only.

`windowKey(forLabel:)` slugifies a localized string. `CreditLimit.shortLabel` is `L("Monthly usage limit")`, so on zh-Hans the slug is empty and the key becomes `"window"`. Switching language drops the baseline and the 30-day pattern. Two non-Latin sibling rows collide and the `seen` set silently drops one. The comment saying it never collides is not true off Latin.

`shortLabel` also appends "· limit reached", so the key flips exactly across at-limit to reset. That is the goodwill reset the changelog leads with, and it cannot fire.

Also: `anchoredToNewCycle` is vacuous when `resetsAt` tracks `now + length`, the inequality holds for any gap. Claude's four fixed windows were the guard. Now anything with a duration goes through. And a Codex spend-cap increase drops percent with `resetsAt` unmoved, which is signal 2. Want tests for both.

For the UI: before and after screenshots of the Capacity Dock detail popover with the early-reset band, the menubar quota hover card with the history caption, and the notification itself.

ozymandiashh:
All three detector points are fixed in d23f0190, the replay numbers follow, and the Codex-only question gets a direct answer below.

**1. windowKey / shortLabel — both real, both fixed.** Storage identity is no longer the display label. `QuotaSummary.Window` gains `storageLabel` (pre-localization, state-free); Codex's credit row sets it (`"Monthly usage limit"`, a constant) while the popover keeps the localized, `· limit reached`-aware label. The monitor keys and names the window from `storageLabel` when present. So: no empty slug on zh-Hans, no baseline dropped on a language switch, no sibling collision, and — the one that matters — no key flip at the limit boundary, which was exactly where the flagship case lived. Pinned by `storageLabelStabilizesTheKey`.

**2. anchoredToNewCycle — you were right that it was vacuous for that shape, and I've stopped pretending otherwise.** The anchor keeps its original semantics (a successor cycle began after the last look at the one it cut short — this is what rejects a same-cycle nudge, and my first replacement attempt broke exactly those tests, which is how I learned the anchor was load-bearing). What a pair of readings cannot prove is now the type doc's first paragraph: a rolling re-anchor's implied start is always "now", so the anchor holds for any gap, and the exclusion of rolling windows is the `windowSeconds` contract itself — an adapter that cannot vouch for a fixed cycle passes no length and the window gets no opinion. Tests pin the contract (`rollingTrackerNeedsTheContract` — the indistinguishable pair stays silent because no duration is vouched), sub-tolerance creep, and a backwards successor.

**3. Codex spend-cap increase → signal 2 — fixed on absolutes.** `CreditLimit` already carries `used`/`limit`; the reading now carries `usedUnits` when the provider reports one, and signal 2 requires the absolute figure to fall too. Your scenario (limit raised, ratio collapsed, usage not) is silent; a real cleared counter falls in both ratio and units and still fires; percent-only providers (Claude) keep the ratio test unchanged. Tests for all three.

**Your replay request — my numbers.** This machine's Claude store: **1,052 snapshots (Sep 8–16), 493 five-hour readings with 14 scheduled rollovers, 559 seven-day with 1 — 0 events, 0 false positives on scheduled resets**, replayed through this branch's detector. (No early reset appears in my store; your Sep 1 incident predates its horizon.) For Codex: there is no on-disk quota-window history to replay anywhere — persisting windows for other providers is what this PR adds, so pre-merge replay for them is not a thing that exists; what I can offer from this machine is that its Codex account does early-reset for real (banked-reset credits recorded Sep 14) and its zcode account gave a 5-hour early reset **today**.

**"Convince me it should not be Codex-only."** Directly: the set that can ever fire is exactly the set whose adapter vouches for fixed cycling — Claude and Codex today, zcode the moment #1347 lands (its payload carries the unit/count enum a fixed duration derives from; follow-up commit there). That is three providers that demonstrably do this in practice, one of which (Claude) already shipped through this exact plumbing in #725 — a Codex-only version would delete nothing but the third provider's future. Everything provider-specific about the feature was already generic in #725's shipped shape (notification, band, hover caption, dedupe ledger); this PR's real content is handing non-Claude windows to that path and the per-provider scoping, not new surface area per provider.

UI screenshots (dock band, hover card with the history caption, the notification) follow as a separate comment on a build with an injected announcement.

ozymandiashh:
1-dock-band
2-hover-card-history

iamtoruk:
Thanks for the screenshots. They show a build from before 13 Sep though: the "Runs out in" and "in reserve" lines were removed from the quota rows on main that day (a858014b), and your branch is cut from before it. So they do not tell me what this PR looks like on the current app.

Also a decision on our side that changes the shape of this PR: the twelve-hour dock band and the 30-day "last 3 resets came early" caption are coming out of main. An early reset is a one-time event and the notification already says it. I am landing that removal today.

What I want from this PR after that:
1. Merge main into your branch.
2. Drop the band and caption plumbing, so this is "Codex gets the same early-reset notification Claude gets". The detector fixes from d23f0190 stay.
3. Cut the changelog entry to two sentences.
4. Fresh screenshots of the notification on the current build, or the test output if you cannot trigger one.

Then it is a small change and I will take it.

iamtoruk:
The band and caption removal is on main now (#1452). Rebase onto that and the rest of my comment above stands.

ozymandiashh:
Merged main (with #1452) and cut it down to what you asked for, in 1837f96d:

1. **Main merged.** Conflicts resolved in main's favour, so the band, the caption and all of their plumbing are gone. That includes this branch's per-provider band and caption state and the cycle ledger that fed the caption.
2. **Codex only.** The one new call site is Codex's refresh and bootstrap. They hand their windows to the same monitor Claude uses, and a Codex disconnect clears Codex's record. The Kimi/Gemini/Copilot/Antigravity/dock-adapter wiring is gone. The detector fixes from d23f0190 stay: `storageLabel` identity for the credit row, the `usedUnits` spend-cap guard and the anchor contract doc.
3. **Changelog** is two sentences.
4. **Test output.** I can't trigger a real early reset on demand, so here are the tests. Three new ones drive `refreshCodexReportingSuccess()` end to end with a stubbed fetcher:
   - `codexRefreshAnnounces`: weekly window at 80% with 18h left, then a new cycle. One notification, `Codex quota reset early` / `Codex's weekly limit reset 18h early. You're back to 100%.`
   - `creditRowSurvivesLimitReached`: the credit row goes from `· limit reached` to cleared with the reset time unchanged. One `Codex quota cleared early`, because the key no longer flips at the boundary.
   - `codexWithoutDurationIsSilent`: the same cleared counter with no validated window length. Nothing.

   I checked that the tests can fail. Removing the Codex refresh call breaks the first and second, and dropping `storageLabel` breaks the second. The full early-reset file passes locally (42/42), and CI's `swift test` is the real gate.

Diff against main: +187/−18 in sources and changelog, the rest is tests.

Issue #1333, opened 2026-09-12
mac CI: ServeConnectionTests is timing-flaky under load (two different tests in two days)


ozymandiashh:
Two unrelated PRs, two different tests in the same suite, both pure timing failures on `macos-latest`, both branches untouched `ServeConnection`:

| date | PR | run | test | failure |
|---|---|---|---|---|
| 2026-09-11 | #1290 (menubar second row) | [34626755506](https://github.com/getagentseal/codeburn/actions/runs/34626755506) | `a request queued during cancelled hydration completes on the same child` (ServeConnectionTests.swift:263) | expected events `[..., "second-replied"]`, got the array without the last event |
| 2026-09-12 | #1289 (zh-Hans localization) | [34725149445](https://github.com/getagentseal/codeburn/actions/runs/34725149445) | `cancelling a hung request returns promptly` (ServeConnectionTests.swift:182, :185) | `error → ServeUnavailable()` where `CancellationError` was expected, and `elapsed → 1.08 s` against a `500 ms` budget |

Both re-runs of the same head went green, and neither branch touches `mac/Sources/CodeBurnMenubar/Data/ServeConnection*`. The second failure is the informative one: a 500 ms wall-clock budget on a shared runner is not a property of the code, and the cancellation arriving as `ServeUnavailable` suggests the child was torn down by the timeout path before the cancellation propagated, i.e. the two races the suite is supposed to distinguish collapse under load.

Suggested fix, for whoever owns `ServeConnection`: drive these tests with an injected clock / continuation instead of real sleeps, or at least widen the wall-clock assertions to something a busy runner can meet (the neighbouring `cli-status-menubar` vitest flake at the 30 s boundary is the same class, see #1216's history). Filing so the next red run on this suite is recognised as this rather than chased.

ozymandiashh:
Third data point, same head as the second row above, re-run of the failed job only ([job 103638897740](https://github.com/getagentseal/codeburn/actions/runs/34725149445/job/103638897740)): `cancelling a hung request returns promptly` again, this time only the wall-clock assertion, `elapsed → 0.516 s` against `.milliseconds(500)`. A 16 ms overshoot on a shared runner is the whole story; the code path is fine.

[...]

PR #1340, opened 2026-09-13
test(menubar): make the localization scan cost milliseconds, not seconds

ozymandiashh:
## Summary

- `LocalizationCoverageTests` was burning ~21 s of CPU in the debug build `swift test` uses. swift-testing runs suites concurrently in one process, so that CPU was taken from every wall-clock assertion running beside it — including `ServeConnectionTests`' 500 ms cancellation budget, which overshot by 16 ms twice in a row on #1289 (#1333).
- The cost was algorithmic, not essential (the tree is 88 files, 1.3 MB): the file was walked once per call-site pattern (26×), each comparison allocated an array slice at every character position, `enclosingTypeName` rescanned from the top for every property (quadratic), and the suite's four whole-tree tests each re-read every file. The scanner now makes one byte-level pass per file behind one- and two-byte dispatch tables, tracks the enclosing type as it walks, builds the line index only when there is a finding, and memoises the scan so the four tests pay for one.
- Adds `scanStaysCheap`, which prints the scan's cost once and fails above 2 s of **thread** CPU (~10× today). Thread CPU rather than wall time, so the ceiling cannot itself become a flake on a shared runner, and rather than process CPU, which would bill it for concurrently running tests. It fails at 20.6 s against the previous scanner.

| | before | after |
| --- | --- | --- |
| thread CPU, whole suite, debug (as CI runs it) | 20.6 – 22.4 s | 0.20 – 0.25 s |
| release build, CPU | 6.2 s | ~25 ms |

Detection is unchanged; this is a performance-only change. All 22 existing coverage tests pass unmodified, and a differential run of the old scanner against the new one — every source file, three mutations of each that force the positive paths, and 25 edge cases (nested comments, multiline/unterminated literals, non-ASCII) — agrees on all 313 findings and 1,406 keys.

Refs #1333

## Testing

- [ ] `npm test` passes

Test-only Swift change; no TypeScript touched. `swift build --package-path mac` succeeds. `swift test` could not run on the machine this was prepared on (Command Line Tools only, no `Testing` module — true on `main` too), so the suite's test bodies were run through a minimal Testing shim: 23/23 pass. Timings above were taken on a heavily loaded machine, so wall time there is contention; thread CPU is the stable figure. CI runs the real suites.

### DIFF
--- a/mac/Tests/CodeBurnMenubarTests/LocalizationCoverageTests.swift
+++ b/mac/Tests/CodeBurnMenubarTests/LocalizationCoverageTests.swift
@@ -104,6 +104,42 @@ struct LocalizationCoverageTests {
         )
     }
 
+    /// A generous ceiling on what the whole-tree scan costs.
+    ///
+    /// swift-testing runs suites concurrently in one process, so this scan's
+    /// cost is paid by every wall-clock assertion running beside it —
+    /// `ServeConnectionTests` checks that cancelling a hung request returns
+    /// inside 500 ms (#1333). In the unoptimised build `swift test` uses, the
+    /// scan once burned about 21 s of CPU across the four tests above; it now
+    /// takes about 0.2 s, once, because the result is memoised. The ceiling is
+    /// ten times that, so it trips on an algorithmic regression — a pattern
+    /// loop rescanning the file, an allocation per byte — and never on a merely
+    /// slow runner.
+    ///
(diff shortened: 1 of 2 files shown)


iamtoruk:
Closing. Six seconds of test time does not justify a 600-line scanner rewrite, and a CPU-budget assertion is a new flake source of its own. #1345 removes the timing dependency that made the scan cost matter, which is the fix we want.

PR #1341, opened 2026-09-13
feat(menubar): say how much of a long quota window the pace leaves unused

ozymandiashh:
## Summary

- On weekly and monthly quota windows that will last, the Capacity Dock caption and the agent-tab hover card now say how much of the window your current pace leaves unused: `Lasts until reset · ~44% unused` (5% used 15 hours into a 7-day window projects to 56% at reset). The tooltip gets a matching sentence, `At this pace about 44% of the window goes unused by the reset.` Both strings are in en and zh-Hans.
- The share is `100 − projected` from the existing `QuotaPace` projection, rounded to a whole percent. It only appears when the rounded figure is at least 10 points and the window is outside the existing two-point on-pace band. Below that the caption stays `Lasts until reset`, because a single-digit leftover is noise on a linear projection. The early-window, validated-duration and freshness/connection guards are unchanged. Windows of six hours or less keep `On pace` / `in deficit` / `in reserve`, and the Plan tab's captions are untouched.
- "At this pace" doesn't fit the dock's two-column cell (155pt; the full sentence sets at ~208pt), so it lives in the tooltip. The hover card's caption (136pt) now shrinks to 0.8 like the dock's, so the share isn't clipped. `codeburn quota` prints no pace caption, so the CLI is unchanged.

Refs #725

## Testing

- [ ] `npm test` passes

- `cd mac && swift build` passes.
- New and updated cases in `CapacityDockPacePresentationTests`: the 44% example, 9 points (plain), exactly 10, rounding (55.6→56, 9.6→10), on-pace band, early-window silence, short-window and 6h boundary, overflow.
- `swift test` can't load `Testing` on a Command Line Tools-only host, so the same cases ran in a standalone `swiftc` harness against the real sources and real `.strings` tables. They all pass on this branch and 12 fail on `main`. Mutating the threshold, the rounding, the sign or either guard fails the harness.
- The localization scanner and catalog checks (unrouted literals, missing or orphaned keys, specifier and `%%` parity) pass on the branch via the same harness approach.
- No TypeScript changes.


iamtoruk:
Closing. The unused share is 100 minus the projected figure the tooltip already shows, so this adds a second sentence for the same fact. The code and tests are clean; the extra surface is not worth carrying.

PR #1343, opened 2026-09-13
fix(menubar): make the quota warning legible and name the window it means

ozymandiashh:
## Summary

- The popover's quota warning painted its text in the same system yellow/orange/red as its 12% pill: 1.23:1 to 1.43:1 for yellow in light mode, 3.33:1 for red in dark mode. The pill keeps its hue; text and glyph now use a per-appearance foreground (`QuotaWarningPalette`) that measures at least 5.29:1, and a test holds every severity at WCAG AA against the composited pill in light and dark mode.
- The banner said "Claude 70% of quota used" without naming a window. That figure is the provider's worst window (here the 5-hour one), so it looked wrong next to a weekly 34%. `AggregateQuotaStatus.warnings` now carries the winning window's label and reset, and the row reads `Claude · 5-hour 71% · resets in 3h 12m`, using the Capacity Dock's countdown and one line per provider. There is no countdown without a reset time or once the reset has passed, and "over limit" now means at or past 100% rather than anything from 90%.
- `"%@ of quota used"` is retired; `"resets in %@"` is added for en and zh-Hans. The flame tint, the compared windows and the 70% threshold are unchanged. The status-item title, its tooltip, the second row and the Capacity Dock do not print the aggregate's percentage, so they are untouched.

## Testing

- [ ] I have tested this locally against real data (not just unit tests). Not done: the popover was not launched; the contrast is modelled against AppKit-resolved backgrounds.
- [x] `swift build` (mac/) succeeds
- [x] New `QuotaWarningRowTests` (16 tests) and the extended `LocalizationCatalogTests.quotaCopyResolves`. This host is CLT-only and cannot run `swift test`, so they were run through a standalone `swiftc -swift-version 6 -strict-concurrency=complete` harness over the real test file: 0 failures, and all 9 mutations caught, including main's colours failing the AA test.
- [x] Localization scanner and catalog rules: 0 findings, 627 keys in both locales
- `npm test` / `npm run build`: not applicable (Swift-only change)


iamtoruk:
Landed via #1349 (merge c2c4844) with your commit intact and credited. One extra commit on top: the `2 * 86_400 + 3 * 3600 + 5 * 60` literal in QuotaWarningRowTests made Swift 6.2.3's type checker give up, so the file did not compile locally; it is now `183_900` with the breakdown in a comment. Verified against real Claude snapshots from this machine: the weekly window crossing 70% renders as `Claude · Weekly 70% · resets in 2d 17h` and turns critical at 75%. Good fix, the contrast numbers were real.

PR #1363, opened 2026-09-14
menubar: click a live session row to raise its terminal tab

iamtoruk:
A live session row in the Capacity Dock card now opens the terminal the session runs in. Claude rows resolve their pid from `<config dir>/sessions/<pid>.json` with a start-time guard against pid reuse; Kimi Code rows find the one running `kimi` whose working directory matches the session's `cwd`. The process tree is walked up to the owning application; Terminal.app and iTerm2 get the exact tab selected by tty, any other terminal is brought to front. The AppleScript runs detached with a three-second cap, so a stuck terminal cannot freeze the app. Rows whose process cannot be found stay inert with no hover affordance.

Verified on this machine: both live rows resolve to Terminal with their ttys, and the generated script followed by activate brings the right window forward. 671 Swift tests pass, 15 new. First click prompts once for Automation permission per terminal app.


ozymandiashh (approved):
Reviewed the full diff. Security posture is right: tty is validated to devname shape before any AppleScript interpolation (with injection tests), scripts are only generated for the two known terminal bundle ids, pid reuse is guarded by start time, osascript is detached with a 3s cap, and ambiguity fails closed (two kimis in one cwd stay inert). Resolution runs once per session-list change off the main actor, rows without a target keep their exact inert layout, and both locale tables gained the strings. Two non-blocking nits: the detached task inside `.task(id:)` doesn't inherit cancellation so a stale result can briefly overwrite a newer one (self-correcting on the next run), and NSCursor push/pop can strand the pointing hand if the row vanishes mid-hover. Fine for a follow-up.

PR #1364, opened 2026-09-14
fix(models): price cliproxy-gateway routed ids as the models they wrap

ozymandiashh:
## Problem

[codex-cliproxy-gateway](https://github.com/aceHubert/codex-cliproxy) keeps Codex's own OpenAI OAuth routing native and forwards only `cliproxy/*` model ids to a local [CLIProxyAPI](https://github.com/luispater/CLIProxyAPI). Every routed Codex session therefore records a `cliproxy/<id>` model name in its rollout — sometimes with a CLIProxyAPI provider path behind the wrapper, e.g. `cliproxy/zcode/glm-5.3-flash`.

None of those names match a pricing row, so routed sessions were counted in the token totals but priced at $0 — and since the models table hides zero-cost rows, routed spend looked absent rather than unpriced. On a real machine routed Claude usage was the single largest line of the day and contributed nothing to the daily cost figure.

## Fix

- Peel the `cliproxy/` wrapper in `ROUTER_PREFIXES`, same class as `orcarouter/`, `cmd/`, etc.
- Add `zcode` to `EXTRA_NAMESPACES` (CLIProxyAPI's provider spelling for the Z.ai coding plans) so `cliproxy/zcode/glm-5.3-flash` strips to its priced `glm-*` leaf.
- An unknown vendor behind the wrapper still fails closed (`cliproxy/nosuchvendor/<id>` stays unpriced), per the existing rule.

## Tests

- Extended the namespace-allowlist and routing-wrapper probe tests (`zzz-*-probe`) with `cliproxy`, `zcode`, the nested `cliproxy/zcode/` spelling, and the fail-closed unknown-vendor case.
- Added `resolveCanonicalModelId` cases: `cliproxy/claude-fable-5-1` → `claude-fable-5-1`, `cliproxy/zcode/glm-5.3-flash` → `glm-5.3-flash`.
- `tests/models.test.ts` plus the five adjacent pricing suites: 271/271 pass.

End-to-end on real data (before → after): the routed `Fable 5.1` row went from hidden/$0 to $218.46 over Sep 8–14, weekly total $2123.95 → $2358.67, and `codeburn models --unpriced` reports a clean period.


iamtoruk:
Landed as #1439 with your commit kept.

Issue #1321, opened 2026-09-12
sharing: current.byBranch[].branch survives sanitizeForSharing and MCP redaction


iamtoruk:
Found while verifying #1319. The menubar payload's current.byBranch rows (from #788) carry raw git branch names, and both sanitizeForSharing and the MCP redactProjectNames pass leave the branch field intact, so a branch such as exp/extraction-arms reaches the shared payload verbatim. Every other identity in the payload (projects, sessions, worktree paths) is stripped or pseudonymized. Branch names can encode ticket ids, customer names and feature codenames, so they should get the same treatment as project names. Pre-existing, not introduced by the branch lens.

PR #1365, opened 2026-09-14
fix(sharing,mcp): drop or pseudonymize pull request rows in redacted payloads

### LINKED ISSUE #1346: Shared and MCP-redacted payloads still carry pull request URLs and owner/repo labels
Follow-up to #1337, which the author scoped out on purpose. `current.pullRequests` survives both `sanitizeForSharing` and `redactProjectNames(include_project_names: false)` verbatim. Each `PrRow` carries the full PR `url` and an `owner/repo#123` label, so the repository owner and name reach a paired peer device and a name-redacted MCP client. Confirmed on a real payload from this machine: the block is present in the week payload.

Expected: sharing drops the block like it drops project and session rows; the MCP pass with names off replaces `url` and `label` with a stable pseudonym and keeps the numbers.

Same shape as #1321 and #1325, one field group, one test per pass.

ozymandiashh:
Closes #1346. Follow-up to #1337, same shape as #1321/#1325: one field group (`current.pullRequests`), one test per pass.

## What leaked

`current.pullRequests` survived both `sanitizeForSharing` and `redactProjectNames(include_project_names: false)` verbatim. Each `PrRow` carries the full PR `url` (the aggregation key) and an `owner/repo#123` label, so the repository owner and name reached a paired peer device and a name-redacted MCP client.

## Fix

- **Sharing** drops the block outright, exactly the way it already drops the project, session and per-branch rows.
- **MCP, names off** replaces `url` and `label` with `pr-<6hex>`, derived from the url via the same salted hash the project/branch/session pseudonyms use, domain-separated (`pr:` prefix) so a PR and a project sharing a name never collide; rows referring to the same PR still line up across calls, and every number (cost, calls, sessions, totals) stays untouched.
- `include_project_names: true` is unchanged: real urls and labels.

## Tests

One per pass, mirroring the #1337 pattern: sharing asserts the block is gone and neither the url nor the label appears anywhere in the serialized payload; the MCP pass asserts the `pr-` shape, label === url pseudonym, preserved numbers, no raw url anywhere, stability across calls, and verbatim pass-through when `include=true`. 21/21 green in the two suites.


iamtoruk:
Merged as #1431 (your commit rebased onto main, authorship unchanged). Thanks.

PR #1370, opened 2026-09-14
fix(menubar): pin multi-argument catalog keys to positional specifiers

### LINKED ISSUE #1331: localization: catalog test cannot see argument-order drift when specifiers share a type
Found while landing #1330. LocalizationCatalogTests compares the specifier list of each key across en and zh-Hans, so a key whose placeholders are all %@ passes even when the Chinese value puts the arguments in the wrong slots. Three such keys shipped with swapped slots in the early-reset copy and were only caught by rendering real data bilingually. Fix: use positional specifiers (%1$@, %2$@) in every multi-argument key so order is explicit and checkable, and have the test require positional form whenever a key has two or more placeholders. Also still unseen by the source scanner: Text(cond ? "A" : "B"), a literal assigned to a let then passed to .help, and prefix + "literal".

ozymandiashh:
Closes #1331.

## The drift class

`LocalizationCatalogTests` compared the specifier *lists* of each key across locales. A key whose placeholders are all `%@` passed even when the Chinese value bound the arguments into the wrong slots — exactly what shipped in the three early-reset keys, caught only by rendering real data bilingually.

## Catalogs

- All **42 multi-argument keys** in both `en` and `zh-Hans` now use positional specifiers (`%1$\@`, `%2$lld`, …). `String(format:)` binds them identically to before, so rendered output is unchanged; the 60 call-site literals moved in lockstep with their keys.
- The parity check now compares **slot → verb**: Chinese word order may reorder slots freely (that is what positional form is *for*), while a slot changing type is still a failure.
- New test: any key with two or more arguments that is not positional, in either locale, fails with the offending tokens named.

## Scanner

The source scanner stopped being blind to the three shapes the issue names — each puts copy on screen looking like machinery to a first-character check:

- ternary branches: `Text(isEarly ? "Came early" : "On time")`
- concatenation: `.help(prefix + " and counting")`
- a literal bound to a `let` and passed by name: `let tip = "…"` then `.help(tip)` (a binding whose right-hand side is `L("…")` is treated as routed)

`Text(verbatim:)` is the explicit do-not-localize opt-out and stays exempt. Running the new rules over `mac/Sources` surfaced exactly four real violations — the `Text(usage.plan ?? "Kimi Code")`-style provider fallbacks in HeatmapSection — now routed through the catalog as verbatim entries per the provider-names policy.

## Verification

This host cannot run `swift test` (no Testing module in the CLT toolchain), so beyond `swift build`:

- a standalone `swiftc` harness exercises the scanner's old and new rules against snippets plus the real `mac/Sources` tree (9/9 pass, 0 findings on the tree)
- a python replica of the catalog tests validates key parity, en identity, positional form, slot parity, `%%` parity, and that every `L(…)` key exists in the table (all pass; 623 keys per locale, 623 requested keys)

CI (`swift test --package-path mac`) is the authoritative gate.


iamtoruk:
Merged as #1435 (your commit rebased onto main, authorship unchanged). Verified by rendering every multi-argument string in both languages against main: no visible change. Thanks.

PR #1434, opened 2026-09-14
fix(menubar): show a Total row under the Trend tooltip's model breakdown

### LINKED ISSUE #1433: Menubar popover: hover tooltip missing Total row since model breakdown (v0.9.24)
## Description

Since v0.9.24, trend/spend charts group models the same way `codeburn models` does (per the release notes: "Trend and Spend charts group models the same way `codeburn models` does"). In the **web dashboard** (`codeburn web`), the hover tooltip still shows a bold **Total** row below the model/session breakdown — so the total for that point stays visible at a glance.

In the **macOS menubar popover** (`CodeBurnMenubar.app`), that Total row is missing entirely from the tooltip since the model breakdown was introduced. Previously, hovering with a single tool selected (e.g. only Claude) showed the total directly; now it only shows the breakdown, with no sum.

## Environment

- CodeBurn 0.9.24 (Homebrew, macOS)
- Affects: CodeBurnMenubar.app, not `codeburn web`

## Steps to reproduce

1. Open the CodeBurn menubar icon
2. Pick a period with daily or hourly buckets (e.g. 7 days / 30 days)
3. Hover over a bar in the trend chart
4. Tooltip shows the model breakdown but no "Total" row beneath it

For comparison: the same hover in the web dashboard (`codeburn web`, same period) correctly shows the Total row.

## Expected behavior

The menubar popover tooltip should show a Total row at the bottom, like the web dashboard does — summing the hovered point regardless of the model/session breakdown above it.

ozymandiashh:
## Summary

- If only Claude is selected, hovering a Trend bar in the menubar popover used to show the day's cost in the tooltip header. Since 0.9.24 it shows a model breakdown and a token count instead, and the day's cost is gone from the tooltip. The web dashboard tooltip still ends with a Total row.
- The model rows are not new. What changed is the data behind the Claude-only view:
  - Up to 0.9.23, provider-filtered `history.daily` rows in `src/usage-aggregator.ts` set `inputTokens`/`outputTokens` to 0 and `topModels` to empty.
  - 38e6e19f (harden provider period and cache contracts, first shipped in 0.9.24) replaced that with `overlayProviderDaySlices`, which reads the real per-provider slices, so those rows now carry tokens and models.
  - `TrendInsight` switches the chart to tokens whenever any bar has tokens. With Claude selected, the chart switched from $ to tokens, and so did the `BarTooltipCard` header. The card never had a Total row, so the cost disappeared.
  - The model rows can't stand in for a total either: the payload carries up to 5 `topModels` and the card renders 4.
- What changed, in `BarTooltipCard` only:
  - When a day has a model breakdown, the rows now end with a 0.5pt separator and a Total row laid out like a model row: `Total`, the day's cost (`bar.cost`, shown when above 0, like the model rows) and `N tok` (`bar.tokens`, always shown, like the model rows).
  - The figures are the payload's day totals, not a sum of the visible rows.
  - The header figure now shows only on days without a breakdown, so no number appears twice. In the All view it moves from the header down to the Total row.
  - New `"Total"` key in en and zh-Hans (`总计`), plus a CHANGELOG entry under Unreleased > Fixed (menubar).
- What did not change:
  - The CLI and the payload. Carrying real per-provider tokens is correct; this only changes how the tooltip shows them.
  - How the chart picks its metric. The Claude-only chart still plots tokens.
  - Days without a model breakdown, which render exactly as before.
  - The web dashboard, desktop app and Windows tray. The hourly buckets and session breakdown mentioned in the issue exist only in the web dashboard; the menubar Trend chart is daily.

Fixes #1433

## Testing

- [ ] `npm test` passes

This is a Swift-only change, so I didn't run the npm checks. I also haven't run the patched app against my own data. What I did run:

- `swift build` in `mac/` passes, both incremental and from a clean build dir, with no warnings from `HeatmapSection.swift`.
- `swift test` can't run on this host (Command Line Tools only, no `Testing` module), so CI is the gate for the Swift tests.
- Localization: I wrote a script that copies the rules from `LocalizationCatalogTests`, `LocalizationCoverageTests` and `LocalizationSourceScanner`. It passes on main (619 keys) and on this branch (620 keys). I also compiled the repo's own `LocalizationSourceScanner.swift` against both trees. It finds 0 unrouted literals and picks up `Total` as a requested key. The new `N tok` text is covered by the scanner's `tok` exemption.
- Payload shape: I ran `codeburn status --format menubar-json --period week` with and without `--provider claude`. The Claude daily rows now have nonzero input/output tokens and a `topModels` list; the All rows have 3 to 5 `topModels`.
- Render harness: I copied `BarTooltipCard` verbatim from main and from this branch and rendered both with ImageRenderer at 332pt wide (popover width minus padding), at 2x, in light and dark:
  - All view with 5 models: the Total row shows the day's cost and tokens, and `Total` lines up with the model names.
  - Claude only: the day's cost is back in the tooltip, in the Total row.
  - Day with tokens but no cost: the Total row shows only tokens.
  - Day with no breakdown: the image is byte-identical to main in both light and dark.
  - Long model names, a 5-digit cost, a 1B+ token count and the zh-Hans label: nothing clips.
  - I can attach the before/after images.
- The card is about 19.5pt taller (one row plus the separator). Its bottom edge is pinned, so it now reaches further up into the chart and can cover a short hovered bar. I left the offset alone to keep the diff small.


iamtoruk (approved):
Checked against my own data. Day totals in the payload match the model rows exactly, same input plus output units, cache excluded on both sides, so the Total row adds up. Days with 5 models correctly show the day total rather than the sum of the four visible rows. Rendered the card with real rows in light and dark, looks right.

Rebased on main locally: build clean, 679/679 tests pass.

One thing. The card is about 20pt taller and its bottom edge is pinned, so it now covers about 30pt of the 90pt chart instead of 11. Short bars disappear under it while you hover. Not blocking. Bump the offset in a follow-up.

Issue #1362, opened 2026-09-13
bug(opencode): total token usage and spend decreases after archiving sessions in OpenCode


nmastereal:
## Summary

In OpenCode, archiving a session is an organizational action that simply sets a timestamp in the `session.time_archived` column of `~/.local/share/opencode/opencode.db`. All session metadata, messages, parts, and token usages remain fully intact in SQLite.

However, Codeburn's discovery query excludes archived sessions (`WHERE time_archived IS NULL`). Because `opencode` is treated as a standard non-durable provider, Codeburn interprets the absence of these sessions during discovery as file deletion, evicting them from the session cache (`session-cache.v9`) and stripping their token usage from historical metrics.

## Environment
- **Codeburn Version:** 0.9.23
- **Provider:** OpenCode (`opencode.db`)
- **OS:** macOS (Apple Silicon)

## Steps to Reproduce
1. Use OpenCode on any project/worktree to accumulate token spend.
2. Run `codeburn overview` or `codeburn sessions -p all` to observe total token usage and cost.
3. In OpenCode, archive the session (or a parent session with child subagents). This sets `time_archived` in `opencode.db`.
4. Run `codeburn overview` or `codeburn sessions -p all` again.

## Expected Behavior
Historical token usage and cost should remain unchanged. OpenCode sessions are still present in the SQLite database; archiving them should not reduce historical metrics or evict them from the cache.

## Actual Behavior
Total token counts and spend decrease. The archived sessions vanish from `session-cache.v9` and from all historical aggregation views.

## Root Cause Analysis
1. **Discovery Query Excludes Archived Sessions:**
   In `src/providers/sqlite-session-parser.ts`:
   ```typescript
   export async function discoverSqliteSessions(
     config: SqliteProviderConfig,
   ): Promise<SessionSource[]> {
     // ...
     const rows = db.query<SessionRow>(
       'SELECT id, CAST(directory AS BLOB) AS directory, CAST(title AS BLOB) AS title, time_created ' +
       'FROM session WHERE time_archived IS NULL AND parent_id IS NULL ORDER BY time_created DESC',
     )
     // ...
   }
   ```
   Archived sessions fail the `WHERE time_archived IS NULL` clause and are omitted from the returned `SessionSource[]`.

2. **Cache Eviction of Undiscovered Sources:**
   In `src/parser.ts`:
   ```typescript
   if (!readOnly && sources.length > 0 && !provider.durableSources) {
[...]

PR #1436, opened 2026-09-15
fix(opencode): read the 2.x session_v2 + session_message generation

### LINKED ISSUE #1293: how can see the usage of the opencode v2?

ozymandiashh:
## Summary

OpenCode 2.x (mainline since 2.0.3, per #1293) writes sessions to `session_v2` and messages to `session_message` (tagged by a `type` column, ordered by `seq`, payload JSON in `data`). The legacy `session`/`message`/`part` tables freeze at upgrade, so every post-upgrade session was invisible to CodeBurn.

- `src/providers/sqlite-session-parser.ts` branches per database on `sqlite_master`: when the v2 tables exist they win and the legacy tables are ignored entirely — the generations are never joined, exactly as ejwill laid out in #1293. Otherwise the legacy path is byte-for-byte unchanged (kilo-code, which shares this parser, is unaffected: it has no v2 tables).
- v2 payloads normalize into the shared message/part shape, so tokens (same `{input, output, reasoning, cache:{read, write}}` object), cost, model refs (`{id, providerID}` → `provider/model`), tools, bash commands, user attribution, the `parent_id` child-session walk, and the session-level cost/token rollup fallback (same columns on `session_v2`) all behave exactly as on legacy.
- `docs/providers/opencode.md` documents the two generations and the branch rule.

## Test plan

- [x] 5 new tests in `tests/providers/opencode.test.ts`: v2-only discovery+parse (tokens/cost/model/tools/user message), non-user/assistant `type` filtering (compaction/idle/model-switched), child-session walk, upgraded DB with both generations (v2 wins, frozen legacy session invisible), and the `session_v2` rollup fallback.
- [x] All 46 pre-existing opencode tests unchanged and passing; full suite 4113 passed / 0 failed.

Fixes #1293.


dcaldr:
Isnt this close contraproductive?

iamtoruk:
Fair question. The cap is not about this PR, it is about fairness across contributors: when one person has five PRs waiting, new ones from them go to the back of the queue instead of the front. Issues and draft PRs are always fine. A sixth ready PR is not, until one of the five lands.

The part that is on us: those five were open because review is slow, not because they are bad. I am going through them today. #1434 just merged, so this one reopens now and the OpenCode 2.x fix goes in next.

iamtoruk (changes requested):
Verified against a real 1.x database: export is byte-identical before and after, $109.91 / 4,009 calls / 177 sessions. Built a 2.x fixture from ejwill's DDL and the v2.0.3 payload schemas; the branch reads it correctly, tokens, cost, model, tools and bash all land. Generations never joined. Good work.

Two things before this lands.

Compaction messages carry cost and tokens in 2.0.3 (CompactionUsage in packages/schema/src/session-message.ts). The `type !== 'assistant'` filter drops them, and the "skips non-user/assistant" test pins that as correct. On 1.x these were role: assistant and we counted them, so this silently undercounts every compacted 2.x session. My fixture lost $0.42 and 181k tokens on one turn. Treat `compaction` like assistant for cost and tokens, and flip that test.

Needs a CHANGELOG bullet under Unreleased / Fixed.

Unrelated note for #1362: the v2 queries add two more `time_archived IS NULL`, so that fix now has five places to strip, not three.

Rebased cleanly onto main, pushed as `pr1436-rebased` if you want to start from that.

ozymandiashh:
Both addressed in da551543, on top of your `pr1436-rebased`:

- `compaction` rows now normalize like assistant messages, so CompactionUsage (cost + tokens, model ref included) lands as it did on 1.x; a `running` compaction carries no usage and still yields nothing. The pinning test is flipped — it now asserts the usage is counted (181k tokens / $0.0042 on the fixture) alongside a running-compaction skip.
- CHANGELOG bullet under Unreleased / Fixed.

opencode + kilo-code suites green (57/57), typecheck clean.

PR #1443, opened 2026-09-15
fix: uncap the Overview model table, billions token rung, unpriced calls dimming (#1318)

### LINKED ISSUE #1318: Overview all-provider model table is capped at 20 rows since it reads topModels
Since #1315 the all-provider Overview table reads current.topModels instead of the per-day history union, which applies TOP_MODELS_LIMIT (20) to that view for the first time. On a machine with 47 models the table drops to 20 and every local or free model (qwen, llama, deepseek, gemini, grok) disappears from it, while the same PR fixed their visibility in the Models tables. The old rows were wrong in both directions, so the trade is defensible, but the cap is a product choice. Knob: raise TOP_MODELS_LIMIT, or give the Overview table its own uncapped list. Also from the same review: compactTokenCount has no billions rung (12.3B renders as 12345.7M), and Models.tsx still dims calls for unpriced rows while the token cells do not.

ozymandiashh:
## What this does

Fixes #1318, and re-lands the three #1315 commits it depends on.

### Part 1 — re-land of #1315 (first three commits, authorship preserved)

As [noted on the issue](https://github.com/getagentseal/codeburn/issues/1318#issuecomment-5649434123), #1315 merged into `fix/dock-pace-followups` minutes after #1314 had already merged that branch into `main`, so `455130f0`, `3d3088f0` and `e6cce142` never reached main — `git merge-base --is-ancestor ef97600d main` fails and `tests/menubar-model-tokens.test.ts` is not on main. #1318 is written against that code.

This PR cherry-picks the three commits onto current main, preserving authors. Conflicts were small and resolved by keeping both sides:
- `src/menubar-json.ts` — main's `rawCategory` on categories + the incoming per-model token fields
- `CHANGELOG.md` — kept main's newer bullets plus the #1265 bullet

If you would rather re-land #1315 your own way, dropping the last commit of this PR leaves you exactly with the re-land.

### Part 2 — the #1318 fix (last commit)

1. **Uncapped period model list.** `buildTopModels` no longer slices at `TOP_MODELS_LIMIT` (20). On the 47-model machine from the issue, every local and free model (qwen, llama, deepseek, gemini, grok) fell out of the Overview table — exactly the rows whose $0 cost ranks them last, and exactly the rows #1315 had just made visible in the Models tables. Every consumer wanting fewer rows already slices its own (desktop hero takes 8, MCP tables take 5 or the caller's limit); the macOS menubar reads a per-session `models` list and never consumed `current.topModels`. The Overview table already renders in a scroll container, so no UI change was needed.
2. **Billions rung for `compactTokenCount`.** The menubar Models row formatter had K/M only, so a period cache-read total past a billion rendered as `12345.7M`; it now renders `12.3B`, matching `MenubarRowFormatter.compactTokens` and the desktop `formatCompact`. Made internal so a test pins the rung.
3. **Calls are no longer dimmed on unpriced rows.** The desktop Models tables dim calls for unpriced models while the token columns (per #1315's own comment) render at full weight — calls are observed usage too, not a pricing artifact. Cost/saved keep their dashes and the add-alias affordance.

### Snapshot semantics

Status snapshot revision 8 → 9: a v8 record holds only the 20 costliest rows, so without the bump a warm snapshot would keep serving the truncated list until the next recompute — the same staleness class v8 itself was bumped for.

## Verification

- CLI suite: 4119 passed (`npm test`)
- Desktop: `tsc --noEmit` clean, 1001 tests passed
- macOS: `ModelEntry token counts` suite 9/9 (incl. the new billions-rung test); module builds via `swift test` with the CLT Swift Testing flags
- The cap test in `menubar-json.test.ts` flipped to pin the uncapped behavior (30 models in → 30 rows out, synthetic still filtered)


iamtoruk:
Thanks for catching the lost #1315 commits, that one is on me.

This changes three screens, so per the new rule it needs before and after screenshots in the PR before review: the desktop Overview model table with more than 20 models, the desktop Models table showing calls no longer dimmed on an unpriced row, and the menubar Models row with a value in the billions rung. Dark and light for the desktop ones. Once they are up I will verify against my own data and take it from there.

ozymandiashh:
The first CI run's mac failure is fixed by dd0f9df8: the model row's secondary token line (from the re-landed #1315) shipped as a raw interpolated `Text(…)`, so the localization coverage scanner flagged it — it never reached the zh-Hans catalog. It now routes through `L(…)` with positional specifiers per the #1331 rule, and both tables carry the key with the Chinese one rebinding the word order. Localized full mac suite passes locally (the only other local failure, ContributionHeatmap's date assertion, is my machine's ro_RO locale — it passes on CI's runners).

ozymandiashh:
Before/after screenshots as requested. **Before** = main (3684f1d5); **after** = this branch (e1b736b9, main merged in). Desktop shots come from the dev build (`vite` + `electron .`) against my own data, all time, provider filter **Claude**.

How I set it up, so you can reproduce:
- **>20 models.** With "All providers", Overview builds this table from `history.daily`, so it's already uncapped on main. The 20-row cap only shows up with a provider filter, which reads `current.topModels`. None of my providers has 20 models, so I added one synthetic Claude transcript through `CLAUDE_CONFIG_DIRS`: 25 extra model ids, one call each, in a sandbox home. Everything else is my real data.
- **Unpriced rows.** `gemini-1.5-pro` and `grok-2` in that transcript have no pricing entry. One thing I found along the way: these rows never reach the desktop Models table on a default run. `codeburn models` drops rows under $0.01 unless `--min-cost`/`--unpriced` is passed, and `codeburn:getModels` passes neither, so the add-alias rows can't show up in practice. For these shots the CLI ran with that default at 0, same change on both builds. Separate issue, not this PR, but worth knowing.
- **Tall table.** The Overview model table sits in a 244px scroll box. For the capture I expanded that box so every row is visible, same on both builds.

### 1. Desktop Overview: "Models this period" (Life, Claude)
Before stops at 20 rows (GPT-4.1 is the last one). After shows all 31.

| | light | dark |
|---|---|---|
| before | before-overview-light | before-overview-dark |
| after | after-overview-light | after-overview-dark |

### 2. Desktop Models: unpriced rows (bottom of the table)
Before: calls are dimmed and tokens show dashes on `<synthetic>`, `gemini-1.5-pro` and `grok-2`. After: calls and tokens are at full weight, and cost/saved keep the dash and "add alias".

| | light | dark |
|---|---|---|
| before | before-models-light | before-models-dark |
| after | after-models-light | after-models-dark |

### 3. Menubar Models rows: billions rung
Main has no per-model token line (#1315 never landed there), so **before** here is this branch at b8bdbb8a, the #1315 re-land without the #1318 fix. **After** is the head. Both render the real `ModelsSection` with ImageRenderer from my real `status --format menubar-json --period all` payload, top six rows, en_US locale.

`10421.6M cache read` → `10.4B cache read`

| before (b8bdbb8a) | after |
|---|---|
| before-menubar-models | after-menubar-models |

iamtoruk (changes requested):
Checked the numbers against my own cache. Froze the window with `--to 2026-09-16` so today's live churn cannot move it: `report --format json` is identical to the cent on three windows including full history, and the frozen menubar payload differs only by the four new token fields and topModels going 20 to 54. Per-model counts sum to the headline exactly. Good.

The uncap is real here: 54 models, and rows 21 to 54 are all the $0 ones, including coder-model at 518 calls and cursor-grok-4.6-high at 749. Billions rung works, Opus 5 reads 10.6B instead of 10576.7M. Swift suite 692 green, localization coverage green, en and zh key sets match. Desktop tsc clean, app tests 29 failures on both sides, same 29.

Two things before it lands.

One, add the B rung to `formatTokens` in `src/format.ts`. Right now `codeburn models` prints a totals row of 29870.8M on my data. That is the exact bug in the title, and the menubar is not the only place it shows.

Two, your getModels finding is right and bigger than you wrote. The default is `minCost: 0.01` at `src/main.ts:2574`, and the filter is cost or savings >= 0.01, so every row your dimming change targets is dropped before it reaches the table. On my data that is 20 rows and 2,160 calls. So the Models.tsx half of this PR does nothing in the shipped app. Keep it, I am opening the issue, but say so in the changelog entry. Not a one-liner to smuggle in here; it changes the table for everyone and needs its own before and after.

Minor: the layout proof test asserts width 360 after setting width to 360. That is not a proof. Assert the token line renders and is not clipped, or rename it.

This lands before the provider-list branch; they collide on one import line only.

ozymandiashh:
All three are addressed in 20712f3e.

1. **`formatTokens` B rung.** Added in `src/format.ts`, so a totals row now reads `29.9B` instead of `29870.8M`. `src/compare.tsx` had its own M/K copy of that formatter, which would still have printed `1234.5M` in `codeburn compare`. It now calls `formatTokens` too; token counts are integers, so output below a billion is unchanged. `tests/format-tokens.test.ts` pins the rungs, including `999_999_999 → 1000.0M` so the boundary can't move silently, and it fails with the B line removed. I checked every other `formatTokens` caller: widths are measured from the formatted string, and sorting and JSON/CSV use raw numbers, so nothing depended on the `M` suffix.
2. **Changelog.** The #1318 entry now says so plainly: in the shipped app unpriced rows don't reach the Models table yet, because `codeburn models` defaults `minCost` to 0.01 and `getModels` passes no override. The dimming change only takes effect once that filter is dealt with. The filter itself is untouched here.
3. **Layout test.** The width expectations are gone. The test now renders `ModelsSection` for two payloads that differ only in whether their rows carry token counts, and asserts the version with counts is taller by at least one text line per row. It fails if the legacy rows get the counts too. The test is renamed to `ModelsSectionTokenLineTests`, and the popover render test no longer claims anything it doesn't check.

Full CLI suite: 4161 passed, `tsc` clean. The Swift test ran locally against the real app module through a swift-testing stub; CI's `swift test` is the real gate.

PR #1448, opened 2026-09-15
feat(models): carry the billing route on the call; key every model row on it

### LINKED ISSUE #1449: models: a Bedrock-routed model renders as its raw id, and aliasing it would merge it into the direct-API row
### The problem, in your words

I run Claude Code and Codex on fixed-fee subscriptions (Claude Max, ChatGPT), and on top of that I dip into API usage — Claude through AWS Bedrock, other models through OpenRouter. The subscriptions cost the same every month whatever I do; the API usage is the variable part, and it is the number I want to watch. CodeBurn cannot separate the two today: the same model billed through Bedrock and through the subscription shows up as one line — or as an unreadable raw id — so a report cannot answer "how much of this month is metered?", let alone "how much of it went through Bedrock?".

The existing knobs each assume the split follows something other than the call: `codeburn plan` treats a whole harness as under the plan, `proxy-path` a whole project, `model-flat-rate` a whole SKU. Hermes records `cost_status = included` for subscription sessions and CodeBurn honours it as $0 (`hermes.ts:483`), but that provenance is not a dimension anywhere. The route (Bedrock / direct / OpenRouter / subscription) is what actually decides whether a call is metered.

Two tools on this machine record Bedrock-routed calls today, and each writes Bedrock's id where a direct call writes Anthropic's / OpenAI's:

| Tool | Direct | Through Bedrock |
|---|---|---|
| Claude Code (`CLAUDE_CODE_USE_BEDROCK=1`) | `claude-haiku-4-5-20251001` | `anthropic.claude-haiku-4-5-20251001-v1:0` (the configured `us.`/`global.` profile prefix is dropped in the transcript) |
| Hermes (`--provider bedrock`) | `claude-fable-5-1` | `anthropic.claude-fable-5-1`, `openai.gpt-5.6-luna` (state.db also stores `billing_provider = bedrock`) |

Evidence (real sessions, this week):

- `codeburn models --provider claude` rendered the Bedrock row as the raw id `anthropic.claude-haiku-4-5-20251001-v1:0` next to `Haiku 4.5`. Pricing was correct (LiteLLM carries the Bedrock rows), so the dollars were right but the row was unreadable.
- `codeburn models --provider hermes` likewise showed `openai.gpt-5.6-luna` and `openai.gpt-5.6-sol` verbatim.
- The obvious fix — aliasing the Bedrock id onto the friendly name so it reads `Haiku 4.5` — would be worse: `modelBreakdown` is keyed by display name (`src/parser.ts:1789`), so the Bedrock spend would silently merge into the direct row and the distinction would be gone.
- Side effect: `looksLikeLocalModel` (`src/models.ts`) treats any `:` in the id as an Ollama tag, so a Bedrock id whose price is missing is excluded from Unpriced as "free local inference" rather than flagged.

### What you tried

- `codeburn model-alias anthropic.claude-haiku-4-5-20251001-v1:0 claude-haiku-4-5` — makes the name readable but merges the rows (see above).
- `--provider bedrock` — rejected; `--provider` is the tool that produced the session, not the billing endpoint.
- Reading `billing_provider` out of `~/.hermes/state.db` by hand.

### Proposed change

Treat the billing door as a **route** on the model, derived from the id shape, and keep it visible in the display name so aggregation keeps the rows apart:

- `getModelRoute(id)` → `{ id: 'bedrock', label: 'Bedrock', baseModel }` for Bedrock's foundation-model ids, `us.`/`eu.`/`global.` inference-profile prefixes, the `-vN:M` version, `:300k` context tags, `arn:aws:bedrock:…` and LiteLLM's `bedrock/` wrappers. Vendor segment from a fixed table so dotted versions (`gpt-4.1-mini`, `glm-4.7`) are never mistaken for a route.
- `getShortModelName` renders a routed id as `"<base short name> (Bedrock)"` — `Haiku 4.5 (Bedrock)`, `GPT-5.6 Luna (Bedrock)`. Every Bedrock spelling of one model folds to one row; direct ids are unchanged; pricing still runs on the full id so cross-region profiles keep their uplifted rate.
- `looksLikeLocalModel` consults the route first.

Bedrock first; the same hook is where OpenRouter / Vertex / Azure spellings belong once a provider is shown to record them. ### Prior art

- #1029 → #1031: `grok-4.5` vs `grok-4.5-build` collapsed to one row; fixed with a display-name suffix (`Grok 4.5 (build)`), ids and pricing untouched. Same mechanism as proposed here.
- #967 → #1053: rows merge by alias-resolved canonical id; "distinct SKUs that only share a label stay two rows." Bedrock ids are distinct SKUs in LiteLLM (a `us.` profile prices above the bare id).
- #1239 → #1241: OpenCode's two spellings of MiniMax-M3 were *merged*, with the split kept as `rawModels`. Different case: same door, same price. Bedrock vs direct is two invoices, so it falls under #1053, not #1241. Since #1241 also made Trend use `getShortModelName`, the desktop and `codeburn models` agree on Bedrock rows for free.
- #1039 / #1118 / #1364: routing wrappers (`cp/`, `orcarouter/`, `cliproxy/`) are peeled on purpose — they are client plumbing with no price or invoice of their own. Bedrock is not that class.
- #968 → #1050: flat-rate is a billing property orthogonal to model identity, handled as a third state. Same shape of problem.
- `looksLikeLocalModel`'s `:` rule dates from #266 (May), before any Bedrock user; it never considered `-v1:0`.

PR: #1448

---

This is the pilot of a three-step series toward a metered-vs-subscription split:

1. **#1449 (this) + #1448** — Bedrock, detected from the model id. Introduces `ModelRoute` / `getModelRoute` and the `<name> (<Route>)` row convention.
2. **#1450** — OpenRouter, and any route that does *not* rename the model: read the endpoint from the provider's own column (Hermes `billing_provider`, OpenCode `providerID`, Cline `apiProvider`, …), carry it on the call. Includes the list of routes others can add after it.
3. **#1451** — surfaces: `--route` on the CLI, route fields in the menubar payload, a "By billing" grouping on the menubar / dock / desktop tab strips beside today's "By tool".

### LINKED ISSUE #1450: Add OpenRouter as a billing route and carry metered/subscription on each call: read the endpoint from the provider's own column, not the model id (with a list of routes to follow)
### The problem, in your words

The question behind this series (#1449): with Claude Code and Codex on fixed-fee subscriptions and API usage on top (Bedrock, OpenRouter), how much *metered* spend am I incurring this month? The route a call took is what decides that — Bedrock and OpenRouter are always metered; a Claude Max or ChatGPT session is covered by the flat fee — so the route has to be a first-class fact about the call, and it has to carry its billing mode.

#1448 makes the *billing route* visible for one endpoint, Bedrock, by recognising the id shape tools write for it (`anthropic.claude-…-v1:0`). That covers Bedrock because Bedrock renames the model. Most other billing endpoints do not: a call billed through **OpenRouter** is recorded as the plain vendor id (`claude-sonnet-4-5`, or `anthropic/claude-sonnet-4.5`), so no id-shape rule can tell it from a direct Anthropic call, and its spend merges into the direct row.

The route *is* on disk, just in a different column than the model:

| Tool | Where the endpoint is recorded | Example |
|---|---|---|
| Hermes | `sessions.billing_provider` (+ `billing_base_url`) in `state.db` | `openrouter`, `anthropic`, `bedrock`, `openai-codex`, `bedrock-mantle` — all seen on one machine this week; `src/providers/hermes.ts` already SELECTs the column and drops it |
| OpenCode / ZCode | session `model` JSON `{ providerID, id }`; `sqlite-session-parser.ts:58` folds it into `providerID/id` and the known-namespace strip then discards it | `openrouter/anthropic/claude-sonnet-4-5` |
| Zerostack | OpenRouter-prefixed model ids (`zerostack.ts:128`) | `deepseek/deepseek-v4-pro` |
| Mux | `provider:model` prefix, stripped at `mux.ts:199` | `openrouter:…` |
| Cline / Roo / Kilo | `apiProvider` in the task's API configuration (not read today) | `openrouter`, `bedrock`, `vertex`, `anthropic` |

So today two things are lost: the OpenRouter route entirely, and any Bedrock/Vertex route on a tool that records the endpoint in a column rather than the id.

### What you tried

- The #1448 mechanism (`getModelRoute` on the id) — cannot work for OpenRouter by construction; the id is the vendor's.
- `codeburn model-alias` — has no notion of a route; aliasing merges rows, the opposite of what is wanted.
- Reading `billing_provider` out of `state.db` by hand.

### Proposed change

Give the route a second source — **the provider's own endpoint column** — feeding the same `ModelRoute` that #1448 introduced, with OpenRouter as the first route that only that source can supply.

1. **Registry of routes**, replacing the Bedrock-only branch in `getModelRoute`: `{ id, label, billing: 'metered' | 'subscription', detect?(modelId), fromProviderField?(value) }`. `bedrock` keeps its id-shape detector, trimmed to what is on disk — `<vendor>.<model>` with an optional `-vN:M` version, vendors `anthropic` and `openai` (the two with coding sessions); ARN / `bedrock/` / `:300k` spellings and the wider vendor table from the first cut of #1448 are dropped until a provider is shown to write them; `openrouter` has none and is recognised from the provider field (`openrouter`, `openrouter/…`, `openrouter:…`, base URL `openrouter.ai`). Both are `metered`. Subscription doors get entries too — `openai-codex` (ChatGPT plan; Hermes writes it as `billing_provider`), `claude-max` / `anthropic-oauth` where a harness can tell OAuth from an API key — as `subscription`. A provider that already records the mode per session (Hermes `cost_status = included`, Copilot's rated/unrated credits) overrides the route's default. Display convention: `Sonnet 4.5 (OpenRouter)`, `Sonnet 4.5 (Bedrock)`, `Sonnet 4.5` for direct.
2. **Carry the field.** `ParsedProviderCall.route?: string` and `billing?: 'metered' | 'subscription'` (resolved from the route's default unless the provider states it). Providers set it where they have it: Hermes from `billing_provider`, OpenCode/ZCode from `providerID` *before* the namespace strip, Mux from the colon prefix before it is dropped, Cline-family from `apiProvider`. Where absent, `getModelRoute(model)?.id` fills it. `parser.ts` keys `modelBreakdown` on `route ? "<name> (<label>)" : name`, so the column-sourced route and the id-sourced route land in the same row shape. Persist on `CachedCall` (bump `CACHE_VERSION`). **The daily cache stores the display key, not raw ids** (maintainer correction on #1448): `day.models` is keyed by the string `getShortModelName` produced at aggregation time, and #1448 needed no bump only because that function was idempotent on the old key. Once the row key becomes `(name, route)`, finalized days must re-derive — `DAILY_CACHE_VERSION` bumps, and the migration ladder at the top of `daily-cache.ts` gets its entry. Every surface must key rows the same way: `parser.ts` `modelBreakdown`, `usage-aggregator` `mergeDayModelsByDisplayName`, `menubar-json` `buildTopModels`, `model-breakdown` `aggregateModelTotals`, and `models-report` (which today keys on `resolveCanonicalModelId` — the mismatch that produced two same-label rows in #1448). One key function, called from all of them. This is the store-at-parse / apply-at-read pattern #890 lays out for reported-vs-computed cost (both are provenance about a call the caches must carry, not recompute).
3. **Pricing stays on the raw id.** LiteLLM already carries `openrouter/…` (103 rows), `bedrock/…`, `vertex_ai/…` (115), `azure/…` (187). A route never changes what a call costs; it changes which row the cost lands on.

Routes others can add after this lands, each a registry entry plus the provider field that names it (id-shape detector only where the endpoint renames the model):

| Route | Id shape? | Provider field | Notes |
|---|---|---|---|
| `openrouter` | no | Hermes `billing_provider`, OpenCode `providerID`, Cline `apiProvider`, Mux prefix | this issue |
| `vertex` | yes: `claude-…@20251001`, `vertex_ai/…` | Cline `apiProvider = vertex` | `@date` is stripped today and merges with direct — `docs/providers/claude.md` |
| `azure` / `azure-ai` | yes: `azure/…`, `azure_ai/…` deployment names | Cline `apiProvider` | deployment names are user-chosen; alias support matters |
| `bedrock-mantle` | partly (`bedrock-mantle.<region>.api.aws`) | Hermes `billing_provider = bedrock-mantle` | Bedrock's OpenAI-compatible endpoint; seen on disk already |
| `openai-codex` (ChatGPT subscription) vs `openai` (API) | no | Hermes `billing_provider` (60 sessions on one machine, all `cost_status = included`), Codex auth mode | the `subscription` side of the split; without it "metered" cannot be computed as a share |
| `claude-max` / Anthropic OAuth vs API key | no | Claude Code transcripts carry no auth marker; Hermes `billing_provider = anthropic` covers both | the biggest gap for the subscription side — may need `codeburn plan` as the tie-break |
| `copilot`, `cline-pass`, `kimi-for-coding`, … | some (`cp/`, `cline-pass/` prefixes already peeled in `models.ts`) | provider-specific | today these peel to the bare model and lose the door |
| `ollama` / local | yes (`:tag`) | — | already handled as "free"; would just gain a label |

Related: #1241's `rawModels` already exposes which raw ids fed a display row; once `route` exists, a row's routes can be listed the same way for consumers that want the merged view. #1420 (unpriced usage hidden) benefits too — a route-tagged unpriced row says which account is missing a price.

Not in scope: filtering/drill-through and the desktop / menubar / web surfaces that show the split — #1451. Pricing changes — none needed.

Depends on #1448 (`ModelRoute`, `getModelRoute`).

ulises-c:
## Summary

Closes #1449, closes #1450. Reworked in place after the reviews below: the route is a field on the call first, rendered second, with one row key across every surface. Series: this → #1451 (`--route` filters, payload fields, "By billing" grouping, and the per-call `metered` / `subscription` split).

- **The billing route rides on the call.** `ParsedProviderCall.route`, threaded through `ParsedApiCall` and `CachedCall`. Two sources feed it: the model id when the door renames the model (`getModelRoute` — Bedrock's `<vendor>.<model>[-vN:M]` with an optional `us.`/`eu.`/`global.` profile prefix, vendors `anthropic` and `openai` only, the shapes on disk and nothing wider), and the provider's own column when it does not (`routeFromProviderField` — Hermes `billing_provider`). Direct doors map to no route: the unsuffixed row *is* the direct row.
- **One row key, everywhere.** `modelRowKey(model, route)` = short name + door label. The cross-region profile is a dearer SKU and keeps its own row, `(Bedrock us)` (#1053). Idempotent, so a display-keyed daily row re-keys to itself. Called from every site that used to key or label rows its own way — `parser.ts` `modelBreakdown`, `day-aggregator` (day *and* provider-slice models), `usage-aggregator`, `menubar-json`, `model-breakdown`, `models-report`, `session-output`, `model-efficiency`, `overview`, `sessions-report`, `telemetry-snapshot`, `compare-cohorts`, `compare-stats`, `live-sessions`. `getShortModelName` is untouched.
- **`models-report` folds on the row key, not the route id.** The fold identity is the alias-resolved id plus the row key's own door suffix (`modelFoldKey`). Two route ids that render one label — one door with two endpoint names — therefore land on one row instead of two rows the reader cannot tell apart, which is what the second review caught.
- **Daily cache v33.** `day.models` is keyed by the row key, so a column-sourced route survives into the finalized day (the raw id alone cannot carry it). `hermes` joins `PENDING_REDERIVE_PROVIDER_VERSIONS` at 33; ladder entry added. Parse version `billing-route-v1` re-parses cached Hermes sessions once.
- Pricing never consults the route: `getModelCosts` runs on the raw id.

**Registered routes:** `bedrock` only. `openrouter` and `bedrock-mantle` are real Hermes `billing_provider` values, but nobody has sessions for them yet, so they are held back under the same rule the id shapes were cut by last round. They map to no route today, exactly like an unknown door.

**One surface deliberately disagrees.** `granular-history.ts` still keys its buckets on the raw model id, so the desktop Trend timeline is the only place where a Bedrock id and its direct sibling are not named the way every other surface names them. Left as is; noted here so it is a known gap and not a surprise.

## Why this shape

On the first review: the first cut put the route in a display string that three surfaces keyed differently (`models-report` on `resolveCanonicalModelId`, the rest on `getShortModelName`) and shipped two same-label rows next to one blended row. This PR makes the route a field first and gives every surface one key function. On the second review: folding on the route id reintroduced the twin-row defect in a narrower form, so the fold now carries the row key's suffix and never the id.

## Testing

- [x] Tested locally against real data (not just unit tests)
- [x] `npm test` — 4203 passed / 5 skipped; `npm run test:locks` (36); `npx tsc --noEmit` clean; `npm run build`
- [x] Merged current `main`; no conflicts left

**The test the first cut lacked.** `tests/cli-route-surfaces.test.ts` runs the real CLI (`tsx src/cli.ts`) over the three Claude sessions the review seeded — direct, `anthropic.…-v1:0`, `us.anthropic.…-v1:0` — and asserts `models --format json`, `status --format menubar-json` (`current.topModels` *and* `history.daily[].topModels`) and the on-disk `daily-cache.v33.json` carry the same three rows: `Haiku 4.5`, `Haiku 4.5 (Bedrock)`, `Haiku 4.5 (Bedrock us)`, at the prices `main` showed for the raw ids (bare Bedrock = direct; `us.` profile above both).

**The fold.** `tests/model-routes.test.ts` pins `modelFoldKey` to the row key: the key carries `(Bedrock)` and never the route id, so a second door added under an existing label folds instead of twinning. `tests/day-aggregator.test.ts` asserts a column-routed call and an id-shaped one of the same SKU land on one row in the finalized day, with the cross-region profile still apart.

**Real data.** On a copy of a real 496 MB cache, `report --period lifetime --format json` and `models --format json` on `main` and on this branch: all 142 finalized days match, $37,898.851498 and 272,985 calls on both sides; 38 lifetime model rows, identical names and costs. Only today moves, in both directions between two consecutive `main` runs — live sessions appending. No Bedrock or column-sourced route exists in that cache, so no row folded or changed name there. The menubar history rows that the v33 bump re-derives (stale v32 days written before the Grok Bot provider landed) were checked against `main` with only `DAILY_CACHE_VERSION` bumped and none of this code: identical, 0 day mismatches.

**Not proven with real data:** any route other than Bedrock, which is why only `bedrock` is registered.

Affiliation: none. I run Bedrock through an employer AWS account alongside Claude Max and ChatGPT plans and wanted the metered share.


iamtoruk (changes requested):
Thanks. Direction is fine, mechanism is not yet.

Seeded three Claude sessions on one machine: direct, `anthropic.claude-haiku-4-5-20251001-v1:0`, and `us.anthropic.claude-haiku-4-5-20251001-v1:0`. `codeburn models` prints two rows both called `Haiku 4.5 (Bedrock)`, $1.65 and $1.50, with nothing to tell them apart. `status --format menubar-json` prints one row at $3.15. Before this PR both surfaces showed three distinct rows. So "every Bedrock spelling folds to one row" and "desktop and codeburn models agree without further work" are not true: models-report keys on `resolveCanonicalModelId`, everything else keys on `getShortModelName`.

The merged row is also the thing you cite #1053 against. `us.` prices 10% above the bare id. Blending them under one label is two SKUs in one row.

The id shapes you can show on disk are `<vendor>.<model>` and `-v1:0`. #1449 says Claude Code drops the profile prefix. So drop the ARN regex, the `bedrock/` prefix, the `:300k` tag, and cut the vendor table to what you have sessions for. Stability and TwelveLabs do not write coding transcripts.

The `looksLikeLocalModel` fix is right and is the only bug in here. It does not need the route machinery; `-v\d+:\d+$` is a version, not a tag. Happy to take that on its own today as a separate small PR.

On the rest: a display suffix puts billing into a string that three surfaces key differently. Put the route on the call as a field first, then render it. That is #1450. Land that shape, and this becomes small.

One correction for later: the daily cache stores the display key, not raw ids. No bump is needed here only because `getShortModelName` is idempotent and the old key happened to be the raw id. Do not carry that assumption into #1450.

Checked: 46 adversarial ids incl. Vertex, Azure, OpenRouter and Ollama tags, zero misclassified; direct ids unchanged; pricing still on the full id; OTLP wire untouched; tsc clean; 4,126 tests pass.

ulises-c:
Reproduced with three seeded Claude sessions (direct, `anthropic.…-v1:0`, `us.anthropic.…-v1:0`): two identical `Haiku 4.5 (Bedrock)` rows in `models`, one blended $3.15 row in the menubar payload, three distinct rows on `main`. `models-report` keys on `resolveCanonicalModelId`, everything else on `getShortModelName`, and I only changed the latter — so the #1053 point stands against this PR. I tested each surface's output but never the same ids side by side across surfaces; that is on me.

Taking your split:

- The `looksLikeLocalModel` fix is up on its own: #1463. One line, `-v\d+:\d+$` exempted before the tag rule, no route machinery. New test is red on `main`, green there; full suite 4148 vs 4147.
- This PR goes back to draft until #1450 lands the route as a field on the call. When it does, the display part here shrinks to rendering that field, keyed once.
- Noted on the daily cache storing the display key. #1450's design will treat the row key as `(name, route)` with a `DAILY_CACHE_VERSION` bump, not assume idempotence carries it.
- Id grammar in #1450 trimmed to what is on disk — `<vendor>.<model>` and `-vN:M`, no ARN / `bedrock/` / `:300k` — and the vendor table cut to `anthropic` and `openai`, the two with real coding sessions. The Hermes `billing_provider` column covers the rest without id shapes.

Thanks for the adversarial-id pass.

ulises-c:
Reworked in place per the review: the route is now a field on the call (`route` / `billing`), fed by the id shape or the provider's own column (Hermes `billing_provider`), and every surface keys model rows on one `modelRowKey`. The branch is a single new commit on current `main` (d5d6a919); the display-suffix commit is gone. Body rewritten to match. The cross-surface test that was missing is `tests/cli-route-surfaces.test.ts`: your three seeded sessions, asserted identical in `models`, the menubar payload and the daily cache, at the prices `main` showed. Daily cache v33 with hermes in the pending-rederive table, per your correction on the display key.

iamtoruk (changes requested):
Numbers check out. On a copy of my real cache, all 142 finalized days match main to the twelfth decimal, 272,985 calls both sides, $37,898.85 both sides. Only today moves, and it moves in both directions between two main runs, so that is drift. Cold rederive costs 3.3 s once.

One thing to know: preserved-from-expired-logs goes from $23,554.65 to $24,839.95. Two days, 15 and 16 Aug, flip to carried, same cost and calls on both sides. I bumped DAILY_CACHE_VERSION to 33 on main with none of your code and got the identical $24,839.95, so it is the version bump, not you. Nothing lost.

The three seeded sessions now agree on models, the menubar payload, the daily cache, overview, sessions, telemetry, live sessions and compare, at the prices main showed. That is what I asked for.

One defect left, and it is the one I rejected the first time in a narrower form. `bedrock` and `bedrock-mantle` are two route ids with one label. models-report folds on the route id, everything else keys on the label. A Hermes user with both doors for one model gets two rows in `models` both called `GPT-5.6 Sol (Bedrock)` with nothing to tell them apart, and one merged row in the menubar. Reproduced through aggregateModels. Fold on the row key, not the route id.

Two cuts while you are in there. `billing` has no readers at all, so hold it for #1451 with the rest. Drop `openrouter` and `bedrock-mantle` until someone has sessions, the same way the id shapes went last time. Also granular-history still keys on the raw id, so the desktop timeline will be the only surface disagreeing; fine to leave, but say so in the body.

Rest is verified: 4,163 tests, locks 36, tsc clean, merges clean.

iamtoruk:
Applied all four.

The fold in models-report now keys on the row key, not the route id: the identity is the alias-resolved id plus the row key's own door suffix, so two route ids that render one label land on one row. `tests/model-routes.test.ts` pins that the fold key carries `(Bedrock)` and never the route id, so the next door added under an existing label cannot reintroduce the twin rows.

`billing` is gone — the field on the call, the type, and the route's own `billing` property, none of which had a reader. It comes back with #1451. `openrouter` and `bedrock-mantle` are dropped until someone has sessions, along with the `aws-bedrock` / `amazon-bedrock` column spellings and the `openrouter/` and `openrouter:` path forms, which existed only for them. `bedrock` is the one registered route. granular-history still keys on the raw id, so the desktop timeline stays the one surface that disagrees; that is now stated in the body.

Also cut an unrelated line in telemetry-snapshot that the rework had touched.

Real cache, 496 MB copy, main vs branch: all 142 finalized days match, $37,898.851498 and 272,985 calls on both sides, 38 lifetime model rows identical in name and cost. Only today moves, and it moves between two consecutive main runs too. No Bedrock or column-sourced route exists in that cache, so nothing folded or renamed there. The menubar history rows the v33 bump re-derives — stale v32 days written before Grok Bot landed — match main with only DAILY_CACHE_VERSION bumped and none of this code, 0 day mismatches.

4203 tests, locks 36, tsc clean, main merged.

PR #1458, opened 2026-09-16
feat: i18n for terminal output surfaces (overview, status, dashboard)

### LINKED ISSUE #1457: Add i18n support: localized CLI output (zh-CN, zh-TW, ja, ko, fr)
## Proposal

Add an internationalization (i18n) layer so CLI output (labels, table headers, section titles, summary lines) can be displayed in languages other than English.

### Requested locales (initial set)
- `en` (default, unchanged behavior)
- `zh-CN` (Simplified Chinese)
- `zh-TW` (Traditional Chinese)
- `ja` (Japanese)
- `ko` (Korean)
- `fr` (French)

### Proposed design (looking for maintainer feedback before writing code)
1. **New module** `src/i18n/`: a hand-rolled, dependency-free locale catalog (one TS file per locale) plus a `t(key, vars?)` lookup with English fallback for missing keys. No new runtime dependencies.
2. **Locale resolution order**: `codeburn language <code>` config (new subcommand, stored in the existing config.json) → `CODEBURN_LANG` env → `LANG`/`LC_ALL` env (e.g. `zh_CN.UTF-8`) → `en`.
3. **Scope of this first PR**: user-facing strings of the read-only report surfaces — `status`, `overview`, `report`/`today`/`month` dashboards (table headers like Cost/Tokens/Share, section titles like By tool / Top models / Daily, totals labels). Parser logic, provider code and the menubar JSON contract are untouched.
4. **Tests**: fixture-based `tests/i18n.test.ts` covering locale resolution, fallback behavior, and that zh-CN output contains no untranslated catalog keys; plus asserting `npm test` parity with main.
5. Numbers, currency symbols and model names stay as data (not translated); currency display remains governed by the existing `currency` config.

### Non-goals for this PR
- Translating commander `--help` text (follow-up PR if the approach is accepted)
- menubar / Electron / GNOME UI strings (follow-up)

Happy to adjust scope or mechanism (e.g. JSON catalogs instead of TS, different env var name) to match maintainer preferences. If this is welcome, I'd implement it against `main` with full-suite verification per CONTRIBUTING.

pixelxiao:
### Linked issues

Closes #1457 — the approach (catalog shape, locale resolution, surface scope) is written up there. No maintainer reply on the issue yet; happy to reshape scope or mechanism (JSON catalogs, different env var, narrower surfaces) to match maintainer preferences before this lands.

### What changed and why

Adds a dependency-free locale catalog (`src/i18n/`, one TS file per locale, typed against `en` so a missing key is a compile error) for six locales: `en`, `zh-CN`, `zh-TW`, `ja`, `ko`, `fr`, and wires the human-facing terminal surfaces:

- `overview`: headings, kv labels, table headers/rows, budget line, footnotes (carried/unattributed/estimated/bottom line)
- `status`: the `Today / Month / N calls` headline
- interactive dashboard: panel titles, empty states, loading/scanning lines, session counts, activity labels

Design decisions:

- **Resolution order**: config `language` (new `codeburn language [code]` / `--reset` command, modeled on `currency`) > `CODEBURN_LANG` > `LANG`/`LC_ALL` > `en`.
- **Machine-readable outputs stay English**: `status --format json|menubar-json`, `export`, and the stdio `serve` protocol force `en`, so the menubar/tray/GNOME clients and contract tests never see translated labels. Shared helpers keep English twins (`formatSessionCount`, `unpricedModelHint`, `SESSION_COUNT_HELP`) for MCP/JSON surfaces; display surfaces use localized twins.
- **CJK table alignment**: `renderTable`/kv padding now use terminal display width (East Asian Wide = 2 columns) instead of codepoint count, so Chinese/Japanese/Korean tables align; long translated keys (e.g. fr `Détail des tokens`) always keep a separating space. English output is byte-identical to before.
- **Hermetic tests**: under vitest the host `LANG`/`LC_ALL` are ignored (CODEBURN_LANG still works), so the suite stays green on CJK developer machines.
- Numbers, currency symbols, model names and command names stay as data; only prose/labels are translated.

Not covered (deliberate follow-ups, noted in #1457): commander `--help` text, web dashboard, and the menubar app UI (upstream `feat/menubar-i18n` already targets that separate surface — no overlap with this change).

### Review follow-ups

(empty)

### Testing

- `npm run build`: clean (only the pre-existing chunk-size warning)
- `npm test` on `main` (baseline, this machine): **303 files / 4116 tests passed, 0 failed**
- `npm test` on this branch: **304 files / 4134 tests passed, 0 failed** (18 new in `tests/i18n.test.ts`: catalog key parity + no-empty across all 6 locales, `normalizeLocale` regional mapping, resolution precedence + machine-format override, display-width/padding, localized render surfaces, CJK table alignment)
- Manual on real data: `overview` / `status` / `language` exercised in all six locales; `codeburn language zh-CN` + `--reset` round-trip; `status --format menubar-json` verified unchanged English under a `zh_CN.UTF-8` host locale

Before / after (`codeburn overview`, same corpus; en unchanged, zh-CN localized, columns aligned):

```
# en (unchanged)                      # CODEBURN_LANG=zh-CN
Totals                                总计
  Cost       ¥42.92                     成本       ¥42.92
  Tokens     22,687,291   (breakdown     Token 明细 22,687,291   (明细见下)
  Calls      194   At least 4 sessions    调用       194   至少 4 个会话
  Cache hit  100.0%                       缓存命中   100.0%
Tokens                                Token 明细
│  Type       │      Tokens  │  Share  │  │  类型      │      Tokens  │  占比  │
│  Cache out  │  21,920,997  │    97%  │  │  缓存读取  │  21,920,997  │   97%  │
```

### Screenshots

Terminal-only change; before/after text captures above (per the readable-output convention, no image attached).

### Notes for reviewers

- `src/i18n/index.ts` reads config.json synchronously (memoized) because locale resolution must stay sync for render helpers; `reloadLocaleConfig()` drops the memo after `codeburn language` writes.
- The kv label column pads to `max(11, width(key)+1)`; English keys are all ≤ 11 columns so en output is untouched.
- Catalog values intentionally keep `{placeholder}` templates; `fmt()` leaves unknown placeholders literal.


ozymandiashh (commented):
Verified on your branch in a scratch worktree: `tsc --noEmit` clean, `tests/i18n.test.ts` 18/18, and the touched suites (`overview`, `format`, `models`, `session-count-label`) 211/211 green.

Things I specifically hunted for, all handled:

- **The serve path.** `forceLocale('en')` on non-terminal `status` (main.ts:1125) and `export` (:1388) would have been insufficient on its own — the resident serve process bypasses commander actions, so a non-English `language` config could have leaked translated labels into menubar-json. The `process.argv[2] === 'serve'` guard (:2977) covers exactly that. Good catch.
- **MCP contract stability.** The localized-twin pattern (`formatSessionCountLocalized`, `unpricedModelHintLocalized`) keeps the originals untouched for MCP/JSON consumers, so no client sees translated labels.
- **Catalog discipline.** The exact-key-set and no-empty-strings tests pin the compile-time typing claim, and the hermetic LANG handling under vitest avoids locale-dependent suites.
- **CJK width.** Display-width padding with ANSI-escape immunity is tested; that's the part most i18n attempts get wrong in terminal tables.

Two non-blocking notes:

1. The PR is currently **conflicting with main** — needs a rebase before it can land.
2. One behavioral question for the maintainer, not a defect: `resolveLocale` reads `config.json`'s new `language` key synchronously and memoizes per process. The desktop app spawns the CLI per poll — that's fine — but the resident serve also resolves once; since serve forces English anyway this is moot today, just worth knowing if serve ever gains a localized surface.

iamtoruk:
Thank you for this, and for the care in the design. Keeping the machine-readable outputs (serve, menubar-json, export) in English is exactly the right call, and the CJK alignment work is nicely done.

We've decided not to localize the CLI. We've put translations into the desktop app and the menu bar, where reading your usage in your own language is the most meaningful, and both are free for everyone. The CLI's audience is developers, and committing to maintain six-language translations for every future terminal string is more than we want to carry on that surface.

So we're going to pass on this one. We really appreciate the effort and the thorough write-up. Anyone who wants CodeBurn in their language can use the desktop app or the menu bar.

PR #1463, opened 2026-09-16
fix(models): a Bedrock `-v1:0` version is not an Ollama tag

ulises-c:
## Summary

- An unpriced Bedrock model now reaches the Unpriced list instead of being classed as free local inference. Bedrock foundation-model ids end in `-v<major>:<minor>` (`anthropic.claude-haiku-4-5-20251001-v1:0`, what Claude Code writes with `CLAUDE_CODE_USE_BEDROCK=1`); `looksLikeLocalModel` read that colon as an Ollama `:tag`, so such a model was excluded from `models --unpriced`, the unpriced warning and the pricing-coverage denominator, with its metered spend at $0 and nothing flagging it.
- One line: exempt `/-v\d+:\d+$/` before the tag rule. Ollama / LM Studio tags are unaffected.
- Split out of #1448 at review, as suggested there. No route or display change; #1448 is back to draft pending #1450.

## Testing

- [x] `npm test` passes — `main` (3684f1d5): 4147 passed / 5 skipped; branch: 4148 passed / 5 skipped, +1, zero new failures. New test is red on `main`, green here.
- [x] `npm run build` succeeds; `npx tsc --noEmit` clean

Real data: on `main`, `isExpectedFreeModel('anthropic.claude-nonexistent-99-v1:0')` returns `true` and `findUnpricedModels` drops the row; with this change it returns `false` and the row is listed. Checked against the Bedrock ids on this machine's Claude Code and Hermes sessions (`anthropic.claude-haiku-4-5-20251001-v1:0`, `us.anthropic.claude-haiku-4-5-20251001-v1:0`, `openai.gpt-5.6-luna`) and the Ollama tags in the existing tests.


ulises-c:
@iamtoruk this is the `looksLikeLocalModel` split you offered to take on its own in #1448: `-v\d+:\d+$` exempted before the tag rule, no route machinery. One test, red on `main`, green here; suite 4148/4147.

Sequencing: this first, then #1448 (now the field-based rework you asked for) rebases onto it — the branches touch different regions of `models.ts` and merge clean, and #1448 does not duplicate these lines.

iamtoruk:
Checked: the test is red on main, green here, real Ollama tags still read as local, 4,148 passing, byte-identical unpriced list on my data.

One gap before I merge: the regex only catches `-v<n>:<n>`. OpenAI's Bedrock ids have no v. `openai.gpt-oss-120b-1:0` is the live id on the Bedrock model card, and `openai.gpt-5.6-luna-1:0` has the same shape; both still read as Ollama tags on this branch. Widen to `-v?\d+:\d+$` and add one of those to the test, then it goes in.

PR #1472, opened 2026-09-17
desktop: token breakdown popover behind dollar amounts

iamtoruk:
## Summary
- Hovering or focusing a dollar amount that has token counts behind it opens a small popover: Input, Output, Cache read, Cache write, and Calls when the source measured them.
- Amounts with no token counts (projected month, savings, retry tax, percentages, plans, by-project / by-PR) stay plain — no fabricated zeros.
- `periodTotals` now carries the four token counts so the Overview hero keeps a breakdown when its number comes from the generation window rather than the payload's own period. Chart buckets sum tokens so a weekly bar's popover matches the bar.

## Testing / Verification
- [x] `npm test` passes (`Usd.test.tsx`, `Overview.test.tsx`, `day-aggregator.test.ts`)
- [ ] Not a UI change

### Test plan
- [ ] Overview hero: hover the big dollar amount → four token rows (+ Calls). Tab to it, Escape closes.
- [ ] Yesterday / 7-day avg / Month to date: popover matches those windows, not the hero.
- [ ] Daily chart bar and Avg/day / Peak / Yesterday chips: same.
- [ ] Sessions and Models cost cells: hover works; grouped Models rows with no token columns stay without a popover.
- [ ] Projected month, savings, retry tax, percentages: no popover.
- [ ] Warm launch with an older payload missing token fields: dollar amounts still render, no popover.


ozymandiashh (approved):
Verified on your branch in a scratch worktree: full app suite green (73 files, 1119 passed + 4 skipped), `tsc --noEmit -p app/tsconfig.json` clean, and read the whole diff.

The parts I specifically checked and like:
- `tokensOf` failing closed to null on any missing count — an older payload shows nothing rather than a zero that reads as measured, and `sumTokens` propagates the same rule to `calls` (one row without a count makes the total unknown instead of smaller). That consistency is what makes the popover trustworthy.
- Hook order in `useUsdPop` is safe: every hook runs before the `!tokens` early return.
- Provider-filtered topModels rows: the payload's optional `cacheWriteTokens` means the popover honestly degrades to nothing when a row's legacy data lacked counts — consistent with the existing dash-not-zero column rules.
- The Overview test rewrite (amount as its own element) pins the new structure instead of loosening the assertion.

No blocking findings. One observation, not a change request: in provider-filtered mode the hero's tokens come from `periodTotals`' all-provider generation only when unscoped — the scoped hero will just have no popover, which matches the absence semantics, just worth knowing it's expected.

PR #1474, opened 2026-09-18
fix(app): pass --min-cost 0 so Models shows sub-cent and unpriced rows

### LINKED ISSUE #1465: Desktop Models table never shows rows under $0.01 or unpriced models
`codeburn models` defaults `minCost` to `0.01` (`src/main.ts:2574`), filtering `costUSD >= minCost || savingsUSD >= minCost`. `codeburn:getModels` (`app/electron/main.ts:573`) passes neither `--min-cost` nor `--unpriced`, so the desktop Models table silently drops every row below a cent.

On a real corpus that is 20 rows and 2,160 calls, including `coder-model` (518 calls, 6.6M input tokens) and `cursor-grok-4.6-high` (749 calls). Because the "unpriced" predicate in `Models.tsx` is `costUSD === 0 && savingsUSD === 0`, no unpriced row can ever survive the filter, so the dimming rules and the "add alias" affordance from #1443 are unreachable in the shipped app.

Fix is `--min-cost 0` in the getModels argv, but it changes the table for every user, so it needs before and after screenshots in both themes and a check that the row count stays sane on a large corpus. Found by ozymandiashh while working on #1443. Related: #1318.

ozymandiashh:
Fixes #1465.

`codeburn:getModels` now passes `--min-cost 0`, so the desktop Models table stops silently dropping every row below a cent — which by construction also excluded every unpriced row (a `0 >= 0.01` predicate), leaving #1443's dimming and add-alias affordances unreachable in the shipped app. The demo-harness argv mirror (`app/demo-bridge.mjs`) gets the same flag so browser demos match.

## Why `--min-cost 0` and not `--unpriced`

`--unpriced` filters to ONLY unpriced rows; the table wants the whole set with unpriced rows included. With `--min-cost 0` the predicate (`costUSD >= 0 || savingsUSD >= 0`) is a no-op — both columns are non-negative sums — so this is exactly "no filter". `opts.unpriced` behaviour, sorting, and `--top` (never passed) are untouched. `--min-cost` is declared in serve's flag allowlist (`src/serve.ts`), so the resident-serve path flows identically, and the in-flight coalescing key includes the full argv.

## Row-count sanity on a large corpus (as requested)

Real machine, lifetime period, all providers — CLI: `models --format json` default = **43 rows**; `--min-cost 0` = **53 rows** (+10 sub-cent/unpriced rows, the exact class #1465 names). No cap is needed: the table has no `--top` and 53 rows render fine.

## Before / after, both themes (as requested)

Captured over CDP against the real Electron app with the real CLI and real corpus, same protocol on both sides (report-snapshot memo wiped, Life period, Models section):

| | before | after |
|---|---|---|
| rows (light) | 43 | 53 |
| rows (dark) | 43 | 53 |

- Before, light: before-light
- Before, dark: before-dark
- After, light: after-light
- After, dark: after-dark

One display note: a row priced strictly between $0.00 and $0.01 renders as "$0.00" without dimming — it is genuinely priced, just below formatUsd's 2-decimal floor. The dim/`—`/add-alias treatment keys off `costUSD === 0 && savingsUSD === 0`, which now actually fires for unpriced rows.

## Tests

- All four `getModels` argv expectations in `app/electron/main.test.ts` updated; full app suite green (72 files, 1112 passed).
- `byTask` rows use the same unpriced predicate and already render observed usage for unpriced models (`ModelTaskRow`), so `--by-task` needs no renderer change.


iamtoruk:
Reviewed on my own corpus. Confirmed, and the reasoning about `--min-cost 0` over `--unpriced` is right: `--unpriced` filters to only the unpriced rows, so it would have replaced the table rather than completed it.

No falsy-zero problem on the CLI side. `models` guards the flag with `typeof opts.minCost === 'number' && Number.isFinite(opts.minCost)`, not a truthiness test, so an explicit 0 survives instead of falling back to the 0.01 default. Verified end to end rather than by reading: on a frozen range (so a live append could not move the totals mid-run), `models --format json` gives 41 rows, 168,141 calls, $22,697.010113, and the same call with `--min-cost 0` gives 62 rows, 170,314 calls, $22,697.018003. Exactly additive: +21 rows, +2,173 calls, +$0.007890, and every one of the 41 default rows comes back with identical calls and cost. 20 of the 21 recovered rows are unpriced (2,172 calls); the odd one out is a genuinely priced sub-cent row (Sonnet 4.5 on Pi, one call, $0.007890). `--by-task` behaves the same, 246 to 330 rows, and lands on the same 170,314 calls as the flat view. The resident `serve --stdio` path returns the identical 41 and 62, so the allowlist entry and the argv-keyed memo both hold.

In the running app, Models, all providers, Life, same window and protocol on both sides: 41 rows before, 62 after, matching the CLI exactly. Before shows 0 dimmed rows and 0 add-alias buttons, which is #1443's treatment being unreachable in the shipped app; after shows 20 dimmed rows with an em dash for cost and saved, and 20 add-alias buttons, with observed calls and tokens still rendering. The recovered rows sort last, so the screenshots are scrolled to the tail where the difference actually is.

On the numbers rule: the Models table does not sum to the Overview headline for Life ($22,754 and 171k calls against $38,444 and 278,530 calls). That gap is the known live-parse shortfall, it is present identically on the main build, and this PR narrows it by 2,173 calls rather than causing it.

Scope is as described. `getModels` has three callers: the Models section, the prefetch that fills the same memo key (so both sides stay consistent, which is what you want), and the once-a-day telemetry fallback that builds `topCategoryByModel`. That last one only does lookups for models already in the payload's `topModels`, so the emitted snapshot shape is unchanged; at most a top model that previously had no `topCategory` because its row was filtered away now gets one. The Overview models table and the menubar come from `status --format menubar-json` and are untouched.

app `tsc --noEmit` clean. App suite 73 files, 1,119 passed, 4 skipped under node 24, and root 313 files, 4,277 passed, both against the branch merged with current main. All CI green.

| | before | after |
|---|---|---|
| rows | 41 | 62 |
| dimmed / add alias | 0 | 20 |

Before, light: before-light
Before, dark: before-dark
After, light: after-light
After, dark: after-dark

PR #1476, opened 2026-09-18
fix(yield): rescue sessions whose branch shipped past the session's window

### LINKED ISSUE #1442: "abandoned" session label is wrong when commits merge to main outside the session's time window
## Bug description

The `codeburn yield` command labels a session as "abandoned" if it can't find commits from that session landing on the main branch. But it only looks for those commits inside a narrow time window tied to the session itself. If the actual merge to main happens later (or gets credited to a different, nearby session), `codeburn` calls the work "abandoned" even though it was fully shipped.

## What I observed

I ran a session that did real work on a feature branch. That branch ended up going through two pull requests, both fully merged into main. But `codeburn yield` still marked the session as "abandoned" — meaning it thinks none of that work ever made it to main.

The README explains that commits are matched to sessions by picking the tightest time window containing them. That's the problem — it's matching commits to sessions purely by time window, not by checking whether the branch/commits from that session ever actually reached main, whenever that happened.

## Expected behavior

If the commits from a session's branch eventually get merged into main — even if that merge happens after the session ended, or even if the time-window logic assigns credit to a different overlapping session — the original session should not be labeled "abandoned." The work was completed and shipped; it just didn't merge inside the same tight time window the tool is checking.

## Likely cause

This looks like a heuristic/methodology issue, not a crash or code bug. The classification logic assumes a session's commits must land on main within (or close to) that session's own time window. That assumption breaks for a very normal workflow: working on a feature branch across multiple sessions or multiple days, then merging later.

## Suggested fix

Instead of relying only on timestamp-window matching, consider also correlating by branch name or commit SHA ancestry. That is: check whether any commit made during a session's branch work later shows up in main's history at all, regardless of when that merge happened or which session's window it falls into. This would avoid falsely calling completed, merged work "abandoned" just because the merge timing didn't line up with the session window.

Happy to help test or provide more detail if useful.

ozymandiashh:
Fixes #1442.

`codeburn yield` labeled a session abandoned when no commit credited to it landed on main — but commits are credited by timestamp window, so two completely normal workflows read as dead work:

- a feature branch merged into main **after** the session's window (or after the period entirely), and
- a **GitHub squash merge**, which lands the work under a brand-new SHA the original SHAs never reach, so `inMain` is false forever.

## The fix: merged-branch evidence, on top of the window

When a session's logs record the git branch it ran on (per-turn `gitBranch` — Claude Code captures this; most providers don't, and those sessions are simply unchanged), yield checks whether that branch demonstrably reached main, **whenever that happened**:

1. **Ancestry** — the branch ref's content is entirely on main (`git branch --merged`). A fully merged branch and a ref parked inside main's history are the same topological object post-merge; both mean the state the session worked toward shipped.
2. **Tree equality** (the squash signature) — the branch carries commits of its own (`rev-list base..ref > 0`) and its tip tree matches a main commit's tree; the squashed tree is byte-identical to the branch tip. The tip must differ from the branch's base tree, so an untouched branch or one that reverted itself back to its base cannot match.

Scope guards, each pinned by a fixture test:
- abandoned **and** ambiguous sessions rescue to productive (an ambiguous session lost its window commits to a tighter window — if its branch shipped, the work landed);
- reverted is untouched (that verdict already describes shipped-then-undone work, and the rescue runs after categorization, which has already settled it);
- the main branch itself never rescues; only the exact name or the single-segment remote-tracking shape (`origin/<branch>`) matches, so an unrelated `topic/<branch>` cannot;
- **confined to `computeYield`**: sync attribution records (`computeAttributionRecords`) keep the pure timestamp-window methodology, and the JSON `methodology: "timestamp-window"` field is unchanged for compat (sync OTLP sends it; the desktop mirror type pins it). The CLI footer and README document the extension honestly.

## Evidence

- 7 fixture tests in `tests/yield-branch-rescue.test.ts` drive real temp git repos through the actual scenarios: squash-merge rescue (the reporter's case), true merge after the window, unmerged branch stays abandoned, main-branch exclusion, parked-ref semantics, a `master`-based repository (main-branch resolution path), and metadata-less sessions untouched.
- All existing yield suites still green (window overlap, repo grouping, JSON report, cache yields) — 17/17 across the four files; root `tsc --noEmit` clean.
- Real-corpus smoke on this machine (288 sessions): no category changes, because this corpus's sessions carry no branch metadata — the honest expected outcome; the rescue requires branch-bearing logs.

Known limits, stated rather than hidden: a branch deleted and pruned after merge leaves no ref to examine (no rescue); a squash-merged commit later reverted out is beyond what tree evidence can see. Both are documented in the tests/comments.


iamtoruk (changes requested):
Thanks for the careful writeup — the squash-merge half of this is a real gap and the tree-equality rule is the right shape for it. But the ancestry half misattributes badly on a real corpus, so I can't take it as it stands.

Your smoke run found no category changes because that corpus carries no branch metadata. Mine does. `codeburn yield` on 9,503 sessions, before → after:

Lifetime
- productive: $7,177.48 / 344 sessions → $14,147.11 / 944 sessions
- abandoned: $12,033.88 / 5,336 → $5,843.00 / 4,772
- ambiguous: $3,555.17 / 3,823 → $2,776.60 / 3,787
- 600 sessions rescued, $6,969.63 moved

30 days
- abandoned: $6,291.95 / 2,673 sessions → $602.11 / 2,228 — 90% of abandoned spend disappears
- productive: $4,815.90 / 197 → $10,647.29 / 659

## The problem: 567 of the 600 rescues, $6,705.84 of the $6,970, come from one branch

`exp/extraction-arms` in one repo. Its ref:

- tip `973611ed`, committed 2026-08-15
- `merge-base` with main IS the tip, so 0 commits of its own
- the tip is a plain commit on main's first-parent line

It is a branch that was created (or left) pointing at an old main commit and never advanced. `git branch --merged` lists it, so the ancestry rule fires, so every session that ever recorded `gitBranch: exp/extraction-arms` becomes productive — forever, and regardless of when it ran.

All 567 of those sessions started *after* the branch's last commit, spanning 2026-08-17 to 2026-09-17. That repo's main tip is `12d52676`, 2026-08-19 — main has not moved in a month. Five by cost:

| session | ran | cost | verdict |
|---|---|---|---|
| 56dea73b | 2026-08-29 → 09-02 | $1,584.46 | abandoned → productive |
| 8ce4b74e | 2026-09-02 → 09-17 | $893.49 | abandoned → productive |
| a2d08318 | 2026-08-17 → 08-19 | $458.54 | ambiguous → productive |
| 32e82536 | 2026-08-22 → 08-24 | $407.68 | abandoned → productive |
| 8dddb2e2 | 2026-08-19 → 08-22 | $395.49 | abandoned → productive |

Four of the five ran entirely after main stopped moving. Nothing they did could have shipped. $3,739.66 of dead work reported as shipped, from those five alone.

A smaller instance of the same thing in this repo: `main-baseline`, 0 own commits, tip `c9ad778c` on main's mainline — a ref someone parked for a before/after comparison. It rescues too.

## Why the guard doesn't hold

`sessionBranchShipped` guards the squash rule with `refOwnCommits` and `tree !== baseTree`, and those guards are good. The ancestry rule has neither, and the comment reasons the guard away:

> a merged branch and a ref parked inside main's history are the same topological object

Topologically yes. Semantically no. A merged branch contributed commits that are in main; a never-advanced branch contributed nothing and its tip is just an old main commit. Both have an empty `base..ref`, which is exactly why `refOwnCommits` can't separate them — but "indistinguishable by this metric" is not "equivalent".

`tests/yield-branch-rescue.test.ts:147` pins this as intended: it creates `feature-parked` from main with no commits and asserts a rescue. That test asserts the bug.

## The second issue, which survives even a perfect branch signal

The rule is per-branch, not per-session. One branch that shipped makes *every* session that ever touched it productive: the one that did the work, the one that opened the branch and did nothing, and the ten that ran on it months later. There is nothing tying a rescued session to the shipped work — no overlap between the session's window and the branch's commits, no check that the session predates the merge, nothing. The report even keeps `commitCount: 0` on the rescued rows, so yield ends up claiming 944 productive sessions of which 600 have no commit behind them.

That's the part I'd want settled before the mechanism lands at all. Some evidence the session actually contributed — its turns overlapping commits that are on the branch, or at minimum the session ending before the branch's last commit — would cut both this and most of the parked-ref damage.

## Smaller things

- `buildMergedBranchIndex` takes the first matching candidate ref and `break`s. If a local ref exists but doesn't qualify, the remote-tracking ref is never examined, so a local stale copy of a branch whose `origin/` ref did merge silently loses the rescue. Collect the candidates instead of breaking.
- `mainBranch` is skipped twice, once in the builder and once in `sessionBranchShipped`. One of them is dead.
- `git log <main> --format=%T` walks all of main's history on every repo group that has any branch metadata. Fine here, worth knowing on a large monorepo.
- Lifetime total drifts $22,766.53 → $22,766.71. Float accumulation order, harmless, but it means the total isn't bit-stable across this change.

## What I'd keep

The tree-equality rule with its two guards is sound and it's the half that actually fixes #1442 — the reporter's case is a squash merge. If you dropped the ancestry rule entirely and shipped only tree equality plus a per-session contribution check, I think that lands.

Not a blocker, for the record: 6 cursor sessions flip abandoned → ambiguous between runs. I reproduced that main-against-main, so it's pre-existing, not yours.

Infra: `tsc --noEmit` clean, the 7 new tests and the existing yield suites pass on your branch.

ozymandiashh:
Reworked exactly on your line: ancestry is gone, and what remains is tree equality plus the per-session cut you asked for.

**The rule is now three conditions, all required:**
1. the branch carries commits of its own (`base..ref` non-empty, tip tree ≠ merge-base tree),
2. its tip's tree hash matches a commit on main — the squash signature, the half you called sound,
3. **at least one of those own commits was made inside the session's window** — the session contributed the work, rather than having run on a branch that later shipped.

Condition 3 is the one that kills both failure classes from your corpus:

- `exp/extraction-arms` (0 own commits, tip = old main commit): fails 1 and 3 — no rescue, ever. The `feature-parked` test now asserts **no** rescue; you were right that it asserted the bug, and the "topologically identical" comment was me reasoning myself into the wrong semantics.
- The 30-day `8ce4b74e`-style sessions that ran Aug–Sep on a branch whose last commit was Aug 15: condition 3 fails — their windows contain no branch commit. Even for a genuinely shipped branch, only the session(s) whose window overlapped the work flip, which is the "tying the rescued session to the shipped work" you wanted.

On your read of the reporter's case: the squash half is what their workflow needs, and it's unchanged in substance from the version you reviewed (same two guards).

The smaller findings, all fixed:
- **Candidate collection**: every matching ref becomes evidence instead of first-wins, with a fixture that parks the local ref at the branch point while `origin/feature-r` holds the merged work — the rescue fires through the remote-tracking ref.
- **Doubled `mainBranch` skip**: the builder's copy is gone; `sessionBranchShipped` keeps the single check.
- **`git log main --format=%T` cost**: still per repo-group-with-branch-metadata, lazily. If monorepo cost ever matters, capping the walk at the branch's oldest commit date would bound it — I left it simple until someone has the monorepo.
- **Float drift ($0.18 on $22,766)**: acknowledged, not addressed — the accumulation order changed because 600 fewer sessions pass through the same additions, and forcing bit-stability would mean sorting sums for no reader's benefit. Say the word if you'd rather it stable.

One honest limitation, stated rather than hidden: a session whose work was committed to the branch *between* sessions (work done in session A, `git commit` made after A's window closed, e.g. committed next morning) still reads abandoned — the window-overlap cut cannot see it, and loosening it (commit-time grace, author-vs-committer dates) reopens exactly the attribution you just closed. The reporter's described case (work and commits inside the session, squash merge later) is covered.

The PR body still describes the old two-rule shape — I'll leave the thread as the record unless you want the body rewritten. 19/19 across the yield suites (7 fixture tests now: squash, parked-ref-no-rescue, late-session-no-rescue, window-overlap true-merge, multi-candidate refs, master repo, no-metadata), root `tsc` clean, rebased on current main.

PR #1484, opened 2026-09-18
feat(providers): add WorkBuddy + proxy-farm providers and Devin sessions.db support

MiloMMIN:
## Summary

- **New provider: WorkBuddy / WorkBuddy AI** — parses `.workbuddy`/`.workbuddy-ai` project stores (sessions + usage), with legacy `codebuddy` path support, model aliases for its house models (hy3/hy3-free observed pricing), and a guard against abnormal Hermes estimated costs.
- **New provider: antigravity-tools** — reads `user_tokens.db` + `proxy_logs.db` so traffic relayed through a local Antigravity Tools instance is attributed to the proxy side. `CODEBURN_ANTIGRAVITY_TOOLS_TOKENS_ONLY` counts only token-authenticated remote traffic; `CODEBURN_ANTIGRAVITY_TOOLS_LOCAL_IPS` excludes self-owned source IPs — both keep local sessions from being double-counted.
- **New provider: cc-switch** — reads `cc-switch.db`, counting only synced sessions whose session file lives outside the local home dir.
- **New provider: new-api** — reads the new-api/ApiProxyFarm `one-api.db`, attributing usage per API token; `CODEBURN_NEWAPI_LOCAL_TOKENS`/`CODEBURN_NEWAPI_LOCAL_USERS` exclude self-owned tokens/users so local calls aren't counted twice.
- **Devin provider: sessions.db support** — current CLI builds keep the whole message tree in `sessions.db` (`message_nodes.chat_message`, no `transcripts/` dir). Each assistant node's `metadata.metrics` prices the call via `calculateCost`, deduplicated by `request_id` (nodes repeat across branches), with SQL-level `dateRange` filtering and transcript-over-DB precedence. Windows `%APPDATA%\devin\cli` is probed first with the Linux XDG path as fallback.
- **Daily cache v34** — this branch's v33 (Hermes guard + WorkBuddy) collided with upstream's billing-route v33; one number can't carry two accountings, so the merge rung bumps to 34 and contracts the Hermes re-derive at 34 so branch-v33 caches still get the route-keyed re-derivation.
- **CI fixes** — drop an unsupported `color` field from the WorkBuddy Provider object (TS2353), and remove stale DeepSeek v4 `BUILTIN_PRICE_OVERRIDES` that shadowed the bundled snapshot's official prices (broke 5 pricing tests).

#### Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npx vitest run tests/providers/` + registry/env suites — 954 tests pass
- [x] New tests: 8 antigravity-tools, 4 cc-switch, 7 new-api, 10 workbuddy, +5 devin sessions.db cases (dedup, hidden sessions, transcript precedence, dateRange)
- [x] GitHub Actions Tests workflow green across Node 22/22.13/24/26, ubuntu/windows/macos, desktop

Generated with [Devin](https://devin.ai)


ozymandiashh (changes requested):
Reviewed against the repo's own parsers. The provider registration, discovery and dedupe design are careful and follow house conventions, but three things block this — two of them move money.

**Blocker 1 — WorkBuddy double-bills cache reads.** `calculateCost` bills `inputTokens` at the full input rate *plus* `cacheReadTokens` at the cache-read rate, so it expects input to exclude cached tokens. `workbuddy.ts` passes `inputTokens: promptTokens` and `calculateCost(model, promptTokens, …, cacheReadTokens, 0)` with the raw prompt total — and WorkBuddy's prompt total is cache-inclusive, as the PR's own fixture shows (`prompt_tokens: 100 = prompt_cache_hit_tokens: 30 + prompt_cache_miss_tokens: 70`). The house convention is the devin transcript parser: `inputTokens: Math.max(0, prompt_tokens - cacheReadInputTokens)` with a comment saying exactly this (`src/providers/devin.ts` on main). The tests currently lock in the wrong numbers. Fix: subtract cache hits from input and update the expectations.

**Blocker 2 — same subset question on the proxy-farm providers.** `antigravity-tools.ts` passes `input_tokens` alongside `cached_tokens` (fixture row 83255/79900 is subset-shaped), `new-api.ts` passes `prompt_tokens` plus `other.cache_tokens`, and the devin `sessions.db` parser passes `metrics.input_tokens` + `cache_read_tokens` directly — while the transcript path of that same provider subtracts. If those sources follow OpenAI cached-subset semantics, they're double-billed the same way. cc-switch's separate `cache_read_tokens`/`cache_creation_tokens` columns look Anthropic-style (exclusive) and are probably fine. Please confirm each source's semantics or subtract.

**Blocker 3 — hermes accepts negative estimates.** The `costStatus === 'estimated'` branch guards `estimated_cost_usd <= 1000` with no lower bound, so a negative estimate is returned as `costUSD` and subtracts from aggregate totals. The legacy fallback branch right below correctly requires `> 0`; `calculateCost` clamps negatives for the same reason. Add the lower bound here too.

CI note: `test (22)` failed on the cache-lock serial-suite 5-minute timeout, not an assertion (main suite 4316/4316 passed; main hit the identical timeout on run 35317649144) — looks like the known flake, but it needs a green re-run before merge.

Non-blocking: the new `SHORT_NAMES` entries (`deepseek-v4.1-flash`, `kimi-k3-1`) are unreachable because `BUILTIN_ALIASES` resolves first; `cachedInputTokens: 0` where the convention mirrors `cacheReadInputTokens` into it; `AgentTabStrip` colors added only for workbuddy/workbuddyai; the new sqlite test files import `node:sqlite` without the `skipUnlessSqlite` guard devin.test.ts uses.

iamtoruk:
Thanks Milo, there is good work in here. We can't take it as one PR though. The antigravity-tools and new-api providers count usage from other people going through a shared quota or relay, and CodeBurn only tracks your own usage, so we won't add those. cc-switch syncs from other machines, same reason. The Hermes cap and the WorkBuddy price guesses change totals, so they'd need official sources.

The Devin sessions.db support is a real gap and we'd love it on its own: a PR with just devin.ts and its tests, rebased on main, plus a redacted sample row from a real sessions.db. If you want WorkBuddy in, send it as its own PR with linked sources for the model prices. Closing this one, thanks again!

PR #1485, opened 2026-09-18
desktop: light theme default, refreshed type and spacing, plugin and plan cards

iamtoruk:
## Type and spacing

A token pass one step up. Body text is 14, titles 20, sidebar rows 36, page padding 32. Sidebar shortcut chips move to hover tooltips. Surfaces run on a three-step ramp, and primary text is near-white or near-black. Fresh installs now default to the light theme. An explicit theme choice is honored.

## Plugins page

The Menu bar and Teams plugins are now cards side by side, each with embossed background art. The Menu bar card shows the Capacity Dock switch and Settings, Quit and Uninstall as icon buttons with tooltips. An info button opens a two-column modal that describes the menu bar and the Capacity Dock with the real supported-provider list. The card can install, update, open, quit and uninstall the menu bar app, versioned so an older menu bar shows Update. A Settings action opens the menu bar app's own settings through a remote-command key.

## Plans page

Redesigned as provider cards in a responsive grid, each with its own logo embossed. Connection status is now stable. A single empty or slow poll no longer flips a connected provider to disconnected. A connected provider keeps its quota bars through a transient refresh miss. A stuck "waiting" resolves to an actionable state after 20 seconds. A login-expired error now offers Connect, matching the menu bar. The Gemini expired-login case was previously stuck on "waiting" and is now a terminal, actionable state like the menu bar.

## Verified

CLI report, models and menubar-json numbers are identical to the cent before and after. Only the Gemini quota error string changes, no figure moves. Root tests, lock tests and the Swift menu bar suite pass.


ozymandiashh (commented):
Verified the "numbers identical" claim holds under inspection — the only src/ file is the menubar-installer fallback (documented, tested), the quota hunks add only a `connectable` flag and the Gemini footer string, and no figure, total or aggregation moves anywhere. The test coverage is genuinely strong. One behavioral gap and a few nits from reading the Plans state machine:

**The theme default is not scoped to fresh installs, and the changelog doesn't say.** `App.tsx` initializes `saved === 'light' | 'dark'` → honored, `'system'` → OS-follow, **else → `data-theme="light"`** — and "else" includes `null`, which on main meant OS-follow (attribute removed, `@media (prefers-color-scheme: dark)` applied). There's no sentinel distinguishing a fresh install from an existing no-choice install, so an existing user on a dark-mode Mac who never touched the setting was seeing dark and flips to light on update. The in-code comment ("only a fresh install with no choice at all defaults to light") and the PR body both describe a restriction the implementation can't deliver. Same conditional in Settings.tsx. The changelog gains three meticulous entries but none mentions the default change — the least-disclosed user-visible change in the PR. Either accepting the flip with a changelog entry that says so, or a one-time sentinel honoring the old OS-follow for pre-existing no-choice installs, would close it.

Plans nits, all in `stabilizeQuota`:

- On escalation, `known` stores the raw `transientFailure` as `last`, not the synthetic terminal row — if the next poll omits the provider entirely (the empty/slow-serve case the trailing loop handles), the retained entry renders "Waiting on the CLI…" again and re-escalates on the following poll. Storing the synthetic row as `last` would stop the terminal↔waiting flap in exactly the scenario the PR set out to stabilize. Not covered by the six new tests.
- A provider that was connected once and then misses forever keeps its frozen bars indefinitely with no staleness marker — the 20s cap only applies to never-connected providers (`waitingSince` is never set for previously-connected ones).
- The "20 seconds" is evaluated only when a new raw array arrives (the steady poll is 30–60s per the code's own comment), so real resolution is ~30–60s.
- `known` never evicts, so a provider uninstalled while the page stays open keeps rendering until remount.

None of these block on their own; the theme gap is the one I'd resolve before merge.

PR #1497, opened 2026-09-19
Add support for multiple Codex, Grok and Hermes data directories

Kaotic:
## Summary

Users with multiple Codex, Grok or Hermes data directories can now  include their sessions in a single Codeburn report by setting `CODEX_HOMES`, `GROK_HOMES` or `HERMES_HOMES`.

- Accept platform-delimited paths (`:` on POSIX, `;` on Windows), expand `~`, and normalize paths.
- Preserve existing single-home fallbacks and explicit factory overrides.
- Deduplicate sessions across homes and continue scanning when a directory is missing or inaccessible.
- Include all configured homes in provider probes and Codex context discovery, and invalidate the cache when home configuration changes.
- Update provider documentation and add tests.

## Testing / Verification

- [X] I have tested this locally against real data (not just unit tests)
- [X] `npm test` passes
- [X] `npm run build` succeeds
- [x] Not a UI change

### For new providers only:


iamtoruk:
Reopening this. Multiple data-home support for Codex, Grok and Hermes is a genuinely useful addition, and the cross-home dedup and cache invalidation are the right instincts. Heads up: #1062 just merged, which adds WSL-distro provider-root discovery on the same providers, so this will need a rebase onto current main and we'll want to confirm the two multi-root paths compose without double-counting. I'll give it a full review.

iamtoruk:
Thanks for this, and for the care in it. The cross-home dedup and the test coverage are genuinely solid.

After thinking it through, we're going to pass on this for now. We'd rather CodeBurn find data folders automatically. The default folder and Linux/WSL folders already are, and a manual list of custom folders via environment variables isn't the direction we want to add right now, since it only helps command-line users and most people never have data outside the standard locations.

Really appreciate the effort and the quality. If we revisit multi-home support, this is a great starting point to build from.

PR #1502, opened 2026-09-19
fix: bound cache-load memory for ranged queries

iliaal:
Fixes #1504.

Fixes the `codeburn overview -p today` heap crash at the cache-I/O layer.

## What was wrong

Ranged loads assembled whole multi-hundred-MB month shards before filtering (`loadShardMemoized` whole-parsed via JSON.parse, including a 578MB shard that exceeds V8's max string length); the pre-lock snapshot overlapped the canonical reload (~787MB duplicate); lock publishes rewrote the entire result file; codex discovery re-decoded the result cache per directory; retained strings were tokenizer slices pinning whole input chunks (~1.2GB phantom heap on this corpus).

## What changed

- Session month shards decode incrementally, one turn at a time, keeping only in-range turns (same kept/dropped/carry contract).
- Codex result cache streams instead of whole-parsing; discovery labels served from calls-free metadata; publishes merge instead of full rewrite.
- Shard loads run serially; pre-lock snapshot released before reload.
- Retained strings detached from tokenizer buffers; dropped dedup keys are 128-bit digests; provider-scoped queries skip unrelated sections.
- Codex single-flight loads retry on publish-generation change; clear path bumps a global epoch so in-flight loads cannot repopulate a cleared memo.

## Verification

- `npm test`: full suite green; `npm run test:locks`: 36/36; `tsc --noEmit` clean.
- Load-phase retention measured 1.46GB -> 232MB on the reporter corpus.
- New suites: range-filter (incl. 30k-turn record test), shard-stream, codex-range-filter, concurrent-publish gate, digest/marker tests.

Stacked follow-up (aggregate parse mode for the bare-today headline) builds on this branch.

PR #1503, opened 2026-09-19
fix: parse bare-today headline in aggregate mode

### LINKED ISSUE #1505: overview -p today still OOMs at 512MB after cache-load bounds (parse retention)
## Problem

Follow-up to #1504. After the cache-load bounds (load retention 1.46GB -> 232MB), `codeburn overview -p today` still exits 134 at a 512MB heap cap after ~5.5 minutes. Heap-floor bisection puts the true requirement between 512MB and 1024MB (a 1GB run survives with heavy GC thrash).

Repro: same command/corpus as #1504.

## Root cause (parse layer)

The same-day parse retains every provider's full turns co-resident through merge, day aggregation, and rendering: `ParsedApiCall` graphs (~600B serialized per call with tool sequences, dedup keys, per-call paths, name arrays) across ~1.2M calls. Per-field census on the reporter corpus: `usage` 114MB, `toolSequence` 105MB, `deduplicationKey` 50MB, per-call paths ~18MB (serialized, ~half corpus; heap is a multiple). Stripping alone is not sufficient while full sessions co-reside, so peak must track one provider instead of the corpus.

## Fix

PR #1503 (stacked on #1502, branch `fix/overview-aggregate-mode`):

- New opt-in `stripForAggregate` parse mode: per-call payloads the headline never reads are stripped per provider right after its summaries resolve (in place, single-owner); shell commands are extracted first (`call.commands`) for PR launch matching; `userMessage` stays (bounded, needed by prompts/corrections).
- Engages only for the narrow shape needing no session payloads (today-only, all-provider, no project filters or day selection); every other caller keeps full sessions.
- Shared parse memo bypassed in both directions; single-pass scope skipped, so stripped graphs never mix with full-session consumers.

## Verification

- `overview -p today` at 512MB: two consecutive exits 0 (was 134).
- New lite-parity suite proves byte-identical rendered output and equal totals/days/models/categories on an isolated corpus; derivation-equivalence test covers the PR command extraction.
- Uncapped lite vs full on the live corpus: identical row structure and mix-consistent totals.
- Full suite + locks green, tsc clean.

iliaal:
Fixes #1505 (stacked on #1509; restacked onto the split stack).

Stacks on #1506 (review that first; this diff includes it until it merges). Fixes the remaining `codeburn overview -p today` heap crash at the parse layer.

## What was wrong

After the load-phase bounds from #1502, the same-day parse still retained ~1-2GB: every provider's full turns co-resident through merge, day aggregation, and rendering. Capped at 512MB the command still exited 134; the heap floor measured between 512MB and 1024MB.

## What changed

- New opt-in `stripForAggregate` parse mode: per-call payloads the headline never reads (tool sequences, dedup keys, per-call paths, bash/skill/mcp/subagent arrays) are stripped per session inside `buildSessionSummary` the moment it resolves (in place, single-owner — the shared memo is bypassed both ways, so peak never holds more than one session's full turns).
- Shell commands are extracted first (`call.commands`); PR launch matching reads them through a derivation-equivalence-tested helper. userMessage is emptied and replaced with a precomputed 160-char `promptPrefix` (the full text averaged ~115B/turn but totaled ~40MB serialized / ~100MB heap); the matcher reads the prefix first with the userMessage fallback, so attribution is identical.
- The lock reload is skipped when the envelope nonce is unchanged: every `saveCache` mints a fresh nonce, so an unchanged nonce proves byte-identical shards and re-streaming them only doubled peak heap for zero new data. Any publication, torn read, or missing envelope keeps the reload (pinned by `parser-cache-refresh-reload-skip.test.ts`).
- Engages only for the narrow shape that needs no session payloads (today-only, all-provider, no project filters or day selection); every other caller keeps full sessions.

The detach-only-retained-strings and size-gate commits now live down the stack in #1506, so no step here regresses on its own.

## Verification

- `overview -p today` completes with real data (was: `FATAL ERROR: JavaScript heap out of memory`, exit 134); lite-parity suite proves byte-identical rendered output and equal totals/days/models/categories on an isolated corpus.
- Uncapped lite vs full on the live corpus: identical row structure, mix-consistent (frozen claude rows, growing omp rows).
- Full suite + locks green, tsc clean.

## Timings (warm cache, medians of 3, 1.2GB shard corpus with one 697MB shard)

| command | main | #1506 (base of stack) | this PR (stack top) |
|---|---|---|---|
| `overview -p today` | 13.8s / 3.1GB | 17.6s / 2.7GB | 26.5s / 1.8GB |
| `report --period today --format json` | 13.2s / 2.5GB | 17.3s / 2.6GB | 26.6s / 1.6GB |

The strip trades ~9s for ~1GB less peak on this corpus (the lite sessions replace full turns through merge/aggregate/render). `report` takes the full-session path and still benefits from the gate plus the skipped reload below it.

## Notes for review

- The strip is a denylist (`toolSequence`/spawn ids/paths/name arrays/dedup key out, everything else spread through), so future `ParsedApiCall` fields stay populated.
- Single gate: `main.ts` requests lite unconditionally for `overview`; `buildDurablePeriod` is the only decider (today-only, all-provider, no filters/selection).
- `serve` dispatches `overview` in-process, so the strip also runs in the resident desktop/menubar process, not only one-shot CLI runs. Verified: a lite `overview -p today` followed by a full `report --period today --format json` in one `serve --stdio` process returns identical full data (memo bypass is per-parse, not process-global).


iamtoruk:
The parse-layer work is good and the numbers are clean.

On a frozen copy of a real corpus I diffed the full JSON export leaf by leaf against main for today, week, 30days, month and all: zero differences from this PR. The single `all`-period difference (`shellCommands` "ls" 6300 vs 6299) is the branch predating main's antigravity `extractBashCommands` change, and it disappears with that one line applied. No contamination either: a cache directory that served three lite `overview -p today` runs produces identical full reports afterwards, and inside one `serve --stdio` process a lite `overview -p today` followed by a full `report --period today --format json` returns identical full data. Suite green (4336 passed), locks 36/36, tsc clean.

The memory result is real: `overview -p today` runs at a 128MB heap cap where main needs more than 256MB.

Before merge:

1. Blocked on #1506, which does not compile on its own. See the review there.
2. Please replace the allowlist in `stripCallForAggregate` with a denylist. It is complete today, but it is 31 hand-copied fields, and the next optional field added to `ParsedApiCall` becomes `undefined` on the today headline with no compile error. Destructure out `toolSequence` and `spawnToolUseIds`, spread the rest, then override the emptied arrays. Shorter, and it cannot silently drop a money field.
3. Say in the PR body that `serve` dispatches `overview` in-process, so the strip runs in the resident desktop and menu bar process, not only in one-shot CLI runs. It behaves correctly there (tested), but the PR reads as CLI-only.
4. The `narrowToday` gate is computed in `main.ts` and re-derived in `buildDurablePeriod`. One is enough.
5. Timing: warm `overview -p today` goes from 2s to 14s here. Part is #1506's streaming, part is the memo bypass making today's sources parse twice (the daily backfill parses them in full, then the lite pass cannot reuse it). Please say which is which, with numbers.
6. Same rebase and indentation points as #1506. Please rebase after 0.9.25 lands.

iamtoruk:
The memory result is real and the output is exact. Thank you.

`overview -p today` goes from 1006MB to 383MB max RSS and `report --period today` from 1007MB to 576MB. The overview text is byte-identical to main for today, week, month, all and today with `--provider claude`, including the shape that engages the strip. The full JSON leaf diff against #1509 is identical on every period, cold and warm.

I re-checked the three hazards. In one resident `serve --stdio`: overview today, then the full today report, then overview again, then the full week report. Both full reports are leaf-identical to main's standalone exports. A cache that served three lite overview runs then produced a full `all` report identical to a full-only cache. No `rangeFiltered` or `droppedKeys` marker reaches any shard on disk. All pass. The denylist is right, and letting `buildDurablePeriod` be the single decider is cleaner than what I suggested.

Still open:

1. The PR body says "userMessage stays (bounded, needed by prompts/corrections)", but commit 8ca6235c empties it and replaces it with a 160-char `promptPrefix` (`src/parser.ts:1795`). The behaviour is correct; the body says the opposite of what the code does.
2. `src/parser.ts:5688` ends mid-sentence again ("Never a correctness gate: on any"). The closing line is present on main, #1506, #1508 and #1509; this PR's own diff deletes it.
3. The timing split. What I measure, warm, medians of 3, `report today`: main 1.83s, #1506 21.19s, #1508 23.08s, #1509 23.14s, this PR 13.37s. `overview today`: main 1.13s, #1506 10.94s, this PR 8.21s. So nearly all the cost is #1506's streaming and this PR gives about 8 seconds back through 1b301772. Even at the top of the stack it is still 7x main on the path the desktop app polls, which we cannot take. The size gate I asked for on #1506 is the fix, and 1b301772 belongs down there.

Conflicts for the rebase. Against main: `CHANGELOG.md`, `src/codex-cache.ts`, `src/parser.ts`, `src/providers/types.ts`, `src/session-cache.ts`. Against 0.9.25 you also pick up `src/providers/warp.ts` and `src/usage-aggregator.ts`, and the usage-aggregator one is semantic: 0.9.25 changes provider exclusion through `buildDurablePeriod`, right where your `stripForAggregate` gate goes. Please rebase after 0.9.25 lands. Nothing in the stack bumps a cache version or changes a cache filename, so that part is safe.

iliaal:
Body fixed: userMessage is emptied, matcher reads the precomputed 160-char promptPrefix first with fallback. The dropped comment closing line is restored (folded into the commit that dropped it). 1b301772 is out of this stack (lives in 1506 now). Timings in body: strip trades about 9s for about 1GB less peak. Full suite plus locks green. Rebase after 0.9.25; noted the warp and usage-aggregator semantic conflicts for then.

iamtoruk:

PR #1508, opened 2026-09-20
fix: stream the codex result cache instead of whole-parsing

iliaal:
Stacks on #1506 (review that first; this diff includes it until it merges). Split out of that PR per review: the codex-cache rewrite rides separately from the session-shard work.

## What changed

- The single result file decodes entry by entry (a top-level decode assembled hundreds of megabytes first; one shard exceeded V8's max string length).
- Discovery labels come from calls-free metadata without loading the calls map.
- Publishes merge dirty entries over the published bytes instead of rewriting from memory.
- Single-flight loads are shared across concurrent readers and retry when a publish lands mid-decode (deterministic gated-stream test); a global clear epoch stops in-flight loads repopulating a cleared memo.

## Verification

- tsc clean; codex suites green (range-filter, concurrent-load, directory-switch).
- No digest machinery in this PR: markers/sets stay raw throughout.


iamtoruk:
Thank you. tsc is clean on this head, the 15 touched tests pass, and the leaf diff against #1506 is identical on every period, cold and warm.

Same blocker as #1506, so this one waits with it. Warm medians, this head vs main: `report today` 23.08s vs 1.83s, `report all` 29.53s vs 3.97s, `overview today` 12.75s vs 1.13s. In the profile stream-json and stream-chain together are 41% of wall time, while plain `JSON.parse` handles the whole 166MB shard set in 333ms (the codex result cache here is 59MB).

The same size gate fixes it: parse normally under a threshold, stream only the oversized file. Nothing else is open on this PR. Once the gate is settled on #1506 this follows straight after.

Conflicts against main: `CHANGELOG.md`, `src/codex-cache.ts`, `src/parser.ts`, `src/providers/types.ts`, `src/session-cache.ts`. The codex-cache one is semantic: main's `evictCachedCodexResults` needs rewriting against your overlay model. Please rebase after 0.9.25.

iliaal:
Follows the #1506 gate: same threshold, timestamp scan reuses the exported `retainCodexEntry` per entry so the paths cannot drift, entry load applies the same retain rule. Parity test covers scan + load on nasty shapes. The mid-flush and concurrent tests pin the stream path explicitly. tsc + full suite green. Rebase with the stack after 0.9.25.

iamtoruk:

PR #1509, opened 2026-09-20
fix: hash cross-file dedup keys with SHA-256 instead of raw strings

iliaal:
Stacks on #1508 (review that first; this diff includes it until it merges). Split out of #1506 per review: the digest wiring rides separately from the session-shard work.

## What changed

- Dropped-key markers and shared dedup sets store full SHA-256 digests (32 one-byte chars, same footprint as before) instead of ~140MB of raw key characters.
- `DedupSet` is a plain class, not a `Set` subclass: iteration yields digests, so unadapted consumers fail at compile time. The provider `createSessionParser` contract takes `DedupSet` throughout.
- Antigravity RPC conversations ride along as paired bare keys (prefix matching cannot work on digests); the key shape lives in the provider-owned `providers/antigravity-keys.ts`.
- Worker results cross threads as digests with explicit hashed-membership checks (no double-hashing).

Measured, not assumed: `createHash` costs the same wall time as the previous inline mixer at ~1M keys, so no weaker bound is needed.

## Verification

- tsc clean; digest/marker/twin/worker suites green; memory delta (~140MB heap on the reporter corpus) measured by field census, not inferred.


iamtoruk:
This one is ready. Thank you for pulling it out.

I verified the digest end to end. It is a full 256-bit SHA-256, not truncated: 32 chars, max code point 243, so a one-byte string at half the footprint of hex. It round-trips through `JSON.stringify` and a UTF-8 file byte for byte.

I checked every place a digest could meet a raw key. `addHashed` and `hasHashed` are only fed by `DedupSet.values()` or `marker.droppedKeys`, both already digests. `add` and `has` are only fed raw keys. `marker.conversations` holds raw bare keys and goes through `add`, which hashes, and `rpcConversationBareKey` returns null for an already-bare key, so nothing re-pairs.

For the no-double-hashing claim I did not want the audit alone, so I ran a cold parse of a real corpus twice, once with workers (27 files through threads) and once with `CODEBURN_PARSE_WORKERS=0`. The `all` reports are leaf-identical. A double hash or a missing hash would have double counted across that boundary.

Leaf diff against #1508 is identical on all five periods, cold and warm. tsc clean, 70 touched tests pass. Cost is 0 to 5% wall.

On the 52 files: that is the right amount of churn and I am not asking you to reduce it. The plain class is what forces each provider signature to change, which is the compile-time safety we wanted.

Two small things I am happy to patch myself: `DedupSet.fromDigests` and `DedupSet.delete` have no call sites in src or tests. And one note, not a blocker: `rpcConversationBareKey` moved to `providers/antigravity-keys.ts`, but `session-cache.ts` still imports it and `DedupSet.add` calls it on every insert, so the generic primitive still knows one provider's key shape.

Blocked only by its parents. Please rebase after 0.9.25.

iliaal:
`fromDigests` and `delete` removed (folded into the digest commit — it never lands dead). Nothing else open. Rebase after 0.9.25.

iamtoruk:

PR #1510, opened 2026-09-20
fix(providers): accept the {"projects":[...]} Crush registry shape

ptstory:
## Summary

- `loadRegistry` doesn't recognize Crush's `{"projects": [...]}` registry wrapper, so a standard upstream Crush registry resolves to zero projects and reports no usage
- Accept the wrapper alongside the array and keyed-object shapes the parser already handles
- Correct the registry-format description in `docs/providers/crush.md`

### Detail

`loadRegistry` accepts a top-level array, or for any other object falls back to
`Object.values(parsed)`. On the wrapper that yields a single element — the array
itself — which fails the per-entry `path`/`data_dir` check. Zero databases are
discovered and `overview` reports "No usage found" for standard upstream Crush
registries.

`doctor` does not surface this: `probeRoots` (#903) confirms `projects.json`
exists without parsing it, so Crush reports healthy while `overview` returns
nothing.

Crush's centralized project registry has used this wrapper since the feature was
introduced:

```go
type ProjectList struct {
    Projects []Project `json:"projects"`
}
```

- Introduced in [`4b86d090`](https://github.com/charmbracelet/crush/blob/4b86d090392a761106d4b33b2937e183af62bcd6/internal/projects/projects.go#L17-L26) (Dec 15 2025)
- Unchanged at [`v0.95.0`](https://github.com/charmbracelet/crush/blob/v0.95.0/internal/projects/projects.go#L17-L26) and on [main](https://github.com/charmbracelet/crush/blob/0bca9525d5f89c6cae830beff0cfbbff77d10c13/internal/projects/projects.go#L17-L26)

I could not find any committed Crush version that writes a bare array or a keyed
object. The provider docs currently say:

> The registry shape is an object keyed by project id (modern Crush) or an array (older builds and tokscale's sample fixtures). The parser accepts both.

Current upstream Crush and tokscale's committed Crush fixtures use the wrapper.
This PR corrects that sentence but keeps both existing branches, since removing
them is a separate decision.

## Testing / Verification

- [x] `npm test` passes
- [x] Not a UI change

Crush `v0.95.0`, CodeBurn `0.9.24`, macOS. Same machine, same registry, same
databases; only the parser differs.

    before:  models --provider crush -p lifetime  -> No usage found
    after:   models --provider crush -p lifetime  -> 40 sessions, 2,447,391 tokens

Three tests added, one per accepted registry shape, each asserting the resolved
session. `npm test` on this branch: 4427 passed, 5 skipped. On `main`: 4425
passed, 5 skipped. `tsc --noEmit` clean on both.


ozymandiashh (approved):
Verified the diff: accepts the {"projects":[...]} registry shape while keeping the array and keyed-object forms working. Tests cover all three shapes. LGTM.

PR #1513, opened 2026-09-20
fix: count every Cursor Agent assistant message instead of one turn per user message

### LINKED ISSUE #1512: Cursor Agent usage severely undercounted: only first assistant message per user message parsed
## Symptom

`codeburn report --provider cursor-agent` shows a fraction of actual Cursor Agent usage. On one real corpus: 240 turns cached vs ~4,900 assistant messages on disk (~4% of assistant text counted).

## Root cause (`src/providers/cursor-agent.ts`)

1. `parseJsonlTranscript` only emits a turn for an assistant message that immediately follows a user message. Agentic loops emit dozens of assistant messages per user message; the rest are dropped. Same bug in the legacy text parser (`flushAssistant` drops when `pendingUsers` is empty).
2. `tool_use` blocks contribute only their name; the input JSON is never counted. `role: 'tool'` content is absent from these transcripts, so nothing else covers it.
3. `extractUserQuery` truncates to `MAX_USER_TEXT_LENGTH` (500) and the truncated length feeds `inputTokens` — prompts embedding whole skill files count as ~125 input tokens.
4. `conversation_summaries` in `ai-code-tracking.db` is empty in practice (0 rows here), and non-UUID conversation IDs are path hashes that can never match a summary row — so model is always `cursor-agent-auto` (priced as Sonnet 4.5) and every turn shares the file-mtime timestamp.

No real token counts exist on disk (transcripts and `chats/*/store.db` carry message text only), so char estimation stays — but it should cover all of the text.

iliaal:
Fixes #1512.

## What changed

- Every assistant message counts as a turn: each carries the last user message forward (JSONL and legacy text parsers).
- `tool_use` inputs are serialized into the output text instead of counting as zero.
- Input tokens use the full user text; the displayed message stays truncated at 500 chars.
- `cursor-agent` parse version bumped (`full-turn-accounting-v1`) so cached sessions re-parse.

## Verification

- 4 new regression tests (carry-forward jsonl/txt, tool-input counting, truncation split); provider suite 21/21.
- Full suite 4317 passed, tsc clean.
- Live corpus, September: 240 turns → 5,004 calls, 2.66M in / 1.36M out tokens, $28.40 (est.).

Still estimates — no real token counts exist on disk — and everything prices as Sonnet 4.5 (`conversation_summaries` is empty in practice).


iamtoruk:
Merged, thanks @iliaal. Your diagnosis held up: on one real corpus main counted 669 of 5,050 Cursor Agent assistant messages. I added three small commits on top: the user's prompt is now billed once per user message rather than once per assistant message (a long pasted prompt was being re-billed on every reply of the loop), the daily cache version is bumped so settled days re-derive with the new counts, and transcripts that end before any assistant reply (usage-limit and error stubs) no longer print an unrecognized-format warning.

